Issue #79, opened 2014-08-19
write go docs


mschoch:
current go docs are insufficient

need function/type level docs for entire top-level API
example code for top-level API
need function/type level docs for all important interfaces in sub-packages

mschoch:
@pmezard did a LOT of work in this area.  But we should do another review as we get closer to 1.0.

PR #137, opened 2015-01-03
adds examples for most query functions

sacheendra:
This has examples for most functions required while building a query. Related to #79

### DIFF
--- a/examples_test.go
+++ b/examples_test.go
@@ -18,9 +18,10 @@ import (
 
 var mapping *IndexMapping
 var example_index Index
+var err error
 
 func TestMain(m *testing.M) {
-	err := os.RemoveAll("path_to_index")
+	err = os.RemoveAll("path_to_index")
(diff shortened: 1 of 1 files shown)


sacheendra:
This is related to #79

mschoch:
Thanks for this.  I'd like to merge this, there are 2 things.

1.  I think its important for the examples to not ignore the returned error value.  I know it makes the examples a bit more verbose, but ignoring helpful error messages has been a frequent source of confusion for new bleve users.  I think the examples should store the error result into a variable, and assert that it is nil.

2.  Can you submit the Couchbase CLA?  There are instructions here: https://github.com/blevesearch/bleve/wiki/Contributing

Thanks.

sacheendra:
assertion of errors has been added.
I did submit the Couchbase SLA.

mschoch:
Thanks for the contribution!

Issue #120, opened 2014-11-18
custom analyzer too picky about types


mschoch:
The custom analyzer looks for attributes of type []interface{}, and its too easy to pass in []string and it silently does the wrong thing.

mschoch:
Fixed by 4abf2a638e93b08eef727876b28cbd6e92f351c5 thanks @sacheendra

PR #140, opened 2015-01-07
makes custom analyzer accept []string type attributes

sacheendra:
#120 Are there any tests for custom analyzer? I was not able to find them. If not how can one go about writing tests for it?


mschoch:
I was just about to review this and see why the build was failing.  Also, I was hoping to tweak it slightly to avoid using the reflect package, and just use type assertions.  But, now I see that you have closed it.  Are you planning to resubmit?

PR #141, opened 2015-01-07
adds handling of []string type attributes to custom analyzer

sacheendra:
#120 Are there any tests for this analyzer? I tried writing one, but it required importing too many components, everything from tokenizers, char filters, etc... Is there a better way of writing tests for this?

### DIFF
--- a/analysis/analyzers/custom_analyzer/custom_analyzer.go
+++ b/analysis/analyzers/custom_analyzer/custom_analyzer.go
@@ -11,6 +11,7 @@ package standard_analyzer
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


sacheendra:
Closed the previous one because there were some really silly mistakes in that one. 
Looked at your message later. Pushed a commit without using reflect.

sacheendra:
What is the problem with using "reflect"?

mschoch:
It just didn't seem necessary in this case.  It also required using string literals like "[]string" and "[]interface{}" instead of directly using the language constructs.

Issue #118, opened 2014-11-17
fix in memory implementation read isolation


mschoch:
The KVReader returned by the KVStore.Reader() method is supposed to be isolated from write operations, however the in memory implementation does not honor this contract.

sacheendra:
This test which checks for reader isolation passes for the in memory option. 
https://github.com/blevesearch/bleve/blob/master/index/store/inmem/store_test.go#L109
I do not understand how isolation is implemented in the in memory store. 
I don't see any locking based or snapshot based isolation in both reader.go and store.go
Can you explain how the isolation works?

sacheendra:
Just found that the the function referenced above is not being called at all.

mschoch:
It is not implemented at all, that is the bug.

Historically we did not have reader isolation at all.  Then later we added it.  Unfortunately, the current in-memory only implementation cannot easily be extended to offer reader isolation.  Sure, we could add a lock so that all reads and writes are serialized, but I'd rather we choose a new in-memory implementation that does some copy-on-write tricks to more easily allow reader isolation.

mschoch:
Yes, the problem was that all these tests were once shared, but it turned out to be inconvenient to try and share this code.

When they were copied into the kvstores directly that method was copied even though the in-memory test doesn't call it, because it is known to be broken.

sacheendra:
As there is an in memory implementation(gtreap), can this be closed. A treap is faster than a skip-list, so shouldn't gtreap be made the default inmem implementation.

mschoch:
Yes probably.  There is some active development going on in this area.  @steveyen has been experimenting with different options and benchmarking their performance.

I belive gtreap does pass the reader isolation test, and is about the same performance-wise as the current in-mem implementation.  So, from my perspective it probably should become the default.

However, we've also observed that at the moment, they are both slower than leveldb.  Leveldb has some advanced memory usage, but it still seems unusual and we're investigating.  In this process @steveyen found the cznic in memory btree.  This is faster than leveldb, but doesn't look like it can easily support reader isolation.

In summary, I would say its still an open topic.  But, trying to add in reader isolation to the current skiplist implementation probably isn't going to yield the best solution.

sacheendra:
@steveyen Can you share how you are benchmarking the stores? Can I help in any way?
I don't understand one thing about the gtreap implementation, it looks like multiple writers can make changes to the store at once. If multiple writers are executing batches at the same time, wont there be loss of data? 
[...]

PR #142, opened 2015-01-09
adds isolation to reader of in memory store using copy on write

sacheendra:
#118 Currently using inbuilt maps instead of a skip list for the copied elements. As the copied elements are temporary, I thought this would be better. Should I change the implementation to use a skip list?

### DIFF
--- a/index/store/inmem/reader.go
+++ b/index/store/inmem/reader.go
@@ -11,23 +11,23 @@ package inmem
 
 import (
 	"github.com/blevesearch/bleve/index/store"
-	"github.com/ryszard/goskiplist/skiplist"
 )
 
 // "newentry" marks a key which was previously empty, but was populated after the creation of the Reader.
 // If the value stored at the key has changed, the value at the time of creation of the Reader is stored in value.
 type readerValue struct {
 	value      []byte
-	newentry   bool
-	deleted    bool
-	firstValue bool
-	prevKey    string
+	newentry   bool   // was there a new entry on same key in the actual list
+	deleted    bool   // was the original key deleted from the actual list
+	firstValue bool   // was this the first value in the list
+	prevKey    string // the key located before the key of this value in the actual list
 }
 
 // This is the readers copy of the data which has changed after the Reader has been created.
 type readerData struct {
-	valueMap                map[string]*readerValue
-	prevValuesOfDeletedKeys *skiplist.SkipList
+	valueMap              map[string]*readerValue
+	prevKeysOfDeletedKeys map[string]string // contains keys located before the key deleted in the actual list mapped to the keys deleted
+	nextKeysOfDeletedKeys map[string]string // contains keys located sfter the key deleted in the actual list mapped to the keys deleted
 }
 
 type Reader struct {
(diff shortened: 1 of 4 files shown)


mschoch:
Could you describe in a few sentences how this adds reader isolation?

sacheendra:
Sure. Whenever a new Reader is created, a separate readerData object is created and passed to the store. The store has a map of readerData objects of all readers. Whenever a write or delete operation takes place, readerData objects of all Readers are checked if the key which is being changed is already  present. If it is not, it means that the key was not changed after the Reader was created. The object is added to the Reader's readerData object (specifically the valueMap in the readerData object). If a new key is created, a flag is set the readerValue object stored in readerData for that specific key so that a query on that key is answered with a nil response. The values in readerData along with the flag are used in Get function of the Reader to check for the existence of the keys in the Reader's readerData before actually checking the main storage. 
The case is similar with the iterator. Several flags are used to keep track of position of the iterator. The deletedKeysList is used in the iterator to iterate over deleted keys. 

In brief, reader isolation is achieved by storing old data in an object in every Reader before every set and delete. A special iterator is used to iterate over the data of the Reader.

mschoch:
Thanks for this.  I have found a problem.  The logic around deleted keys does not seem sufficient, specifically the way ```posOnDeletedList``` is used to keep track of calls to Next().  The simplest way to reproduce the problem is to delete keys out of order (though I don't think out of order deletion is the only issue here)  The following test method illustrates the problem.  Add a call to ```CommonTestReaderIsolationIteratorNext(t, s)``` to the end of ```TestReaderIsolation()``` right after ```CommonTestReaderIsolation()```

```
func CommonTestReaderIsolationIteratorNext(t *testing.T, s store.KVStore) {
	// store already has keys a and b

	// add key c
	writer, err := s.Writer()
	if err != nil {
		t.Error(err)
	}
	err = writer.Set([]byte("c"), []byte("val-c"))
	if err != nil {
		t.Fatal(err)
	}
	writer.Close()

	// create an isolated reader
	reader, err := s.Reader()
	if err != nil {
		t.Error(err)
	}
	defer reader.Close()

	// delete keys c then b
	writer, err = s.Writer()
	if err != nil {
		t.Error(err)
	}
	err = writer.Delete([]byte("c"))
	if err != nil {
		t.Fatal(err)
	}
	err = writer.Delete([]byte("b"))
	if err != nil {
		t.Fatal(err)
	}
	writer.Close()

	// get iterator from reader
	iter := reader.Iterator([]byte(""))
	// seek to a
	iter.Seek([]byte("a"))
	k, v, valid := iter.Current()
	if !valid {
		t.Errorf("expected iterator valid, got invalid")
	}
	if string(k) != "a" {
		t.Errorf("expected key 'a', got '%s'", k)
	}
	if string(v) != "val-a" {
		t.Errorf("expected value 'val-a', got '%s'", v)
	}

	// now call next
	iter.Next()

	// now check where we are
	k, v, valid = iter.Current()
	if !valid {
		t.Errorf("expected iterator valid, got invalid")
	}
	if string(k) != "b" {
		t.Errorf("expected key 'b', got '%s'", k)
	}
	if string(v) != "val-b" {
		t.Errorf("expected value 'val-b', got '%s'", v)
	}
}
```

PR #146, opened 2015-01-15
Fix parsing of the ignored ("-") json tag

pdf:
The current logic will return `-` as the field name for:

```go
type Demo struct {
	IgnoredField string `json:"-"`
}
```

This small change returns an empty string instead, so the that field is 
populated in the index with it's original name.

### DIFF
--- a/reflect.go
+++ b/reflect.go
@@ -72,6 +72,9 @@ func mustString(data interface{}) (string, bool) {
 
 // parseJSONTagName extracts the JSON field name from a struct tag
 func parseJSONTagName(tag string) string {
+	if tag == "-" {
+		return ""
+	}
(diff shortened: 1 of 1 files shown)


mschoch:
I agree that indexing the field with name "-" is wrong, but I don't agree with the fix.  To be consistent with the original intention of our support of the JSON struct tags, we should also ignore the field for indexing purposes.

I also agree with you that this isn't always what you want.  And while I need to review the details, your suggestion for some "bleve" struct tags is probably something we'll move forward with.  These would be treated with higher priority than the the JSON struct tags.  There is a bit more to this though, which I'll comment on in the other issues/pull-requests.

I will be pushing a commit to change this behavior and close the issue.

PR #147, opened 2015-01-15
Process pointer fields, rather than trying to walk them

pdf:
Pointers may be references to any type, existing logic will only handle types
understood by `walkDocument`, instead pass the deref'd element to
`processProperty`, where it can get passed back to `walkDocument` if 
necessary, or be processed as a regular field.

I've done some cursory testing and this has fixed issues I was having, but my
testing has not been extensive and it's possible I've missed something.


mschoch:
Thanks for this fix.  I forgot to ask, did you sign the Couchbase CLA?  https://github.com/blevesearch/bleve/wiki/Contributing

pdf:
CLA is signed, thanks Marty

PR #148, opened 2015-01-15
Handle numeric fields other than float64 in processProperty

pdf:
Assert non-float types to type float64, then re-process.

### DIFF
--- a/mapping_document.go
+++ b/mapping_document.go
@@ -294,7 +294,13 @@ func (dm *DocumentMapping) processProperty(property interface{}, path []string, 
 				}
 			}
 		}
-	case reflect.Float64:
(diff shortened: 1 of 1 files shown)


mschoch:
I'm starting to be persuaded that this is the right way to go for now.  We would need to test and document that integer precision is limited by conversion through float64.  But that this is acceptable for many use cases, so we should not let that block integrating the feature.  Fixing this correctly can be put off to bleve2, etc.

murz:
I need to index a struct with int fields so I just tried this out and it works great for my needs. Thanks @pdf

slavikm:
@mschoch - would really love to get this merged. Just my +1 :-)

PR #177, opened 2015-03-18
Added support for setting a timeout on a search request.

typerandom:


typerandom:
Also, one thing that I missed. Probably good to move the TimeoutError to error.go and change it so that it is consistent with the errors that are defined there.

mschoch:
So, I like the idea here, but I have some concerns with the implementation.  My biggest concern is that it lacks the ability to actually interrupt a long running call to Next().  Instead, it returns after the first such call.  So, if your timeout is 5 seconds, and the first call to Next() takes 30, you won't return until after 30 seconds.  Now, arguably this is better than what we have today, which would keep going after 30 seconds, but I worry that its not what users will expect of a timeout.

My preference is that we add a Context object to the request, add pass that down through all children seachers, possibly even index and kvstore iterators (see https://blog.golang.org/context)  The contract would be for all operations to honor the Done channel.

This would allow timeouts to be built at the top-level, but also allow for non-timeout based cancellation, say if the client closes the connection after issuing a query.

Thoughts?

typerandom:
I fully agree. The reason why I went the way I did was because it was the easiest and least invasive fix. But it's a very limited solution. I did not know about the Context interface, but definitely think that it's the way to go. Will look more into a solution like that.

mschoch:
we just added support for this here: https://github.com/blevesearch/bleve/commit/0b2380d9bf4cc7714f34f9dd61a02cb17e1ce2c0

Issue #202, opened 2015-05-14
panic in highlight fragmenter


dtynn:
this is my code 
https://gist.github.com/dtynn/17ce8d32b9284bebbd7e
and this is the panic info
https://gist.github.com/dtynn/7d7fd0003ec6c43ce952

https://github.com/blevesearch/bleve/blob/master/search/highlight/fragmenters/simple/fragmenter_simple.go#L39
it seems that the method here should check the orig's length?

dtynn:
https://gist.github.com/dtynn/8be6ed079fa6df8f0270
well , the code get the result 

2015/05/14 14:48:27 result: 123: 1 matches, showing 1 through 1, took 340.655µs
    1. highlight (0.137229)
	Tags
		<span class="highlight">abc</span>

which is obviously incorrect.

maybe it should do something more to make sure the docField is the right one.
https://github.com/blevesearch/bleve/blob/master/search/highlight/highlighters/simple/highlighter_simple.go#L81

mschoch:
Thanks, I believe this is a duplicate of #170

mschoch:
I just pushed a fix for this issue and retried your example code.  I now get:

```
2015/07/31 14:40:53 result: 123: 1 matches, showing 1 through 1, took 102.987µs
    1. highlight (0.137229)
	Tags
		<mark>123</mark> 45 678
```

Please reopen it if you continue to observe the problem.

dtynn:
thanks marty, it seems good

PR #203, opened 2015-05-14
add index to the path when walking the document

dtynn:
for the issue: https://github.com/blevesearch/bleve/issues/202
for slices and arrays, I add a mongodb-style index to the path
and it works.


mschoch:
I don't think this the best fix for this problem.  We are already storing an array of index offsets corresponding to where in an array the stored field came from.  The problem is that the highlighter is not checking/using that.

PR #206, opened 2015-05-17
update TermVector

dtynn:
update TermFieldVector struct and TermVector struct
update the index format for TermVector
add a test case

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -61,10 +61,11 @@ type IndexReader interface {
 type FieldTerms map[string][]string
 
 type TermFieldVector struct {
-	Field string
-	Pos   uint64
-	Start uint64
-	End   uint64
+	Field          string
+	ArrayPositions []uint64
+	Pos            uint64
+	Start          uint64
+	End            uint64
 }
 
 type TermFieldDoc struct {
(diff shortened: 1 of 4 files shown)


dtynn:
1. Yes. The formatted value contains at least the length of the arrayPositions for each vector
2. Just did it.
3. Just did it, too. BTW, google says that the OpenID 2.0 is no longer supported when I tried to sign in with my google account.

Have a good vacation :)

mschoch:
OK, I'm back to looking at this again.  I think my concern with item 1 is that many documents don't contain any arrays at all, so I'd like to use an encoding scheme where those don't take up any additional space at all.

I'll review your implementation and see what we can do.

dtynn:
Hi marty,
I found another problem when I was looking into the item 1.

It might take 10 bytes to contain an uint64 using binary.PutUvarint.
18446744073709551615 => []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x01}
(see http://play.golang.org/p/OoVpTq9qEg)

https://github.com/blevesearch/bleve/blob/master/index/upside_down/row.go#L322
So, are 8s here mistakes, or they are used on purpose?

mschoch:
Hmmn, yes I've made that mistake elsewhere as well.  The 8 is only correct for fixed size 64-bit integers, the 64bit var ints we should use the constant binary.MaxVarintLen64.

mschoch:
I went ahead and merged this along with additional changes to support highlighting fields containing arrays.  That you for this contribution.

PR #215, opened 2015-06-24
moves goleveldb to a build tag

drewwells:
The default k/v store is now boltdb. So it made sense to me to isolate the building of goleveldb to a build tag for those not using goleveldb.

We recently encountered problems using bleve, because of this https://github.com/syndtr/goleveldb/commit/a065095 and this https://github.com/syndtr/goleveldb/commit/ca34d8a. The dependency changes in goleveldb were being picked up by `go get` but confused godeps in ways causing our builds to fail. Since we were not familiar with goleveldb, this caused a lot of build headaches.

That's the problem I'm trying to solve here. I tried to adhere best to the patterns found in the repo about putting this kv store behind a build tag. Please let me know if this makes sense to you!

### DIFF
--- a/config.go
+++ b/config.go
@@ -102,7 +102,6 @@ import (
 
 	// kv stores
 	_ "github.com/blevesearch/bleve/index/store/boltdb"
-	_ "github.com/blevesearch/bleve/index/store/goleveldb"
 	_ "github.com/blevesearch/bleve/index/store/gtreap"
 	_ "github.com/blevesearch/bleve/index/store/inmem"
 
--- a/config_goleveldb.go
+++ b/config_goleveldb.go
@@ -1,0 +1,21 @@
+//  Copyright (c) 2014 Couchbase, Inc.
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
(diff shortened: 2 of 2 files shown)


drewwells:
build time for bleve went from 4.5sec to 3.8sec on my system

drewwells:
@mschoch have you had a chance to look at this?

mschoch:
Yeah, the reason I didn't act on it yet is that I'm considering doing it as a part of a larger change.  You sort of alluded to it with your comment about the build time.  Right now bleve includes a lot of things by default (using imports in config.go).  Initially I liked this idea of lots of things included by default... But more and more I'm leaning towards a very lean default, and have users explicitly import additional things they need.  General purpose apps (like bleve_explorer) that don't know what users will need at runtime can continue to import things like config.go does today.  But that should be app decision, not a library default.

As an extension of this change, I'd lean towards apps explicitly including the packages instead of using build tags.  It just seems like we should prefer using the language features to only actually include what we're using.

I was planning on posting to the google group about this soon, but feel free to share your thoughts on this change here if you want.

drewwells:
:+1: that is a great idea. If there's a register function where we pick our backend, that would be even better than sliding various build tags in there. I'm glad to see you're looking into this, I'll close this pull request and look forward to seeing the discussion on google groups.

PR #226, opened 2015-08-11
Allow providing a token filter config in custom analyzer

donhcd:
this makes it possible to add a config for token filters when using a custom analyzer, which is necessary for filters such as ngram and edge_ngram


mschoch:
I'm not sure I understand what problem this solves.  You already can define custom token filters, and then use them in custom analyzers, it is done like this: https://github.com/blevesearch/beer-search/blob/master/mapping_example1.go

donhcd:
oops, I didn't see that. thanks for the tip

mschoch:
In the future, don't hesitate to ask in IRC or on the google group first, we usually respond pretty quickly.

PR #233, opened 2015-09-21
reducing allocation count

kevgs:
Hello.

Found this by benchmarking my code which puts documents into bleve index. mallocgc and its friends are at the top of my CPU profile and it looks like bleve does too much allocations. This simple path significantly decrease allocations count in my test case.

### DIFF
--- a/index/upside_down/row.go
+++ b/index/upside_down/row.go
@@ -276,7 +276,7 @@ type TermFrequencyRow struct {
 	doc     []byte
 	freq    uint64
 	norm    float32
-	vectors []*TermVector
+	vectors []TermVector
 }
 
 func (tfr *TermFrequencyRow) ScanPrefixForField() []byte {
@@ -321,8 +321,9 @@ func (tfr *TermFrequencyRow) DictionaryRowKey() []byte {
 func (tfr *TermFrequencyRow) Value() []byte {
 	used := 0
 	bufLen := binary.MaxVarintLen64 + binary.MaxVarintLen64
-	for _, vector := range tfr.vectors {
-		bufLen += (binary.MaxVarintLen64 * 4) + (1+len(vector.arrayPositions))*binary.MaxVarintLen64
+	for i := range tfr.vectors {
+		l := len(tfr.vectors[i].arrayPositions)
+		bufLen += (binary.MaxVarintLen64 * 4) + (1+l)*binary.MaxVarintLen64
 	}
 	buf := make([]byte, bufLen)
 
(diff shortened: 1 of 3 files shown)


mschoch:
Thanks for this.  I'm in the middle of some other performance optimizations.  So I will be testing this as well.

Two things:

1.  Can you sign the CLA?  http://www.blevesearch.com/docs/Contributing/
2.  Can you share the code or scenario you benchmarked?  It will help me measure/validate the improvement and also help me ensure we don't hurt any other use cases.

Thanks in advance for this contribution.

kevgs:
CLA is signed and here is my code https://gist.github.com/kevgs/42769be6bb90db1e3797

kevgs:
Hi. Any issues with that PR?

mschoch:
Sorry for the delay.  I don't have any issues with it, I've just been in the middle of some other performance improvements.  I'll try to review it this weekend.

mschoch:
OK, so I've been running the code you sent, both with and without the PR.  It might help if you share the numbers you're seeing as an improvement.

First, I was looking at the memory profile, before the patch I pprof reported 613MB used after the patch pprof reported 950MB.  Then I realized you were concerned with fewer allocations, not less memory allocated.  So, I ran both again, with no profiling, just to measure the overall time, to see if the program was faster with the patch.  Before the patch it takes 3m11.994s and after the patch it takes 3m9.354s as measured simply using the time command to run the program.  Thats a pretty small gain, and possibly not even significant if I were to run both multiple times.

Can you share more information about the improvement you observed with this PR?

kevgs:
Running my script I found that most of CPU time is spent in GC.

This patch significantly reduces allocations. But after applying it most of CPU time is still in GC. Because we still have a lot of allocations.

Overall time with patch is the same for me too. But it's not serious to measure it with that script which does an HTTP request for every inserted document. I will provide a better bench test some time later. For now I just can share this https://gist.github.com/kevgs/e4b8be49d78beb1f9d1b

613MB and 950MB is a peak memory, I guess. If so, it surprises me - in theory second value should be lesser that first due to smaller heap fragmentation. I will check this when a can.

kevgs:
The patch isn't worth committing. Sorry for bothering your time.

mschoch:
No bother, I appreciate you trying to make it better.  We are working on improving performance, a good part of which is reducing allocations.  Many of the seemingly obvious improvements like using sync.Pool to reuse objects have not yielded gains.  At the moment it appears that if we apply sync.Pool on too fine-grained things we trade allocation time for lock contention with no net win.  So we're having to look at reusing things at a more coarse-grained level.

I mention this just so that you know we are working on it, but at times it is slow going.  Hopefully we'll have concrete improvements soon.

PR #238, opened 2015-09-28
fix compliation with the latest changes to kagome

ikawaha:


analysis/language/ja/ja_morph_kagome.go
```diff
@@ -15,23 +15,27 @@ import (
...
 
 const TokenizerName = "kagome"
 
 type KagomeMorphTokenizer struct {
-	tok *kagome.Tokenizer
+	tok tokenizer.Tokenizer
+}
+
+func init() {
+	_ = tokenizer.SysDic() // prepare system dictionary
```
Shugyousha: On error this will panic but I think that is okay. It does not make sense to recover if the analysis tool chain initialization failed.

Shugyousha:
Thanks for the pull request! That way the bleve people don't have to make code adjustments themselves :)

I applied the patch locally and the tests pass so :+1: from me.

mschoch:
Thanks for contributing this.  I agree panic inside init method is acceptable.

PR #249, opened 2015-10-13
Add search for dates and times.

zaccone:
This patch extends parser capabilities for date searching.
Parser's grammar has been extended for following cases:

* tLESS tPHRASE (<)
* tLESS tEQUAL tPHRASE (<=)
* tGREATER tPHRASE (>)
* tGREATER tEQUAL tPHRASE (>=)

As currently there is no logic for interpreting strings with operators
for anything but dates we assume user provided input for searching on
date/time attributes in their mappings. Bleve supports few basic formats
of date and/or datetime:

* date only (US format): 02-01-2006
* date only (European format): 01-02-2006
* RFC822: 02 Jan 06 15:04 MST
* UnixDate: Mon Jan _2 15:04:05 MST 2006
* RFC3339: 2006-01-02T15:04:05Z07:00
* RFC3339Nano: 2006-01-02T15:04:05.999999999Z07:00

### DIFF
--- a/query_string.nn.go
+++ b/query_string.nn.go
@@ -26,6 +26,8 @@ type lexer struct {
 	// Since then, I introduced the built-in Line() and Column() functions.
 	l, c int
 
+	parseResult interface{}
+
 	// The following line makes it easy for scripts to insert fields in the
 	// generated code.
 	// [NEX_END_OF_LEXER_STRUCT]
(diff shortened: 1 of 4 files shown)


mschoch:
This looks good, a couple of things:

1.  Is it possible to submit without all the white-space changes to query_string.y?  This makes it hard to figure out what were the actual important changes.
2.  Unfortunately the output of nex is not deterministic.  It looks like you didn't change the nex rules, so in that case I would not add the .nn.go file to the commit.

mschoch:
Thanks for this contribution, I had to rework it a little bit to use the existing datetime parser, but the functionality works largely as you proposed.

PR #254, opened 2015-10-18
tweak queryparser 'go generate' lines (portablity)

bcampbell:
sed is used to massage the .go files output from nex and yacc.
However, the -i (in-place) option works slightly differently
between gnu (eg Linux) and bsd (eg OSX) sed.
This change should make it work on both flavours.


mschoch:
This looks good, but can't we also just 'rm' the .tmp files as well?

telendt:
The most portable way is to use in-place replace in sed is to call it like:

    sed -e COMMAND -i'' -- path/to/file

mschoch:
Using single quotes worked from the shell, but when executing it through go generate I would end up with filename.'' on my Mac.

telendt:
Oh yes, you're right (looks that haven't tested it properly before sending my  last message).

bcampbell:
For some reason I don't think it ever occurred to me to add `rm` lines, but it seems like a good plan. I just made the change in my fork. Works fine on my Linux box, but not tried it yet on windows or osx.

Issue #109, opened 2014-10-09
support indexing and specialized querying of boolean fields


mschoch:

mmindenhall:
I just ran into the case where I have some `bool` fields in structs that aren't getting indexed.  I checked the source code and didn't see where that would be happening, then found this issue.  I see this has been open for over a year...is this especially difficult to do for some reason?  If not, I might give it a shot.

mmindenhall:
As a start, I'd be happy with indexing bools as string "true" or "false", or as numeric 0 or 1.

PR #286, opened 2015-11-19
Index boolean fields as strings ("true"/"false")

mmindenhall:
I think this should solve #109.  I don't know what sort of "specialized querying" you had in mind...for my purposes being able to do a field specific query is good enough.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@
 /utils/bleve_registry/bleve_registry
 /y.output
 *.test
+
+# IDE metadata
+**/.idea/
+**/*.iml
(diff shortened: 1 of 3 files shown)


mmindenhall:
Any comments or thoughts on this one?

mschoch:
I did review some background material today.  I think what I'd prefer to do is introduce a BooleanField.  This will handle mapping not just booleans, but also other types into boolean.  So, cases where you have a string or number you want to interpret as boolean.  In the end it will index it very similar to this, Elasticsearch uses terms T and F instead of "true" and "false".  But there would be no special searches, you just use normal term searches.

mschoch:
I think we may still want to support a user explicitly choosing to index a boolean value as a string, like "true" or "false".

PR #287, opened 2015-11-21
Fixes datetime mapping from JSON, using DateTimeFieldMapping

mmindenhall:
In the following case, datetime fields are not getting indexed:
- When using an explicit `DateTimeFieldMapping` instead of relying on dynamic mapping
- When indexing from JSON instead of a struct

Test included that demonstrates the problem, and passes with the fix.


mschoch:
Thanks for the bug fix!

Issue #296, opened 2015-12-25
firestorm panics with divide by zero when no documents


aybabtme:
The firestorm index panics with a 'divide by zero' error when the GC runs and no documents have been indexed so far.

mschoch:
Do you have the details of the panic?  While disabling GC when doc count is 0 is fine, I'm wondering where the existing code fails.

aybabtme:
sure!

```
panic: runtime error: integer divide by zero
[signal 0x8 code=0x7 addr=0x40d18e2 pc=0x40d18e2]

goroutine 40 [running]:
github.com/blevesearch/bleve/index/firestorm.(*GarbageCollector).run(0xc820168d20)
	/Users/antoine/gocode/src/github.com/blevesearch/bleve/index/firestorm/garbage.go:79 +0x2e2
created by github.com/blevesearch/bleve/index/firestorm.(*GarbageCollector).Start
	/Users/antoine/gocode/src/github.com/blevesearch/bleve/index/firestorm/garbage.go:55 +0x58
```

PR #297, opened 2015-12-25
Firestorm: dont gc if no documents

aybabtme:
This fixes https://github.com/blevesearch/bleve/issues/296 and adds a reproducer so that the problem doesn't show up again later.


mschoch:
Thanks for the fix.

aybabtme:
:+1: :santa: :gift:

PR #318, opened 2016-01-12
Implemented boolean field support

### LINKED ISSUE #109: support indexing and specialized querying of boolean fields

slavikm:
First go at implementing bool field support - fixed #109

### DIFF
--- a/mapping_field.go
+++ b/mapping_field.go
@@ -73,6 +73,16 @@ func NewDateTimeFieldMapping() *FieldMapping {
 	}
 }
 
+// NewBooleanFieldMapping returns a default field mapping for dates
+func NewBooleanFieldMapping() *FieldMapping {
+	return &FieldMapping{
+		Type:         "boolean",
+		Store:        true,
+		Index:        true,
+		IncludeInAll: true,
+	}
+}
+
 // Options returns the indexing options for this field.
 func (fm *FieldMapping) Options() document.IndexingOptions {
 	var rv document.IndexingOptions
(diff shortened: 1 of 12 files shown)


mapping_field.go
```diff
@@ -73,6 +73,16 @@ func NewDateTimeFieldMapping() *FieldMapping {
 	}
 }
 
+// NewBooleanFieldMapping returns a default field mapping for dates
+func NewBooleanFieldMapping() *FieldMapping {
```
Shugyousha: copy/paste error. Should be "... a default field mapping for booleans"

mschoch:
Thanks for this for this I'll try to review this afternoon.  Have you had a chance to sign the CLA?  http://www.blevesearch.com/docs/Contributing/

Unfortunately right now it requires OpenID to authenticate.  If you run into any problems let me know and I'll try to help.  (we are working on making this process easier)

mapping_field.go
slavikm: Thanks, done.

slavikm:
@mschoch where do I sign the CLA? I logged in via my old Yahoo ID. How do I proceed from there?

mschoch:
It should be under  `Settings > Agreements`

slavikm:
@mschoch found it and signed.

mschoch:
Thanks, I've confirmed the signed CLA and first review looks good.  Tomorrow I'll pull it down and do some more testing.

mschoch:
OK, I reviewed more carefully today and I think it looks good.  There are a couple of minor things I need to fix up and rebase locally.  But I'll basically be merging this as is.

Thanks again for chipping in and helping with this!

slavikm:
@mschoch - great!

Not directly related but how would you like to support the various number types? I would like Bleve to index all of them (int, int32, int64, etc.)

There is a simple way to just cast to float64 and continue from there - any reason not to do so?

mschoch:
There is PR that does that already: https://github.com/blevesearch/bleve/pull/148

I think there are 2 problems:

1.  The implementation internally requires having the number inside a float64.  But float64 can only accurately represent 56-bit integers.  So, while it will appear to just work, it will fail in some cases.  I think everything other than 64bit integers is fine.

2.  We lose the original type information.  So when loading the stored field version of this, we'll always return it as float64.  I think this is probably OK in the short term.  Our implementation of stored fields needs to be overhauled in general, but we shouldn't let that hold up this issue.

So how can we address 1?  I think we'd either need to create yet another type just for 64bit integer (i'm not a fan of this) or we'd have to find a way to use more bits to ensure we preserve things correctly.  I'm open to other ideas/suggestions here.

slavikm:
Not sure I follow how numbers are stored and searched. Is there a wiki entry or design doc that explains this? I think I follow the code but want to make sure I understand correctly.

mschoch:
The approach is basically the one used by Lucene and described here:

https://lucene.apache.org/core/5_4_0/core/org/apache/lucene/util/NumericUtils.html

slavikm:
Well, it might not be a great solution but I would go with the casting and just document the precision issue you specified above. It definitely works for our use case :-)

PR #326, opened 2016-01-15
Scorers allocs

kevgs:
benchmark                       old ns/op     new ns/op     delta
BenchmarkConjunctionScore-2     142           26.4          -81.41%
BenchmarkTermScore-2            1373          1226          -10.71%

benchmark                       old allocs     new allocs     delta
BenchmarkConjunctionScore-2     1              0              -100.00%
BenchmarkTermScore-2            7              6              -14.29%

benchmark                       old bytes     new bytes     delta
BenchmarkConjunctionScore-2     80            0             -100.00%
BenchmarkTermScore-2            792           712           -10.10%

### DIFF
--- a/search/scorers/scorer_conjunction.go
+++ b/search/scorers/scorer_conjunction.go
@@ -15,6 +15,7 @@ import (
 
 type ConjunctionQueryScorer struct {
 	explain bool
+	rv      search.DocumentMatch
 }
 
 func NewConjunctionQueryScorer(explain bool) *ConjunctionQueryScorer {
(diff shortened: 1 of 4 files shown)


mschoch:
I have some concerns with this, as there is now state in the scorer which was not the intended design.  I'm not saying it can't work, but that wasn't how it was expected to work.

Also, the integration tests fail.  I'm not sure if its related to that or not.

kevgs:
I see no problems with properly encapsulated state. But this patch add a restriction on scorer users - it's now disallowed to change value returned by Score(). It's possible to document this restriction on all scorers. The benefit is a better performance. Do you thinks it's worth it?

Integration test failure is probably not related to patch.

mschoch:
Well the tests are run on every push and have been passing.  However, when run against this pull-request they fail.  So this patch is the most likely cause at the moment.

The part I don't understand is that the same TermQueryScorer is used to score multiple matches.  But, since you only ever allocate one DocumentMatch, aren't you constantly overwriting the same DocumentMatch, and returning references to the same one each time?

mschoch:
I've reviewed this a final time.  Reducing allocations is great, but unfortunately this change breaks the intended usage of Scorers.  Currently searchers instantiate a single scorer, and use it to score all the hits found by the searcher.  This change only ever creates one DocumentMatch object, and all calls to Score keep modifying this single instance and returning references to it.

The unit tests still pass because they check the contents of the DocumentMatch after each call to Score, but the problem is that we collect all the DocumentMatches and percolate them back up the searcher hierarchy.  This problem is exposed by the integration tests, which is why they are failing.

PR #336, opened 2016-02-03
pass the boost value into the term searcher

### LINKED ISSUE #333: Boost values are not used for anything outside term queries
For example, NewRegexpSearcher just uses 1.0 instead of the boost parameter it receives.

slavikm:
Fixed #333


mschoch:
Working on this now.  Adding some tests to verify correctness.

PR #338, opened 2016-02-09
Partial results for search

nimishzynga:


steveyen:
0.02cents...

I like the stop on error idea.

To generalize, I guess there would be StopOnNumErrors?  Not sure if it's worth it, though, and the extra tracking.

On success vs failed counts, if you (also? instead?) provide that array (or map?) of errors, then as I user, I can do the length() count myself if I wanted, but would have all the granular error msgs on hand?  (To log or display them, I guess if length of errors > 0).

If there's a timeout (future feature?), would that one day be expressed as just yet another error?

mschoch:
We've gone ahead and implemented this an alternate way.

PR #340, opened 2016-02-17
Fix masking of error when creating index (minor error reporting fix)

mmindenhall:
Our continuous integration tests are failing with `cannot create new index, path already exists` when creating a test index.  I can't reproduce this error on my laptop (macbook pro) -- it only happens in CI (shippable.com, ubuntu in a Docker container).  I added a line to explicitly remove the path before creating the index, and am seeing the same failure.  I think the code I changed was masking the actual error by always returning the "path already exists" error without first checking the error type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 .#*
 .project
 .settings
+**/.idea/
+**/*.iml
 .DS_Store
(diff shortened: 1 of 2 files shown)


mschoch:
Thanks

PR #370, opened 2016-04-15
do not put +/-Inf or NaN values into the stats map

mschoch:


steveyen:
Hi Marty - this looks good to me

PR #372, opened 2016-04-28
Load the document only once for both fields and highlighter

slavikm:


mschoch:
Thanks for the great performance contributions.

Issue #374, opened 2016-05-18
Data leak with goleveldb backend?


mmindenhall:
The application I'm working on runs on IoT gateways with limited resources (~100MB storage free).  Therefore, I submitted #373 to be able to proactively reclaim disk space after deleting "expired" documents from indexes.

I wrote a test where I do the following:

1.  Create and initialize a new index, take snapshot of size of index folder
1.  Do the following 5 times:
    1.  Add 1000 documents to the index
    1.  Take snapshot of size of index folder
    1.  Delete all documents from the index
    1.  Call the new `Compact` method I added in #373 
    1.  Take snapshot of size of index folder

Here are the sizes of the index folder:

Size at start:  52K

| Iteration | After adding 1K docs | After delete / compact |
| ---------- | ----------------------: | -----------------------: |
| 1 | 6.7M | 876K |
| 2 | 7.4M | 1.6M |
| 3 | 8.1M | 2.2M |
| 4 | 8.7M | 2.9M |
| 5 | 9.4M | 3.5M |

Just to be sure the `Compact` method was actually doing something, I commented out that line and ran again (create 1000, delete 1000, no compact).  Without the call to `Compact`, deleting the 1000 documents actually _increases_ the size of the index:

| Iteration | After adding 1K docs | After delete only |
| ---------- | ----------------------: | -----------------------: |
| 1 | 6.6M | 8.5M |
| 2 | 7.4M | 9.3M |
| 3 | 8.0M | 9.9M |
| 4 | 8.6M | 11M |
| 5 | 9.4M | 11M |

Is this a bug?  I'm wondering if there's some set of keys that gets created when a document is indexed that is not getting deleted when the document is removed?

mschoch:
Can you run the bleve_dump utility to see what data (if any) is still there?

If bleve_dump doesn't show anything, you can use the lower-level 'leveldbutil' that comes with leveldb to dump just the contents of the leveldb file.

Running those after one of the iterations should give some insight.

mschoch:
There are a few things that are expected to be left around.

1.  The index mapping itself is stored. (frequently this is less than 1k)
2.  The field definitions exist independent of any documents being indexed.  So there should be 1 row for each field, again in practice even if you had hundreds of these they are just a few bytes each.

mmindenhall:
I ran the test with just one iteration, and with just 10 documents.  After delete and compact, I ran bleve_dump.  There are a huge number (1750, to be exact) of `Dictionary Term` entries:

```
[...]

PR #373, opened 2016-05-17
Add compact method to goleveldb store

mmindenhall:
This is as described in #363.

PR #376, opened 2016-05-26
Remove DictionaryTerm with count 0 during compact (workaround for #374)

mmindenhall:
I spent a couple hours on this last night as a workaround for #374.

The implementation removes all DictionaryTerm entries with Count=0 from the index, in configurable batches, within a transaction.  Originally I did this all in one large transaction, but settled on this approach to avoid locking out other writers for an extended period of time.  A batch size of 250 seems like a good default number for a server-based implementation...I'll be using batch sizes of more like 50-100 in my app (running on a single core ARMv7).

I ran this test, that I wrote before submitting #374:

>1.  Create and initialize a new index, take snapshot of size of index folder
>1.  Do the following 5 times:
>    1.  Add 1000 documents to the index
>    1.  Take snapshot of size of index folder
>    1.  Delete all documents from the index
>    1.  Call the new `Compact` method I added in #373 
>    1.  Take snapshot of size of index folder

I verified that there were no dictionary term entries present after the test, and the `.ldb` file contained only the document mapping and fields.

I also ran a variation of the test above to ensure that documents could still be indexed during the execution of `CompactWithBatchSize`.  Just before calling `CompactWithBatchSize`, I started another goroutine to index documents during the compact.  I verified that the documents were indexed, document count after finishing was correct, and there were no dictionary term entries with count 0.

### DIFF
--- a/index/store/goleveldb/store.go
+++ b/index/store/goleveldb/store.go
@@ -10,6 +10,7 @@
 package goleveldb
 
 import (
+	"bytes"
 	"fmt"
 
 	"github.com/blevesearch/bleve/index/store"
@@ -19,7 +20,10 @@ import (
 	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
-const Name = "goleveldb"
+const (
+	Name             = "goleveldb"
+	defaultBatchSize = 250
+)
 
 type Store struct {
 	path string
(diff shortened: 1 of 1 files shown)


mmindenhall:
Yes...definitely makes more sense that way.  I made the change, and ran my tests again.

PR #380, opened 2016-06-01
fix pagination bug introduced by collector optimization

### LINKED ISSUE #378: search: .From bug
Haven't looked into this too too much but i think the bug is [here](https://github.com/blevesearch/bleve/blob/105626269cf398e9de7c2ffeac0cfa1cf60c2f2b/search/collectors/collector_top_score.go#L42), `k: k` should be `k: k + skip` or something.

cheers

mschoch:
fixes #378

this bug was introduced by:
https://github.com/blevesearch/bleve/commit/f2aba116c49ea51b27bc9afd3bf15305ef04883c

theory of operation for this collector (top N, skip K)

- collect the highest scoring N+K results
- if K > 0, skip K and return the next N

internal details

- the top N+K are kept in a list
- the list is ordered from lowest scoring (first) to highest scoring (last)
- as a hit comes in, we find where this new hit would fit into this list
- if this caused the list to get too big, trim off the head (lowest scoring hit)

theory of the optimization

- we were not tracking the lowest score in the list
- so if the score was lower than the lowest score, we would add/remove it
- by keeping track of the lowest score in the list, we can avoid these ops

problem with the optimization
- the optimization worked by returning early
- by returning early there was a subtle change to documents which had the same score
- the reason is that which docs end up in the top N+K changed by returning early
- why was that? docs are coming in, in order by key ascending
- when finding the correct position to insert a hit into the list, we checked <, not <= the score
- this has the subtle effect that docs with the same score end up in reverse order

for example consider the following in progress list:

doc ids [   c    a    b  ]
scores  [   1    5    9  ]

if we now see doc d with score 5, we get:

doc ids [   c    a    d    b  ]
scores  [   1    5    5    9  ]

While that appears in order (a, d) it is actually reverse order, because when we
produce the top N we start at the end.

theory of the fix

- previous pagination depended on later hits with the same score "bumping" earlier
hits with the same score off the bottom of the list
- however, if we change the logic to <= instead of <, now the list in the previous
example would look like:

doc ids [   c    d    a    b  ]
scores  [   1    5    5    9  ]

- this small change means that now earlier (lower id) will score higher, and
thus we no longer depend on later hits bumping things down, which means returning
early is a valid thing to do

NOTE: this does depend on the hits coming back in order by ID.  this is not
something strictly guaranteed, but it was the same assumption that allowed the
original behavior

This also has the side-effect that 2 hits with the same score come back in
ascending ID order, which is somehow more pleasing to me than reverse order.


mschoch:
Would appreciate review from @steveyen and/or @slavikm 

See commit message for explanation of behavior and theory behind fix.

mschoch:
Oh, sadly this breaks the order of results in **ALL** the integration tests.  So I have more work to do...

steveyen:
This looks good to me.

Side note: this is the best example of explanation to LoC change ratio (1 character) I've seen in a long time.

slavikm:
👍

PR #381, opened 2016-06-01
Compact for boltdb (workaround for #374)

mmindenhall:
This is similar to PR #376 (workaround for issue #374), except for boltdb.  

Sorry I didn't reach this conclusion sooner, but it looks like goleveldb won't work for our application even with the new compact logic.  In continued stress testing on a gateway device (128MB RAM, ~100MB available flash), we saw the following:

1.  Within `Compact`, can successfully remove unused `DictionaryTerm` entries.
1.  Application crashes (out of memory) during call to `ldbs.db.CompactRange(util.Range{nil, nil})`.  Apparently `CompactRange` uses a lot of extra memory to remap the file.

I swapped out goleveldb for boltdb and tried all of this again.  With boltdb, we can do the following to manage our filesystem resources without crashing:

1.  Calculate the maximum size we can allow the DB file to reach based on partition free space
1.  When the DB reaches maximum allowable size, begin monitoring the size of the free page list
1.  When the free page list reaches a threshold (prior to where boltdb would try to grow the file):
    1.  Purge oldest documents to free up space
    1.  Call this `Compact` to free up unused `DictionaryTerm` entries

Again, the code from both of these PRs can be removed once #374 is solved.

### DIFF
--- a/index/store/boltdb/store.go
+++ b/index/store/boltdb/store.go
@@ -28,8 +28,8 @@ import (
 )
 
 const (
-	Name             = "boltdb"
-	defaultBatchSize = 100
+	Name                    = "boltdb"
+	defaultCompactBatchSize = 100
 )
 
 type Store struct {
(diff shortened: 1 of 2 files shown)


mmindenhall:
Note that the default batch size for removing `DictionaryTerm` entries is 100 (vs 250 for goleveldb), as writes are much slower with boltdb.

mschoch:
Can you change `defaultBatchSize` to something like `defaultCompactBatchSize` since thats currently the only thing its used for?

mmindenhall:
Done, for both goleveldb and boltdb implementations.

PR #383, opened 2016-06-03
init. simple camel case parser.

a-little-srdjan:
The code is geared towards readability (a simple push-down automata), and making it easy to change the token classes. People often have strong, differing, opinions on exactly how to slice a string :)

### DIFF
--- a/analysis/token_filters/camelcase_filter/parser.go
+++ b/analysis/token_filters/camelcase_filter/parser.go
@@ -1,0 +1,93 @@
+package camelcase_filter
+
+import (
+	"unicode/utf8"
+
+	"github.com/blevesearch/bleve/analysis"
+)
+
+func buildTokenFromTerm(buffer []rune) *analysis.Token {
+	return &analysis.Token{
+		Term: buildTermFromRunes(buffer),
+	}
+}
+
+// TODO: Lifted from ngram_filter. Expose as public and re-use?
+func buildTermFromRunes(runes []rune) []byte {
+	rv := make([]byte, 0, len(runes)*4)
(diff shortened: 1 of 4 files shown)


analysis/token_filters/camelcase_filter/parser.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/blevesearch/bleve/analysis"
+)
+
+func buildTokenFromTerm(buffer []rune) *analysis.Token {
+	return &analysis.Token{
+		Term: buildTermFromRunes(buffer),
+	}
+}
+
+// TODO: Lifted from ngram_filter. Expose as public and re-use?
```
mschoch: This has been made public here:

https://github.com/blevesearch/bleve/blob/master/analysis/util.go#L37

mschoch:
Just had that one comment, will merge now anyway, we can clean that up later.

a-little-srdjan:
Bugger, always look at util :) I will make another pull request to clean it up. Thanks for the pointer

mschoch:
Well, since you copied from ngram_filter, it sounds like we're not even using it throughout the codebase.  We should probably grep through and find similar cases.

a-little-srdjan:
I've gone through the token filters. I found that the following filters can be refactored:
EdgeNgram
Ngram
CamelCase :)
Truncate (this one can benefit directly from util.TruncateRunes)

I can go ahead and start refactoring these filters?

mschoch:
Yep, that would be awesome.

a-little-srdjan:
Submitted as the pull request #386

PR #388, opened 2016-06-14
Add bucket fillPercent option for boltdb

mmindenhall:
Hi again,

I couldn't resist experimenting with this setting when I saw it in the boltdb `Bucket` documentation:

```go
    // Sets the threshold for filling nodes when they split. By default,
    // the bucket will fill to 50% but it can be useful to increase this
    // amount if you know that your write workloads are mostly append-only.
    //
    // This is non-persisted across transactions so it must be set in every Tx.
    FillPercent float64
```

Documents in our IoT "device reports" index are immutable for as long as they're retained.  I found that page utilization on this index is very much improved with higher `FillPercent` values.  For example, using a value of `0.8` (instead of the default `0.5`), we're able to index 20-25% more reports within a given index file size.

This PR adds `fillPercent` as an option when using the boltdb store.

### DIFF
--- a/index/store/boltdb/store.go
+++ b/index/store/boltdb/store.go
@@ -33,11 +33,12 @@ const (
 )
 
 type Store struct {
-	path   string
-	bucket string
-	db     *bolt.DB
-	noSync bool
-	mo     store.MergeOperator
+	path        string
+	bucket      string
+	db          *bolt.DB
+	noSync      bool
+	fillPercent float64
+	mo          store.MergeOperator
 }
 
 func New(mo store.MergeOperator, config map[string]interface{}) (store.KVStore, error) {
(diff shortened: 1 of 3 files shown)


benbjohnson:
Heads up that `bolt.Bucket.FillPercent` can cause a headache if you are not doing append-only writes into the bucket. If you're randomly inserting data then you can end up with lots of mostly empty pages.

mschoch:
Thanks, yes I am aware of this setting, and have been using it successfully on an experimental project which builds an LSM on top of bolt – https://github.com/couchbaselabs/cellar

I think it does make sense to expose it as an option, but as @benbjohnson  notes, it should only be used by people that know what they're doing.

@mmindenhall I just took a quick look, the one concern I have is the behavior when the option isn't set.  It looks like it would use the zero value, which would not only be a bad default, I don't think its valid.  We should be sure that if the config option is not specified, bolt continues to use it's own default value.

mmindenhall:
@benbjohnson, thanks for the heads up!  I did test with different values ranging from 0.3 up to 1.0.  I chose 0.8 just to leave a bit of extra room for the random inserts.

@mschoch, within New (store.go:59), I'm setting the value to the bolt default (currently 0.5) if the option is not set.  If someone were to create Store directly instead of via New, that could be a problem, but in that case I don't think _anything_ would work since none of the fields are exposed.

mschoch:
@mmindenhall cool, yeah I missed that.  Adding the test is nice too.  Thanks.

PR #394, opened 2016-07-06
Added match query operator

jakubkulhan:
Match query operator, either `or` (default), or `and` controls whether analyzed terms construct disjunction, or conjunction query.


mschoch:
This looks great, thanks for the contribution, and thanks for signing the CLA.

PR #398, opened 2016-07-17
Make facets much faster

slavikm:
1. Eliminate field cache lookup for each document match times number of fields and terms
2. Make sure to return only relevant fields in the FieldTerms

On a sample of a few 1000s documents with back index rows ranging from 10s to 10000s single facet search returns in less than a second instead of almost 4 min of 100% CPU.

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -77,10 +77,9 @@ type IndexReader interface {
 
 	Document(id string) (*document.Document, error)
 	DocumentFieldTerms(id string) (FieldTerms, error)
-	DocumentFieldTermsForFields(id string, fields map[string]uint16) (FieldTerms, error)
+	DocumentFieldTermsForFields(id string, fields []string) (FieldTerms, error)
 
 	Fields() ([]string, error)
-	FieldIDs(fields []string) ([]uint16, error)
 
 	GetInternal(key []byte) ([]byte, error)
 
--- a/index/index.go
+++ b/index/index.go
@@ -77,8 +77,10 @@ type IndexReader interface {
 
 	Document(id string) (*document.Document, error)
 	DocumentFieldTerms(id string) (FieldTerms, error)
+	DocumentFieldTermsForFields(id string, fieldIDs []uint16, fields []string) (FieldTerms, error)
 
 	Fields() ([]string, error)
+	FieldIDs(fields []string) ([]uint16, error)
 
 	GetInternal(key []byte) ([]byte, error)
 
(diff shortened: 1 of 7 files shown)


slavikm:
@mschoch any chance of taking a look? Blocking us on a feature depending on facets.

mschoch:
Sorry for the delay.  I don't think I'm going to be able to rush this one through.

Obviously we want to make facets faster, and some of this makes sense, but I do have some concerns.

1.  I don't really like the idea that the numeric field IDs become exposed with this change.  I was hoping that at the index tier, everything external was done with field names, and field ids were an internal implementation detail.  This is important since we're separately working on optimizing the indexing tier and these abstractions/boundaries are what give us flexibility to make changes internally.

2.  I'm not exactly sure why keeping the map of field IDs in the facet builder helps, as internal to the indexing tier we have the "field cache", which as its name implies is supposed to be the fast/cached fieldName/fieldID mapping.

Separately, but related I've realized there are other features (sorting results on something other than score) than can take advantage of loading these field terms.  Obviously, if you create a facet on a field and later want to sort the hits on that same field, you shouldn't have to look it up twice.  I was thinking that perhaps these already looked up values should be put into DocumentMatch structure.  This isn't a fully-baked solution yet, but that was my thinking.

slavikm:
@mschoch Understood regarding field IDs BUT going to the field cache for every match for potentially 100,000s of matches is one of the reasons this is pretty slow.

On the other hand, just loading the relevant fields and resolving them outside of the internal loop over termEntries is going to increase performance by a lot.

Would that make sense? I can change it to be a much smaller change - remove the exposed field IDs but keep the `DocumentFieldTermsForFields`.

mschoch:
Yeah I understand what you're saying, I just don't understand why from reading the code.  The cache was itself put in to speed this up.  This feels like we're putting a cache in front of the cache, but they're both hash maps.  I don't doubt what you're saying, I just don't understand it.

I'll take a closer look, but I'm offline most of tomorrow so it may not be right away.

slavikm:
One resolve happens just once (when building the facets). The other resolve happens for each match. Field cache is doing a full loop on all fields because the map is name->id and you need id->name. On the other hand, I did a bit of performance testing and definitely the biggest bang is removing the resolve from internal loop as I said above.
`fieldName := i.index.fieldCache.FieldIndexed(uint16(*entry.Field))`

slavikm:
@mschoch made a quick change to just make sure we are not doing the loop on all fields inside the loop of termEntries. This way, the field IDs are not exposed but there are still major performance gains.

slavikm:
@mschoch just to add a bit of color in our use case of a single facet on about 6000 fairly complex documents we moved from 190 sec 100% CPU to 1 sec, with the change as it is now (without exposing field IDs)

mschoch:
OK, this looks good now.

PR #402, opened 2016-07-31
Index/Search API work

mschoch:
hey @steveyen this is now ready for review

i hate it, but the tests pass, and it seems to accomplish the original goals

to recap, the original goals were
- make index implementation's id representation opaque (flexibility for future indexing schemes)
- avoid converting from internal representation to external representation for hits that don't matter

why do i hate it?

- new docid reader mode to only return ones also in a list of ids, i'm ok with this existing, but i suspect it may still be buggy, more edge cases need testing, right now relying on higher-level tests in DocIDSearcher

- searchers package relies on "upside_down" package for testing.  these commits didn't introduce the problem, but as we move towards new implementations, this dependency is more glaringly bad

- having the "collector" do the final conversion required it to have the IndexReader, which i didn't anticipate, this feels wrong...  maybe ways to clean/tighten this up

- now i'm not sure if having separate DocumentMatch and DocumentMatchInternal really makes sense.  i wanted it to prevent accidentally depending on the ID field as string somewhere, and it does accomplish that.  but... i dunno maybe one struct would have been fine and meant less changes overall.

- while one of the original goals was perf oriented (avoid unnecessary copying), it turns out that the other goal is performance penalty (indirection around doc id implementation is potentially all over hot path of query)  i really hope it's not as bad as i think, because this gives us lots of flexibility, but i suspect its worse than i think, and a deal breaker.

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -122,9 +122,13 @@ type TermFieldDoc struct {
 	Vectors []*TermFieldVector
 }
 
+// Reset allows an already allocated TermFieldDoc to be reused
 func (tfd *TermFieldDoc) Reset() *TermFieldDoc {
+	// remember the []byte used for the ID
 	id := tfd.ID
+	// idiom to copy over from empty TermFieldDoc (0 allocations)
 	*tfd = TermFieldDoc{}
+	// reuse the []byte already allocated (and reset len to 0)
 	tfd.ID = id[:0]
 	return tfd
 }
--- a/index/index.go
+++ b/index/index.go
@@ -60,7 +60,7 @@ type AsyncIndex interface {
 }
 
 type IndexReader interface {
-	TermFieldReader(term []byte, field string) (TermFieldReader, error)
+	TermFieldReader(term []byte, field string, includeFreq, includeNorm, includeTermVectors bool) (TermFieldReader, error)
 
 	// DocIDReader returns an iterator over documents which identifiers are
 	// greater than or equal to start and smaller than end. Set start to the
--- a/index/index.go
+++ b/index/index.go
@@ -123,7 +123,9 @@ type TermFieldDoc struct {
 }
 
 func (tfd *TermFieldDoc) Reset() *TermFieldDoc {
+	id := tfd.ID
 	*tfd = TermFieldDoc{}
+	tfd.ID = id[:0]
 	return tfd
 }
 
--- a/index/index.go
+++ b/index/index.go
@@ -69,6 +69,8 @@ type IndexReader interface {
 	// The caller must close returned instance to release associated resources.
 	DocIDReader(start, end string) (DocIDReader, error)
 
+	DocIDReaderOnly(ids []string) (DocIDReader, error)
+
 	FieldDict(field string) (FieldDict, error)
 
 	// FieldDictRange is currently defined to include the start and end terms
@@ -76,14 +78,16 @@ type IndexReader interface {
 	FieldDictPrefix(field string, termPrefix []byte) (FieldDict, error)
 
 	Document(id string) (*document.Document, error)
-	DocumentFieldTerms(id string) (FieldTerms, error)
-	DocumentFieldTermsForFields(id string, fields []string) (FieldTerms, error)
+	DocumentFieldTerms(id IndexInternalID) (FieldTerms, error)
+	DocumentFieldTermsForFields(id IndexInternalID, fields []string) (FieldTerms, error)
 
 	Fields() ([]string, error)
 
 	GetInternal(key []byte) ([]byte, error)
 
 	DocCount() uint64
+
+	FinalizeDocID(id IndexInternalID) (string, error)
 
 	Close() error
 }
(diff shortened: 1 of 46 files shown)


mschoch:
i tried beer search while still on this branch this morning and was getting some panics, i'll switch back and track it down.  definitely points to some missing unit tests.

mschoch:
Pushed 2 more commits that move this closer to a direction I think that might work.  I addresses some of the things I hate above.

@steveyen see below for important realization/observation on the attempted perf gain

Separately, I went back to why I observed the beer-search application failing.  It turns out it was really important.  The bleve store iterator API says that the current observed key/value (each []byte) can only relied on in the moment of the iteration.  If you move to another key with Next/Seek or called Close, you had to copy the previous value if you still wanted it.

It turns out that previously, our converting of the doc ID ([]byte) to string, was making the essential copy.  This set of changes, which attempts to delay the copy, now exposes the problem that we're not making the copy required by the API.

Simple, so now in the TermFieldReader, we have to make a copy of the docID before returning it....

BUT, now this whole changeset might again be a net negative for performance, because then later when we ultimately do convert it to a string, that will be yet another copy.

Why didn't we see this in the original unit tests?  Well they run with bolt, not goleveldb.  Each KV store ultimately offers different guarantees here.  I think it could have failed with bolt to, it just didn't happen to in the unit tests, whereas with goleveldb it fails right away all the time.

I have another commit which adds the copies back in, so that this PR is at least functionally correct.

mschoch:
Well, delay on that next commit, since somehow it appears that highlighting is broken as well.

mschoch:
Found the issue, actually introduced in the earlier set of patches from @steveyen.  When we reuse the TermFrequencyRow, we call parseK() and parseV(), logically these need to "reset" every time, but there was some lazy initialization code which meant term vectors were not reset, just kept getting appended to each time.

mschoch:
So, let me see if I can recap again:

The original goals were:

- make index implementation's id representation opaque (flexibility for future indexing schemes)
- avoid converting from internal representation to external representation for hits that don't matter

But, we now know that the conversion from []byte to string (deemed expensive and unnecessary) was actually serving an important function.  It created a safe copy of the docID that could be used outside of the life of the iteration step that produced it.

I've updated this PR to make a copy of the []byte, for safe usage after iteration, but then we end up making *another* copy at the very end for results in the final result set.  I suspect this won't make things worse, but means the prospective gain was never there in the first place.

All that being said, we did achieve a more opaque internal document ID for the indexing tier, which I see as win.

Also I merged DocumentMatchInternal and DocumentMatch back together again, removing another one of my concerns.

The change to Collector API is still irksome, but not going to block this because of that.

Finally, the fact that the testing of the searcher package is still tightly coupled to the upside_down index scheme remains, but this is probably best cleaned up in a separate PR.  The problem is two fold:

1.  We want to be able to test the searcher package against any new indexing scheme as well.
2.  We artificially create IndexInternalIDs for testing like index.IndexInternalID("myid") that is making assumptions about the underlying implementation, and is not generic enough.  Probably we should just have 2-way lookup methods on the IndexReader.  Something like:

ExternalID(index.IndexInternalID) string
InternalID(string) index.InternalID

The ExternalID method exists today with the name FinalizeID (probably not a good name).
The InternalID method would really only be needed for testing, but probably not harmful either.

In summary, @steveyen I think we should review this, clean up any remaining concerns, and test that its at least no worse than before performance-wise.  If so, the flexibility it offers for new indexing schemes is desirable.

mschoch:
And I just realized some unit tests are still failing :(

steveyen:
Structurally, looks good to me.  Still reading through it, and don't see anything wrong yet.  And, very much agree with the aims.

I did run it through the profiler, and the "new" chunk of garbage creation that shows up more prominently, unsurprisingly, is w.r.t. making copies of the ID during term iteration...

https://github.com/mschoch/bleve/blob/indexapiwork/index/upside_down/reader.go#L86

I have some ideas on that, along the lines that perhaps we could arrange it so that the preAlloc'ed index.TermFieldDoc parameter might have an ID slice that has preallocated, ready-to-use capacity.

But, I wouldn't hold up getting this in based on that, so you can reach bleve 1.0 API milestone sooner.  I like the improved internal API, and we can just do another / future round of more perf-related stuff as more, future pull-requests.

steveyen:
btw, I did some commented out code near the DocIDSearcher / search_docid.go, so perhaps some of that is more in-flight / experimental?

mschoch:
pushed another commit to remove commented out code

mschoch:
that was the previous implementation, which was moved into the index reader itself, i just didn't remove it completely until the new version worked, gone now

mschoch:
also, regarding the increased garbage, im surprised its more than before when we copied it to a string (though it might look different in the profile)

also, i used this idiom:

    b = append([]T(nil), a...)

because its one line, and assuming append gets inlined... but i can imagine it still might be worse than:

    b = make([]T, len(a))
    copy(b, a)

when we know the final size.  But still seems like it would only matter if the length is large.

steveyen:
Cool - yeah, with removed dead code, looks good!

I also treat append([]T(nil), a...) as a shorthand synonym in my mind for make()/copy().

mschoch:
hi @steveyen i went ahead and pushed the commit anyway, even though i haven't yet checked the perf.  i figured there is enough to review already, and we can start bike shedding over method names and api stuff now, while we check perf in parallel.

mschoch:
Hi @steveyen 

I have done some benchmarking locally, and this PR is faster than master for me, but the numbers are not comparable to yours.  Would it be possible for you to run it on your machine (as well as reviewing the contents of the most recent commit?

Thanks

steveyen:
looks good to me - much cleaner than the previous "pre-allocated" approach

steveyen:
also, local perf runs on my macbook show a ~15% improvement in perf (200K wiki docs, high-freq term search via bleve-query)

and profiling looks cleaner... no memory alloc's in hot path: https://gist.github.com/steveyen/c53e53ed69ffb92bda6991eac374ceb9

PR #407, opened 2016-08-10
option to sort by other fields

### LINKED ISSUE #110: option to sort by other fields
Sometimes its useful for applications to want the search results sorted by another field, often a date or numeric value.

dtylman:
closes #110

option to sort by other fields

Usage:
```go
req := NewSearchRequest(NewMatchQuery("one")
req.SortBy("Day", true)
req.SortBy("Name", true)
sr, err := index.Search(req)
.
.
.
```

Will match "one", and will sort results by: Day, Name and Score.


mschoch:
Thanks for this, unfortunately a big API change like this won't be able to be merged right away.  But it is a great start thanks.  Here are some of the things we need to work through:

1.  Need you to sign the CLA  http://www.blevesearch.com/docs/Contributing/
2.  There appears to be one unchecked error reported by the errcheck utility:
`search/collectors/collector_heap.go:66:19:	hc.collectSingle(next)`
3.  The indexapi work is going to be rebased today, and that has some changes to how the collector interacts with the searcher (not your problem, I will handle rebasing this part).
4.  I see your tests use two new external packages and this requires some review.  First, I need to understand what they do and what the license terms are.  Second, I'd really rather not introduce new dependencies at point, even in the test packages.  But, let me look closer before we decide.
5.  I think we may need to tweak the top-level API where users specify their sort.  When I was looking into this I saw that ES also let's your sort by things other than just field values.  In particular it lets you describe what to do when fields have multiple values etc.   Finally, I'd also like the API to let you specify sort by document ID and by score, and integrate that naturally into the API.  So, there may be some changes here as well.

But, please don't misinterpret the fact that there will be some changes, with the fact that this is a great contribution.  Thanks.

dtylman:
First step I'll handle 1,2 & 4 (don't really need the imports, just creates mock data for testing)

mschoch:
Great, yeah if you can take care of 1, 2 & 4, I'll take care of 3 and I'm already reviewing ideas for 5.  I think we can get this cleaned up and merged pretty quickly.  Lots of people want it, so we're motivated to finally get it done.

dtylman:
AFAIK, 1,2 & 4 done :)

Tell me if you have any further issues to fix on this.

mschoch:
Thanks, I'm working on the other issues now.

One additional problem with this approach is that it requires loading stored fields for all the search hits.  I think in larger indexes this is problematic.  It will be more efficient to sort on indexed fields, we still have to load them for head hit, but by combining this with a searcher at the top-level we can at least read out the values in sequence.

But, I want to get this merged before I write that code, so I'm going to code it such that sorting on stored fields is explicit, and then we can add support for sorting on indexed fields later.

mschoch:
Another tweak, the intention of MaxScore in the SearchResults is be the maximum score across the whole result set.  I think this PR made it the the highest score on the heap (size+skip).  So I'll be changing that back.

mschoch:
I'm going to close this PR, these commits, plus some add-ons are now being worked on and reviewed here: #410

PR #410, opened 2016-08-12
Sort by field try2

mschoch:
This, includes commits from #407 and obsoletes it.

It offers the same capabilities, a slightly different API, and some code reorganization.  It is compatible with master, and the tests pass on my machine.  However, it is still not quite ready for merging.  Some issues:

1.  The SearchRequest structure has a custom Unmarshal method that does not yet know anything about the Sort field.
2.  I introduced ability to sort on doc id, as well as score and stored field.  Need another test there.
3.  Need to add Example of custom sort.
4.  Maybe can add custom sort to the integration test suite to get a little more confidence in the new capabilities.


mschoch:
Ready for some initial review from @dtylman and @steveyen 

One API aspect I didn't like, I wanted it to be possible to do custom sort without requiring additional import, but only straightforward to do that AND have the collector be able to test with the same sort behaviors would be to duplicate/copy code.

So, for now, one downside of this API is that you have import the search package to do custom sort.

mschoch:
Alright, nevermind on the review, for some reason tests are failing (though not here)

mschoch:
Ah, the integration tests read query from JSON, which with the current Unmarshal method means they end up not having a sort order (which defaults to only the index natural order).  Let me address that now.

mschoch:
I pushed a fix to just restore the default sort order we had before when parsing from JSON.  Not a full fix to support parsing the sort parameter yet.

Also, on the todo list is to compare the benchmarks with the heap vs list approach.  The last time someone tried the heap it was slower (but we never looked to see if the benchmark was flawed or something else was going on)

mschoch:
Added JSON serialization/deserialization capabilities and played with it using the beer-search example app.  It seems to work pretty well, but needs more testing/docs.

document/field.go
```diff
@@ -32,3 +32,86 @@ type Field interface {
...
+
+// CompareFieldValues provides ordering amongst stored field values
+// when trying compare field values of different types,
+// we impose the following order:
+// - nil (missing field)
+// - boolean
+// - number
+// - text
+// - date
+func CompareFieldValues(i, j Field, descending bool) int {
```
dtylman: I really like this 🎉

search/collectors/collector_heap.go
```diff
@@ -0,0 +1,241 @@
...
+		d.Document, err = reader.Document(d.ID)
+		if err != nil {
+			return err
+		}
+	}
+
+	// optimization, we track lowest sorting hit already removed from heap
+	// with this one comparision, we can avoid all heap operations if
+	// this hit would have been added and then immediately removed
+	if hc.lowestMatchOutsideResults != nil {
```
dtylman: I removed the lower bound optimisation that was here before...

Great that it's back!

dtylman:
👍

search/collectors/collector_heap.go
```diff
@@ -0,0 +1,241 @@
...
+	// see if we need to load the stored fields
+	if len(hc.sort.RequiredStoredFields()) > 0 {
+		if d.ID == "" {
+			// look up the id since we need it for lookup
+			d.ID, err = reader.FinalizeDocID(d.IndexInternalID)
+			if err != nil {
+				return err
+			}
+		}
+		d.Document, err = reader.Document(d.ID)
```
slavikm: Do we need to load the entire document? Can't we just read the relevant fields using `DocumentFieldTermsForFields`? Wouldn't it be faster? Not sure it's going to be as clean, though...

mschoch: No, we don't.  The original PR worked by using stored fields.  DocumentFieldTermsForFields works on the indexed values.  And I do think that is more correct.  But I wanted to get the PR functional first.  So yes, probably tomorrow sometime I'll introduce a version that uses DocumentFieldTermsForFields.

mschoch:
Just a quick update, I'm still working on this.  I have it sorting by indexed fields now.  There were some minor complications, but I think its a much cleaner approach.  I hope to push these improvements by tomorrow.

mschoch:
OK, I've pushed a few commits with the additional changes I've been working on.  The biggest change is that now when you sort on a field, it uses the indexed value, not the stored value.  I think this is better for several reasons, but it isn't perfect either, so I'd like to outline some of the pros/cons.

Pros:
-  Generally, our philosophy is that we don't "do anything" with the stored fields.  We just store them for you, and return them to you later.  This makes it easier for users to understand, do I need to index a field or store it?  The answer is clear, both faceting and sorting require the field to be indexed.  (but what about highlighting? doesn't that violate this rule?  depending on your point of view yes, but I would argue that the highlighting is done only after the search is finished, none of the core search capabilities require a field to be stored)  In summary, this is an API clarity/simplicity rule.
- Sorting on indexed fields can be faster.  At the moment it's probably not much different, but the fact that the field is indexed means more things are possible in the future.  With stored fields, it's a dead-end, we aren't supposed to be doing anything extra with them.
- The requirement that a field be indexed is already a common one in Lucene/ES world.
- Facets/Sorting share a cache of field values looked up already, so sorting on a field that you also built a facet on can save additional lookups.

Cons: 
- Stored fields retain a bit more "type" information, so we can do things like sort numbers before strings, etc.  Indexed fields lose this as everything in the index is just a term.  This means sorting heterogeneous values can lose some capability.  But, this seems like a highly unusual case anyway.

I suspect there still may be people who disagree, and I welcome them to share their input.  But at the moment I think this is a better way forward.

index/upside_down/index_reader.go
```diff
@@ -125,10 +105,9 @@ func (i *IndexReader) DocumentFieldTermsForFields(id index.IndexInternalID, fiel
 	fieldsMap := make(map[uint16]string, len(fields))
 	for _, f := range fields {
 		id, ok := i.index.fieldCache.FieldNamed(f, false)
-		if !ok {
-			return nil, fmt.Errorf("Field %s was not found in cache", f)
```
slavikm: Yeah, I added the error and eventually regretted it. It should just add fields that exist as you do.

mschoch:
I added some examples and integration tests on my flight yesterday.  So, we're getting close to finishing this up.  But I found one problem.

As currently implemented, it doesn't work for MultiSearch (across multiple indexes).  When we merge the results, we incorrectly always sort by score.  This is made worse by the fact that we no longer have access to the underlying documents, so they can't be merged into the correct order.  What Elasticsearch does is return the actual sort values in the response, that way when aggregating results from multiple shards, you can still apply the right sort.  But, that does mean a little bit more work....

slavikm:
From my perspective, this is fine adding piece by piece. We personally have our own way of doing cross-index searches and for us the solution as is will work fine.

mschoch:
So, this commit is now functionally where I'd like it and the API is OK.  But, we have performance problems.  The collector benchmarks, which while not perfect still represent something, show that we're about 4x slower and do considerably more memory allocations, even when not doing any new sorting, just return results ordered by score descending.

So, now I'll shift gears to understanding this.

mschoch:
Also, in case someone else is inclined to run the benchmarks they actually fail because the reuse the collector instance.  I'm pretty sure that was never correct, so I'll be pushing a commit to master to change that, and rebase that into this PR for fair comparison.

mschoch:
This shows the situation:

```
$ go test -run=xxx -bench=. -benchmem
BenchmarkTop10of100000Scores-4     	    3000	    453269 ns/op	  360161 B/op	      42 allocs/op
BenchmarkTop100of100000Scores-4    	    2000	    519131 ns/op	  388275 B/op	     219 allocs/op
BenchmarkTop10of1000000Scores-4    	     200	   7459004 ns/op	 4628236 B/op	      52 allocs/op
BenchmarkTop100of1000000Scores-4   	     200	   8064864 ns/op	 4656596 B/op	     232 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/collectors	7.385s

$ go test -run=xxx -bench=. -benchmem
BenchmarkTop10of100000Scores-4     	    1000	   2079277 ns/op	  841793 B/op	   30028 allocs/op
BenchmarkTop100of100000Scores-4    	     500	   2262132 ns/op	  870976 B/op	   30028 allocs/op
BenchmarkTop10of1000000Scores-4    	      50	  22456473 ns/op	 9429828 B/op	  300038 allocs/op
BenchmarkTop100of1000000Scores-4   	      50	  22349645 ns/op	 9459018 B/op	  300040 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/collectors	6.143s
```

The numbers in the benchmarks are mislabeled, the first 2 are 10k and the second 2 are 100k.  This corresponds to 3 new allocations for each matching document.

mschoch:
Playing with a slightly different implementation locally I got it to be:

```
$ go test -run=xxx -bench=. -benchmem
BenchmarkTop10of100000Scores-4     	    1000	   1362301 ns/op	  442323 B/op	   10038 allocs/op
BenchmarkTop100of100000Scores-4    	    1000	   1479354 ns/op	  474451 B/op	   10039 allocs/op
BenchmarkTop10of1000000Scores-4    	     100	  14365423 ns/op	 5446281 B/op	  101038 allocs/op
BenchmarkTop100of1000000Scores-4   	     100	  14719631 ns/op	 5479128 B/op	  101046 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/collectors	7.779s
```

In this version numbers/dates are left as numbers in the returned sort values.  This lets the default case of sorting by score avoid some allocations.  But, it means we need to use reflection in some other cases.  Not totally sold on it yet, but inching closer to acceptable performance.

mschoch:
I went ahead and pushed those changes as they do perform considerably better this this case:

```
$ go test -run=xxx -bench=. -benchmem -cpuprofile=cpu.out
BenchmarkTop10of100000Scores-4     	    1000	   1162698 ns/op	  442361 B/op	   10036 allocs/op
BenchmarkTop100of100000Scores-4    	    1000	   1251414 ns/op	  476459 B/op	   10124 allocs/op
BenchmarkTop10of1000000Scores-4    	     100	  12870209 ns/op	 5430481 B/op	  100046 allocs/op
BenchmarkTop100of1000000Scores-4   	     100	  13655071 ns/op	 5465295 B/op	  100141 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/collectors	6.754s
```

Could use some review from @steveyen at this point.  I think the remaining allocation is caused by the conversion of a value to an interface{}.

mschoch:
OK, I just pushed the perf fixes for this.  We have gone back to using []string for the sort values, but we now return "_score" in place of the actual score.  This avoids expensive allocations (either from converting the numeric score into a term, or from converting all the sort terms into interface{})

The updated numbers are:

```
$ go test -run=xxx -bench=. -benchmem -cpuprofile=cpu.out
BenchmarkTop10of100000Scores-4     	    3000	    465809 ns/op	    2548 B/op	      33 allocs/op
BenchmarkTop100of100000Scores-4    	    2000	    626488 ns/op	   21484 B/op	     213 allocs/op
BenchmarkTop10of1000000Scores-4    	     300	   5107658 ns/op	    2560 B/op	      33 allocs/op
BenchmarkTop100of1000000Scores-4   	     300	   5275403 ns/op	   21624 B/op	     213 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/collectors	7.188s
```

Perf on the smaller case is very close, and we scale better on the larger cases.
We also have fewer allocations and alloc'd bytes, though some of this is also related to benchmark cleanup.

If @steveyen can review on the bleve-bench side, I think we're close to merging this.

mschoch:
On my bleve-bench setup with an index of 100k wiki docs, master gets:

41.263670 query/sec

On this PR, I get:

36.709359 query/sec

Pretty close, but I have one more thing to try.

mschoch:
Pushed small fix to avoid dynamic dispatch on the hot loop, gets us up to:

37.726271

slavikm:
When I previously tried playing with heap instead of the previous array method, I thought that heap will be faster but it turned out to be a bit slower than the arrays...

mschoch:
As it currently looks, it appears that this current implementation with a heap would be faster, were it not for the additional cost of search.SearchSort being an interface.  We avoid the dynamic dispatch on calls where the value does not change, but when using it to actually find the sort value the dynamic call is required.

We could eliminate the interface, but that is of course key to making it possible to introduce new things to sort on in the future.  So, at the moment I'm a bit reluctant to give that up, and see this as simply the cost of the new feature.

Also, the benchmarks seem to show that the heap performs much better in cases where we see a lot more hits.  Master has 8M ops/sec and this PR is now only 5M ops/sec.  If these numbers are valid, thats significant to me.

mschoch:
Pushed one more commit which further optimizes the default case where we're only sorting on score.  In this case it completely avoids dynamic dispatch.  This lessens the blow a bit for those users not interesting in more advanced sorting.

```
$ go test -run=xxx -bench=. -benchmem -cpuprofile=cpu.out
BenchmarkTop10of100000Scores-4     	    5000	    354262 ns/op	    2600 B/op	      34 allocs/op
BenchmarkTop100of100000Scores-4    	    3000	    505730 ns/op	   21536 B/op	     214 allocs/op
BenchmarkTop10of1000000Scores-4    	     300	   4110797 ns/op	    2616 B/op	      34 allocs/op
BenchmarkTop100of1000000Scores-4   	     300	   4292229 ns/op	   21688 B/op	     214 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/collectors	6.972s
```

And in the bleve-query tests I now get 39 query/sec.

mschoch:
So, even with all the latest changes, we were still a little bit slower than before.  So, I wanted to go back to try the list again, but I didn't want to throw away the heap.

Instead, what I have done is separate the data structure from the rest of the logic of the collector.  However, I did not design it to use an interface since this is a very hot loop during searching.  Rather, we have 2 structs (one for heap, one for list) and developers can change 2 lines to toggle between them (or experiment with other data structures).

The list is faster in the benchmark, but makes more allocations.

In the bleve-query tests I run, they seem to perform about the same (but still not quite as fast as master).

mschoch:
Here is the benchmark, using the list impl:

```
$ go test -run=xxx -bench=. -benchmem -cpuprofile=cpu.out
BenchmarkTop10of100000Scores-4     	    5000	    317485 ns/op	    6888 B/op	     126 allocs/op
BenchmarkTop100of100000Scores-4    	    2000	    666260 ns/op	   50561 B/op	     838 allocs/op
BenchmarkTop10of1000000Scores-4    	     500	   3589816 ns/op	    7520 B/op	     139 allocs/op
BenchmarkTop100of1000000Scores-4   	     300	   4136913 ns/op	   59816 B/op	    1028 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/collectors	8.047s
```

search/collectors/slice.go
```diff
@@ -0,0 +1,63 @@
...
+func (c *collectStoreSlice) Add(doc *search.DocumentMatch) {
+	// find where to insert, starting at end (lowest)
+	i := len(c.slice)
+	for ; i > 0; i-- {
+		cmp := c.compare(doc, c.slice[i-1])
+		if cmp >= 0 {
+			break
+		}
+	}
+	if i < 0 {
```
steveyen: not really important, but i'm thinking the iterator i value never goes below 0?

mschoch: yep good catch, the logic changed and that is no longer needed

search/collectors/topn_test.go
```diff
@@ -398,21 +399,28 @@ func TestPaginationSameScores(t *testing.T) {
 			t.Errorf("doc ID %s is in top 5 and next 5 result sets", hit.ID)
 		}
 	}
-
 }
 
 func BenchmarkTop10of100000Scores(b *testing.B) {
-	benchHelper(10000, NewTopScorerCollector(10), b)
+	benchHelper(10000, func() search.Collector {
```
steveyen: the 10000 here doesn't seem to match the func name of "BenchmarkTop10of100000Scores"?

search/collectors/topn_test.go
```diff
@@ -398,21 +399,28 @@ func TestPaginationSameScores(t *testing.T) {
...
+		return NewTopNCollector(10, 0, search.SortOrder{&search.SortScore{Desc: true}})
+	}, b)
 }
 
 func BenchmarkTop100of100000Scores(b *testing.B) {
-	benchHelper(10000, NewTopScorerCollector(100), b)
+	benchHelper(10000, func() search.Collector {
```
steveyen: Same here, and with the other benchmark funcs here?

search/collectors/topn_test.go
mschoch: The count mismatches were noted earlier in the thread.  Initially I didn't want to change them here because it was going to complicate the merge.  Since then I've already complicated the merge.  So yes I will correct this either during the merge, or in a commit after the merge.

steveyen:
A wide ranging change that looks good to me.  Only saw a couple minor "bike shed color" things, but can fix them later, imho.

An exciting feature.  Plus, I learned about json.RawMessage from this!

PR #414, opened 2016-08-25
Add http endpoint for batch import.

### LINKED ISSUE #171: add http handler for batch

daschl:
This is a work in progress of adding batch import support as a HTTP endpoint based on the current functionality.

The proposed format is line delimited first a meta json with a defined format and then optionally a document block (depending on the type).

Proposed meta:

``` json
{"index": "<indexname>", "id": "<docid>", "action": "<index or delete>"}
```

So for example:

A file with

``` json
{"index": "foo", "id": "airline_10", "action": "index"}
{"type": "airline","name": "40-Mile Air","iata": "Q5","icao": "MLA","callsign": "MILE-AIR","country": "United States"}
{"index": "foo", "id": "enoent", "action": "delete"}
{"index": "foo", "id": "airline_10123", "action": "index"}
{"type": "airline","name": "Texas Wings","iata": "TQ","icao": "TXW","callsign": "TXW","country": "United States" }
{"index": "foo", "id": "airline_10226", "action": "index"}
{"type": "airline","name": "Texas Wings","iata": "TQ","icao": "TXW","callsign": "TXW","country": "United States" }
{"index": "foo", "id": "airline_10226", "action": "delete"}
```

will end up with 2 docs stored, the first delete will be ignored (enoent) and the second delete will remove one of the docs. 

I tested with a simple server

``` go
package main

import(
    "github.com/gorilla/mux"
    bleveHttp "github.com/blevesearch/bleve/http"
    "net/http"
    "log"
    "flag"
    "fmt"
)

var bindAddr = flag.String("addr", ":8095", "http listen address")

func main() {
    router := mux.NewRouter()

    index, err := bleve.New("foo.bleve", mapping)

    fmt.Println("ERROR!!", err)

    bleveHttp.RegisterIndexName("foo", index)

    bulkIndexHandler := bleveHttp.NewBatchIndexHandler()
    router.Handle("/_bulk", bulkIndexHandler).Methods("POST")

    http.Handle("/", router)
    log.Printf("Listening on %v", *bindAddr)
    log.Fatal(http.ListenAndServe(*bindAddr, nil))
}
```

and storing the file shown above with `curl -vv -XPOST http://localhost:8095/_bulk --data-binary "@requests"`.

`bleve_dump` shows me I think it did the right thing. 

So:
- functionality/semantics wise is there anything I missed?
- there is lots of duplication in there, and any goisms I didn't follow please let me know and I'll clean it up.
- where is the best place to add tests?
- any other things I missed?

Once this one is sane I'll make an independent PR to the bleve-explorer to take advantage of it.

fixes #171

### DIFF
--- a/http/batch_index.go
+++ b/http/batch_index.go
@@ -1,0 +1,114 @@
+//  Copyright (c) 2016 Couchbase, Inc.
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+package http
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/blevesearch/bleve"
+	"io"
+	"net/http"
+)
+
+type MetaData struct {
+	Action string `json:"action"`
+	Index  string `json:"index"`
+	Id     string `json:"id"`
+}
+
+type BatchIndexHandler struct {
+}
+
+func NewBatchIndexHandler() *BatchIndexHandler {
+	return &BatchIndexHandler{}
+}
+
+func (h *BatchIndexHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	openBatches := map[string]*bleve.Batch{}
+
+	dec := json.NewDecoder(req.Body)
+	isMeta := true
+	var currentMeta MetaData
+
+	for {
+		if isMeta {
+			var meta MetaData
+			if err := dec.Decode(&meta); err == io.EOF {
+				break
+			} else if err != nil {
+				showError(w, req, fmt.Sprintf("error parsing meta JSON: %v", err), 400)
+				return
+			}
+			currentMeta = meta
+
+			batch, present := openBatches[currentMeta.Index]
+
+			if !present {
+				index := IndexByName(currentMeta.Index)
+				if index == nil {
+					showError(w, req, fmt.Sprintf("no such  index '%s'", currentMeta.Index), 404)
+					return
+				}
+				batch = index.NewBatch()
+				openBatches[currentMeta.Index] = batch
+			}
+
+			switch currentMeta.Action {
+			case "index":
+				isMeta = false
+			case "delete":
+				batch.Delete(currentMeta.Id)
+			default:
+				showError(w, req, fmt.Sprintf("unknown action: %v", currentMeta.Action), 400)
+				return
+			}
+
+		} else {
+			var doc interface{}
+			if err := dec.Decode(&doc); err == io.EOF {
+				break
+			} else if err != nil {
+				showError(w, req, fmt.Sprintf("error parsing document JSON: %v", err), 400)
+				return
+			}
+
+			batch, _ := openBatches[currentMeta.Index]
+
+			switch currentMeta.Action {
+			case "index":
+				err := batch.Index(currentMeta.Id, doc)
+				if err != nil {
+					showError(w, req, fmt.Sprintf("error indexing document in batch '%s': %v", currentMeta.Id, err), 500)
+					return
+				}
+			default:
+				showError(w, req, fmt.Sprintf("unknown action: %v", currentMeta.Action), 400)
+				return
+			}
+			isMeta = true
+		}
+
+	}
+
+	for idx, batch := range openBatches {
+		index := IndexByName(idx)
+		err := index.Batch(batch)
+		if err != nil {
+			showError(w, req, fmt.Sprintf("error while storing batch: %v", err), 400)
+			return
+		}
+	}
+
+	rv := struct {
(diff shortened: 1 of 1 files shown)


mschoch:
Hi Michael, I haven't looked at the code directly yet, but our continuous integration runs several tools.  One is called errcheck, which checks for errors returned that are being ignored.  It found:

$ errcheck $(go list ./... | grep -v vendor/)
http/batch_index.go:85:16:	batch.Index(currentMeta.Id, doc)

mschoch:
One high-level concern I have is the way batch execution errors are reported.  Consider the case where I have changes affecting 2 or more indexes.  If while executing the 2nd batch I get an error, I don't know that the first one was already applied successfully.  Further, the user has no control over the order, so we don't even know if it attempted any of the others or not.

It seems to me like once we get to this execution phase, the response should be JSON, indication a per-index success/fail.  This also raises the question of whether or not we should continue to try the other batches when one fails.

mschoch:
Here is some related info about how ES responds to bulk requests:

https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/action/bulk/BulkResponse.java#L31-L35

daschl:
So we could, as soon as we get to the batch indexing, return a JSON object that lists the indexes as key and a boolean of success as the value? and if at least one of them failed http 500, otherwise 200 (or 201?).  And yes if we go down the great length of separate reporting we should keep on going if an individual batch fails.

that brings up two more questions:

 - are there any errors returned from an index batch that would make us bail out completely (ala disk full or something?)
 - do we need to do some special handling when a batch index fails? or can we just move on to the next one

otoolep:
While this seems like useful functionality in general (I have built this type of thing myself on bleve) is this the right place for it?

bleve is like Lucene - a library to build other search systems with, not an application itself, right?

mschoch:
We've always offered an optional http handler sub-package.  I think these are general purpose enough, I use them on 2 different projects.  But no one is required to use them.  One could argue that they should be outside of the main bleve package, but I'm kind of tired of having too many repos as well.  This package is small, and doesn't have any external dependencies, so I don't see the harm in keeping it here.

With that in mind, a bulk indexing handler seems to fit well with the other handlers in this package.

otoolep:
Ah, ok, I must have missed that. Thanks for clearing it up. If it makes sense with what we've done before, then I withdraw my objection.

http/batch_index.go
```diff
@@ -0,0 +1,114 @@
...
+	isMeta := true
+	var currentMeta MetaData
+
+	for {
+		if isMeta {
+			var meta MetaData
+			if err := dec.Decode(&meta); err == io.EOF {
+				break
+			} else if err != nil {
+				showError(w, req, fmt.Sprintf("error parsing meta JSON: %v", err), 400)
```
mschoch: I generally prefer using the status names exported by the http package, mainly because I forget what the less common numbers mean.  In this case we could use:

http.StatusBadRequest instead of 400

http/batch_index.go
```diff
@@ -0,0 +1,114 @@
...
+		} else {
+			var doc interface{}
+			if err := dec.Decode(&doc); err == io.EOF {
+				break
+			} else if err != nil {
+				showError(w, req, fmt.Sprintf("error parsing document JSON: %v", err), 400)
+				return
+			}
+
+			batch, _ := openBatches[currentMeta.Index]
```
mschoch: In cases where you don't care about the second return value, you can omit it rather than include ", _".  Something like:

batch := openBatches[currentMeta.Index]

http/batch_index.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/blevesearch/bleve"
+	"io"
+	"net/http"
+)
+
+type MetaData struct {
```
mschoch: since we probably only use this data structure internally to this package, i would prefer we make it lower case:

type metaData {...

http/batch_index.go
```diff
@@ -0,0 +1,114 @@
...
+	openBatches := map[string]*bleve.Batch{}
+
+	dec := json.NewDecoder(req.Body)
+	isMeta := true
+	var currentMeta MetaData
+
+	for {
+		if isMeta {
+			var meta MetaData
+			if err := dec.Decode(&meta); err == io.EOF {
```
mschoch: seems like we can decode directly into currentMeta right? rather than decode into meta first and assign back to currentMeta?

mschoch:
@daschl I did some code level review, even though we still have open questions.  I'll circle back to those now.

mschoch:
I guess my preference would be to just stop on the first batch error encountered, and let the user decide if they want to keep going by retrying.  I'd rather not have this package decide if errors are fatal or retry-able, etc.

But, yeah we'd still need a result structure which captures a map of index and status.  Maybe (and this  just a proposal):

{
  "indexes": {
    "beer-search": {
      "status": "ok"
    },
    "travel-sample": {
      "status": "failed",
      "error": "full error message"
    },
    "default": {
      "status": "not executed"
    }
  }
}
    
Possible statuses:
-  "ok" - executed successfully
- "failed" or "error" - executed unsuccessfuly, additional "error" key with details
- "not executed" or something similar but shorter means we didn't even try this one?

Was thinking maybe have top-level status of "ok"/"fail"/"partial" but maybe we're over thinking it and just keep it simple.

mschoch:
Closing due to inactivity.

PR #415, opened 2016-08-31
cap preallocation by the collector to reasonable value

### LINKED ISSUE #408: TopNCollector unsuitable for retrieving all hits
Sorry I can't dig in and provide more info right now, but commit 24a2b57 introduced a bug that causes the Search function to hang and never return (it works with previous commit).  I'm using an in-memory index (gtreap), and am doing the following:

1.  Creating the in-memory index with an index mapping that includes a DateTimeMapping field.
1.  Batch index some documents using the mapping.
1.  Create a search request with `bleve.NewDateRangeInclusiveQuery`.
1.  Execute the search -- it hangs until the test times out.

If I can, I'll try to write a test later tonight to reproduce the problem.

mschoch:
the collector has optimizations to avoid allocation and reslicing
during the common case of searching for top hits

however, in some cases users request an a very large number of
search hits to be returned (attempting to get them all)  this
caused unnecessary allocation of ram.

to address this we introduce a new constant PreAllocSizeSkipCap
it defaults the value of 1000.  if your search+skip is less than
this constant, you get the optimized behavior.  if your
search+skip is greater than this, we cap the preallcations to
this lower value.  additional space is acquired on an as needed
basis by growing the DocumentMatchPool and reslicing the
collector backing slice

applications can change the value of PreAllocSizeSkipCap to suit
their own needs

fixes #408

### DIFF
--- a/index_test.go
+++ b/index_test.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math"
 	"os"
 	"reflect"
 	"sort"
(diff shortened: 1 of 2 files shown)


mschoch:
Review welcome @mmindenhall and @steveyen

steveyen:
looks good to me -- makes sense

mmindenhall:
I used `math.MaxInt32` as the size because it's somewhat of a convention (maybe not a good one) for when you want "all" of something from an API that specifies a result limit.

I think if I use this convention to search an index that contains only a few documents, this change will preallocate space for 1000 results?  If so, is a further optimization possible to reduce the preallocation cap when `index.DocCount() < PreAllocSizeSkipCap`?

mschoch:
I honestly think using math.MaxInt32 is an anti-pattern.  But at the moment you have no other option.  A better fix would be a telling bleve you want all hits, then it can choose an more appropriate collector.

I don't think this collector should be doing any optimizations based on the doc count of the index.  If your use case is one where you always have small indexes AND you always want all of the hits, then you should just adjust `PreAllocSizeSkipCap` in your application.

I just realized I had incorrectly made `PreAllocSizeSkipCap` a const, when I intended it to be a var that applications can change.

mmindenhall:
OK...makes sense.  Thanks!

PR #421, opened 2016-09-09
Smolder

mschoch:
An alternate indexing scheme which uses numbers instead of full doc IDs internally.  This saves some space if you use wide doc IDs like UUIDs.

There is some additional cost to maintain/lookup this new ID mapping.  So we've also added a cuckoo filter to avoid lookups when we know the documents don't exist.

It was also observed that our backindex rows were unnecessarily verbose, so the format of these rows was tightened up slightly.

If you regularly delete older docs and add new ones, the width of the internal IDs grows over time which makes the index less compact over time.

This is currently experimental and should not be used.


mschoch:
Confirmed that the failing unit test is some sort of bug.  I had seen it before but thought it was related to other tests not cleaning up properly.  Seeing it here and now confirming it happens locally as well...

mschoch:
As a data point, I just rebased the smolder branch with all the other fixups, and for the first time averaged > 5MB/s on my laptop.  The exact value was 5.120403.

The upside_down indexing scheme with all the same changes gets 4.608407MB/s.

CC @steveyen would like to have the discussion about merging this tomorrow.

PR #423, opened 2016-09-11
avoid allocation in stop token filter

mschoch:
the token stream resulting from the removal of stop words must
be shorter or the same length as the original, so we just
reuse it and truncate it at the end.


mschoch:
Review @steveyen

steveyen:
I kind of forget where TokenStreams come from.  Any chance that the current callers of this filter would have a cached, reused token stream that's provided as input?

If not, looks good to me!

mschoch:
Every token filter is passed in a TokenStream and has to return one.  It owns the stream for the duration of the request and cannot hold reference to it beyond that.  Many of the other filters already reuse the original stream (when its not growing in size).

PR #424, opened 2016-09-11
speed up english possessive filter

mschoch:
previous impl always did full utf8 decode of rune
if we assume most tokens are not possessive this is unnecessary
and even if they are, we only need to chop off last to runes
so, now we only decode last rune of token, and if it looks like
s/S then we proceed to decode second to last rune, and then
only if it looks like any form of apostrophe, do we make any
changes to token, again by just reslicing original to chop
off the possessive extension

### DIFF
--- a/analysis/language/en/possessive_filter_en.go
+++ b/analysis/language/en/possessive_filter_en.go
@@ -10,7 +10,7 @@
 package en
 
 import (
-	"bytes"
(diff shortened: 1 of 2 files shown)


mschoch:
Review @steveyen

steveyen:
Nice!  looks good to me

PR #425, opened 2016-09-11
improve perf of porter stemmer

mschoch:
1.  porter stemmer offers method to NOT do lowercasing, however
to use this we must convert to runes first ourself, so we did this

2.  now we can invoke the version that skips lowercasing, we
already do this ourselves before stemming through separate filter

due to the fact that the stemmer modifies the runes in place
we have no way to know if there were changes, thus we must
always encode back into the term byte slice

added unit test which catches the problem found

NOTE this uses analysis.BuildTermFromRunes so perf gain is
only visible with other PR also merged

future gains are possible if we udpate the stemmer to let us
know if changes were made, thus skipping re-encoding to
[]byte when no changes were actually made

### DIFF
--- a/analysis/token_filters/porter/porter.go
+++ b/analysis/token_filters/porter/porter.go
@@ -10,6 +10,8 @@
 package porter
 
 import (
+	"bytes"
+
 	"github.com/blevesearch/bleve/analysis"
 	"github.com/blevesearch/bleve/registry"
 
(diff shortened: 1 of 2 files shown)


mschoch:
Review @steveyen

mschoch:
Oh, there are failing tests, looks like there may be some problem here.

mschoch:
Ah, it looks like the stemmer transforms the original rune slice, defeating some of what I was trying to do.  Let me rethink...

mschoch:
Updated the commit, this is now a smaller gain (only skips lowercasing).  Also because this uses BuildTermFromRunes, the perf gains are only visible once the other PR is merged.  Finally, I added a unit test to catch the problem found by the integration test.

steveyen:
looks good to me.

btw, typo in commit mtg "anlysis" -- hehe, starting to suspect you're sneaking in some typos into the ancillary parts of your PR's just to keep the fanbase on its toes :-)

mschoch:
No secret plan to insert typos, just weekend coding which usually means I'm doing other things at the same time.

PR #426, opened 2016-09-11
encode runes directly into buffer

mschoch:
avoid allocating unnecessary intermediate buffer

### DIFF
--- a/analysis/util.go
+++ b/analysis/util.go
@@ -34,14 +34,32 @@ func InsertRune(in []rune, pos int, r rune) []rune {
 	return rv
 }
 
-func BuildTermFromRunes(runes []rune) []byte {
-	rv := make([]byte, 0, len(runes)*4)
+// BuildTermFromRunesOptimistic will build a term from the provided runes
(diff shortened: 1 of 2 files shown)


mschoch:
Review @steveyen

mschoch:
I've updated this to include a new variant method which can optimistically attempt to encode back into an existing []byte.  The thinking is that in a lot of cases, we start with a []byte, go into runes, do some work then come back to []byte.  Often the new one is the same or shorter, thus we can encode back into the original buffer and avoid new allocation.  This method intentionally takes an optimistic approach, it checks that there is enough space one rune at a time so that it can do everything in a single pass, but it means that if we run out of space, we have more bytes to copy.

The thinking is that this will be a good tradeoff, but we could update it to check the full length needed first (makes 2 passes).  Or we could make that into a separate method if we think both are useful.

For backwards compat, the existing BuildTermFromRunes allocates a new buffer with sufficient size.  This way we can move users over to the new function as needed.

steveyen:
looks good to me.

the only thing i might note is a commit title typo

PR #429, opened 2016-09-13
BREAKING CHANGE - removed DumpXXX() methods from bleve.Index

mschoch:
The DumpXXX() methods were always documented as internal and
unsupported.  However, now they are being removed from the
public top-level API.  They are still available on the internal
IndexReader, which can be accessed using the Advanced() method.

The DocCount() and DumpXXX() methods on the internal index
have moved to the internal index reader, since they logically
operate on a snapshot of an index.


mschoch:
Review @steveyen

steveyen:
looks good to me

PR #430, opened 2016-09-14
migrated all bleve utils into single bleve command

mschoch:
used spf13/cobra to make it awesome
and attempting to vendor this new dep


mschoch:
Review @steveyen

mschoch:
Will also need to tweak our custom code coverage scripts to ignore vendor code.

steveyen:
looks good to me

PR #440, opened 2016-09-21
Make sure that the transaction is closed if there is an error

slavikm:
We tried to index some terms that were too big and bolt failed writing them (key too large) but now the index hangs because the transaction remained open.

### DIFF
--- a/index/store/boltdb/writer.go
+++ b/index/store/boltdb/writer.go
@@ -51,7 +51,6 @@ func (w *Writer) ExecuteBatch(batch store.KVBatch) (err error) {
 			tx.Rollback()
 		}
 	}()
-	defer tx.Rollback()
 
 	bucket := tx.Bucket([]byte(w.store.bucket))
 	bucket.FillPercent = w.store.fillPercent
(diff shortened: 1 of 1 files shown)


index/store/boltdb/writer.go
```diff
@@ -38,7 +38,8 @@ func (w *Writer) ExecuteBatch(batch store.KVBatch) error {
 	if err != nil {
 		return err
 	}
-
+	defer tx.Rollback()
```
daschl: I am not super familiar with the code, but shouldn't this go in the if block above?

mschoch: @daschl unfortunately it depends.  I generally follow the pattern where a constructor for a Closable thing returns nil, and not a half-open thing, for the case where an error occurs while creating/opening.  Thus the case of Rolling back the thing you *just* failed to open isn't required (because we did any necessary clean up for you).  So, for the me contract is, you **MUST** call Close() on anything that was successfully opened.  And if it was not successfully opened in the first place, that isn't required.

But, in the end you have to check, because not everyone follows that.

slavikm:
@mschoch I like the suggested change. Would you like me to add this or are you going to?

mschoch:
It might be faster if you do it, I won't be able to get to it for a few hours.

PR #444, opened 2016-09-23
Move IndexMapping and Query into separate packages

mschoch:
## Overview

The top-level bleve package was always envisioned as glueing together functionality offered by the sub-packages into an easy to use layer.  Over time, a lot of logic has crept into the top-level package that possibly could exist elsewhere.  Also, its important to remember that while we want to ensure that you *can* use bleve with just the top-level package, it *is* OK for more complex applications to need to import the sub-packages.

### Why?

The IndexMapping, DocumentMapping, and the 4 FieldMapping structures all contain somewhat complex logic.  However, the interactions thats the top-level bleve package has with mappings are straight-forward.  If we introduce a new interface `IndexMapping` and move the existing implementation structure to IndexMappingImpl then we can carve this out into a separate package.

By putting this logic into a separate package, we make it easier for people to understand the top-level package because these unimportant details are moved out.

**NOTE**: Although the introduction of this interface means it would be possible to have alternate implementations in the future, currently that is not possible because mappings are stored in the index, and upon deserializing them we must know which implementation was used.  This too could be enhanced in the future, but is outside the scope of the current proposal.

### Wrapper functions

Because the top-level bleve package is intended to be usable on its own, we provide wrapper functions to create the mapping strucutres:

```
NewIndexMapping()
NewDocumentMapping()
NewDocumentStaticMapping()
NewDocumentDisabledMapping()
NewTextFieldMapping()
NewNumericFieldMapping()
NewDateTimeFieldMapping()
NewBooleanFieldMapping()
```

### Changes Required

Even with the wrapper functions, some old code would need to be changed.  If you made explicit references to `bleve.IndexMapping` like:

```
var mapping bleve.IndexMapping
mapping = bleve.NewIndexMapping()
```

Then you would either need to import the new `mapping` package and write:

```
var mapping mapping.IndexMapping
mapping = bleve.NewIndexMapping()
```

or, rewrite to the style:

```
```

### Why?

We have many different query structures, and although they are not complex objects, the end up cluttering the top-level bleve API.  The bleve top-level API only works with queries in a very limited way, one that is already captured by the Query interface.  By moving the Query interface definition and the implementation structures into a new package, we simplify the top-level package and make it easier to understand.

### Query interface changes

The Query interface previously had several methods, it has now been reduced to a single method Searcher(...).

All other features, Boostable, Fieldable, and Validatable have been broken off into separate interfaces.  This allows these optional capabilities to be discoverable, and only used when needed.  It also allowed us to remove code that were simply stub implementations that did nothing.

### Query structures now exported

Previously the query structures themselves were not exported.  This was always consider a bad practice in Go, and now as a part of the moving to their own package, they are exported.  Along with this change, the field members have more sane names (Boost, not BoostVal).  This means that is also more reasonable to work with these structures directly and avoid the New...() methods entirely.

### Query methods no longer chainable

Previously we support constructs like:

```
q := NewTermQuery("water").SetBoost(2.5)
```

Along with these changes we lose this capability.  The reason is that the return type of the interface methods is limited, so by embracing the interface for these capabilities, we lose the ability to chain.  I actually find this isn't such bad thing, though it does require some rewriting.
Now this same operation would need to be written on two lines as:

```
q := NewTermQuery("water")
q.SetBoost(2.5)
```

### DateTimeQuery changed to work with time.Time not string

This is unrelated to the other changes, but it made sense to do it at the same time.  Previously, DateTimeQuery always involved parsing strings.  Now, for Go applications you represent endpoints using the standard time.Time structures.  Unspecified endpoints use the zero Time.  This is much more natural for Go applications.

For applications using JSON serialization/deserializaiton of these structures, the package level variables `QueryDateTimeParser` and `QueryDateTimeFormat` control the supported formats on input and output.

### Wrapper functions

Because the top-level bleve package is intended to be usable on its own, we provide wrapper functions to create the query strucutres:

```
NewBoolFieldQuery()
NewBooleanQuery()
NewConjunctionQuery()
NewDateRangeQuery()
NewDateRangeInclusiveQuery()
NewDisjunctionQuery()
NewDisjunctionQueryMin()
NewDocIDQuery()
NewFuzzyQuery()
NewMatchAllQuery()
NewMatchNoneQuery()
NewMatchPhraseQuery()
NewMatchQuery()
NewMatchQueryOperator()
NewNumericRangeQuery()
NewNumericRangeInclusiveQuery()
NewPhraseQuery()
NewPrefixQuery()
NewRegexpQuery()
NewQueryStringQuery()
NewTermQuery()
NewWildcardQuery()
```

### Changes Required

Even with the wrapper functions, some old code would need to be changed.  If you made explicit references to `bleve.Query` like:

```
var q bleve.Query
q = bleve.NewTermQuery("water")
```

Then you would need to change that to:

```
var q query.Query
q = bleve.NewTermQuery("water")
```

or rewrite it in the form:

```
q := bleve.NewTermQuery("water")
```

Another case requiring change is with queries that refer to other queries.  As an example we'll consider the conjunction query.  Previously you created these as:

```
q1 := NewTermQuery("water")
q2 := NewTermQuery("soda")
cq := NewConjunctionQuery([]bleve.Query{q1, q2})
```

Writing the construct in exactly that form would have required importing the new query package, so instead we've changed the constructor in these wrappers to use the ...query.Query notation.  This lets you pass in a slice of queries, without explicitly instantiating a literal slice.  The new version looks like:

```
q1 := NewTermQuery("water")
q2 := NewTermQuery("soda")
cq := NewConjunctionQuery(q1, q2)
```

A similar problem afflicts the BooleanQuery, which took 3 separate slices, one each for the MUST, SHOULD and MUST NOT clauses.  Previously you would write:

```
q1 := NewTermQuery("water")
q2 := NewTermQuery("soda")
q3 := NewTermQuery("butter")
cq := NewBooleanQuery([]bleve.Query{q1, q2}, nil, []bleve.Query{q3})
```

The approach we used above only works for a single slice, so instead we have remove all those queries from the constructor.  Instead you add them after constructing the query.  So in the new version you'd write:

```
q1 := NewTermQuery("water")
q2 := NewTermQuery("soda")
q3 := NewTermQuery("butter")
cq := NewBooleanQuery()
cq.AddMust(q1, q2)
cq.AddMustNot(q3)
```

Although more verbose, I actually think it increases the readability of the query.  At a glance you can identify which queries are being added to which clauses.


mschoch:
Ready for initial review @steveyen

mschoch:
Actually @steveyen it seems like current moss has broken bleve.  I'm not exactly sure how as the manifest was supposed to prevent this.  Any ideas?

steveyen:
First quickie reaction from the PR description (which is one of the best written things I've read all week)...

Not sure if these deserve their own top-level helper API's...
- NewDocumentStaticMapping()
- NewDocumentDisabledMapping()

Still need to dig into the actual code.

One other 0.01$ thought is you'd probably want yet another tag (like v0.5.0?) right before this goes in.

mschoch:
@steveyen regarding NewDocumentStaticMapping and NewDocumentDisabledMapping these two both also have helpers in the top-level.

And I rewrote the top-section to more clearly explain the changes, the motivation, and how I expect users may have to update their code.

examples_test.go
```diff
@@ -464,50 +447,6 @@ func ExampleNewDisjunctionQueryMin() {
...
-
-	fieldMapping := NewTextFieldMapping()
-	fieldMapping.Analyzer = "en"
-	documentMapping.AddFieldMappingsAt("NestedProperty", fieldMapping)
-
-	fmt.Println(len(documentMapping.Properties["NestedProperty"].Fields))
-	// Output:
-	// 1
-}
-
```
steveyen: On the removal of these example document mappings... wondering if the functionality went away?

mschoch: No, but they had to move to the package where those functions actually are (mapping).  The go vet tool complained that the documentation was for functions that were no longer here.

On the one hand, that seems unfortunate as you have to dig deeper.  On the other hand, most users need not worry about the details of mappings at first.  And when they are ready to, those details are in one place, the "mapping" package.

mschoch: Oh, so if that wasn't clear, the documentation had to move along with the functions they describe.

index_impl.go
```diff
@@ -228,7 +229,7 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
 		return nil, err
 	}
 
-	var im IndexMapping
+	var im *mapping.IndexMappingImpl
```
steveyen: wondering here offhand if the mapping unmarshaling might or is able to be moved fully into the mapping package, so that way mapping.IndexMappingImpl doesn't necessarily have to be exposed?

examples_test.go
steveyen: cool, thx

index_impl.go
mschoch: I think it has to be exposed anyway, because it has lots of public members that users set directly.  So, in order to not break that, I don't return the interface, I return the actual implementation.

I'm not sure how you mean about moving the mapping unmarshaling...

index_test.go
```diff
@@ -1297,9 +1280,9 @@ func TestDateTimeFieldMappingIssue287(t *testing.T) {
 	}
 
 	// search range across all docs
-	start := now.Add(-4 * time.Hour).Format(time.RFC3339)
-	end := now.Format(time.RFC3339)
-	sreq := NewSearchRequest(NewDateRangeQuery(&start, &end))
+	start := now.Add(-4 * time.Hour)
+	end := now
+	sreq := NewSearchRequest(NewDateRangeQuery(start, end))
```
steveyen: Looks much better!  As a user, I don't have to remember/lookup the RFC-what-what-something-something anymore.

search/query/boolean.go
```diff
@@ -33,19 +34,17 @@ type booleanQuery struct {
...
+func NewBooleanQuery(must []Query, should []Query, mustNot []Query) *BooleanQuery {
 	return NewBooleanQueryMinShould(must, should, mustNot, 0.0)
 }
 
 // NewBooleanQueryMinShould is the same as
 // NewBooleanQuery, only it offers control of the
 // minimum number of should queries that must be
 // satisfied.
-func NewBooleanQueryMinShould(must []Query, should []Query, mustNot []Query, minShould float64) *booleanQuery {
+func NewBooleanQueryMinShould(must []Query, should []Query, mustNot []Query, minShould float64) *BooleanQuery {
```
steveyen: This might be a separate, independent issue than the API subpackage move...

But, I was wondering whether NewBooleanQueryMinShould() was really popular?  That is, I was thinking that NewBooleanQuery() + SetMinShould() is equivalent?  And we often try to limit the API porcelain "surface area" when we can?

search/query/boolean.go
```diff
@@ -61,41 +60,43 @@ func NewBooleanQueryMinShould(must []Query, should []Query, mustNot []Query, min
...
 // SetMinShould requires that at least minShould of the
 // should Queries must be satisfied.
-func (q *booleanQuery) SetMinShould(minShould float64) {
-	q.Should.(*disjunctionQuery).SetMin(minShould)
+func (q *BooleanQuery) SetMinShould(minShould float64) {
+	q.Should.(*DisjunctionQuery).SetMin(minShould)
 }
 
-func (q *booleanQuery) AddMust(m Query) {
+func (q *BooleanQuery) AddMust(m ...Query) {
```
steveyen: For folks that did used to have an slice of child queries...
qrys := []Query{q1, q2, q3, q4}

...might want to have an example somewhere that they can use the '...' idiom?
booleanQuery.AddMust(...qrys)

search/query/boolean.go
mschoch: Yeah, I'd be in favor of that, there is a SetMinShould() method on the BooleanQuery, so it can always be done in 2 steps.

Similarly there is NewMatchQueryOperator() which does not have a setter method to toggle, but I think if it did, that constructor could go away too.  I'll see about packaging this up into another commit.

search/query/boolean.go
mschoch: Yeah this was introduced to address the opposite case.  Previously, they have []bleve.Query{}, now that has to be come []query.Query{} thus necessessitating the import of the "query" package, which I assume many would rather avoid.  By writing it this way you can pass (q1, q2, q3) without needing to create the array first.

But your point is valid, just not sure where to document/demonstrate that.

search/query/disjunction.go
```diff
@@ -7,65 +7,55 @@
...
+	q.Boost = &boost
 }
 
-func (q *disjunctionQuery) SetBoost(b float64) Query {
-	q.BoostVal = b
-	return q
-}
-
-func (q *disjunctionQuery) AddQuery(aq Query) Query {
+func (q *DisjunctionQuery) AddQuery(aq Query) {
```
steveyen: One thought is if this took the multi-value '...' idiom as an input type then folks could then use...

   disjQuery.AddQuery(q1, q2, q3)

or

   disjQuery.AddQuery(...queryParts)

mschoch: Yeah I'll make that change for consistency with BooleanQuery

PR #452, opened 2016-09-27
BREAKING CHANGE - new method to create memory only index

### LINKED ISSUE #427: make creating in-memory only indexes more obvious
Currently, the way to create a new in-memory only index is to pass "" as the first argument to New, like:

```
index, err := bleve.New("", NewIndexMapping())
```

I received feedback from @a-little-srdjan a few weeks back that this was neither obvious, nor very clear to people later reading the code.  At the time had proposed a new method to create in-memory only indexes.

I'm a bit reluctant to add a new method for this as it seems a bit wasteful to me.  But I agree that it is not obvious, and many users have asked this question (most recently from @AlexanderDaw on gitter).

My proposal is to add a constant to the bleve package, like:

```
const InMemory = ""
```

Then the code to create a new in-memory index would look like:

```
index, err := bleve.New(bleve.InMemory, NewIndexMapping())
```

I think this does address the issue of code readability, but I'm willing to consider other ideas.

Thoughts?

mschoch:
Previously bleve allowed you to create a memory-only index by
simply passing "" as the path argument to the New() method.

This was not clear when reading the code, and led to some
problematic error cases as well.

Now, to create a memory-only index one should use the
NewMemOnly() method.  Passing "" as the path argument
to the New() method will now return os.ErrInvalid.

Advanced users calling NewUsing() can create disk-based or
memory-only indexes, but the change here is that passing ""
as the path argument no longer defaults you into getting
a memory-only index.  Instead, the KV store is selected
manually, just as it is for the disk-based solutions.

Here is an example use of the NewUsing() method to create
a memory-only index:

```
NewUsing("", indexMapping, Config.DefaultIndexType,
         Config.DefaultMemKVStore, nil)
```

`Config.DefaultMemKVStore` is just a new default value
added to the configuration, it currently points to
`gtreap.Name` (which could have been used directly
instead for more control)

closes #427

### DIFF
--- a/config.go
+++ b/config.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/blevesearch/bleve/analysis/datetime_parsers/datetime_optional"
 	"github.com/blevesearch/bleve/index"
+	"github.com/blevesearch/bleve/index/store/gtreap"
 	"github.com/blevesearch/bleve/index/upside_down"
 	"github.com/blevesearch/bleve/registry"
 	"github.com/blevesearch/bleve/search/highlight/highlighters/html"
@@ -28,6 +29,7 @@ type configuration struct {
 	Cache                  *registry.Cache
 	DefaultHighlighter     string
 	DefaultKVStore         string
+	DefaultMemKVStore      string
 	DefaultIndexType       string
 	QueryDateTimeParser    string
 	SlowSearchLogThreshold time.Duration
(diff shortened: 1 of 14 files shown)


mschoch:
Review @a-little-srdjan @daschl @AlexanderDaw @steveyen

daschl:
:+1: from an API perspective, I leave the code review to the more experiencenced gophers!

PR #458, opened 2016-09-30
Metrics Work

mschoch:
First, we no longer export the internal metrics timers (opens door for vendoring go-metrics in the future)

~~Second, we aggregate stats on an iterator, the record those numbers on Close() of the iterator.  Internally, the metrics library maintains a mutex on the stat, which must be acquired/released, so we want avoid that in the hottest paths.~~

### DIFF
--- a/index/store/metrics/iterator.go
+++ b/index/store/metrics/iterator.go
@@ -8,13 +8,13 @@ type Iterator struct {
 }
 
 func (i *Iterator) Seek(x []byte) {
-	i.s.TimerIteratorSeek.Time(func() {
+	i.s.timerIteratorSeek.Time(func() {
 		i.o.Seek(x)
 	})
 }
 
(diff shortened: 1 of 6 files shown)


index/store/metrics/iterator.go
```diff
@@ -36,6 +42,9 @@ func (i *Iterator) Valid() bool {
 }
 
 func (i *Iterator) Close() error {
+	// record time spent in aggregate
+	i.s.timerIteratorSeek.Update(i.seekTime)
+	i.s.timerIteratorNext.Update(i.nextTime)
```
steveyen: I'm thinking this changes the semantics of the seek time and next time stats.

Previously, it use to be the time of each individual seek() or next().

With this change, it's now different numbers... the totals.

mschoch:
Yeah, removing that second commit.

PR #470, opened 2016-10-11
Address special unicode sigma at end of term when lowercasing.

### LINKED ISSUE #468: lower case token filter incorrectly handles Σ at the end of a word
@daschl noticed that we incorrectly handle this case:

https://doc.rust-lang.org/src/collections/up/src/libcollections/str.rs.html#1634

daschl:
Σ maps to σ, except at the end of a word where it maps to ς.
This is the only conditional (contextual) but language-independent
mapping in unicode.

fixes #468

### DIFF
--- a/analysis/token/lowercase/lowercase.go
+++ b/analysis/token/lowercase/lowercase.go
@@ -75,6 +75,13 @@ func toLowerDeferredCopy(s []byte) []byte {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


daschl:
Note that since this conditional is behind the "don't do anything" optimization the additional check does not measurably (with the current benchmarks) have a negative effect.

PR #484, opened 2016-10-22
add the bleve check tool

mschoch:
bleve check was a consistency checking tool originally developed
as a part of cbft.  currently it checks that the term dictionary
counts match the number of postings for the term.  in the future
additional checks could be added.  this tool has been back
ported to bleve as we've now adopted a single common tool for
both cbft and bleve.


mschoch:
Review @steveyen

PR #485, opened 2016-10-24
vendor mousetrap

mschoch:
our policy has been to vendor the libraries required for the
command-line portion of the bleve command-line tool.  this
included spf13/cobra etc, unfortunately we missed this library
as it is only required for windows users.


mschoch:
review @steveyen

PR #489, opened 2016-10-27
Refactor PhraseSearcher

Shugyousha:
In the first patch we use early continues to reduce the level of nesting. The following two patches fix some minor issues in the code and in the last patch we add checking of the DocumentMatch.Location values to the phrase searcher test.

If you prefer to have some of the patches squashed, just let me know.

### DIFF
--- a/search/searcher/search_phrase.go
+++ b/search/searcher/search_phrase.go
@@ -104,37 +104,38 @@ func (s *PhraseSearcher) Next(ctx *search.SearchContext) (*search.DocumentMatch,
 		for field, termLocMap := range s.currMust.Locations {
 			rvtlm := make(search.TermLocationMap, 0)
 			locations, ok := termLocMap[firstTerm]
-			if ok {
-			OUTER:
-				for _, location := range locations {
-					crvtlm := make(search.TermLocationMap, 0)
-				INNER:
-					for i := 0; i < len(s.terms); i++ {
-						nextTerm := s.terms[i]
-						if nextTerm != "" {
-							// look through all these term locations
-							// to try and find the correct offsets
-							nextLocations, ok := termLocMap[nextTerm]
-							if ok {
-								for _, nextLocation := range nextLocations {
-									if nextLocation.Pos == location.Pos+float64(i) && nextLocation.SameArrayElement(location) {
-										// found a location match for this term
-										crvtlm.AddLocation(nextTerm, nextLocation)
-										continue INNER
-									}
-								}
-								// if we got here we didn't find a location match for this term
-								continue OUTER
-							} else {
-								continue OUTER
-							}
+			if !ok {
+				continue
+			}
+		OUTER:
+			for _, location := range locations {
+				crvtlm := make(search.TermLocationMap, 0)
+			INNER:
+				for i := 0; i < len(s.terms); i++ {
+					nextTerm := s.terms[i]
+					if nextTerm == "" {
+						continue
+					}
+					// look through all these term locations
+					// to try and find the correct offsets
+					nextLocations, ok := termLocMap[nextTerm]
+					if !ok {
+						continue OUTER
+					}
+					for _, nextLocation := range nextLocations {
+						if nextLocation.Pos == location.Pos+float64(i) && nextLocation.SameArrayElement(location) {
+							// found a location match for this term
+							crvtlm.AddLocation(nextTerm, nextLocation)
+							continue INNER
 						}
 					}
-					// if we got here all the terms matched
-					freq++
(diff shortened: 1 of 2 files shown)


mschoch:
Sorry for the delay, I wanted to actually review the loop changes.  This looks good to me.

I'd also like to ask @steveyen for review as he was recently reading the phrase searcher.

While reading it I think there is another opportunity for simplification.  I think we can restructure the code such that the handling of the first term isn't a special case, but put into the same loop.  Until we try it though its hard to say if the result is more/less readable.

Shugyousha:
On Mon, Oct 31, 2016 at 3:38 PM, Marty Schoch <notifications@github.com> wrote:
> While reading it I think there is another opportunity for simplification. I
> think we can restructure the code such that the handling of the first term
> isn't a special case, but put into the same loop. Until we try it though its
> hard to say if the result is more/less readable.

Yes, that was my hunch as well. I will look into simplifying further
but it may take some time so I thought I would send these patches
first.

search/searcher/search_phrase.go
```diff
@@ -102,39 +102,42 @@ func (s *PhraseSearcher) Next(ctx *search.SearchContext) (*search.DocumentMatch,
...
+					// look through all these term locations
+					// to try and find the correct offsets
+					nextLocations, ok := termLocMap[nextTerm]
+					if !ok {
+						continue OUTER
+					}
+					for _, nextLocation := range nextLocations {
+						if nextLocation.Pos == location.Pos+float64(i) && nextLocation.SameArrayElement(location) {
+							// found a location match for this term
+							crvtlm.AddLocation(nextTerm, nextLocation)
```
steveyen: I'm not as familiar with the phrase searcher, but to my eyes, this change looks almost functionally the same, except with the change/optimization where the INNER for-loop starts at 1 instead of 0.

On that, does that mean the crvtlm.AddLocation(nextTerm, nextLocation) here won't happen for the first (or 0'th) term?  And, if so, is that an important difference in behavior?

Shugyousha:
On Mon, Oct 31, 2016 at 8:10 PM, Steve Yen <notifications@github.com> wrote:
> @steveyen commented on this pull request.
>
> ________________________________
>
> In search/searcher/search_phrase.go:
>
>> + INNER:
> + for i := 1; i < len(s.terms); i++ {
> + nextTerm := s.terms[i]
> + if nextTerm == "" {
> + continue
> + }
> + // look through all these term locations
> + // to try and find the correct offsets
> + nextLocations, ok := termLocMap[nextTerm]
> + if !ok {
> + continue OUTER
> + }
> + for _, nextLocation := range nextLocations {
> + if nextLocation.Pos == location.Pos+float64(i) &&
> nextLocation.SameArrayElement(location) {
> + // found a location match for this term
> + crvtlm.AddLocation(nextTerm, nextLocation)
>
> I'm not as familiar with the phrase searcher, but to my eyes, this change
> looks almost functionally the same, except with the change/optimization
> where the INNER for-loop starts at 1 instead of 0.

Yes, this patch set is supposed to decrease the nesting while keeping
the behavior the same, except for the changed offset.


> On that, does that mean the crvtlm.AddLocation(nextTerm, nextLocation) here
> won't happen for the first (or 0'th) term? And, if so, is that an important
> difference in behavior?

Good catch!

Now that you mention it, I think that's the reason why I wanted to
check the Locations and changed the test. Looking at my test again, I
only check for the second term. That's why there wasn't a difference
when I ran the test before and after applying the change to the INNER
for-loop.

In the original code, we should have two locations in the TermLocation
map, one for "angst" and one for "beer". I will have another look at
the code, change the test to more accurately check the behavior of the
original code and then try to replicate it exactly while simplifying
the for-loops.

Is it okay if I force changes to this PR or do you want me to close
this one and open a new one?

mschoch:
Force push to this PR is fine.

Shugyousha:
I changed the test to find the missing entry and changed the PhraseSearcher code to add the location of the first term.

I have experimented with a recursive version of the position-checking code for the PhraseSearcher which you can find here: https://github.com/Shugyousha/bleve/blob/phrasesearcherrecursive/search/searcher/search_phrase.go

With this approach I have to do a copy of the return value which makes it less efficient and by now I am also not sure if it is more readable or not. If you think it would be worth investigating that approach further I will have another look otherwise we can just keep this refactored version.

steveyen:
happy new year!  Latest version (non-recursive) looks ok to me.

I also took a look at the recursive version, and normally I do like a nice recursive solution.  But at this point, the non-recursive version does look "closer" to the original code to me, so has the benefit of being more mechanically understandable for me.

PR #491, opened 2016-10-28
IndexBatcher: wrapper to automatically batch modifications

robmccoll:
IndexBatcher aggregates modifications (Index(), Delete(), SetInternal(), DeleteInternal()) that occur within close proximity into a batch execution for increased throughput at the cost of an amortized period / 2 latency increase.  It is a fairly transparent wrapper around the Index interface that is itself an Index.  The period is adjustable.  When experimenting with Cassandra as a backend, I ran into throughput bottlenecks as a result of the number of keys being inserted at the KVStore layer.  Guiding individual insertions into batches via this layer nearly doubled throughput.

### DIFF
--- a/index_batcher.go
+++ b/index_batcher.go
@@ -9,9 +9,31 @@ import (
 // NOTE XXX As this approach uses carefully orchestrated interactions
 // between a mutex, channels, and reference swapping, be sure to carefully
 // consider the impact of reordering statements or making changes to the
-// patterns present in the code.
+// patterns present in the code. Specifically, individual operations follow the
+// pattern of:
+// - Construct empty references for result error, signal channel
+// - Acquire lock
+// - Assign references to current result error and signal channel
+// - Add operation to batch
+// - Release lock
+// - Wait on signal channel to close
+// - Return result error
+// While the batch loop follows the pattern on timer event:
+// - Acquire lock
+// - Execute batch
+// - Store result in current result error
+// - Make new current result error for use by operations in the next batch
+// - Close signal channel so that operations waiting on the completion of this
+//   batch will return the result pointed to by the previous result error that
+//   they are still holding.
+// - Make new current signal channel for use by operations in the next batch
+// - Release lock
+// This design minimizes the need for allocating channels, uses only one go
+// routine for the batching, and doesn't require tracking the number of
+// operations waiting on a response or looping to notify each waiting operation.
 
-// bleveIndex is an alias that allows the IndexBatcher to embed a Index
+// bleveIndex is an alias for Index used by the IndexBatcher to avoid a between
+// the embedded Index field and the overrideen Index method
 type bleveIndex Index
 
 // IndexBatcher can be wrapped around a Index to aggregate operations
@@ -29,13 +51,8 @@ type IndexBatcher struct {
 	// and adding operations to the batch.
 	lock   sync.Mutex
 	batch  *Batch
-	result *rslt
+	result *error
 	signal chan bool
-}
-
-// rslt is used as a kind of double pointer for errors
-type rslt struct {
-	err error
 }
 
 // NewIndexBatcher returns an index that will aggregate and fire modifying
@@ -52,7 +69,7 @@ func NewIndexBatcher(index Index, period time.Duration) Index {
 		period: period,
 		closer: make(chan bool),
 		signal: make(chan bool),
-		result: &rslt{},
+		result: new(error),
 	}
 
 	go ib.batchloop()
@@ -72,13 +89,13 @@ BatchLoop:
 			func() {
 				defer func() {
 					if r := recover(); r != nil {
-						ib.result.err = fmt.Errorf("IndexBatcher caught a panic: %v", r)
+						(*ib.result) = fmt.Errorf("IndexBatcher caught a panic: %v", r)
 					}
 				}()
-				ib.result.err = ib.Batch(ib.batch)
+				(*ib.result) = ib.Batch(ib.batch)
 			}()
 			ib.batch.Reset()
-			ib.result = &rslt{}
+			ib.result = new(error)
 			close(ib.signal)
 			ib.signal = make(chan bool)
 			ib.lock.Unlock()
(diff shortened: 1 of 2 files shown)


index_batcher.go
```diff
@@ -0,0 +1,164 @@
+package bleve
+
+import (
+	"fmt"
+	"sync"
+	"time"
+)
+
+// NOTE XXX As this approach uses carefully orchestrated interactions
```
mschoch: I think we should rewrite this comment with some concrete explanations.  Just saying 'be careful' isn't helpful if users aren't sure what the issues are.

index_batcher.go
```diff
@@ -0,0 +1,164 @@
...
+// between a mutex, channels, and reference swapping, be sure to carefully
+// consider the impact of reordering statements or making changes to the
+// patterns present in the code.
+
+// bleveIndex is an alias that allows the IndexBatcher to embed a Index
+type bleveIndex Index
```
mschoch: I would change the comment to explain that using Index directly as an embedded field then conflicts with the Index method.  It wasn't obvious to me and I had to try it to find out.

index_batcher.go
```diff
@@ -0,0 +1,164 @@
...
+	// the batch loop.  Elsewhere it is used for getting a reference to the
+	// current result, getting a reference to the current signal channel,
+	// and adding operations to the batch.
+	lock   sync.Mutex
+	batch  *Batch
+	result *rslt
+	signal chan bool
+}
+
+// rslt is used as a kind of double pointer for errors
```
mschoch: I still don't quite get the need for this.  Locally I just changed it to `error` and it compiles and the test passes.  Can you explain?

mschoch:
I put some line comments in, but I actually also have a larger question.  I'm not sure whether this belongs in the core bleve repo, or in blevex.  I'm a bit reluctant to increase the surface area of the API at this late stage, but I could be persuaded.  I'll try to ask around and get some other opinions.

index_batcher.go
robmccoll: Yeah I gave up to quickly on trying to figure out how to make Go dynamically allocate a nil error interface (turns out new(error)) and instead thought "well I'll just wrap a concrete type around the interface since its easier to work with a pointer to a concrete type than a pointer to an interface anyway".  Your comment drove me to fix it though.

index_batcher.go
robmccoll: Good point - added a description of the protocol for synchronization between the batch loop and the operation methods along with a bit of explanation as to why.

index_batcher.go
robmccoll: Yeah I gave up to quickly on trying to figure out how to make Go dynamically allocate a nil error interface (turns out new(error)) and instead thought "well I'll just wrap a concrete type around the interface since its easier to work with a pointer to a concrete type than a pointer to an interface anyway". Your comment drove me to fix it though.

robmccoll:
That's a fair question to ask - no pressure to include it in one place, the other, or either.  I just wrote this up for my own use and figured if it helps anyone else get better throughput in a concurrent operations scenario, why not contribute it back?

index_batcher.go
```diff
@@ -29,15 +51,10 @@ type IndexBatcher struct {
 	// and adding operations to the batch.
 	lock   sync.Mutex
 	batch  *Batch
-	result *rslt
+	result *error
```
mschoch: I think this can be the much more common and even simpler `error` and not pointer to error.  Error is an interface, so pointers to them are not very common.

index_batcher.go
```diff
@@ -52,7 +69,7 @@ func NewIndexBatcher(index Index, period time.Duration) Index {
 		period: period,
 		closer: make(chan bool),
 		signal: make(chan bool),
-		result: &rslt{},
+		result: new(error),
```
mschoch: this line can just be removed, the 0 value will be nil.

index_batcher.go
```diff
@@ -72,13 +89,13 @@ BatchLoop:
 			func() {
 				defer func() {
 					if r := recover(); r != nil {
-						ib.result.err = fmt.Errorf("IndexBatcher caught a panic: %v", r)
+						(*ib.result) = fmt.Errorf("IndexBatcher caught a panic: %v", r)
```
mschoch: all of the lines of this form can just 
```
ib.result =
```

index_batcher.go
```diff
@@ -72,13 +89,13 @@ BatchLoop:
...
 					}
 				}()
-				ib.result.err = ib.Batch(ib.batch)
+				(*ib.result) = ib.Batch(ib.batch)
 			}()
 			ib.batch.Reset()
-			ib.result = &rslt{}
+			ib.result = new(error)
```
mschoch: this can just be
```
ib.result = nil
```

index_batcher.go
```diff
@@ -9,9 +9,31 @@ import (
...
+// - Close signal channel so that operations waiting on the completion of this
+//   batch will return the result pointed to by the previous result error that
+//   they are still holding.
+// - Make new current signal channel for use by operations in the next batch
+// - Release lock
+// This design minimizes the need for allocating channels, uses only one go
+// routine for the batching, and doesn't require tracking the number of
+// operations waiting on a response or looping to notify each waiting operation.
+
+// bleveIndex is an alias for Index used by the IndexBatcher to avoid a between
```
mschoch: missing word in `avoid a between`

index_batcher.go
robmccoll: You are correct that error is an interface and that pointers are very uncommon; however, in this circumstance, the pointer to error is intentional.  Otherwise there is a race between the scheduler releasing previous calls to Index(), Delete() etc. picking up the result of the previous batch and the batchLoop() firing again and overwriting the result that those operations were supposed to have received.  Basically what's happening is described in this PDF (unless I'm missing something): 
[IndexBatcher.pdf](https://github.com/blevesearch/bleve/files/566710/IndexBatcher.pdf)

mschoch:
I've put this copy of the IndexBatcher into the blevex repo.  If any further changes/improvements are needed let's discuss it there.

https://github.com/blevesearch/blevex/tree/master/batcher

PR #493, opened 2016-11-03
numeric range facet merging compare range values not pointers

### LINKED ISSUE #492: multisearch does not correctly merge numeric facet results
A sample multisearch results is here:

https://gist.github.com/mschoch/f076de6c42567adaf8e1efacdb14b646

As you can see there are many duplicated facet result buckets that are identical, but are not properly merged together.

It appears the problematic line is here:

https://gist.github.com/mschoch/f076de6c42567adaf8e1efacdb14b646

We compare pointers, not the values they point to.  Notice how for date facets we use a separate function Same() which appears to take the pointers into account.  Likely we encountered this before with dates, but did not appreciate it affected numeric ranges as well.

https://github.com/blevesearch/bleve/blob/master/search/facets_builder.go#L138-L159

mschoch:
fix #492

### DIFF
--- a/search/facets_builder.go
+++ b/search/facets_builder.go
@@ -105,11 +105,34 @@ type NumericRangeFacet struct {
 	Count int      `json:"count"`
 }
 
+func (nrf *NumericRangeFacet) Same(other *NumericRangeFacet) bool {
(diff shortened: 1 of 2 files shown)


mschoch:
review @steveyen

PR #496, opened 2016-11-04
fix date facets when using MultiSearch

mschoch:
changed date parsing to NOT update internal state of the date
range object (avoids races)

second, when marshaling a facet date range, we now use the
string version, if the time.Time is zero and the string version
is not ""


mschoch:
review @steveyen

PR #506, opened 2016-11-30
Query String - fix less/greater than on dates

porjo:
logic for less than / greater than for dates in query string shouldn't be inclusive


porjo:
Sorry, I realised my fix wasn't required after all!

PR #515, opened 2017-01-12
Add recover to the multi-search

slavikm:
Not sure how to add a dependency to the vendor manifest

### DIFF
--- a/index_alias_impl.go
+++ b/index_alias_impl.go
@@ -26,6 +26,7 @@ import (
 	"github.com/blevesearch/bleve/index/store"
 	"github.com/blevesearch/bleve/mapping"
 	"github.com/blevesearch/bleve/search"
+	"github.com/go-errors/errors"
 )
 
 type indexAliasImpl struct {
@@ -455,10 +456,15 @@ func MultiSearch(ctx context.Context, req *SearchRequest, indexes ...Index) (*Se
 	var waitGroup sync.WaitGroup
 
 	var searchChildIndex = func(in Index, childReq *SearchRequest) {
-		rv := asyncSearchResult{Name: in.Name()}
+		rv := &asyncSearchResult{Name: in.Name()}
+		defer func() {
+			if err := recover(); err != nil {
+				rv.Err = errors.Wrap(err, 2)
+			}
+			asyncResults <- rv
+			waitGroup.Done()
+		}()
 		rv.Result, rv.Err = in.SearchInContext(ctx, childReq)
-		asyncResults <- &rv
-		waitGroup.Done()
 	}
 
 	waitGroup.Add(len(indexes))


aviadl:
This is the stack trace we got
```
goroutine 136 [running]:
panic(0x11142e0, 0xc4200120b0)
        /usr/local/go/src/runtime/panic.go:500 +0x1a1
sort.(*StringSlice).Less(0xc4211460c0, 0x17, 0x24, 0x1)
        <autogenerated>:13 +0x93
sort.doPivot(0x1d159c0, 0xc4211460c0, 0x17, 0x2c, 0x17, 0x17)
        /usr/local/go/src/sort/sort.go:121 +0x1be
sort.quickSort(0x1d159c0, 0xc4211460c0, 0x17, 0x2c, 0xa)
        /usr/local/go/src/sort/sort.go:188 +0x83
sort.quickSort(0x1d159c0, 0xc4211460c0, 0x0, 0x2c, 0xb)
        /usr/local/go/src/sort/sort.go:195 +0x130
sort.Sort(0x1d159c0, 0xc4211460c0)
        /usr/local/go/src/sort/sort.go:222 +0x80
sort.Strings(0xc4200d2b00, 0x2c, 0x2c)
        /usr/local/go/src/sort/sort.go:298 +0x76
github.com/blevesearch/bleve/index/upsidedown.newUpsideDownCouchDocIDReaderOnly(0xc421146060, 0xc4200d2b00, 0x2c, 0x2c, 0x135ce, 0xc421146020, 0x519340)
        github.com/blevesearch/bleve/index/upsidedown/reader.go:183 +0x5d
github.com/blevesearch/bleve/index/upsidedown.(*IndexReader).DocIDReaderOnly(0xc421146060, 0xc4200d2b00, 0x2c, 0x2c, 0x1196820, 0x38, 0x14a966c, 0x10)
        github.com/blevesearch/bleve/index/upsidedown/index_reader.go:58 +0x49
github.com/blevesearch/bleve/search/searcher.NewDocIDSearcher(0x1d1eb00, 0xc421146060, 0xc4200d2b00, 0x2c, 0x2c, 0x3ff0000000000000, 0x0, 0x0, 0x0, 0x38)
        github.com/blevesearch/bleve/search/searcher/search_docid.go:33 +0x70
github.com/blevesearch/bleve/search/query.(*DocIDQuery).Searcher(0xc42136e280, 0x1d1eb00, 0xc421146060, 0x1d18c00, 0xc4238a2480, 0x0, 0xc42003c6f0, 0x135ce, 0xc4211460a0, 0x20)
        github.com/blevesearch/bleve/search/query/docid.go:48 +0x7b
github.com/blevesearch/bleve/search/query.(*DisjunctionQuery).Searcher(0xc42378dda0, 0x1d1eb00, 0xc421146060, 0x1d18c00, 0xc4238a2480, 0xc42003c700, 0x4359b, 0x20, 0x1196820, 0x1)
        github.com/blevesearch/bleve/search/query/disjunction.go:65 +0xf4
github.com/blevesearch/bleve/search/query.(*BooleanQuery).Searcher(0xc423780800, 0x1d1eb00, 0xc421146060, 0x1d18c00, 0xc4238a2480, 0x0, 0xc42003c828, 0x135ce, 0xc421146080, 0x20)
        github.com/blevesearch/bleve/search/query/boolean.go:104 +0x337
github.com/blevesearch/bleve/search/query.(*ConjunctionQuery).Searcher(0xc42136e2c0, 0x1d1eb00, 0xc421146060, 0x1d18c00, 0xc4238a2480, 0xc42003c800, 0x4359b, 0x20, 0x1196820, 0xc4203a2c01)
        github.com/blevesearch/bleve/search/query/conjunction.go:58 +0xf4
github.com/blevesearch/bleve/search/query.(*BooleanQuery).Searcher(0xc423780880, 0x1d1eb00, 0xc421146060, 0x1d18c00, 0xc4238a2480, 0x0, 0xc4238a22d8, 0xc42003c928, 0x2d053, 0xc42003c8f8)
        github.com/blevesearch/bleve/search/query/boolean.go:115 +0x27d
github.com/blevesearch/bleve/search/query.(*ConjunctionQuery).Searcher(0xc42136e320, 0x1d1eb00, 0xc421146060, 0x1d18c00, 0xc4238a2480, 0xc4209ae000, 0xc4206880f0, 0x10e9000, 0xc42061e480, 0x16)
        github.com/blevesearch/bleve/search/query/conjunction.go:58 +0xf4
github.com/blevesearch/bleve.(*indexImpl).SearchInContext(0xc422eda070, 0x2e48000, 0xc420012430, 0xc422ed5110, 0x0, 0x0, 0x0)
        github.com/blevesearch/bleve/index_impl.go:373 +0x227
github.com/blevesearch/bleve.MultiSearch.func1(0x1d1fd40, 0xc422eda070, 0xc422ed5110)
        github.com/blevesearch/bleve/index_alias_impl.go:458 +0x154
created by github.com/blevesearch/bleve.MultiSearch
        github.com/blevesearch/bleve/index_alias_impl.go:465 +0x1c4
```

mschoch:
A couple of things:

1) I'm not strictly opposed to adding a recover here for the case of attempting to return partial results in the event of a panic on one of the members of the alias...
2) ...but this is clearly a case where we shouldn't panic.  There is a bug to be fixed that we can't just paper over
3) I don't want to introduce a new package at this late stage for error handling, and if we do this isn't the one I'd choose.
4) Can you provide git SHA of bleve you're using (line number 183 in reader.go is now whitespace) and also the version of Go you're using.

Thanks

slavikm:
Go version 1.7.x (mine is 1.7.4)
Bleve - c351931701d7d4aa58e41cb735a68400a4e88b29
We need to translate the recover interface somehow - we can create a local function in the error.go but to get the most of the error it is useful to capture the stack trace as well.

aviadl:
@mschoch any updates on this issue ?
Thanks

mschoch:
To be honest I haven't thought about it much, but I'm still not keen on adding code to recover panics.

In general, the code should not panic, even if something unexpected happens, it should return an error, and in this case of MultiSearch, you should get partial results.

aviadl:
There is no argue that the code should not panic
The problem is, that using bleve as a library for a production environment, we want to make sure that the product is up and running even if some extreme use case has failed 
Go architecture prevents it if you don't protect your gonroutines
So I would say, if you provide a library that you expect to be used in production, you need to be protective ...
My 2c

mschoch:
We do use bleve in production.

Our philosophy is to try and fix any panics reported, so that you can trust that bleve will not panic.

If you can't trust bleve not to panic, and this affects your production application, you'll have to add recovery code outside of bleve anyway.  Adding a recover call in just this one place just happens to work in this case, but is somewhat arbitrary.  Another use not using MultiSearch might encounter the same panic, and not benefit from the recover.

aviadl:
Right, and we recover on any go routine spawned on our app level

meirwah:
@mschoch can you reconsider merging this patch?
We prefer not to fork bleve, but it seems that without this fix, we are forced to fork bleve, in order to make sure it will not crush our server in production...

mschoch:
There is still no compelling reason to add a recover in this one particular case.

In this library we try to fix any reports of a panic.

If you are concerned that bleve will crash your server in production, you need not fork it, you simply should have a recover set up in the goroutine calling into bleve.

I'm going to go ahead and close this issue, as I don't see us proceeding on this path.  However I will leave it open for comments so everyone can voice their opinion.

aviadl:
@mschoch the problem is that this particular place spawns another go routine.
So there is no way we can protect it in the application level (otherwise we would have done it already)
And this is (at least for my knowledge is currently the only place where bleve spawns a go routine.
If there are more, then they should be protected as well.

mschoch:
There are about 5-10 non-test usages of goroutines.  MultiSearch is unique in that it is a case where we actually wait for the goroutines to do some work, so its conceivable that we could convert the panic into an error and return it.  Most of the cases however are work done asynchronously with no convenient way, or even obvious place to return an error to.

I'm struggling with the idea that a library should guard every single goroutine with a recover statement, just so that the application never panics.  I believe that recover should definitely be used to protected against explicit calls to panic crossing a library boundary, but that is not the case here.

As a datapoint I have looked at the standard library, and I find a few hundred uses of goroutines, most of which have no recover.  Possibly you'll object, and say that the standard library is different, and achieves a level of quality higher than bleve, thus is need not have such protections.  I guess my only counter-argument is that they have achieved that quality by people reporting the crashes and fixing them.

So I guess I remain unpersuaded to make a change.  Panics caused by the runtime should crash the application, those bugs should be reported and fixed.

mschoch:
Also, although now unrelated to the discussion, the source of the original panic seems like it was likely related to this: https://github.com/blevesearch/bleve/issues/518

PR #517, opened 2017-01-14
fix the confusing variable name

minagawa-sho:
The variable has UnicodeTokenizer object, but its name is "whitespaceTokenizer".


mschoch:
Thanks

PR #520, opened 2017-01-18
improve performance of regular expression and wildcard queries

mschoch:
While researching an observed performance issue with wildcard
queries, it was observed that the LiteralPrefix() method on
the regexp.Regexp struct did not always behave as expected.

In particular, when the pattern starts with ^, AND involves
some backtracking, the LiteralPrefix() seems to always be the
empty string.

The side-effect of this is that we rely on having a helpful
prefix, to reduce the number of terms in the term dictionary
that need to be visited.

This change now makes the searcher enforce start/end on the term
directly, by using FindStringIndex() instead of Match().
Next, we also modified WildcardQuery and RegexpQuery to no
longer include the ^ and $ modifiers.

Documentation was also udpated to instruct users that they should
not include the ^ and $ modifiers in their patterns.

### DIFF
--- a/search/query/regexp.go
+++ b/search/query/regexp.go
@@ -33,7 +33,9 @@ type RegexpQuery struct {
 
 // NewRegexpQuery creates a new Query which finds
 // documents containing terms that match the
-// specified regular expression.
+// specified regular expression.  The regexp pattern
+// SHOULD NOT include ^ or $ modifiers, the search
+// will only match entire terms even without them.
 func NewRegexpQuery(regexp string) *RegexpQuery {
 	return &RegexpQuery{
 		Regexp: regexp,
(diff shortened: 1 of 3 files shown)


search/query/regexp.go
```diff
@@ -76,13 +78,13 @@ func (q *RegexpQuery) Validate() error {
...
-		if !strings.HasPrefix(actualRegexp, "^") {
-			actualRegexp = "^" + actualRegexp
+		if strings.HasPrefix(actualRegexp, "^") {
+			actualRegexp = actualRegexp[1:] // remove leading ^
 		}
-		if !strings.HasSuffix(actualRegexp, "$") {
-			actualRegexp = actualRegexp + "$"
+		if strings.HasSuffix(actualRegexp, "$") {
+			actualRegexp = actualRegexp[:len(actualRegexp)-1] // remove trailing $
 		}
```
steveyen: One thought is perhaps the user is looking for '$' (many folks are looking for U.S. dollars?), so wondering if there's some backslash edge case around this?  This probably hardly ever occurs, if at all.

mschoch: Actually I think its a reasonable concern.  In particular I remember one index encoding which put $ by itself at the end of terms.  Totally unrelated to this case, but still makes me nervous.

I think instead I'll remove the code that trims off $ from the end.  In the cases we know about the presence of $ does not interfere with LiteralPrefix().  I had removed it to handle it consistently with ^, but issue of possible escaping makes it trickier for sure.

PR #526, opened 2017-02-01
MB-20793: Validation for min/max/start/end params for numeric/date ra…

sreekanth-cb:
…nge facets

Improved the validations for date and numeric range queries for facets

@mschoch , @steveyen , please review the changes aiming to improve the range validations, 
[I am not a collaborator for the project & couldn't push my local branch to remote repo for a PR creation. Hence forked the original repo and creating a PR from there]

### DIFF
--- a/search.go
+++ b/search.go
@@ -116,7 +116,7 @@ func (fr *FacetRequest) Validate() error {
 	if nrCount > 0 && drCount > 0 {
 		return fmt.Errorf("facet can only conain numeric ranges or date ranges, not both")
 	}
-	rangeSet := false
+
 	if nrCount > 0 {
 		nrNames := map[string]interface{}{}
 		for _, nr := range fr.NumericRanges {
@@ -124,12 +124,9 @@ func (fr *FacetRequest) Validate() error {
 				return fmt.Errorf("numeric ranges contains duplicate name '%s'", nr.Name)
 			}
 			nrNames[nr.Name] = struct{}{}
-			if !rangeSet && (nr.Min != nil || nr.Max != nil) {
-				rangeSet = true
+			if nr.Min == nil && nr.Max == nil {
+				return fmt.Errorf("numeric range query must specify either min, max or both for range name '%s'", nr.Name)
 			}
-		}
-		if !rangeSet {
-			return fmt.Errorf("numeric range query must specify either min, max or both")
 		}
 
 	} else {
@@ -143,15 +140,10 @@ func (fr *FacetRequest) Validate() error {
 				return fmt.Errorf("date ranges contains duplicate name '%s'", dr.Name)
 			}
 			drNames[dr.Name] = struct{}{}
-			if !rangeSet {
-				start, end := dr.ParseDates(dateTimeParser)
-				if !start.IsZero() || !end.IsZero() {
-					rangeSet = true
-				}
+			start, end := dr.ParseDates(dateTimeParser)
+			if start.IsZero() && end.IsZero() {
+				return fmt.Errorf("date range query must specify either start, end or both for range name '%s'", dr.Name)
 			}
-		}
-		if !rangeSet {
-			return fmt.Errorf("date range query must specify either start, end or both")
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


steveyen:
> [I am not a collaborator for the project & couldn't push my local branch to remote repo for a PR creation. Hence forked the original repo and creating a PR from there]

Indeed.  That is the same kind of steps that I follow too for bleve, where I have my own fork to push up PR's.

search.go
```diff
@@ -105,26 +111,49 @@ type FacetRequest struct {
...
+			if _, ok := nrNames[nr.Name]; ok {
+				return fmt.Errorf("numeric ranges contains duplicate name '%s'", nr.Name)
+			}
+			nrNames[nr.Name] = struct{}{}
+			if !rangeSet && (nr.Min != nil || nr.Max != nil) {
+				rangeSet = true
+			}
+		}
+		if !rangeSet {
+			return fmt.Errorf("numeric range query must specify either min, max or both")
```
mschoch: I think this logic can miss cases where the ranges are missing.  For example, consider a case where there are 2 numeric ranges.  Initially rangeSet is false, we loop over the numeric ranges.  Let's say the first range does have a minimum set.  Now rangeSet is set to true.  Now we look at the second range, and let imagine it doesn't have either range set.  But, we won't even check that because !rangeSet is false now.  After exiting the loop, since !rangeSet is false, no error message is returned... but one of the ranges was improperly specified.  And I think the same problem exists for dates below.

Of course I could be reading it wrong, but to @steveyen point, additional tests would help make it clear that we verified these cases.

mschoch: Second, you asked me about integration tests, but I think we should first use some unit testing here to check that certain structures return the correct error from Validate().

sreekanth-cb: Yes Marty, I wasn't sure of all the valid facet nr/dr formats. So does it mean the above mentioned query would like below Or I unable to envision how  multiple numeric ranges works against a single field.
```
curl -XPOST -H "Content-Type: application/json" \
 http://127.0.0.1:9201/api/index/idx/query \
 -d '{
  "size": 10,
  "query": {
    "boost": 1,
    "query": "water"
  },
  "facets": {
    "abv": {
      "size": 5,
      "field": "abv",
      "numeric_ranges": [
        {
          "name": "high",
          "max1": 7
        },
        {
          "name": "low",
          "min1": 7
        }
      ]
    },
    "abv_second": {
      "size": 15,
      "field": "abv",
      "numeric_ranges": [
        {
          "name": "high",
          "max1": 7
        },
        {
          "name": "low",
          "min1": 7
        }
      ]
    }
  }
}'
```

sreekanth-cb: Oh I see, you mean to say that 
```
 "numeric_ranges": [
        {
          "name": "high"
        },
        {
          "name": "low",
          "min": 7
        }
      ]
``` .  
is an invalid query  ?  [I was thinking in the whole numeric_range [] array we have to specify the boundaries for any of the entries]

Comment (sreekanth-cb) [inline search.go]: Yes, I got it now.

File: search_test.go [hunk 2]
```diff
@@ -183,3 +184,81 @@ func TestUnmarshalingSearchResult(t *testing.T) {
...
+func TestFacetNumericDateRangeRequests(t *testing.T) {
+	var drMissingErr = fmt.Errorf("date range query must specify either start, end or both for range name 'testName'")
+	var nrMissingErr = fmt.Errorf("numeric range query must specify either min, max or both for range name 'testName'")
+	var drNrErr = fmt.Errorf("facet can only conain numeric ranges or date ranges, not both")
+	value := float64(5)
+
+	facet := NewFacetRequest("Date_Range_Success_With_StartEnd", 1)
+	facet.DateTimeRanges = make([]*dateTimeRange, 0, 1)
+	facet.DateTimeRanges = append(facet.DateTimeRanges, &dateTimeRange{Name: "testName", Start: time.Unix(0, 0), End: time.Now()})
+	if err := facet.Validate(); err != nil {
```
mschoch: As discussed in the meeting, instead of making repeated calls to Validate(), we should use the Go test pattern of defining test cases up front, with slice of anonymous structs, like:

https://github.com/couchbase/go-couchbase/blob/master/vbmap.go#L70-L77

And then have a for loop over those test cases like:
https://github.com/blevesearch/bleve/blob/master/search_test.go#L72

PR #527, opened 2017-02-03
refactor phrase search to be recursive

mschoch:
a more correct solution that will enable us to extend in two
important ways:

1) support slop
2) support multi-phrase

### DIFF
--- a/search/searcher/search_phrase.go
+++ b/search/searcher/search_phrase.go
@@ -126,9 +126,9 @@ func (s *PhraseSearcher) checkCurrMustMatch(ctx *search.SearchContext) *search.D
 	// one field, but we allow for this to not be the case
 	// but, we note that phrase constraints can only be satisfied within
 	// a single field, so we can check them each independently
-	for field := range s.currMust.Locations {
+	for field, tlm := range s.currMust.Locations {
 
-		f, rvtlm := s.checkCurrMustMatchField(ctx, field)
+		f, rvtlm := s.checkCurrMustMatchField(ctx, tlm)
 		if f > 0 {
 			freq += f
 			rvftlm[field] = rvtlm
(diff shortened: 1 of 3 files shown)


search/searcher/search_phrase.go
```diff
@@ -150,49 +150,68 @@ func (s *PhraseSearcher) checkCurrMustMatch(ctx *search.SearchContext) *search.D
...
+
+	// locations for this term
+	locations := tlm[car]
+	var rv []phrasePath
+	for _, loc := range locations {
+		// check each location against prev (nil treated as match, the initial case)
+		if prev == nil || (prev.Pos+float64(offset) == loc.Pos && prev.SameArrayElement(loc)) {
+			// this location works, add it to the path (but not for empty term)
+			px := append(p, &phrasePart{term: car, loc: loc})
+			rv = append(rv, findPhrasePaths(loc, cdr, tlm, 1, px)...)
```
steveyen: Still trying to grok this code, but wondering if offset+1 should be used instead of just 1 for this recursive call to findPhrasePaths()

search/searcher/search_phrase.go
```diff
@@ -150,49 +150,68 @@ func (s *PhraseSearcher) checkCurrMustMatch(ctx *search.SearchContext) *search.D
...
+// returns slice of paths, or nil if invocation did not find any successul paths
+func findPhrasePaths(prev *search.Location, phraseTerms []string, tlm search.TermLocationMap, offset int, p phrasePath) []phrasePath {
+
+	// no more terms
+	if len(phraseTerms) < 1 {
+		return []phrasePath{p}
+	}
+
+	car := phraseTerms[0]
+	cdr := phraseTerms[1:]
```
steveyen: car and cdr! :-)

search/searcher/search_phrase.go
mschoch: well, for the normal case it's one each time.  because we expect the next match to be 1 more than the previous.

it was offset+1 in the case where we skipped over a space, and we want to the next match to be 2 more than the previous (or 3, 4 etc if there are multiple gaps)

mschoch: I actually think this part of the code will change if I add support for slop, because then there will be like a total slop value that can be used up.  And each match uses up some slop (or not if its in the exact place you expect), until you run out of slop.

mschoch: Also, unlike the previous one where I wanted to walk you through it, this one I wanted you to see blind.  If it's not grokkable, or could be improved to be more grokkable, I'd like to find that.

steveyen: i see, so the recursive is driven more by advancing the loc (or prev parameter) search.Location every time

search/searcher/search_phrase.go
```diff
@@ -150,49 +150,68 @@ func (s *PhraseSearcher) checkCurrMustMatch(ctx *search.SearchContext) *search.D
...
-			}
-			// if we got here we didn't find a location match for this term
-			continue OUTER
-		}
-		// if we got here all the terms matched
-		freq++
-		search.MergeTermLocationMaps(rvtlm, crvtlm)
+func (s *PhraseSearcher) checkCurrMustMatchField(ctx *search.SearchContext, tlm search.TermLocationMap) (int, search.TermLocationMap) {
+	paths := findPhrasePaths(nil, s.terms, tlm, 1, nil)
+	rv := make(search.TermLocationMap, 0)
```
Shugyousha: This map can be initialised to the length of s.terms, I think.

mschoch: Yes, I think you're right.  I'll incorporate that in the next commit.

PR #532, opened 2017-02-10
move phrase search logic into phrase searcher

mschoch:
the logic of how a phrase search works should be an internal
detail of the phrase searcher.  further, these changes will
allow proper scoring of phrase matches, which require access
to the underlying searcher objects, which were hidden in the
previous approach.

### DIFF
--- a/search/searcher/search_phrase.go
+++ b/search/searcher/search_phrase.go
@@ -169,7 +169,12 @@ func (s *PhraseSearcher) checkCurrMustMatch(ctx *search.SearchContext) *search.D
 // satisfied, and these locations are returned.  otherwise 0 and either
 // a nil or empty TermLocationMap
 func (s *PhraseSearcher) checkCurrMustMatchField(ctx *search.SearchContext, tlm search.TermLocationMap) (int, search.TermLocationMap) {
-	paths := findPhrasePaths(0, nil, s.terms, tlm, nil, 0)
+	// temporarily turn flat terms []sting into [][]string
+	terms := make([][]string, len(s.terms))
+	for i, term := range s.terms {
+		terms[i] = []string{term}
+	}
+	paths := findPhrasePaths(0, nil, terms, tlm, nil, 0)
 	rv := make(search.TermLocationMap, len(s.terms))
 	for _, p := range paths {
 		p.MergeInto(rv)
@@ -203,7 +208,7 @@ func (p phrasePath) MergeInto(in search.TermLocationMap) {
 //     this is the primary state being built during the traversal
 //
 // returns slice of paths, or nil if invocation did not find any successul paths
-func findPhrasePaths(prevPos uint64, ap search.ArrayPositions, phraseTerms []string, tlm search.TermLocationMap, p phrasePath, remainingSlop int) []phrasePath {
+func findPhrasePaths(prevPos uint64, ap search.ArrayPositions, phraseTerms [][]string, tlm search.TermLocationMap, p phrasePath, remainingSlop int) []phrasePath {
 
 	// no more terms
 	if len(phraseTerms) < 1 {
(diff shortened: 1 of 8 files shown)


search/searcher/search_phrase.go
```diff
@@ -31,7 +32,24 @@ type PhraseSearcher struct {
...
 }
 
-func NewPhraseSearcher(indexReader index.IndexReader, mustSearcher *ConjunctionSearcher, terms []string) (*PhraseSearcher, error) {
+func NewPhraseSearcher(indexReader index.IndexReader, terms []string, field string, options search.SearcherOptions) (*PhraseSearcher, error) {
+	options.IncludeTermVectors = true
+	termSearchers := make([]search.Searcher, 0)
+	for _, term := range terms {
+		if term != "" {
+			ts, err := NewTermSearcher(indexReader, term, field, 1.0, options)
+			if err != nil {
```
steveyen: When an error's encountered, I'm thinking there should be a Close()'ing of all the child termSearchers that have been gathered so far.

search/searcher/search_phrase.go
```diff
@@ -31,7 +32,24 @@ type PhraseSearcher struct {
...
+				return nil, fmt.Errorf("phrase searcher error building term searcher: %v", err)
+			}
+			termSearchers = append(termSearchers, ts)
+		}
+	}
+
+	mustSearcher, err := NewConjunctionSearcher(indexReader, termSearchers, options)
+	if err != nil {
```
steveyen: Probably here too.  Maybe the original code had this issue, too.

search/searcher/search_phrase.go
```diff
@@ -169,7 +169,12 @@ func (s *PhraseSearcher) checkCurrMustMatch(ctx *search.SearchContext) *search.D
 // satisfied, and these locations are returned.  otherwise 0 and either
 // a nil or empty TermLocationMap
 func (s *PhraseSearcher) checkCurrMustMatchField(ctx *search.SearchContext, tlm search.TermLocationMap) (int, search.TermLocationMap) {
-	paths := findPhrasePaths(0, nil, s.terms, tlm, nil, 0)
+	// temporarily turn flat terms []sting into [][]string
```
steveyen: minor typo on the []string comment

search/searcher/search_phrase.go
mschoch: ya, we always had that problem, suspect we'll need your deferred closure trampoline megajump trick

search/searcher/search_phrase.go
```diff
@@ -169,7 +169,12 @@ func (s *PhraseSearcher) checkCurrMustMatch(ctx *search.SearchContext) *search.D
...
+	terms := make([][]string, len(s.terms))
```
steveyen: didn't look deeply on this, but an optimization thought... wondering if this can be pulled into a pre-computed or reused field of the PhraseSearcher to avoid a recomputation/re'allocation on every Next()

mschoch: this is temporary and goes away in the next commit, i just broke it out so it was easier to follow

mschoch:
ready for review when you have a chance @steveyen

search/searcher/search_phrase.go
```diff
@@ -196,7 +258,7 @@ func findPhrasePaths(prevPos uint64, ap search.ArrayPositions, phraseTerms []str
 	cdr := phraseTerms[1:]
 
 	// empty term is treated as match (continue)
-	if car == "" {
+	if len(car) == 0 || len(car) == 1 && car[0] == "" {
```
steveyen: I googled the operator precedence spec to double check this, and it's right.

On the other hand, I indeed had to look it up, so perhaps some parens might help the drive-by reader here (unless if go fmt "helpfully" strips them out for you).

mschoch: yep, i agree – i'm typically in favor of more parentheses.  i think i added this in at the last minute and didn't really look at it very carefully.

PR #534, opened 2017-02-14
Apache2 license badge

tleyden:
Moar badges


mschoch:
I kind of want to say no because github already identifies our project as Apache 2 at the top.  But, I also feel like you're making fun of all our badges, so therefore we'll go ahead and add one more.

tleyden:
How about a metabadge that assigns repositories a "badge score"?

- Lines of code to badge ratio
   - 10 badges for a few hundred lines of code is just ridiculous.  But 100K lines of code and not a single badge is just as bad.
- Badge-to-Github redundancy rating
   - This PR would get dinged a bit, since the information is already present on the github page.   Also I see badges like "100% Pure Java" which is also redundant to info in github.
- Badge-to-Badge redundancy rating
   - Do you really need 10 badges related to code metrics?
- Missing badges
   - If you have no code coverage badge, you're automatically downloaded to a C.  No build badge either?  D.
- Frivolous badges
   - I really don't care if you wrote the code with `vim`, `spacemacs`, or `med`
- Color clashing
   - Red badges and orange badges just don't go together
- Broken badges
   - Immediate F, regardless of all other badge states.  Drone.io's implosion just broke a ton of badges .. it's a badge massacre

Issue #283, opened 2015-11-19
Consider renaming bleve.Classifier Type() method


mmindenhall:
I have multiple struct types I'm trying to index that have `Type` fields.  This results in a compiler error when I implement the `bleve.Classifier` interface:

```
type MyStruct has both field and method named Type
```

How about something like this to avoid namespace collision:

``` go
type Classifier interface {
    BleveType() string
}
```

mmindenhall:
Actually it would be bad to just rename the method, as that would break indexing for current implementers of the interface.

How about adding an additional interface, and trying both to determine the type.  E.g.:

```go
// (Deprecated:  use BleveClassifier).  A Classifier is an interface describing any object
// which knows how to identify its own type.
type Classifier interface {
	Type() string
}

// A BleveClassifier is an interface describing any object
// which knows how to identify its own document type.
type BleveClassifier interface {
	BleveType() string
}
```
Then within the IndexMapping.determineType method, add a check for the new type.

pmezard:
FWIW, you can work around it using IndexMapping.TypeField. See Index interface documentation for details about how the document type is resolved.

mmindenhall:
Yes...that's what I ended up doing.  It was that or change my field names from Type to something else.

mschoch:
Yeah, I'd like to think about this and get additional input from others.  I had originally chosen the names Classifier and Type because the idea was that we index your structures, and we don't force you to clutter them with a lot of Bleve specific stuff.

But, conflicting with a Type field is a real issue to consider.  Another option would be to offer both interfaces, without explicitly deprecating Classifier.

I'll ask around today and try to get additional input, also @pmezard do you have opinion on this?

pmezard:
I am afraid my opinion is to avoid adding more magic, but that is not constructive.
If I had to design the API from scratch, I would do something explicit like:
```
func Index(kind string, id string, doc interface{}) error
```
and get rid of the Classifier, TypeField, etc.

I have no strong opinion about adding a BleveType(), clearly Type() is too generic.

pdf:
What about running with @pmezard's option by providing an additional index method, eg:
```go
func IndexKind(kind string, id string, doc interface{}) error
```
It's not ideal, but it does at least provide a method for people to side-step the issue where other options are infeasible.

pdf:
See f9354ec on my fork for a POC, still a breaking change (though not for most users) with the modification of the index interface, but pretty minimal surface area.

PR #535, opened 2017-02-14
Remove document type determination and add kind to index function

sokolovstas:
This pull request improved #283


mschoch:
Thanks,

1)  This is a rather big change to the API, so I need to get some feedback from others in the community to decide what we want to do.  I will email the google-group about this now and solicit feedback/comments.

2)  There are unit tests failing in the http (and possibly other) packages.  We need to get those fixed before it can move forward.

3) You would need to sign CLA.  See http://www.blevesearch.com/docs/Contributing/

mschoch:
Could you move the Russian analyzer change into a separate PR.  The review and decision process for these two changes should be independent.  Thanks.

sokolovstas:
Moved to #539

PR #537, opened 2017-02-15
Update bleve cmd

mschoch:


cmd/bleve/cmd/bulk.go
```diff
@@ -31,6 +31,9 @@ var bulkCmd = &cobra.Command{
 	Use:   "bulk [index path] [data paths ...]",
 	Short: "bulk loads from newline delimited JSON files",
 	Long:  `The bulk command will perform batch loading of documents in one or more newline delimited JSON files.`,
+	Annotations: map[string]string{
+		canMutateBleveIndex: "true",
```
steveyen: cool! - didn't know you could initialize a map like this in golang with a variable key

PR #538, opened 2017-02-16
Added Russian analyzer with snowball stemmer

sokolovstas:


sokolovstas:
Is there any problems to merge this?

mschoch:
Sorry for the delay.  What happened is that in the interim we've made progress getting a Go generator for the official snowballstem project.  See:

https://github.com/snowballstem/snowball/pull/57

The advantage of this is that we can get pure Go support for a much longer list of languages from only one new dependency (the snowball runtime support package).

Unfortunately, it has taken longer to get those changes merged upstream.  My hope is that we can get those merged, then make some tweaks to the PR to use that stemmer, and around the same time, re-add support for a long list of languages that are currently only supported with cgo in the blevex package.

sokolovstas:
@mschoch found small typo in your go generator commit in readme

mschoch:
Thanks for taking a look at it.  I just pushed a commit to fix the typo.

sokolovstas:
Implemented using snowball go generator #604

PR #539, opened 2017-02-17
Remove document type determination and add kind to index function

sokolovstas:
This pull request improved #283

### DIFF
--- a/examples_test.go
+++ b/examples_test.go
@@ -77,11 +77,11 @@ func ExampleIndex_indexing() {
 	}{Name: "great nameless one", Created: time.Now(), Age: 25}
 
 	// index some data
-	err = exampleIndex.Index("document id 1", data)
+	err = exampleIndex.Index("document id 1", "default", data)
 	if err != nil {
 		panic(err)
 	}
-	err = exampleIndex.Index("document id 2", data2)
+	err = exampleIndex.Index("document id 2", "default", data2)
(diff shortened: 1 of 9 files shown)


joeblew99:
What's the status of this ? It looked good. It will at least for me, make building large code bases with bleve easier

mschoch:
@joeblew99 the status is that we haven't made a final decision.  But I can update everyone on my thinking.

1.  I emailed the google group and tweeted for input.  I got exactly 1 reply, which suggests to me that either no one cares, or this change isn't a big deal for people.

2.  This change will break EVERY single bleve program every written.  I'm really not comfortable making this change at this time.  I do think it's something we should consider for bleve 2.0 when we introduce the next set of breaking changes.

joeblew99:
I can certainly see why v2 makes sense now.
Am fine with it going into v2

joeblew99:
Would it be possible that what is up for discussion for v2 goes into a Google Doc so use developers can help feedback ?

mschoch:
I think we have pretty strong agreement on this change going into v2.  The IndexMapping is way too complicated, and making the type detection manual (if we even keep types as a concept) is an obvious way to reduce the complexity.

I don't have any ideas organized enough yet to start a google doc, but please feel free to post them to the Google Group, and we can open a document in the near future.

I'm going to close this for now as I'm cleaning up the project.

PR #541, opened 2017-02-20
Add Snapshot interface methods for moss child collections feature

hisundar:
moss store will soon support child collections. This change adds corresponding implementations of the new moss.Snapshot interface methods ChildCollectionNames and ChildCollectionSnapshot.

### DIFF
--- a/index/store/moss/lower.go
+++ b/index/store/moss/lower.go
@@ -102,12 +102,12 @@ type llStore struct {
 // llSnapshot represents a lower-level snapshot, wrapping a bleve
 // store.KVReader, and implements the moss.Snapshot interface.
 type llSnapshot struct {
-	llStore  *llStore // Holds 1 refs on the llStore.
-	kvReader store.KVReader
-
-	m              sync.Mutex // Protects fields that follow.
-	refs           int
+	llStore        *llStore // Holds 1 refs on the llStore.
+	kvReader       store.KVReader
 	childSnapshots map[string]*llSnapshot
+
+	m    sync.Mutex // Protects fields that follow.
+	refs int
 }
 
 // llIterator represents a lower-level iterator, wrapping a bleve
--- a/index/store/moss/lower.go
+++ b/index/store/moss/lower.go
@@ -105,8 +105,9 @@ type llSnapshot struct {
 	llStore  *llStore // Holds 1 refs on the llStore.
 	kvReader store.KVReader
 
-	m    sync.Mutex // Protects fields that follow.
-	refs int
+	m              sync.Mutex // Protects fields that follow.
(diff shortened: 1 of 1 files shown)


hisundar:
I guess any implementation of the moss.Snapshot needs to also implement these methods. So far I could only find this llSnapshot which needed to be fixed in bleve, please let me know if there are other places that would need fixing.

steveyen:
Hmmm, another approach that would be non-breaking... don't add new methods to the moss.Snapshot interface, but instead add a brand new interface, like moss.SnapshotWithChildCollections (horrid name, I know), that contains the new ChildCollectionNames() / ChildCollectionSnapshot() methods.

Maybe moss.SnapshotParent?  Also, since a new interface kinda gives you a blank, new namespace, perhaps you can now have shorter method names?  Like, rename ChildCollectionNames() -to-> ChildNames()?  ChildCollectionSnapshot(childName) -to-> ChildSnapshot(childName)?

This thread probably belongs now over on some moss thread, but not sure where.

hisundar:
Sure, we could do that but won't that just end up adding a lot more changes to the consumer once it begins to use the feature? Because now all its Snapshot interface calls would need to be "ported" to use the new moss.SnapshotParent. Not to mention the changes within mossStore itself.
By just fixing up this one odd implementation of moss.Snapshot within bleve, we have automagically added new functionality without much regression while simultaneously making future feature migration simpler right?

steveyen:
It's a bit tricky -- need to sequence an interface change across two different projects.  The curse of library backwards compatibility!

steveyen:
Whoops - I misunderstood.  Spoke with Sundar offline and now I understand that this change doesn't need to be submitted simultaneously with a moss project change.  Instead, this change can go in first, without breaking anything.  Then later, when moss project picks up changes that add the child collection methods to the Snapshot interface, it all works, as long as we do the two-step dance in that order.

index/store/moss/lower.go
```diff
@@ -105,8 +105,9 @@ type llSnapshot struct {
 	llStore  *llStore // Holds 1 refs on the llStore.
 	kvReader store.KVReader
 
-	m    sync.Mutex // Protects fields that follow.
-	refs int
+	m              sync.Mutex // Protects fields that follow.
+	refs           int
+	childSnapshots map[string]*llSnapshot
```
mschoch: It looks lik childSnapshots is in the same paragraph (not separated by newline) of the section with mutex labeled "protects fields that follow", but the operations that use this field do not acquire the mutex.

hisundar: Thanks let me move the field above the mutex definition.

PR #543, opened 2017-02-23
improve query string compatibility

mschoch:
1) disjunction and conjunction queries now support a
"query string mode".  By default they do not operate
in this mode.  When in this mode, any disjunct/conjunct
which evaluates to MatchNone searcher, will be removed
from the disjunction/conjunction.  If the query ends
up with NO conjuncts/disjuncts, it will itself
return the MatchNone seacher.

2) boolean query also supports a query string mode.  when in
this mode, the Must, Should and MustNot searchers are all put
into query string mode.

3) rewriting of negation only queries (like -foo) now take into
account the rewriting rules above, and those are handled first.
this means that we rewrite correctly in case of +stoword -foo

4) the empty query string is now valid, and returns 0 hits.
previously this was considered a validation error.

### DIFF
--- a/search/query/conjunction.go
+++ b/search/query/conjunction.go
@@ -24,8 +24,9 @@ import (
 )
 
 type ConjunctionQuery struct {
-	Conjuncts []Query `json:"conjuncts"`
-	BoostVal  *Boost  `json:"boost,omitempty"`
+	Conjuncts       []Query `json:"conjuncts"`
+	BoostVal        *Boost  `json:"boost,omitempty"`
+	queryStringMode bool
 }
 
 // NewConjunctionQuery creates a new compound Query.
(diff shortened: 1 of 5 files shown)


search/query/conjunction.go
```diff
@@ -52,10 +53,10 @@ func (q *ConjunctionQuery) AddQuery(aq ...Query) {
 }
 
 func (q *ConjunctionQuery) Searcher(i index.IndexReader, m mapping.IndexMapping, options search.SearcherOptions) (search.Searcher, error) {
-	ss := make([]search.Searcher, len(q.Conjuncts))
-	for in, conjunct := range q.Conjuncts {
+	ss := make([]search.Searcher, 0, len(q.Conjuncts))
+	for _, conjunct := range q.Conjuncts {
 		var err error
-		ss[in], err = conjunct.Searcher(i, m, options)
+		sr, err := conjunct.Searcher(i, m, options)
```
steveyen: Might not need line 58 anymore due to the := on line 59?

PR #544, opened 2017-02-24
add ability to perform exact numeric matches

mschoch:


search/query/query_string.y
```diff
@@ -127,7 +127,12 @@ tSTRING tCOLON tSTRING tTILDE {
 tNUMBER {
 	str := $1
 	logDebugGrammar("STRING - %s", str)
-	q := NewMatchQuery(str)
+	q1 := NewMatchQuery(str)
+	val, _ := strconv.ParseFloat($1, 64)
```
steveyen: If err handling/propagation supported in these foo.y files?  I guess ParseFloat() might be able to return a range/syntax error, even if it looks numeric'ish from the parser's point of view...

https://golang.org/src/strconv/atof.go?s=11079:11134#L523

mschoch: ya, we have this problem throughout the file though.  i'll add another commit with that, plus we need new category of tests for it

steveyen: yep, no prob - not a priority either, imho

PR #546, opened 2017-03-01
Store abort close

sreekanth-cb:
@mschoch , @steveyen , Attempt to make configurable CloseEx/Abort for mossStore.
[The store level close would happen twice due to the following collection close which is there in the original code. But by adding an extra check at the Store level close, we may avoid code level regressions(nil dereferencing) on it, but not sure whether thats  a hack or a proper way to do it]
Reference: http://review.couchbase.org/#/c/74385/

### DIFF
--- a/index/store/moss/store.go
+++ b/index/store/moss/store.go
@@ -41,7 +41,8 @@ type Store struct {
 	llstore store.KVStore // May be nil.
 	llstats statsFunc     // May be nil.
 
-	s *stats
+	s      *stats
+	config map[string]interface{}
 }
 
 type statsFunc func() map[string]interface{}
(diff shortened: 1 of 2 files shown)


steveyen:
side note: this change should go in (if it's approved / haven't looked deeply yet) only after this change in moss gets through first... http://review.couchbase.org/#/c/74385/

mschoch:
The CI build is failing because the new methods of moss are missing.  I know that PR was merged, but here we also have to update the vendor SHA.

See: https://github.com/blevesearch/bleve/blob/master/vendor/manifest#L28-L35

mschoch:
CI is now failing because of an unchecked error:

index/store/moss/store.go:181:14:	s.CloseEx(moss.StoreCloseExOptions{Abort: true})

If this truly is a case where you explicitly want to ignore the error, you must use this syntax:

_ = funcReturningErrorWeWantToIgnore()

That is the signal to the errcheck tool that we know the method returns error, but we're not ignoring it by accident.  But in general, the purpose is to ensure we check/return errors.

sreekanth-cb:
Thanks a lot @mschoch ,  I didn't know the manifest & couldn't figure out immediately on the second errChk tool failure.

mschoch:
This is OK for me now, but I'd like @steveyen  to review as well.

Also for the future @sreekanth-cb, github now supports requesting code review on PRs.  In the upper-right hand corner of this page is a section for reviewers, where you can add Steve and myself in the future.

index/store/moss/store.go
```diff
@@ -167,12 +168,21 @@ func New(mo store.MergeOperator, config map[string]interface{}) (
...
 		llstore: llStore,
 		llstats: llStats,
+		config:  config,
 	}
 	rv.s = &stats{s: &rv}
 	return &rv, nil
 }
 
 func (s *Store) Close() error {
+	if v, ok := s.config["mossAbortCloseEnabled"]; ok && v.(bool) == true {
```
steveyen: Probably need to do a second, additional if-check instead of doing the cast to bool on a single line.

The reason why is a failed cast can panic() the process / evil input can take down the server.  Quickie test of that here... https://play.golang.org/p/ErHjReFC72

PR #548, opened 2017-03-04
fix perf regression, unnecessarily loading backindex

mschoch:


search/collector/topn.go
```diff
@@ -138,11 +138,14 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
 var sortByScoreOpt = []string{"_score"}
 
 func (hc *TopNCollector) collectSingle(ctx *search.SearchContext, reader index.IndexReader, d *search.DocumentMatch) error {
+	var err error
 
 	// visit field terms for features that require it (sort, facets)
-	err := hc.visitFieldTerms(reader, d)
-	if err != nil {
-		return err
+	if len(hc.neededFields) > 0 {
```
steveyen: wondering if also need to check if facetBuilder != nil? or is that covered as part of the neededFields?

mschoch: nope, if you are sorting, we need to do this even if facet builder is nil

mschoch: needed fields is the total list of fields required for sorting and facets, if it is empty, no lookup is needed.

PR #558, opened 2017-03-28
Added name to copyright notice

MTecknology:
I noticed these while looking through licensing of this project.
(everything else looked correct)


mschoch:
Thanks

PR #565, opened 2017-03-31
refactor searchers

mschoch:
- TermSearcher has alternate constructor if term is []byte, this can avoid
  copying in some cases.  TermScorer updated to accept []byte term. Also
  removed a few struct fields which were not being used.

- New MultiTermSearcher searches for documents containing any of a list of
  terms.  Current implementation simply uses DisjunctionSearcher.

- Several other searcher constructors now simply build a list of terms and
  then delegate to the MultiTermSearcher
  - NewPrefixSearcher
  - NewRegexpSearcher
  - NewFuzzySearcher
  - NewNumericRangeSearcher

- NewGeoBoundingBoxSearcher and NewGeoPointDistanceSearcher make use of
  the MultiTermSearcher internally, and follow the pattern of returning
  an existing search.Searcher, as opposed to their own wrapping struct.

- Callback filter functions used in NewGeoBoundingBoxSearcher and
  NewGeoPointDistanceSearcher have been extracted into separate functions
  which makes the code much easier to read.

### DIFF
--- a/search/searcher/search_term.go
+++ b/search/searcher/search_term.go
@@ -22,12 +22,9 @@ import (
 
 type TermSearcher struct {
 	indexReader index.IndexReader
-	term        string
-	field       string
 	reader      index.TermFieldReader
 	scorer      *scorer.TermQueryScorer
 	tfd         index.TermFieldDoc
-	options     search.SearcherOptions
 }
 
 func NewTermSearcher(indexReader index.IndexReader, term string, field string, boost float64, options search.SearcherOptions) (*TermSearcher, error) {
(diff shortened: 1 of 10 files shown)


search/searcher/search_term.go
```diff
@@ -40,12 +37,27 @@ func NewTermSearcher(indexReader index.IndexReader, term string, field string, b
...
+	scorer := scorer.NewTermQueryScorer([]byte(term), field, boost, count, reader.Count(), options)
+	return &TermSearcher{
+		indexReader: indexReader,
+		reader:      reader,
+		scorer:      scorer,
+	}, nil
+}
+
+func NewTermSearcherBytes(indexReader index.IndexReader, term []byte, field string, boost float64, options search.SearcherOptions) (*TermSearcher, error) {
+	reader, err := indexReader.TermFieldReader([]byte(term), field, true, true, options.IncludeTermVectors)
```
steveyen: extra []byte cast not needed here?

mschoch:
good catch on the extra []byte() conv, apparently my push -f removes that comment

PR #566, opened 2017-04-01
introduce new query TermRange

mschoch:
The term range query is not often used in full-text queries, but
can be useful when filtering on keyword indexed text terms in
the index.

The JSON syntax to do a TermRange query is the same as for
NumericRange, but the min/max values must be string and not
float64.

### DIFF
--- a/search/searcher/search_term_range.go
+++ b/search/searcher/search_term_range.go
@@ -1,0 +1,75 @@
+//  Copyright (c) 2017 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


search/searcher/search_term_range.go
```diff
@@ -0,0 +1,73 @@
...
+	// find the terms with this prefix
+	fieldDict, err := indexReader.FieldDictRange(field, min, rangeMax)
+	if err != nil {
+		return nil, err
+	}
+
+	var terms []string
+	tfd, err := fieldDict.Next()
+	for err == nil && tfd != nil {
+
```
steveyen: obsessively minor thing (or, if you're leaving whitespace to see if i'm paying attention), i'd remove line 54 whitespace

search/searcher/search_term_range.go
```diff
@@ -0,0 +1,73 @@
...
+		terms = append(terms, tfd.Term)
+		tfd, err = fieldDict.Next()
+	}
+
```
steveyen: wondering if we bailed from the loop because err != nil, might as well return the err here?

search/searcher/search_term_range.go
mschoch: no, i think i had a log printf there that got removed, and i left the newline by accident

search/searcher/search_term_range.go
mschoch: yeah thats a bug

PR #578, opened 2017-04-11
Add new IndexAdvanced function

seiflotfy:
IndexAdvanced bypasses the mapping matching and adds the doc directly


mschoch:
@steveyen do you have thoughts on the naming? Part of me likes IndexDocument, as that is more descriptive of what it is, but IndexAdvanced implies that we don't really want people to do this...

PR #590, opened 2017-05-11
MB-23979 - perf: better json marshal/unmarshalling

sreekanth-cb:
Exporting struct fields to support custom json parsing


mschoch:
I thought we had discussed moving this change into cbft so that there wouldn't be any changes required in bleve.

PR #594, opened 2017-05-18
add support for BleveType() alternative for type detection

### LINKED ISSUE #283: Consider renaming bleve.Classifier Type() method

mschoch:
Many existing structs already have a Type field or method which
conflicts with the bleve Classifier interface.  To address this
without breaking existing applications, we introduce an
alternate BleveType() method which will be checked first.  The
interface describing this method is private, as it should never
need to be referenced outside this package.

fixes #283

### DIFF
--- a/index.go
+++ b/index.go
@@ -110,12 +110,15 @@ func (b *Batch) Reset() {
 // them.
 //
 // The DocumentMapping used to index a value is deduced by the following rules:
-// 1) If value implements Classifier interface, resolve the mapping from Type().
-// 2) If value has a string field or value at IndexMapping.TypeField.
+// 1) If value implements mapping.bleveClassifier interface, resolve the mapping
+//    from BleveType().
+// 2) If value implements mapping.Classifier interface, resolve the mapping
+//    from Type().
+// 3) If value has a string field or value at IndexMapping.TypeField.
 // (defaulting to "_type"), use it to resolve the mapping. Fields addressing
 // is described below.
-// 3) If IndexMapping.DefaultType is registered, return it.
-// 4) Return IndexMapping.DefaultMapping.
+// 4) If IndexMapping.DefaultType is registered, return it.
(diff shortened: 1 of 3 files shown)


index.go
```diff
@@ -110,12 +110,15 @@ func (b *Batch) Reset() {
 // them.
 //
 // The DocumentMapping used to index a value is deduced by the following rules:
-// 1) If value implements Classifier interface, resolve the mapping from Type().
-// 2) If value has a string field or value at IndexMapping.TypeField.
+// 1) If value implements mapping.bleveClassifier interface, resolve the mapping
+//    from BleveType().
+// 2) If value implements mapping. Classifier interface, resolve the mapping
```
steveyen: extra space here in mapping.Classifier

mapping/mapping.go
```diff
@@ -22,12 +22,21 @@ import (
...
-// A Classifier is an interface describing any object
-// which knows how to identify its own type.
+// A Classifier is an interface describing any object which knows how to
+// identify its own type.  Alternatively, if a struct already has a Type
+// field or method in conflict, one can use BleveType instead.
 type Classifier interface {
 	Type() string
 }
 
+// A BleveClassifier is an interface describing any object which knows how
```
steveyen: I guess bleveClassifier interface is private, so lowercase.

Neat trick with the private interface, btw.

mapping/mapping.go
```diff
@@ -22,12 +22,21 @@ import (
...
+// to identify its own type.  This is introuced as an alternative to the
```
steveyen: sp... introduced

PR #597, opened 2017-05-25
Add comments to Location struct

### LINKED ISSUE #596: Any way to determine which subdocuments for hit locations?
Hello, thank you for Bleve!

When using IncludeLocations and sub-documents, I haven't been able to find a way to determine which sub-document the Hit.Location applies to (or more generally which sub-document(s) triggered the hit, which could be useful to know even when not using locations).

Here's a contrived example: https://play.golang.org/p/-QKGWUQkfO

When run, it produces:

```
Field Emails.Subject = [horse cat monkey apple pear banana]
Field Id = thread1
Field Emails.Id = [email1 email2]
    Field Emails.Subject has term cat from 6 to 9 (Pos 2?)
```

What I'd like to know is that "cat" was found in the Subject of "email1", not "email2".

Currently the only way I see to do it is to go back and look at each Subject myself, checking positions 6-9 until I find which email has a matching Subject.

Based on the way Emails.Subject contains the subjects concatenated, I don't have high hopes that the data on which sub-document is accessible when constructing the Location, but since the returned positions are relative to the original strings I thought I'd double-check.

Thanks for any insights!

caseymrm:
Closes #596

### DIFF
--- a/search/search.go
+++ b/search/search.go
@@ -36,9 +36,14 @@ func (ap ArrayPositions) Equals(other ArrayPositions) bool {
 }
 
 type Location struct {
+	// Pos is the position of the term within the field, starting at 1
 	Pos            uint64         `json:"pos"`
+	
+	// Start and End are the character offsets of the term in the field
 	Start          uint64         `json:"start"`
 	End            uint64         `json:"end"`
+	
+	// ArrayPositions contains the positions of the term within each array of subdocuments.
 	ArrayPositions ArrayPositions `json:"array_positions"`
 }
 
--- a/search/search.go
+++ b/search/search.go
@@ -39,11 +39,11 @@ type Location struct {
 	// Pos is the position of the term within the field, starting at 1
 	Pos            uint64         `json:"pos"`
 	
-	// Start and End are the character offsets of the term in the field
+	// Start and End are the byte offsets of the term in the field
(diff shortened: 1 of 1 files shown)


search/search.go
```diff
@@ -36,9 +36,14 @@ func (ap ArrayPositions) Equals(other ArrayPositions) bool {
 }
 
 type Location struct {
+	// Pos is the position of the term within the field, starting at 1
 	Pos            uint64         `json:"pos"`
+	
+	// Start and End are the character offsets of the term in the field
```
mschoch: We should use `byte` not `character` here to be more precise.

search/search.go
```diff
@@ -36,9 +36,14 @@ func (ap ArrayPositions) Equals(other ArrayPositions) bool {
...
 	Start          uint64         `json:"start"`
 	End            uint64         `json:"end"`
+	
+	// ArrayPositions contains the positions of the term within each array of subdocuments.
```
mschoch: I'd prefer we just say "...positions of the term within any elements."  The reason is that it can be for simple values as well, not just subdocuments, and the term sub-documents seems to be used  in different ways by different people, so I try to avoid it.

mschoch:
Thanks!

PR #602, opened 2017-05-31
filter numeric range terms against the term dictionary

mschoch:
previously, all numeric terms required to implement a numeric
range search were passed to the disjunction query (possibly
exceeding the disjunction clause limit)

now, after producing the list of terms, we filter them against
the terms which actually exist in the term dictionary.  the
theory is that this will often greatly reduce the number of terms
and therefore reduce the liklihood that you would run into the
disjunction term limit in practice.

because the term dictionary interface does not have a seek API
and we're reluctant to add that now, i chose to do a binary
search of the terms, which either finds the term, or not. then
subsequent binary searches can proceed from that position,
since both the list of terms and the term dictionary are sorted.

### DIFF
--- a/search/searcher/search_numeric_range.go
+++ b/search/searcher/search_numeric_range.go
@@ -17,6 +17,7 @@ package searcher
 import (
 	"bytes"
 	"math"
+	"sort"
 
 	"github.com/blevesearch/bleve/index"
 	"github.com/blevesearch/bleve/numeric"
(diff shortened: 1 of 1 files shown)


search/searcher/search_numeric_range.go
```diff
@@ -63,6 +75,34 @@ func NewNumericRangeSearcher(indexReader index.IndexReader,
...
 }
 
+func filterCandidateTerms(indexReader index.IndexReader,
+	terms [][]byte, field string) (rv [][]byte, err error) {
+	rv = make([][]byte, 0)
+	fieldDict, err := indexReader.FieldDictRange(field, terms[0], terms[len(terms)-1])
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
```
steveyen: Here, on the one hand defer near the resource creation feels idiomatic go'ish.

On the other hand, I guess the rest of the function doesn't exit early, so moving this to the end can skip a defer

PR #608, opened 2017-06-06
MB-24680: Buffer up to 20 segments by default before compacting

hisundar:
Triggering full compactions on every single small update,
exacerbates the index building stall in moss.
To mitigate this, allow the file to buffer upto 20 segments, by default
before re-writing the file.
Standalone testing shows much lower blocking times without severe
impacts to query latencies.


mschoch:
In general we don't reference MB-XYZ in Bleve commits, since they don't mean anything to the Bleve community.

hisundar:
We do have positive numbers from stand alone testing. Let me send up another pull request once 2 more heavy DGM fixes are merged.

PR #611, opened 2017-06-09
Allow synchronous writing for moss and set mossStore default for cmd

### LINKED ISSUE #529: How to use 'moss' store with cmd/bleve?
The following works:
```
> ./bleve create -s boltdb boltdb.data
> ./bleve index boltdb.data/ -j ../../test/tests/basic/data/a.json
Indexing: a
> ./bleve query boltdb.data/ marty
1 matches, showing 1 through 1, took 205.996µs
    1. a (0.068614)
	name
		marty
 ```

However, replacing boltdb with moss yields an error.
```
> ./bleve create -s moss moss.data
> ./bleve index moss.data/ -j ../../test/tests/basic/data/a.json
Error: error opening bleve index: error parsing mapping JSON: unexpected end of JSON input
mapping contents:

Usage:
  bleve index [index path] [data paths ...] [flags]

Flags:
  -j, --json      Parse the contents as JSON, defaults true. (default true)
  -d, --keepDir   Keep the directory in the dodcument id, defaults false.
  -x, --keepExt   Keep the extension in the document id, defaults false.

error opening bleve index: error parsing mapping JSON: unexpected end of JSON input
mapping contents:
 ```

The moss store meta file did persist fine, though.

```
> cat moss.data/index_meta.json 
{"storage":"moss","index_type":"upside_down"}
```

bt:
Should be good to go now! 🚀 

This should fix #529 by allowing synchronous writes for moss.

With this PR and the example as provided by @simonz05:

```
$ ./bleve create -s moss moss.data

$ ./bleve index moss.data -j blevesearch/bleve/test/tests/basic/data/a.json
Indexing: a

$ ./bleve query moss.data marty
1 matches, showing 1 through 1, took 199.289µs
    1. a (0.068614)
	name
		marty
```

To enable synchronous writing, set mossPersistSync in KVConfig to true. This has been done automatically in the bleve command: https://github.com/bt/bleve/blob/3dbeb082216a91e0464ed67f962ade897dc0d054/cmd/bleve/cmd/create.go#L57


mschoch:
So there are a few problems with this:

1) if moss supports durable persistence, it should be the default, as that is what the bleve API contract specifies.  this means no code checking if we're using moss and toggling a flag, it should default to durable behavior, and advanced users have to turn it off if they dont need it

2) this implementation seems to only work if there is a single batch at a time, execute batch is settingup/resetting state.  the bleve API expects multiple concurrent calls to execute batch.

3) i don't think the approach of checking the stats counters will ever work with multiple concurrent batches, because the dirty bytes counter may never go to 0.

4) even if i'm wrong about 3, the correct place for this logic to exist is in moss, not in bleve

bt:
Hey @mschoch,

So I've tested this briefly with my app and it seems to work. However I don't have multiple calls to ExecuteBatch at the same time so I can't confirm that. 

I understand that this should probably go into mods but from what I can see, I don't see much activity and it doesn't look like anyone's actually actively working on solving a durable persistency issue. 

This was a suggestion based on code as used in tests for moss, so I based the code off that. There is no real other way of doing this without some serious refactoring of the moss code. 

Happy to hear your thoughts.

PR #622, opened 2017-08-07
Better Document.GoString

cristaloleg:


mschoch:
Thanks, could you update the commit message to show sample old/new output generated?

cristaloleg:
They're the same. I've compared old/new output in tests with diff - no changes.
My changes just reduce allocations and improve speed.
Am I misunderstood your question?

mschoch:
Ah, OK, I had assumed "better" meant, "more useful for debugging" as thats really the only time I ever use this output.

I'm not opposed to improving perf, but I'm curious do you have any use case where the performance of this method matters?

cristaloleg:
Sorry for unclear commit message.
Check this out: [How to efficiently concatenate strings in Go?](https://stackoverflow.com/a/1766304)

mschoch:
Right, I understand it's more efficient.  I'm saying it's in a method where clarity may be more important than efficiency.

In this case I'm fine with merging it as it isn't really any less clear, but I'm just pointing out that in real applications this method never gets called, so the gain is 0.

In order to merge this you need to do 2 things:

1) update commit message to say "more efficient" not "better"
2) please sign the CLA as described here: https://github.com/blevesearch/bleve/blob/master/CONTRIBUTING.md

cristaloleg:
I feel sorry for disturbing you.

mschoch:
Having accurate messages and everyone signing the CLA are requirements for the project.  If you change your mind let me know.

PR #644, opened 2017-11-23
Added support for IP field.

cv711:
Hey!

Since the discussion on the mailing list went a bit stale, I took a first stab on adding support for IP fields reusing most of the preexisting components. Hence a 128 bit IP is being split into 5 fields (4 uint32s to not loose any precision and one bool to distinguish v4 from v6).

I might also do some work on the Lexer to parse IP looking values.

Let me know what you think.

Regards,
Chris


cv711:
Any feedback on this one?

gilliek:
I'm also interested in the support of IP addresse type. Is there any chance for this PR to be merged?

titusjaka:
@mschoch any thoughts?

amnonbc:
@mschoch wrote on the mailing list on 3 Jan 2018, 21:09:01

I did look at the PR.  I think the reuse of existing fields is clever, but not really consistent with the rest of bleve.  I think it would be better to create a new IP field type (in bleve/document package) which supports a 128-bit number directly in a single field.  Then support for IPv4 is handled in the same way as lucene:

"IPv4 addresses are converted to IPv4-Mapped IPv6 Addresses: indexing 1.2.3.4 is the same as indexing ::FFFF:1.2.3.4."

The 128-bit field would work just like our existing numeric field (multiple terms indexed for a single value).  Then we support range queries on the 128-bit field, and CIDR queries are just bit-prefix queries which get converted to range queries.

--------

After a delay of 3 years, I have picked up this PR and followed Marty's advice.

See https://github.com/blevesearch/bleve/pull/1536 for the new version.

I think we can close this PR.

PR #652, opened 2017-12-13
mapping: Fix closestDocMapping selecting wrong mapping

damz:
With a document like:

```
{
  "foo": "value",
  "bar": {
    "foo": "value2",
    "baz": "value3"
  }
}
```

And a mapping like:

```
mapping.IndexDynamic = false
mapping.DefaultMapping = bleve.NewDocumentStaticMapping()
mapping.DefaultMapping.AddFieldMappingsAt("foo", bleve.NewTextFieldMapping())
```

The default mapper will incorrectly index `bar.foo` as a dynamic field, but will correctly skip `bar.baz`.

### DIFF
--- a/mapping/document.go
+++ b/mapping/document.go
@@ -179,6 +179,7 @@ OUTER:
 				continue OUTER
 			}
 		}
+		break
 	}
 	return current
 }
--- a/mapping/mapping_test.go
+++ b/mapping/mapping_test.go
@@ -991,3 +991,26 @@ func TestMappingForNilTextMarshaler(t *testing.T) {
 	}
 
 }
+
+func TestClosestDocDynamicMapping(t *testing.T) {
+	mapping := NewIndexMapping()
+	mapping.IndexDynamic = false
+	mapping.DefaultMapping = NewDocumentStaticMapping()
+	mapping.DefaultMapping.AddFieldMappingsAt("foo", NewTextFieldMapping())
+
+	doc := document.NewDocument("x")
+	err := mapping.MapDocument(doc, map[string]interface{}{
+		"foo": "value",
+		"bar": map[string]string{
+			"foo": "value2",
+			"baz": "value3",
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(doc.Fields) != 1 {
+		t.Fatalf("expected 1 field, got: %d", len(doc.Fields))
+	}
+}


mschoch:
Do you have a test case that illustrates the problem you've observed?

I'm reluctant to modify this function because the "findClosest..." was supposed to do exactly that, it was supposed to find the deepest "foo" from some starting point.  So, changing that will break other things.  Though I can also see how that isn't what you want in all cases.

damz:
@mschoch That doesn't sound right, but what I described above is wrong too. The reason `bar.foo` becomes indexed is that it considers `foo` as the closest mapping, and it turns out that its `Dynamic` property is `true` (which is the default), so it indexes it as a dynamic field.

This concept of "closest mapping" is strictly only used to figure out whether a given field should be dynamically mapped, which makes me think that the only mappings that should be considered for `bar.foo` are `bar` (if it exists) and the root mapping.

damz:
I updated the description above.

damz:
Without this fix, a mapping for `bar.qux` would match a `foo.bar.baz.qux` field... that cannot possibly be the expected behavior.

mschoch:
Thanks for clarifying, I was confusing this with another similarly named function.

I've created this function which seems to make a very simple attempt at confirming what you describe (it fails without the fix and passes with it):

```
func TestClosestDocMapping(t *testing.T) {

	mapping := NewIndexMapping()
	mapping.IndexDynamic = false
	mapping.DefaultMapping = NewDocumentStaticMapping()
	mapping.DefaultMapping.AddFieldMappingsAt("foo", NewTextFieldMapping())

	closestMapping := mapping.DefaultMapping.closestDocMapping("bar.foo")
	if closestMapping != mapping.DefaultMapping {
		t.Fatalf("expected to get default mapping %p, got %p",
			mapping.DefaultMapping, closestMapping)
	}
}
```

Do you agree?

damz:
@mschoch Yes, makes perfect sense. Alternatively, this would be testing the actual bug as described:

```go
func TestClosestDocDynamicMapping(t *testing.T) {
	mapping := NewIndexMapping()
	mapping.IndexDynamic = false
	mapping.DefaultMapping = NewDocumentStaticMapping()
	mapping.DefaultMapping.AddFieldMappingsAt("foo", NewTextFieldMapping())

	doc := document.NewDocument("x")
	err := mapping.MapDocument(doc, map[string]interface{}{
		"foo": "value",
		"bar": map[string]string{
			"foo": "value2",
			"baz": "value3",
		},
	})
	if err != nil {
		t.Fatal(err)
	}

	if len(doc.Fields) != 1 {
		t.Fatalf("expected 1 field, got: %d", len(doc.Fields))
	}
}
```

I'm happy to update the PR with either of these.

mschoch:
Yes, yours is a bit cleaner and not relying so heavily on the internals.  If you can add that to the PR, I'll merge.

Also can I ask you to sign the CLA, as described here: https://github.com/blevesearch/bleve/blob/master/CONTRIBUTING.md

damz:
@mschoch PR updated, CLA signed.

mschoch:
Thanks @damz

damz:
:+1: Thank you for this awesome library. We are currently looking into leveraging it a bit outside of its confort zone, as part of a new log indexing pipeline we are building.

PR #654, opened 2017-12-14
DocumentVisitFieldTerms Scorch implementation level1

sreekanth-cb:
Creating a fresh PR incorporating the review comments from the older PR, 
Yes @mschoch , I have verified that subsequent calls to the same snapshot doesn't recompute the data again.


mschoch:
This was merged even though we know there are integration issues with segment merging/removal.

PR #663, opened 2017-12-16

abhinavdangeti:
- Snapshot
- SnapshotPrevious
- SnapshotRevert

+ unit test

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -32,6 +32,11 @@ type segmentIntroduction struct {
 	persisted chan error
 }
 
+type snapshotRevert struct {
+	snapshot *IndexSnapshot
+	applied  chan error
+}
+
 func (s *Scorch) mainLoop() {
 	var notify notificationChan
 OUTER:
@@ -48,6 +53,12 @@ OUTER:
 
 		case next := <-s.introductions:
 			err := s.introduceSegment(next)
+			if err != nil {
+				continue OUTER
+			}
+
+		case revertTo := <-s.revertToSnapshots:
+			err := s.revertToSnapshot(revertTo)
 			if err != nil {
 				continue OUTER
 			}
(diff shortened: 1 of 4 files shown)


index/scorch/introducer.go
```diff
@@ -230,3 +241,50 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
...
+		epoch:    s.nextSnapshotEpoch,
+		refs:     1,
+	}
+	s.nextSnapshotEpoch++
+
+	// iterate through segments
+	for i, segmentSnapshot := range revertTo.snapshot.segment {
+		newSnapshot.segment[i] = &SegmentSnapshot{
+			id:         segmentSnapshot.id,
+			segment:    segmentSnapshot.segment,
```
steveyen: pls double-check the ref-counting here... suspecting you'll need to AddRef() on the shared segment here

index/scorch/introducer.go
```diff
@@ -230,3 +241,50 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
...
+
+func (s *Scorch) revertToSnapshot(revertTo *snapshotRevert) error {
+	// acquire lock
+	s.rootLock.Lock()
+
+	// prepare a new index snapshot, based on next snapshot
+	newSnapshot := &IndexSnapshot{
+		parent:   s,
+		segment:  make([]*SegmentSnapshot, len(revertTo.snapshot.segment)),
+		offsets:  make([]uint64, len(revertTo.snapshot.offsets)),
```
steveyen: as an optimization - i suspect you don't  actually need to copy the offsets here?  or copy the internal?  can just share the same instance, and let GC do all the hard work of releasing memory if unused?

index/scorch/snapshot_rollback.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/blevesearch/bleve/index/scorch/segment"
+)
+
+type indexSnapErr struct {
+	i *IndexSnapshot
+	e error
+}
+
+// Snapshot fetches the latest persisted IndexSnapshot.
+func (s *Scorch) Snapshot() (*IndexSnapshot, error) {
```
steveyen: The method name here is probably too general -- easy to inadvertently entice folks into thinking that they're getting the latest snapshot?  Perhaps LastPersistedSnapshot() is a clearer name of what the method's doing?

index/scorch/snapshot_rollback.go
```diff
@@ -0,0 +1,183 @@
...
+
+	return rv.i, rv.e
+}
+
+// SnapshotPrevious returns the next older, previous IndexSnapshot
+// based on the provided IndexSnapshot, allowing the application to
+// walk backwards into the history of a store at previous points in
+// time. A nil returned snapshot means no previous snapshots are
+// available.
+func (s *Scorch) SnapshotPrevious(is *IndexSnapshot) (*IndexSnapshot, error) {
```
steveyen: Same here with this method name?  Perhaps PreviousPersistedSnapshot()?

Also, this code here, at a first glance, looks strikingly similar to the new Snapshot() method that you've proposed above?  Maybe there's a way to consolidate, and only offer a single new method?
  Let's try... PreviousPersistedSnapshot(is *IndexSnapshot), but it allows the callers to pass in a nil value, which means "just give me the last persisted snapshot, if any"?

index/scorch/snapshot_rollback.go

index/scorch/snapshot_rollback.go

index/scorch/introducer.go
abhinavdangeti: Will do.

index/scorch/introducer.go
abhinavdangeti: Yeah I guess, I'll change this in the next patch.

PR #669, opened 2017-12-18

abhinavdangeti:
- PreviousPersistedSnapshot
- SnapshotRevert

+ unit test

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -37,6 +37,11 @@ type epochWatcher struct {
 	notifyCh notificationChan
 }
 
+type snapshotReversion struct {
+	snapshot *IndexSnapshot
+	applied  chan error
+}
+
 func (s *Scorch) mainLoop() {
 	var epochWatchers []*epochWatcher
 OUTER:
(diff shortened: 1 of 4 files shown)


index/scorch/snapshot_rollback.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+type indexSnapErr struct {
+	i *IndexSnapshot
+	e error
+}
+
+// PreviousPersistedSnapshot returns the next older, previous
+// IndexSnapshot based on the provided IndexSnapshot. If the provided
+// argument were nil, the last persisted indexsnapshot is returned.
```
steveyen: minor grammar / clarity tweak...

   argument is nil, the newest persisted IndexSnapshot is returned

index/scorch/introducer.go
```diff
@@ -241,3 +252,44 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 	// notify merger we incorporated this
 	close(nextMerge.notify)
 }
+
+func (s *Scorch) revertToSnapshot(revertTo *snapshotReversion) error {
```
steveyen: interestingly, this implementation is very general... which is nice.  As long as the caller can somehow get their hands on an IndexSnapshot, this method actually just writes (a copy of) that IndexSnapshot as the new root.

So, there's perhaps an opportunity to get a better name for this method / functionality...

It's almost like WriteSnapshot(*IndexSnapshot) error or AcceptSnapshot().  Naming is hard.

steveyen: ForceSnapshot()?  PushSnapshot()?

PR #670, opened 2017-12-19

abhinavdangeti:
- PreviousPersistedSnapshot
- SnapshotRevert

+ unit test


index/scorch/snapshot_rollback.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			indexSnapshot.epoch = snapshotEpoch
+			return indexSnapshot, nil
+		}
+
+		return nil, nil
+	}
+
+	indexSnapshot, err := getFromBolt()
```
steveyen: wondering why an internal helper func getFromBolt is needed?

PR #672, opened 2017-12-19

abhinavdangeti:
- PreviousPersistedSnapshot
- SnapshotRevert

+ unit test

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -37,6 +37,11 @@ type epochWatcher struct {
 	notifyCh notificationChan
 }
 
+type snapshotReversion struct {
+	snapshot *IndexSnapshot
+	applied  chan error
+}
+
 func (s *Scorch) mainLoop() {
 	var epochWatchers []*epochWatcher
 OUTER:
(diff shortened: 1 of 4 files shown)


index/scorch/introducer.go
```diff
@@ -241,3 +252,48 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 	// notify merger we incorporated this
 	close(nextMerge.notify)
 }
+
+func (s *Scorch) revertToSnapshot(revertTo *snapshotReversion) error {
+	defer close(revertTo.applied)
+
+	if revertTo.snapshot == nil {
+		return fmt.Errorf("Cannot revert to a nil snapshot")
```
steveyen: need to also send this error onto the revertTo.applied channel?

index/scorch/snapshot_rollback.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+
+			indexSnapshot.epoch = snapshotEpoch
+			return indexSnapshot, nil
+		}
+
+		return nil, nil
+	}
+
+	return getFromBolt()
```
steveyen: looks like a helper getFromBolt() func isn't really needed here?  might as well just make it a straight set of code?

index/scorch/introducer.go
abhinavdangeti: Done.

index/scorch/snapshot_rollback.go

PR #673, opened 2017-12-19

abhinavdangeti:
- PreviousPersistedSnapshot
- SnapshotRevert

+ unit test

### DIFF
--- a/index/scorch/snapshot_rollback.go
+++ b/index/scorch/snapshot_rollback.go
@@ -1,0 +1,120 @@
+//  Copyright (c) 2017 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package scorch
+
+import (
+	"bytes"
+	"log"
+
+	"github.com/blevesearch/bleve/index/scorch/segment"
+)
+
+// PreviousPersistedSnapshot returns the next older, previous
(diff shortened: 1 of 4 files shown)


index/scorch/snapshot_rollback.go
```diff
@@ -0,0 +1,120 @@
...
+	revert := &snapshotReversion{
+		snapshot: revertTo,
+		applied:  make(chan error),
+	}
+
+	s.revertToSnapshots <- revert
+
+	// block until this IndexSnapshot is applied
+	err := <-revert.applied
+	return err
```
steveyen: minor thing -- feels like this can be a 1-liner:

    return <-revert.applied

index/scorch/snapshot_rollback.go
```diff
@@ -0,0 +1,120 @@
...
+				continue
+			} else if comp == 0 {
+				// Found the entry that is being searched for,
+				// return a valid IndexSnapshot found in the next iteration.
+				returnNextSnapshot = true
+				continue
+			} else { // comp < 0
+				// Found an entry that is smaller than the one being searched for,
+				// return a valid IndexSnapshot found right away.
+				returnNextSnapshot = true
```
steveyen: the extra returnNextSnapshot var tracking is a bit harder to follow -- can be simplified to not need this extra var perhaps?  one attempt: i guess it might just want to try loading the snapshot if (pos == nil || pos < k)?

index/scorch/snapshot_rollback.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+		}
+
+		_, snapshotEpoch, err := segment.DecodeUvarintAscending(k)
+		if err != nil {
+			log.Printf("PreviousPersistedSnapshot:"+
+				" unable to parse segment epoch %x, continuing", k)
+			continue
```
steveyen: when there's an error, wondering instead of continuing and trying the next entry, whether the error handling if-tests should just bail and return the error?

index/scorch/snapshot_rollback.go

index/scorch/snapshot_rollback.go
abhinavdangeti: Hmm, then there's a few places we'll need to edit to reflect the same behavior - perhaps a separate change?

index/scorch/snapshot_rollback.go

PR #684, opened 2017-12-20
Adding a num_items_to_persist stat

abhinavdangeti:
+ For this stat's purpose, adding a new item to the
  IndexSnapshot's struct - call it numItems .. which
  is updated whenever a segment is added to it - be it
  an introduction or a merge.
+ Whenever the Stats API for scorch is invoked, it
  firstly calls an internal API to updateStats which
  applies the root's stats, thereby letting the Stats
  API return the most updated values.

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -16,6 +16,7 @@ package scorch
 
 import (
 	"fmt"
+	"sync/atomic"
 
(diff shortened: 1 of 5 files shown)


abhinavdangeti:
Yeah in upsideDown the stats holds a pointer to the index, so as to access stats directly, let me push up a new patch that does things the same way.

PR #688, opened 2017-12-21
Scorch specific stats

abhinavdangeti:


index/scorch/introducer.go
```diff
@@ -166,6 +172,10 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 	// swap in new index snapshot
 	rootPrev := s.root
 	s.root = newSnapshot
+
+	// Update the numItems count of the root
+	atomic.StoreUint64(&s.root.numItems, numItems)
```
steveyen: From a quick scan, the numItems looks immutable?  If so, then the sync/atomic isn't need to access it?

abhinavdangeti: Oh but this atomic is so that there isn't any data race in the setting and reading of this variable. The stats() APIs are the ones that read off this variable.

PR #691, opened 2017-12-22
Scorch specific stats

abhinavdangeti:


index/scorch/stats.go
```diff
@@ -21,21 +21,27 @@ import (
...
+	m["errors"] = atomic.LoadUint64(&s.errors)
 	m["analysis_time"] = atomic.LoadUint64(&s.analysisTime)
 	m["index_time"] = atomic.LoadUint64(&s.indexTime)
-	// m["term_searchers_started"] = atomic.LoadUint64(&i.termSearchersStarted)
-	// m["term_searchers_finished"] = atomic.LoadUint64(&i.termSearchersFinished)
-	// m["num_plain_text_bytes_indexed"] = atomic.LoadUint64(&i.numPlainTextBytesIndexed)
+	m["term_searchers_started"] = atomic.LoadUint64(&s.termSearchersStarted)
+	m["term_searchers_finished"] = atomic.LoadUint64(&s.termSearchersFinished)
+	m["num_plain_text_bytes_indexed"] = atomic.LoadUint64(&s.numPlainTextBytesIndexed)
+	m["num_items_to_persist"] = rootStats.numItems
```
mschoch: What is numItems a count of?  From the code it seems like a total count of all items (regardless if deleted or not and with no relation to persistence).  But, here we return it as num_items_to_persist, which I thought was a count of items that were not yet persisted.

index/scorch/scorch.go
```diff
@@ -303,6 +318,7 @@ func (s *Scorch) Reader() (index.IndexReader, error) {
 	rv := &Reader{root: s.root}
 	rv.root.AddRef()
 	s.rootLock.RUnlock()
+	atomic.AddUint64(&s.stats.termSearchersStarted, 1)
```
mschoch: termSearchersStarted/Finished is supposed to count TermFrequencyReaders, not IndexReaders

index/scorch/stats.go
abhinavdangeti: Shoot, I went with the assumption that the segments in the root snapshot are always unpersisted. I know exactly what I'll need to change this to, let me push up a new patch.

PR #697, opened 2017-12-26
Wait for rollback'ed snapshot to persist

abhinavdangeti:


index/scorch/snapshot_rollback.go
```diff
@@ -91,12 +91,22 @@ func (s *Scorch) PreviousPersistedSnapshot(is *IndexSnapshot) (*IndexSnapshot, e
 // only be passed an IndexSnapshot that came from the same store.
 func (s *Scorch) SnapshotRevert(revertTo *IndexSnapshot) error {
 	revert := &snapshotReversion{
-		snapshot: revertTo,
-		applied:  make(chan error),
+		snapshot:  revertTo,
+		applied:   make(chan error),
+		persisted: make(chan error, 1),
```
mschoch: Can you explain why we needed a buffered channel here and not anywhere else?

abhinavdangeti: Looking into it again, I guess we don't really need  a buffered channel here as the sender doesn't need to block when there is no empty slot available - there is always the one slot. I just followed the way segment introduction is done. Let me make this an unbuffered channel. 

Also I see that the persisted channel (in segment introduction) is used only when unsafeBatch is set to false, I'll apply that check here as well.

PR #698, opened 2017-12-27
Adding onEvent callback support for scorch

abhinavdangeti:
Event types:
- EventKindCloseStart
- EventKindClose
- EventKindMergerProgress
- EventKindPersisterProgress

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -19,6 +19,7 @@ import (
 	"log"
 	"os"
 	"sync/atomic"
+	"time"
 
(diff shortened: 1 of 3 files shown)


abhinavdangeti:
This change isn't complete yet. I will have some more changes here for actually tracking the memory consumption. This change was to support back pressure incase of high memory consumption.

Here are things that I will be adding alongside here:
- Memory consumption at a segment level.
- Aggregated memory usage across an index snapshot
- Event firing on segment introduction.

I will close this pull request for now, as its still work-in-progress. I will have another one up soon.

PR #700, opened 2017-12-27
phrase searcher don't allow advance after end

mschoch:


mschoch:
ping @steveyen with this change, the bolt seeking backwards behavior is no longer observed

again, not necessarily convinced this is the last of these...

PR #703, opened 2017-12-28
docValue persist changes

sreekanth-cb:
docValues are persisted along with the index,
in a columnar fashion per field with variable
sized chunking for quick look up.
-naive chunk level caching is added per field
-data part inside a chunk is snappy compressed
-metaHeader inside the chunk index the dv values
 inside the uncompressed data part
-all the fields are docValue persisted in this iteration

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -72,9 +72,44 @@ type Scorch struct {
 
 	eligibleForRemoval   []uint64        // Index snapshot epochs that are safe to GC.
 	ineligibleForRemoval map[string]bool // Filenames that should not be GC'ed yet.
+
+	onEvent func(event Event)
 }
 
-func NewScorch(storeName string, config map[string]interface{}, analysisQueue *index.AnalysisQueue) (index.Index, error) {
+// Event represents the information provided in an OnEvent() callback.
+type Event struct {
+	Kind     EventKind
+	Scorch   *Scorch
+	Duration time.Duration
+}
+
+// EventKind represents an event code for OnEvent() callbacks.
+type EventKind int
+
+// EventKindCLoseStart is fired when a Scorch.Close() has begun.
+var EventKindCloseStart = EventKind(1)
+
+// EventKindClose is fired when a scorch index has been fully closed.
+var EventKindClose = EventKind(2)
+
+// EventKindMergerProgress is fired when the merger has completed a
+// round of merge processing.
+var EventKindMergerProgress = EventKind(3)
+
+// EventKindPersisterProgress is fired when the persister has completed
+// a round of persistence processing.
+var EventKindPersisterProgress = EventKind(4)
+
+// EventKindBatchIntroductionStart is fired when Batch() is invoked which
+// introduces a new segment.
+var EventKindBatchIntroductionStart = EventKind(5)
+
+// EventKindBatchIntroduction is fired when Batch() completes.
+var EventKindBatchIntroduction = EventKind(6)
+
+func NewScorch(storeName string,
+	config map[string]interface{},
+	analysisQueue *index.AnalysisQueue) (index.Index, error) {
 	rv := &Scorch{
 		version:              Version,
 		config:               config,
@@ -94,6 +129,16 @@ func NewScorch(storeName string, config map[string]interface{}, analysisQueue *i
 		rv.unsafeBatch = ub
 	}
 	return rv, nil
+}
+
+func (s *Scorch) SetEventCallback(f func(Event)) {
+	s.onEvent = f
+}
+
+func (s *Scorch) fireEvent(kind EventKind, dur time.Duration) {
+	if s.onEvent != nil {
+		s.onEvent(Event{Kind: kind, Scorch: s, Duration: dur})
+	}
 }
 
 func (s *Scorch) Open() error {
@@ -163,6 +208,13 @@ func (s *Scorch) Open() error {
 }
 
 func (s *Scorch) Close() (err error) {
+	startTime := time.Now()
+	defer func() {
+		s.fireEvent(EventKindClose, time.Since(startTime))
+	}()
+
+	s.fireEvent(EventKindCloseStart, 0)
+
 	// signal to async tasks we want to close
 	close(s.closeCh)
 	// wait for them to close
@@ -195,11 +247,16 @@ func (s *Scorch) Delete(id string) error {
 
 // Batch applices a batch of changes to the index atomically
 func (s *Scorch) Batch(batch *index.Batch) error {
-	analysisStart := time.Now()
+	start := time.Now()
+
+	defer func() {
+		s.fireEvent(EventKindBatchIntroduction, time.Since(start))
+	}()
 
 	resultChan := make(chan *index.AnalysisResult, len(batch.IndexOps))
 
 	var numUpdates uint64
+	var numDeletes uint64
 	var numPlainTextBytes uint64
 	var ids []string
 	for docID, doc := range batch.IndexOps {
@@ -208,6 +265,8 @@ func (s *Scorch) Batch(batch *index.Batch) error {
 			doc.AddField(document.NewTextFieldCustom("_id", nil, []byte(doc.ID), document.IndexField|document.StoreField, nil))
 			numUpdates++
 			numPlainTextBytes += doc.NumPlainTextBytes()
+		} else {
+			numDeletes++
 		}
 		ids = append(ids, docID)
 	}
@@ -234,7 +293,10 @@ func (s *Scorch) Batch(batch *index.Batch) error {
 	}
 	close(resultChan)
 
-	atomic.AddUint64(&s.stats.analysisTime, uint64(time.Since(analysisStart)))
+	atomic.AddUint64(&s.stats.analysisTime, uint64(time.Since(start)))
+
+	// notify handlers that we're about to introduce a segment
+	s.fireEvent(EventKindBatchIntroductionStart, 0)
 
 	var newSegment segment.Segment
 	if len(analysisResults) > 0 {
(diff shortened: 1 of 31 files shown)


index/scorch/scorch.go
```diff
@@ -37,6 +37,14 @@ const Name = "scorch"
 
 const Version uint8 = 1
 
+// UnInvertIndex is implemented by various scorch index implementations
```
mschoch: there is only one scorch index implementation, did you mean various segment implementations? in that case, could this interface definition be moved to index/scorch/segment/segment.go? with the rest of the segment related interfaces?

index/scorch/scorch.go
```diff
@@ -37,6 +37,14 @@ const Name = "scorch"
...
+// to provide the un inverting of the postings or other indexed values.
+type UnInvertIndex interface {
+	// apparently need better namings here..
```
mschoch: any specific names you don't like or want suggestions for? or can we remove this comment?

index/scorch/segment/mem/segment.go
```diff
@@ -87,12 +87,17 @@ type Segment struct {
 	// stored field array positions
 	//  docNum -> field id -> slice of array positions (each is []uint64)
 	StoredPos []map[uint16][][]uint64
+
+	// for marking the docValue override status
+	// field id -> status
+	DocValueFields map[uint16]bool
```
mschoch: since these are just bools (not that large) and the key of the map is field id (integer, and not that many of them) could we just use a slice instead?

index/scorch/segment/zap/build.go
```diff
@@ -419,3 +426,104 @@ func persistDictionary(memSegment *mem.Segment, w *CountHashWriter, postingsLocs
...
+		if err != nil {
+			return fieldChunkOffsets, err
+		}
+
+		dictItr := dict.Iterator()
+		next, err := dictItr.Next()
+		for err == nil && next != nil {
+			postings, err1 := dict.PostingsList(next.Term, nil)
+			if err1 != nil {
+				return fieldChunkOffsets, err
```
mschoch: usually clearer to return 0,err
if you forget to check err somewhere, 0 more likely to stand out as incorrect offset

index/scorch/segment/zap/build.go
```diff
@@ -419,3 +426,104 @@ func persistDictionary(memSegment *mem.Segment, w *CountHashWriter, postingsLocs
...
+
+	return fieldChunkOffsets, nil
+}
+
+func persistFieldDocValues(w *CountHashWriter, chunkFactor uint32,
+	memSegment *mem.Segment) (uint64, error) {
+
+	fieldDvOffsets, err := persistDocValues(memSegment, w, chunkFactor)
+	if err != nil {
+		return math.MaxUint64, err
```
mschoch: any significance to this return value? if not i'd probably suggest just using 0 which is more common in Go when returning an error

index/scorch/segment/zap/build.go
```diff
@@ -419,3 +426,104 @@ func persistDictionary(memSegment *mem.Segment, w *CountHashWriter, postingsLocs
...
+
+	fieldDocValuesOffset := uint64(w.Count())
+	buf := make([]byte, binary.MaxVarintLen64)
+	offset := uint64(math.MaxUint64)
+	ok := true
+	for fieldID := range memSegment.FieldsInv {
+		// if the field isn't configured for docValue, then mark
+		// the offset accordingly
+		if offset, ok = fieldDvOffsets[uint16(fieldID)]; !ok {
+			offset = math.MaxUint64
```
mschoch: consider defining a constant with a more explanatory name for this value, and then giving the constant the value math.MaxUint64, something like

const fieldNotUninverted = math.maxUint64

it should make the code read more clearly

index/scorch/segment/zap/contentcoder.go
```diff
@@ -0,0 +1,181 @@
...
+	buf := make([]byte, binary.MaxVarintLen64)
+	n := binary.PutUvarint(buf, uint64(len(c.chunkMeta)))
+	_, err := c.chunkMetaBuf.Write(buf[:n])
+	if err != nil {
+		return err
+	}
+
+	// write out the metaData slice
+	for _, meta := range c.chunkMeta {
+		n := binary.PutUvarint(buf, meta.docID)
```
mschoch: consider using this helper method i wrote here:

https://github.com/blevesearch/bleve/blob/scorch/index/scorch/segment/zap/write.go#L127

it would let you write these 3 uvarints with a single call

_, err = writeUvarints(meta.docID, meta.docDvLoc, meta.docDvLen)
if err != nil {
   return err
}

index/scorch/segment/zap/contentcoder.go
```diff
@@ -0,0 +1,181 @@
...
+	c.chunkMetaBuf.Reset()
+	for i := range c.chunkLens {
+		c.chunkLens[i] = 0
+	}
+	c.chunkMeta = []metaData{}
+}
+
+// Close indicates you are done calling Add() this allows
+// the final chunk to be encoded.
+func (c *chunkedContentCoder) Close() {
```
mschoch: consider returning the error here, in general Go convention is signature like

func (s *Something) Close() error

(and yes I see that my version didn't return error, but i really think it should )

index/scorch/snapshot_index.go
```diff
@@ -401,24 +404,35 @@ func (i *IndexSnapshot) DocumentVisitFieldTerms(id index.IndexInternalID,
 
 	ss := i.segment[segmentIndex]
 
-	err = ss.cachedDocs.prepareFields(fields, ss)
-	if err != nil {
-		return err
-	}
+	switch seg := ss.segment.(type) {
+	case *mem.Segment:
```
mschoch: instead of doing type assertion matching the implementation structure, should we just be checking to see if it supports UnInvertIndex.  If it does, we use that, if it doesn't we fallback to the existing mechanism (which should work for any segment regardless of implementation)

index/scorch/segment/zap/merge.go
```diff
@@ -306,19 +314,51 @@ func persistMergedRest(segments []*Segment, drops []*roaring.Bitmap,
...
 		// write this vellum to disk
 		_, err = w.Write(vellumData)
 		if err != nil {
-			return nil, err
+			return nil, fieldDvLocsOffset, err
+		}
+
+		rv1[fieldID] = dictOffset
+
+		// update teh doc value
```
mschoch: typo 'teh'

index/scorch/segment/zap/build.go
sreekanth-cb: yea, right.. as it is a slice may be a nil seems the right choice..

index/scorch/segment/zap/build.go
sreekanth-cb: hm.. It was all zero's in my original code. For some reason I felt zero being a default value, it is better to specify an explicit value  through a different code., Hence initialised to "math.MaxUint64" .. But let me  change to '0' ..

index/scorch/scorch.go
sreekanth-cb: any suggestions on the interface & method name ? :) I don't like both of them.. just given some names for me to proceed..

mschoch: Here is one suggestion:

https://golang.org/doc/effective_go.html#interface-names

"By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: Reader, Writer, Formatter, CloseNotifier etc."

That doesn't quite fit in this case, because the closest to that would be `DocumentFieldTermVisitor` which is already taken to define the type for the callback itself.  Maybe we go with `DocumentFieldTermVisitable`?  I don't love that since it doesn't follow the convention, but I don't really want to rename the higher-level code at this time.  Does that work for you?

index/scorch/segment/zap/build.go
mschoch: Oops looks like its actually a map (right now, i had suggested considering a slice instead).  But in either case nil would be more appropriate, since I don't think it makes sense to proceed with a partial result if we encounter an error.

index/scorch/segment/zap/build.go
mschoch: I don't think we ever check that value though in the case of an error, I think for people reading the code `0, err` makes more sense.  Think of it this way, everyone reading Go code has some idea what `return 0, err` means.  When I read `return math.MaxUint64, err` I immediately had to stop and think, I wonder why he's returning a non-zero value with the error.  But, when I looked closer, it's not used anywhere, so it seems like just using `0,err` reduces the amount of thinking for the person reading the code.

index/scorch/segment/zap/segment.go
```diff
@@ -355,3 +374,20 @@ func (s *Segment) DictAddr(field string) (uint64, error) {
 
 	return s.fieldsOffsets[fieldID-1], nil
 }
+
+func (s *Segment) loadDvIterators() error {
+	if s.docValueOffset == fieldNotUninverted || s.docValueOffset == 0 {
```
mschoch: why is 0 an invalid location of the docValueOffset?

index/scorch/segment/zap/contentcoder.go
```diff
@@ -0,0 +1,173 @@
...
+func (c *chunkedContentCoder) flushContents() error {
+	// flush the contents, with meta information at first
+	buf := make([]byte, binary.MaxVarintLen64)
+	n := binary.PutUvarint(buf, uint64(len(c.chunkMeta)))
+	_, err := c.chunkMetaBuf.Write(buf[:n])
+	if err != nil {
+		return err
+	}
+
+	w := bufio.NewWriter(&c.chunkMetaBuf)
```
mschoch: why did you decide to introduce a new buffered writer here?  i dont think the writeUvarints method requires it (any writer is fine, including the existing buffer) and it seems costly to add it just to use the helper method.

sreekanth-cb: yep..its redundant. reverting..

index/scorch/segment/zap/segment.go
sreekanth-cb: assumption was that it can't begin at zero address as there are other valid entries before that at that address location?

mschoch: i think thats true today, but depends on other things unrelated to the implementation of the doc values themselves.  for example, if we decided to move the stored field data to a different section of the file, this might be first and always start at zero, and no one would think to check for this line.  i would prefer that this code not make this assumption.

sreekanth-cb:
hi Marty, good point..I didn't thought of it yet... One naive approach comes to mind atm is to add another method to new UninvertIndex interface to know the list of fields persisted with docValues. And make two separate (parallel?)calls with those two mutually exclusive field lists....need to think more.

mschoch:
Hey Sreekanth, sorry for the delay.  I wanted to gather some before/after numbers on the time to index and the size of the index using the bleve beer-search example before merging.  Unfortunately, when I pull your changes locally, the beer-search app seems to hang during indexing.  I'll try to dig deeper, but could you maybe give that a try on your end as well?

mschoch:
This looks good now.  Is it possible to not include the commits from master (74b8826) I'm worried it may complicate the merge to master later.

marty

PR #704, opened 2017-12-28
Scorch specific stats

abhinavdangeti:


index/scorch/persister.go
```diff
@@ -59,6 +61,8 @@ OUTER:
 			ourSnapshot.AddRef()
 			ourPersisted = s.rootPersisted
 			s.rootPersisted = nil
+			numItemsToPersist = atomic.LoadUint64(&s.stats.numItemsIntroduced) -
```
mschoch: this doesn't feel like the best way to compute how many items we're persisting, because it assumes that we will persist everything introduced so far, which happens to be true in the current design, but doesn't feel like something essential.

shouldn't we just have some independent way of knowing how many items we persisted?  that way if numItemsIntroduced - numItemsPersisted = 0, it actually means the system is working right.  in the current design it always goes to 0 because of the way we compute it.

abhinavdangeti: Hmm, yea I agree. As we already have the information within the persister on whether a segment snapshot is new or a replacement, I can increment the numItemsPersisted if in case of a new segment. Let me push up a new patch.

PR #708, opened 2017-12-29
Adding onEvent callback support for scorch

abhinavdangeti:
Event types:
- EventKindCloseStart
- EventKindClose
- EventKindMergerProgress
- EventKindPersisterProgress
- EventKindBatchIntroduction

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -19,6 +19,7 @@ import (
 	"log"
 	"os"
 	"sync/atomic"
+	"time"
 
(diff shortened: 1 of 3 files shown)


mschoch:
If the EventKindBatchIntroduction fires after a batch has been introduced, how will it allow us to block incoming batches when too much memory is currently being used?

abhinavdangeti:
Crap, I missed an event - BatchIntroductionStart that is to fire at the start. Let me push up a new batch.

PR #709, opened 2017-12-29
Tracking memory consumption for a scorch index

abhinavdangeti:
+ Track memory usage at a segment level
+ Add a new scorch API: MemoryUsed()
    - Aggregate the memory consumption across
      segments when API is invoked.

### DIFF
--- a/index/scorch/segment/mem/segment.go
+++ b/index/scorch/segment/mem/segment.go
@@ -87,6 +87,10 @@ type Segment struct {
 	// stored field array positions
 	//  docNum -> field id -> slice of array positions (each is []uint64)
 	StoredPos []map[uint16][][]uint64
+
+	// footprint of the segment, updated when analyzed document mutations
+	// are added into the segment
+	sizeInBytes uint64
 }
 
 // New builds a new empty Segment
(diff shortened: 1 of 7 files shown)


index/scorch/segment/mem/segment.go
```diff
@@ -96,6 +100,84 @@ func New() *Segment {
 	}
 }
 
+func (s *Segment) updateSizeInBytes() {
```
mschoch: a couple of general concerns

1) iterating through all this data a second time seems expensive, and that it would be cheaper to just increment the sizes while we do the building.  i get that it will be a lot messier, and more prone to missing something, so lets not change this now, but it may be something we have to revisit later.

2) we need to consider the overhead as well.  pointers take up 4 or 8 bytes depending on the architecture word size, slices take up 3 more words, and maps have some overhead we need to consider as well.  obviously this is smaller than the data, so i'd be ok deferring this to a later commit, as this gets us going in the right direciton.

3)  some of these loops seem unnecessary, we know that FieldsMap and FieldsInv are built with corresponding entries, so no need to iterate both, likewise for Dicts and DictKeys.

abhinavdangeti: Yes incrementing the sizes while building was what I did first, and it looked super messy.
Ah yes i missed the pointer overheads, let me add that as a separate commit.
Let me push up another batch that gets rid of some of these loops.

PR #711, opened 2017-12-29
Tracking memory consumption for a scorch index

abhinavdangeti:
+ Track memory usage at a segment level
+ Add a new scorch API: MemoryUsed()
    - Aggregate the memory consumption across
      segments when API is invoked.

+ TODO:
    - Revisit the second iteration if it can be gotten
      rid off, and the size accounted for during the first
      run while building an in-mem segment.
    - Accounting for pointer and slice overhead.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -361,6 +361,21 @@ func (s *Scorch) AddEligibleForRemoval(epoch uint64) {
 	s.rootLock.Unlock()
 }
 
+func (s *Scorch) MemoryUsed() uint64 {
+	var memUsed uint64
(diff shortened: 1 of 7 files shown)


mschoch:
@abhinavdangeti do you to work on these TODO items before merging?  Or should we proceed with this for now?

mschoch:
I'm going to go ahead and merge this, it's good enough for now, and we have other chunks of code landing, so I want to avoid merge pain by getting things in when they're clean.

abhinavdangeti:
@mschoch thanks! I'll work on the TODOs as separate changes.

PR #715, opened 2017-12-30
Rollback: GC'ing segments referenced by an indexSnapshot

abhinavdangeti:
+ When an indexSnapshot is returned through the API:
  PreviousPersistedSnapshot(...), add all the segment files
  that fall within the snapshot as inEligibleForRemoval.
  This is so that these files aren't removed when midway
  through the rollback operation.
+ Update rollback test to better test the scenario

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -289,7 +289,7 @@ func (s *Scorch) revertToSnapshot(revertTo *snapshotReversion) error {
 			deleted:    segmentSnapshot.deleted,
 			cachedDocs: segmentSnapshot.cachedDocs,
 		}
-		segmentSnapshot.segment.AddRef()
+		newSnapshot.segment[i].segment.AddRef()
 	}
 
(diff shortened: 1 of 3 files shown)


mschoch:
i think this has a new problem.  the original intended usage of markIneligibleForRemoval works because it is paired with unmarkIneligibleForRemoval that will guaranteed to eventually be called.  but in this case there seems to be no such guarantee.  it seems like once protected by a rollback, it gets stuck forever on the ineligible list.

it still seems to me like the deeper issue is that we end up opening the second an additional time.  using the ineligible list hides the most obvious side-effect, which was file deletion, but the real issue is we never should have opened the file a second time.  i have no idea how we could make that happen, but that feels more correct to me...

abhinavdangeti:
Yeah, that sounds right to me too.. While I continue to figure out how we can do it, let me split this change up .. so we can get the right test for rollback in..

PR #716, opened 2018-01-02
Adapting rollback test to better test the scenario

abhinavdangeti:
The rollback test only checks if the new root has an epoch that is less than or equal to the one reverted to .. which is pretty weak. Improving this test to write a couple of batches and retrieve the snapshot between the two and rollback to that. Now we can better validate that rollback was successful by ensuring that we only see the data before the second batch.


mschoch:
Can you update the commit message with a short description of how this is a better test?

PR #720, opened 2018-01-03
Updated Rollback APIs

abhinavdangeti:
New APIs:
+ RollbackPoints()
    - Retrieves the available list of rollback points: epoch+meta.
    - The application will need to check with the meta to decide
    on the rollback point.
+ Rollback()
    - API requires a rollback point identified by the first API.
    - Atomically & Durably rolls back the index to specified point,
    provided the specified rollback point is still available.
+ Unit test: TestIndexRollback
    - Writes a batch.
    - Sets the rollback point.
    - Writes second batch.
    - Rollback to previously decided point.
    - Ensure that data is as is before the second batch.

### DIFF
--- a/index/scorch/snapshot_rollback.go
+++ b/index/scorch/snapshot_rollback.go
@@ -15,96 +15,162 @@
 package scorch
 
 import (
-	"bytes"
+	"fmt"
 	"log"
 
 	"github.com/blevesearch/bleve/index/scorch/segment"
+	"github.com/boltdb/bolt"
 )
 
-// PreviousPersistedSnapshot returns the next older, previous
-// IndexSnapshot based on the provided IndexSnapshot. If the provided
-// argument is nil, the most recently persisted IndexSnapshot is returned.
-// This API allows the application to walk backwards into the history
-// of a store to previous points in time. A nil return value indicates
-// that no previous snapshots are available.
-func (s *Scorch) PreviousPersistedSnapshot(is *IndexSnapshot) (*IndexSnapshot, error) {
+type RollbackPoint struct {
+	epoch uint64
+	meta  map[string][]byte
+}
+
+func (r *RollbackPoint) GetInternal(key []byte) []byte {
+	return r.meta[string(key)]
+}
+
+// RollbackPoints returns an array of rollback points available
+// for the application to make a decision on where to rollback
+// to. A nil return value indicates that there are no available
+// rollback points.
+func (s *Scorch) RollbackPoints() ([]*RollbackPoint, error) {
 	if s.rootBolt == nil {
-		return nil, nil
+		return nil, fmt.Errorf("RollbackPoints: root is nil")
 	}
 
-	// start a read-only transaction
+	// start a read-only bolt transaction
 	tx, err := s.rootBolt.Begin(false)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("RollbackPoints: failed to start" +
+			" read-only transaction")
 	}
 
-	// Read-only bolt transactions to be rolled back.
+	// read-only bolt transactions to be rolled back
 	defer func() {
 		_ = tx.Rollback()
 	}()
 
 	snapshots := tx.Bucket(boltSnapshotsBucket)
 	if snapshots == nil {
-		return nil, nil
+		return nil, fmt.Errorf("RollbackPoints: no snapshots available")
 	}
 
-	pos := []byte(nil)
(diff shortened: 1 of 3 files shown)


index/scorch/snapshot_rollback.go
```diff
@@ -15,81 +15,142 @@
...
-					" unable to load snapshot, %v, continuing", err)
-				continue
+		meta := map[string][]byte{}
+		c2 := snapshot.Cursor()
+		for j, _ := c2.First(); j != nil; j, _ = c2.Next() {
+			if j[0] == boltInternalKey[0] {
+				internalBucket := snapshot.Bucket(j)
+				err = internalBucket.ForEach(func(key []byte, val []byte) error {
+					copiedVal := append([]byte(nil), val...)
+					meta[string(key)] = copiedVal
```
sreekanth-cb: not understanding the reason here , can't we assign the val directly to the meta map?
may I know, why do we need this step "copiedVal := append([]byte(nil), val...)" ?

index/scorch/snapshot_rollback.go
```diff
@@ -15,81 +15,142 @@
...
+					break
+				}
 			}
+		}
 
-			indexSnapshot.epoch = snapshotEpoch
-			return indexSnapshot, nil
+		if err != nil {
+			log.Printf("RollbackPoints:"+
+				" failed in fetching interal data: %v", err)
```
sreekanth-cb: "internal"?

index/scorch/snapshot_rollback.go
```diff
@@ -15,81 +15,142 @@
...
+	return rollbackPoints, nil
 }
 
-// SnapshotRevert atomically brings the store back to the point in time
-// as represented by the revertTo IndexSnapshot. SnapshotRevert() should
-// only be passed an IndexSnapshot that came from the same store.
-func (s *Scorch) SnapshotRevert(revertTo *IndexSnapshot) error {
+// Rollback atomically and durably (if unsafeBatch is unset) brings
+// the store back to the point in time as represented by the
+// RollbackPoint. Rollbac() should only be passed a RollbackPoint
```
sreekanth-cb: "Rollback()"?

index/scorch/snapshot_rollback.go
mschoch: Values returned by BoltDB are only guaranteed to be valid for the duration of the transaction.

index/scorch/snapshot_rollback.go
```diff
@@ -15,81 +15,142 @@
...
+
+	// add segments referenced by loaded index snapshot to the
+	// ineligibleForRemoval map
+	for _, segSnap := range revertTo.segment {
+		filename := zapFileName(segSnap.id)
+		s.markIneligibleForRemoval(filename)
+	}
+
+	// read-only bolt transactions to be rolled back
+	_ = tx.Rollback()
```
mschoch: should we use the defer pattern we used before? i see other return paths that do not clean up the transaction

index/scorch/snapshot_rollback.go
```diff
@@ -15,81 +15,142 @@
...
+	revertTo, err := s.loadSnapshot(snapshot)
+	if err != nil {
+		return fmt.Errorf("Rollback: unable to load snapshot: %v", err)
+	}
+
+	// add segments referenced by loaded index snapshot to the
+	// ineligibleForRemoval map
+	for _, segSnap := range revertTo.segment {
+		filename := zapFileName(segSnap.id)
+		s.markIneligibleForRemoval(filename)
```
mschoch: wondering if we should acquire the lock at the start and hold it for the duration, instead of relying on the acquisition and release of the lock in each call here.  since its rollback we're not worried about the extra cost here, and it feels safer.

index/scorch/snapshot_rollback.go
abhinavdangeti: You're right - I did miss a few return paths. The reason I didn't use defer here is because of persistence - If unsafeBatch is unset we wait for the persister to complete and there'd be a deadlock situation with the persister waiting to grab the boltdb write lock to complete the transaction. I'll address this.

index/scorch/snapshot_rollback.go
abhinavdangeti: Yeah makes sense.

index/scorch/snapshot_rollback.go

index/scorch/snapshot_rollback.go

PR #721, opened 2018-01-05
docValue configs change

sreekanth-cb:
draft implementation for
-making dv configs overridable at field level
-enabling on the fly/runtime un inverting of doc values
-few UT updates

Creating the PR to set a context for further discussions..

### DIFF
--- a/index/scorch/segment/zap/docvalues.go
+++ b/index/scorch/segment/zap/docvalues.go
@@ -184,14 +184,14 @@ func (s *Segment) VisitDocumentFieldTerms(localDocNum uint64, fields []string,
 // persisted doc value terms ready to be visitable using the
 // VisitDocumentFieldTerms method.
 func (s *Segment) VisitableDocValueFields() ([]string, error) {
-	if len(s.fieldsMap) == 0 {
-		return nil, fmt.Errorf("VisitableDocValueFields: no eligible fields" +
-			" with doc value persisted")
+	if len(s.fieldsInv) == 0 {
+		return nil, fmt.Errorf("VisitableDocValueFields: no fields" +
+			" for doc value persistence")
 	}
 
 	var rv []string
-	for field, fieldID := range s.fieldsMap {
-		if dvIter, ok := s.fieldDvIterMap[fieldID-1]; ok {
+	for fieldID, field := range s.fieldsInv {
+		if dvIter, ok := s.fieldDvIterMap[uint16(fieldID)]; ok {
 			if dvIter != nil {
 				rv = append(rv, field)
 			}
--- a/index/scorch/segment/zap/segment_test.go
+++ b/index/scorch/segment/zap/segment_test.go
@@ -575,7 +575,7 @@ func TestSegmentVisitableDocValueFieldsList(t *testing.T) {
 		}
 
 		if !reflect.DeepEqual(fields, expectedFields) {
-			t.Errorf("expected field terms: %#v, got: %#v", fields, expectedFields)
+			t.Errorf("expected field terms: %#v, got: %#v", expectedFields, fields)
 		}
 	}
 }


index/scorch/snapshot_index.go
```diff
@@ -404,11 +404,42 @@ func (i *IndexSnapshot) DocumentVisitFieldTerms(id index.IndexInternalID,
...
+		if len(fieldsDvPending) == 0 {
+			// all fields are doc value persisted
+			return zaps.VisitDocumentFieldTerms(localDocNum, fields, visitor)
+		}
+
+		// concurrently trigger the runtime doc value computations for pending fields
+		// as well as the visit of the persisted doc values
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go visitDocumentFieldCacheAsync(localDocNum, fieldsDvPending, ss, &wg, visitor)
```
sreekanth-cb: ha...may be not a good idea to call the visitor callback from different routines. Looks like the FieldCache preparation can be done concurrently and the invoking the visitor callback can be in sequential..

index/scorch/segment/zap/docvalues.go
```diff
@@ -178,3 +179,24 @@ func (s *Segment) VisitDocumentFieldTerms(localDocNum uint64, fields []string,
 	}
 	return nil
 }
+
+// VisitableDocValueFields returns the list of fields with
+// persisted doc value terms ready to be visitable using the
+// VisitDocumentFieldTerms method.
+func (s *Segment) VisitableDocValueFields() ([]string, error) {
+	if len(s.fieldsInv) == 0 {
```
mschoch: seems like it isn't necessary to treat this as an error, if there are no fields, none of them are visitable for doc values, so can't we just return nil

index/scorch/segment/zap/docvalues.go
```diff
@@ -178,3 +179,24 @@ func (s *Segment) VisitDocumentFieldTerms(localDocNum uint64, fields []string,
...
+		return nil, fmt.Errorf("VisitableDocValueFields: no fields" +
+			" for doc value persistence")
+	}
+
+	var rv []string
+	for fieldID, field := range s.fieldsInv {
+		if dvIter, ok := s.fieldDvIterMap[uint16(fieldID)]; ok {
```
mschoch: could write this one line instead of 2 if conditions as:
...; ok && dvIter != nil {

index/scorch/segment/zap/segment_test.go
```diff
@@ -515,3 +517,65 @@ func TestOpenMultiWithTwoChunks(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("segment VisitableDocValueFields err: %v", err)
+		}
+		// no persisted doc value fields
+		if len(fields) != 0 {
+			t.Errorf("expected no persisted fields for doc values, got: %#v", fields)
+		}
+	}
+
+	_ = os.RemoveAll("/tmp/scorch.zap")
```
mschoch: the Close() on the previous segment was deferred, and won't have run before you remove the file.  it may not cause problems, but this is incorrect.

index/scorch/snapshot_index.go
```diff
@@ -430,3 +461,25 @@ func (i *IndexSnapshot) DocumentVisitFieldTerms(id index.IndexInternalID,
 
 	return nil
 }
+
+func stringsToMap(strsArr []string) map[string]bool {
+	if strsArr == nil {
+		return nil
+	}
+	strs := map[string]bool{}
```
mschoch: in go, when tracking existence of things in a map, it is recommend to use empty struct as a value instead of bool:

strs := map[string]struct{}{}

when checking, you do
if _, ok := strs[key]; ok {
  // key is in the map
}

index/scorch/snapshot_index.go
mschoch: another problem here is that you ignore errors that happen because it's async and you have no easy way to deal with them.  this is usually a bad sign, and done right, you'd need a channel to return the errors on.

index/scorch/snapshot_index.go
```diff
@@ -404,11 +404,42 @@ func (i *IndexSnapshot) DocumentVisitFieldTerms(id index.IndexInternalID,
...
-	// else fallback to the in memory fieldCache
-	err = ss.cachedDocs.prepareFields(fields, ss)
+	return visitDocumentFieldCache(localDocNum, fields, ss, visitor)
+}
+
+func visitDocumentFieldCacheAsync(localDocNum uint64, fields []string,
+	ss *SegmentSnapshot, wg *sync.WaitGroup,
+	visitor index.DocumentFieldTermVisitor) {
+	defer wg.Done()
+	_ = visitDocumentFieldCache(localDocNum, fields, ss, visitor)
```
mschoch: this is the ignored error i mentioned earlier

mapping/field.go
```diff
@@ -26,8 +26,9 @@ import (
 
 // control the default behavior for dynamic fields (those not explicitly mapped)
 var (
-	IndexDynamic = true
-	StoreDynamic = true
+	IndexDynamic     = true
+	StoreDynamic     = true
+	DocValuesDynamic = true // may be default disabled?
```
mschoch: we should be consistent with the naming, you chose IncludeDocValues earlier, we have to stick with it, or change them all to just "DocValues"

mapping/field.go
```diff
@@ -54,6 +55,10 @@ type FieldMapping struct {
 	IncludeTermVectors bool   `json:"include_term_vectors,omitempty"`
 	IncludeInAll       bool   `json:"include_in_all,omitempty"`
 	DateFormat         string `json:"date_format,omitempty"`
+
+	// DocValues, if true makes the index uninverting possible for this field
+	// It is useful for faceting and sorting queries.
+	DocValues bool `json:"docvalues,omitempty"`
```
mschoch: Again, switched from IncludeDocValues earlier to just DocValues now?

document/field_boolean.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/blevesearch/bleve/analysis"
 )
 
-const DefaultBooleanIndexingOptions = StoreField | IndexField
+const DefaultBooleanIndexingOptions = StoreField | IndexField | IncludeDocValues
```
mschoch: I thought we were going to go with the shorter "DocValues".

edit: yes I see later you use just "DocValues" in the mapping.  Is there any reason we can't use the same everywhere, otherwise we have to remember which one to use where.

index/scorch/snapshot_index.go
mschoch: reading this more carefully, yes, we must only invoke the callback from a single goroutine.  the reason is that the callbacks are written in a simple way expecting this, if multiple goroutines call them, they would need locking/synchronization on all of their data.

index/scorch/snapshot_index.go
```diff
@@ -430,3 +461,25 @@ func (i *IndexSnapshot) DocumentVisitFieldTerms(id index.IndexInternalID,
...
+		return nil
+	}
+	strs := map[string]bool{}
+	for _, str := range strsArr {
+		strs[str] = true
+	}
+	return strs
+}
+
+func stringsRemoveStrings(stringArr, removeArr []string) []string {
```
mschoch: i find these string functions ugly, i would combine them into a single function.  while stringsToMap seems common, its also pretty trvial and probably clear to just manually inline everywhere.

mschoch:
Sorry Sreekanth, we decided to merge Scorch into master.  Hopefully you can rebase your changes against master without too much difficulty.

sreekanth-cb:
Sure Marty, Closing this PR and opening another afresh..

PR #722, opened 2018-01-05
fix race condition in setting up event callbacks

mschoch:
previous approach used SetEventCallback method which allowed
you to change the callback, unfotunately that also included
times after the goroutines were started and potentially firing
the callback.

checking lock on this would be too expensive, so instead we go
for an approach that allows callbacks to be registered by name
during process init(), then upon opening up an index a string
config key 'eventCallbackName' is used to look up the
appropriate callback function.  also, since this strinc config
name is serializable, it fits into the existing bleve index
metadata without any new issues.

### DIFF
--- a/index/scorch/event.go
+++ b/index/scorch/event.go
@@ -1,0 +1,52 @@
+//  Copyright (c) 2018 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


index/scorch/event.go
```diff
@@ -0,0 +1,52 @@
+//  Copyright (c) 2017 Couchbase, Inc.
```
abhinavdangeti: 2018 :)

mschoch: damn....

PR #723, opened 2018-01-05
add initial support for async error callback

mschoch:


index/scorch/scorch.go
```diff
@@ -100,6 +101,10 @@ func NewScorch(storeName string,
 	if ok {
 		rv.onEvent = RegistryEventCallbacks[ecbName]
 	}
+	aecbName, ok := config["ascyncErrorCallbackName"].(string)
```
abhinavdangeti: asyncErrorCallbackName

PR #726, opened 2018-01-08
DocValue Config, new API Changes

sreekanth-cb:
-new `VisitableDocValueFields` API for persisted DV field list
-making dv configs overridable at field level
-enabling on the fly/runtime un inverting of doc values
-few UT updates

### DIFF
--- a/index/scorch/segment/zap/docvalues.go
+++ b/index/scorch/segment/zap/docvalues.go
@@ -184,10 +184,6 @@ func (s *Segment) VisitDocumentFieldTerms(localDocNum uint64, fields []string,
 // persisted doc value terms ready to be visitable using the
 // VisitDocumentFieldTerms method.
 func (s *Segment) VisitableDocValueFields() ([]string, error) {
-	if len(s.fieldsInv) == 0 {
-		return nil, nil
-	}
-
 	var rv []string
 	for fieldID, field := range s.fieldsInv {
 		if dvIter, ok := s.fieldDvIterMap[uint16(fieldID)]; ok &&
@@ -195,6 +191,5 @@ func (s *Segment) VisitableDocValueFields() ([]string, error) {
 			rv = append(rv, field)
 		}
 	}
-
 	return rv, nil
 }
--- a/index/scorch/segment/zap/docvalues.go
+++ b/index/scorch/segment/zap/docvalues.go
@@ -151,7 +151,8 @@ func (di *docValueIterator) getDocValueLocs(docID uint64) (uint64, uint64) {
 	return math.MaxUint64, math.MaxUint64
 }
 
-// VisitDocumentFieldTerms is an implementation of the UnInvertIndex interface
+// VisitDocumentFieldTerms is an implementation of the
+// DocumentFieldTermVisitable interface
 func (s *Segment) VisitDocumentFieldTerms(localDocNum uint64, fields []string,
 	visitor index.DocumentFieldTermVisitor) error {
 	fieldID := uint16(0)
@@ -177,4 +178,23 @@ func (s *Segment) VisitDocumentFieldTerms(localDocNum uint64, fields []string,
 		}
 	}
 	return nil
+}
+
+// VisitableDocValueFields returns the list of fields with
+// persisted doc value terms ready to be visitable using the
+// VisitDocumentFieldTerms method.
+func (s *Segment) VisitableDocValueFields() ([]string, error) {
+	if len(s.fieldsInv) == 0 {
+		return nil, nil
+	}
+
+	var rv []string
+	for fieldID, field := range s.fieldsInv {
+		if dvIter, ok := s.fieldDvIterMap[uint16(fieldID)]; ok &&
+			dvIter != nil {
+			rv = append(rv, field)
+		}
+	}
+
+	return rv, nil
 }
--- a/index/scorch/segment/zap/segment_test.go
+++ b/index/scorch/segment/zap/segment_test.go
@@ -40,7 +40,7 @@ func TestOpen(t *testing.T) {
 	defer func() {
 		cerr := segment.Close()
 		if cerr != nil {
-			t.Fatalf("error closing segment: %v", err)
+			t.Fatalf("error closing segment: %v", cerr)
 		}
 	}()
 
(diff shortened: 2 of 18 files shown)


index/scorch/segment/zap/docvalues.go
```diff
@@ -178,3 +179,22 @@ func (s *Segment) VisitDocumentFieldTerms(localDocNum uint64, fields []string,
 	}
 	return nil
 }
+
+// VisitableDocValueFields returns the list of fields with
+// persisted doc value terms ready to be visitable using the
+// VisitDocumentFieldTerms method.
+func (s *Segment) VisitableDocValueFields() ([]string, error) {
+	if len(s.fieldsInv) == 0 {
```
mschoch: i think we can just remove this if check, if we leave it out, we return nil, nil anyway and i don't think there is any significant gain by doing it explicitly.

index/scorch/segment/zap/segment_test.go
```diff
@@ -515,3 +518,83 @@ func TestOpenMultiWithTwoChunks(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("error persisting segment: %v", err)
+	}
+
+	seg, err := Open("/tmp/scorch.zap")
+	if err != nil {
+		t.Fatalf("error opening segment: %v", err)
+	}
+
+	cerr := seg.Close()
```
mschoch: why do we close the segment we just opened, and then keep using it in the next block?

edit: i see what happened, you copy/pasted it from a defer and just removed the deferred part, but now that code is the wrong place (defer executes out of the sequential flow of normal code)

index/scorch/segment/zap/segment_test.go
```diff
@@ -515,3 +518,83 @@ func TestOpenMultiWithTwoChunks(t *testing.T) {
...
+	if cerr != nil {
+		t.Fatalf("error closing segment: %v", err)
```
mschoch: we checked cerr, but then we pass err to Fatalf

index/scorch/segment/zap/segment_test.go
```diff
@@ -515,3 +518,83 @@ func TestOpenMultiWithTwoChunks(t *testing.T) {
...
+	}
+
+	seg, err = Open("/tmp/scorch.zap")
+	if err != nil {
+		t.Fatalf("error opening segment: %v", err)
+	}
+	defer func() {
+		cerr := seg.Close()
+		if cerr != nil {
+			t.Fatalf("error closing segment: %v", err)
```
mschoch: also should be cerr (i should review for all of these, since i'm guessing these all originated elsewhere and were copy/pasted)

mapping/field.go
```diff
@@ -28,6 +28,7 @@ import (
 var (
 	IndexDynamic = true
 	StoreDynamic = true
+	DocValues    = true // TODO revisit default?
```
mschoch: Shouldn't this be DocValuesDynamic?

mapping/index.go
```diff
@@ -50,6 +50,7 @@ type IndexMappingImpl struct {
 	DefaultField          string                      `json:"default_field"`
 	StoreDynamic          bool                        `json:"store_dynamic"`
 	IndexDynamic          bool                        `json:"index_dynamic"`
+	DocValues             bool                        `json:"docvalues,omitempty"`
```
mschoch: Should be DocValuesDynamic

mapping/field.go
```diff
@@ -64,13 +69,15 @@ func NewTextFieldMapping() *FieldMapping {
...
 		IncludeInAll:       true,
+		DocValues:          true,
 	}
 }
 
 func newTextFieldMappingDynamic(im *IndexMappingImpl) *FieldMapping {
 	rv := NewTextFieldMapping()
 	rv.Store = im.StoreDynamic
 	rv.Index = im.IndexDynamic
+	rv.DocValues = im.DocValues
```
mschoch: should refer to im.DocValuesDynamic

mapping/field.go
```diff
@@ -81,13 +88,15 @@ func NewNumericFieldMapping() *FieldMapping {
...
 		IncludeInAll: true,
+		DocValues:    true,
 	}
 }
 
 func newNumericFieldMappingDynamic(im *IndexMappingImpl) *FieldMapping {
 	rv := NewNumericFieldMapping()
 	rv.Store = im.StoreDynamic
 	rv.Index = im.IndexDynamic
+	rv.DocValues = im.DocValues
```
mschoch: im.DocValuesDynamic

mapping/field.go
```diff
@@ -98,13 +107,15 @@ func NewDateTimeFieldMapping() *FieldMapping {
...
 		IncludeInAll: true,
+		DocValues:    true,
 	}
 }
 
 func newDateTimeFieldMappingDynamic(im *IndexMappingImpl) *FieldMapping {
 	rv := NewDateTimeFieldMapping()
 	rv.Store = im.StoreDynamic
 	rv.Index = im.IndexDynamic
+	rv.DocValues = im.DocValues
```
mschoch: im.DocValuesDynamic

mapping/field.go
```diff
@@ -115,13 +126,15 @@ func NewBooleanFieldMapping() *FieldMapping {
...
 		IncludeInAll: true,
+		DocValues:    true,
 	}
 }
 
 func newBooleanFieldMappingDynamic(im *IndexMappingImpl) *FieldMapping {
 	rv := NewBooleanFieldMapping()
 	rv.Store = im.StoreDynamic
 	rv.Index = im.IndexDynamic
+	rv.DocValues = im.DocValues
```
mschoch: im.DocValuesDynamic

mapping/field.go
sreekanth-cb: It was "DocValuesDynamic" in last PR , but you had a comment there suggesting that everywhere we move it to "DocValues". At least that is what I understood there..
Yes, I prefer `DocValuesDynamic`.

mschoch: There is some confusion then.  In the previous version some of them had "Include" and that is what I wanted removed.

index/scorch/segment/zap/segment_test.go
sreekanth-cb: Yikes...Its a total copy pasted carelessness error.. :(

PR #728, opened 2018-01-08
segment SizeInBytes  updates

sreekanth-cb:
-segment size computations updated with docValue fields

@abhinavdangeti ,sending the changes over a PR as it is easier for you to suggest/fix if anything went wrong (as I am about to sleep atm..)

@mschoch , one doubt I have regarding the SizeInBytes API is, 
Is it for computing the exact size of the fields/pointers embedded within the Segment structs, not accounting for size of items which are referenced through pointers.) (which may not be the case)
, Else if it is for giving an estimate about the actual memory consumption size, then we need to somehow compute the FieldCache size for mem segments (though its not a part of the segment, its a part of the snapshot atm and for persisted segment docValueIterator sizes needs to be accounted to be precise about the size figures. If so, I shall work on it tomorrow to make the size computations accommodate those changes?

### DIFF
--- a/index/scorch/segment/zap/docvalues.go
+++ b/index/scorch/segment/zap/docvalues.go
@@ -40,7 +40,6 @@ func (di *docValueIterator) sizeInBytes() uint64 {
 	sizeInBytes := 24
 	sizeInBytes += len(di.field)
 	sizeInBytes += len(di.chunkLens) * 8
-	sizeInBytes += len(di.curChunkData)
 	sizeInBytes += len(di.curChunkHeader) * 24
 	return uint64(sizeInBytes)
 }
--- a/index/scorch/segment/zap/docvalues.go
+++ b/index/scorch/segment/zap/docvalues.go
@@ -35,6 +35,16 @@ type docValueIterator struct {
 	curChunkData   []byte // compressed data cache
 }
 
+func (di *docValueIterator) sizeInBytes() uint64 {
+	// curChunkNum, numChunks, dvDataLoc
+	sizeInBytes := 24
+	sizeInBytes += len(di.field)
+	sizeInBytes += len(di.chunkLens) * 8
+	sizeInBytes += len(di.curChunkData)
+	sizeInBytes += len(di.curChunkHeader) * 24
+	return uint64(sizeInBytes)
+}
+
 func (di *docValueIterator) fieldName() string {
 	return di.field
 }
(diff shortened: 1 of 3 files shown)


mschoch:
@sreekanth-cb SizeInBytes should account for all memory that is on the heap.  This is why the on-disk segments generally avoid creating a lot of structure on the heap, and mostly work with references []byte that are accessed via mmap.

index/scorch/segment/zap/segment.go
```diff
@@ -118,6 +120,13 @@ func (s *Segment) SizeInBytes() uint64 {
 	sizeInBytes += len(s.fieldsOffsets) * 8 /* size of uint64 */
 	sizeInBytes += 8                        /* size of refs -> int64 */
 
+	for _, entry := range s.fieldDvIterMap {
+		sizeInBytes += int(unsafe.Sizeof(entry)) + 2 /* size of uint16 */
```
mschoch: I believe since the values in the fidldDvIterMap are pointers, then unsae.Sizeof(entry) gives you the size of the pointer and not the struct itself.  Is that what you intended?

sreekanth-cb: yes, and then upon valid pointers compute the size of the referenced.

mschoch: in that case, i'd prefer we just assume 8 bytes for a pointer, rather than using unsafe

index/scorch/segment/zap/docvalues.go
```diff
@@ -35,6 +35,16 @@ type docValueIterator struct {
 	curChunkData   []byte // compressed data cache
 }
 
+func (di *docValueIterator) sizeInBytes() uint64 {
+	// curChunkNum, numChunks, dvDataLoc
+	sizeInBytes := 24
+	sizeInBytes += len(di.field)
+	sizeInBytes += len(di.chunkLens) * 8
+	sizeInBytes += len(di.curChunkData)
```
mschoch: since curChunkData is initialized from s.mm which is mmap'd, then we shouldn't count this in the calculation which is intended to count space on the heap

abhinavdangeti:
Hey @sreekanth-cb, @mschoch , I've created a PR that accounts for overhead from all the data structures within mem.Segment and zap.Segment here: https://github.com/blevesearch/bleve/pull/733, I've included changes from this one as well.

mschoch:
@abhinavdangeti can we close this now?

PR #729, opened 2018-01-09
Refactor MemoryUsed to CurDirtyBytes

abhinavdangeti:
This API is designed to only emit the dirty "unpersisted"
bytes only. This does not included mmap'ed zap segments.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -407,16 +407,20 @@ func (s *Scorch) AddEligibleForRemoval(epoch uint64) {
 	s.rootLock.Unlock()
 }
 
-func (s *Scorch) MemoryUsed() uint64 {
+func (s *Scorch) CurDirtyBytes() uint64 {
 	var memUsed uint64
 	s.rootLock.RLock()
-	for _, segmentSnapshot := range s.root.segment {
-		memUsed += 8 /* size of id -> uint64 */ +
-			segmentSnapshot.segment.SizeInBytes()
-		if segmentSnapshot.deleted != nil {
-			memUsed += segmentSnapshot.deleted.GetSizeInBytes()
+	if s.root != nil {
+		for _, segmentSnapshot := range s.root.segment {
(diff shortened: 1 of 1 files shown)


index/scorch/scorch.go
```diff
@@ -407,16 +407,20 @@ func (s *Scorch) AddEligibleForRemoval(epoch uint64) {
...
 	s.rootLock.RLock()
-	for _, segmentSnapshot := range s.root.segment {
-		memUsed += 8 /* size of id -> uint64 */ +
-			segmentSnapshot.segment.SizeInBytes()
-		if segmentSnapshot.deleted != nil {
-			memUsed += segmentSnapshot.deleted.GetSizeInBytes()
+	if s.root != nil {
+		for _, segmentSnapshot := range s.root.segment {
+			memUsed += 8 /* size of id -> uint64 */
+			if _, ok := segmentSnapshot.segment.(*mem.Segment); ok {
```
sreekanth-cb: sorry, I am not able to follow the logic here,  why are we restricting the memory used computations only to mem segments? At any point in time , there could be a mix of both mem and zap segments within an index snapshot right? 
I could see in the earlier logic, we were calling the segment.SizeInBytes() API which was implemented by both mem and zap segments..

sreekanth-cb: Whoops...Nevermind, I read the commit msg now.. :)

mschoch:
So, I'm still not 100% sold on ignoring the heap space used by the disk segments (not mmap'd space).  It seems to me like the sizes we're talking about here are tiny.  Like 100 bytes + the size of the field data.  The tests we're doing are really small scale, so even with 512MB, we should be able to have LOTS of disk segments before this is a concern.

Can't we just do a simpler change to stop include 'len(s.mm)' in the calculation, as that was not something I intended us to include and missed in the review phase.

abhinavdangeti:
Yeah that makes sense, I'll push up another patch - also to the right branch.

PR #731, opened 2018-01-09
Avoid creating new instances of IndexSnapshots where possible

abhinavdangeti:


mschoch:
It seems like you're modifying the root IndexSnapshot.  But a key part of the design is that IndexSnapshots are immutable.  That is what us safely use them as snapshots.

abhinavdangeti:
Yea you're right. But since the updates to root and even access to the root indexSnapshot is always within rootLock's scope - is there any danger in changing the way we do things? I didn't see any fallout (at least not yet).

mschoch:
So, when you get an index snapshot via the Reader, even though we get a lock, we just a pointer to the root IndexSnapshot.  That is only safe if IndexSnapshots are immutable.  Otherwise, the contents of the IndexSnapshot can change, which is precisely what would happen with this PR.  So, the data accessible by the Reader() is not a stable snapshot, but can actually change over time now.

abhinavdangeti:
Shoot, okay. I'll need to re-think this again.

mschoch:
I'm going to close this for now, I think we need to revisit garbage holistically, not in commits that try to chip away at it.

PR #732, opened 2018-01-11
 MB-27498 - date range facet query panics

sreekanth-cb:
Initialise the facet results map in case of an
 empty partial hits with a multi node cluster

### DIFF
--- a/search.go
+++ b/search.go
@@ -481,7 +481,7 @@ func (sr *SearchResult) Merge(other *SearchResult) {
 	if other.MaxScore > sr.MaxScore {
 		sr.MaxScore = other.MaxScore
 	}
-	if len(sr.Facets) == 0 {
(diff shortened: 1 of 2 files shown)


search.go
```diff
@@ -481,5 +481,8 @@ func (sr *SearchResult) Merge(other *SearchResult) {
 	if other.MaxScore > sr.MaxScore {
 		sr.MaxScore = other.MaxScore
 	}
+	if len(sr.Facets) == 0 && len(other.Facets) != 0 {
```
mschoch: it could be len 0, but not nil, and we would create yet another map

would it be simpler if sr.Facets == nil && len(other.Facets) != 0 to simply use the right-hand-side of the expression, sr.Facets = other.Facets?  or does this introduce new problems?

sreekanth-cb: done!

PR #733, opened 2018-01-12
Include overhead from data structures in segment's SizeInBytes

abhinavdangeti:
+ Account for all the overhead incurred from the data structures
  within mem.Segment and zap.Segment.
+ Include overhead from certain new fields as well.

### DIFF
--- a/index/scorch/segment/mem/segment.go
+++ b/index/scorch/segment/mem/segment.go
@@ -107,27 +107,41 @@ func New() *Segment {
 func (s *Segment) updateSizeInBytes() {
 	var sizeInBytes uint64
 
+	// FieldsMap, FieldsInv
 	for k, _ := range s.FieldsMap {
-		sizeInBytes += uint64(len(k)*2 /* FieldsMap + FieldsInv */ +
+		sizeInBytes += uint64((len(k)+int(segment.SizeOfString))*2 +
 			2 /* size of uint16 */)
 	}
+	// overhead from the data structures
+	sizeInBytes += (segment.SizeOfMap + segment.SizeOfSlice)
 
+	// Dicts, DictKeys
 	for _, entry := range s.Dicts {
 		for k, _ := range entry {
-			sizeInBytes += uint64(len(k)*2 /* Dicts + DictKeys */ +
+			sizeInBytes += uint64((len(k)+int(segment.SizeOfString))*2 +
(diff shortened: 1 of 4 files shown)


abhinavdangeti:
Ok, I'll have another patch up soon.

index/scorch/segment/segment.go
```diff
@@ -19,6 +19,11 @@ import (
 	"github.com/blevesearch/bleve/index"
 )
 
+// Overhead from golang data structures when deployed on a 64-bit system.
```
mschoch: Go, not golang

index/scorch/segment/mem/segment.go
```diff
@@ -107,59 +107,90 @@ func New() *Segment {
 func (s *Segment) updateSizeInBytes() {
 	var sizeInBytes uint64
 
+	// FieldsMap, FieldsInv
 	for k, _ := range s.FieldsMap {
 		sizeInBytes += uint64(len(k)*2 /* FieldsMap + FieldsInv */ +
```
mschoch: need additional constant SizeOfString to account for string header (16 bytes on 64bit)

then we need (len(k)+SizeOfString)*2

index/scorch/segment/mem/segment.go
```diff
@@ -107,59 +107,90 @@ func New() *Segment {
...
 	}
+	// overhead from the data structures
+	sizeInBytes += (segment.SizeOfMap + segment.SizeOfSlice)
 
+	// Dicts, DictKeys
 	for _, entry := range s.Dicts {
 		for k, _ := range entry {
-			sizeInBytes += uint64(len(k)*2 /* Dicts + DictKeys */ +
-				8 /* size of uint64 */)
+			sizeInBytes += uint64(len(k)*2 + 8 /* size of uint64 */)
```
mschoch: account for string header here as well

index/scorch/segment/mem/segment.go
abhinavdangeti: Done.

index/scorch/segment/mem/segment.go
abhinavdangeti: Done.

index/scorch/segment/segment.go
abhinavdangeti: Done.

PR #738, opened 2018-01-17
Add support for certain disk stats

abhinavdangeti:


abhinavdangeti:
Ah crap, my mistake!

abhinavdangeti:
Hmm, this would be a problem on a warmup, needs one more change.

abhinavdangeti:
I did update the commit message, turns out the title of the PR wasn't updated.
Upon merge, we'd notify the persister, so I think that case should be covered. I'll need to handle the case on deleting files however.

mschoch:
It looks like in moss we just compute the size of the files at the time the stats are requested:

https://github.com/couchbase/moss/blob/5fdf85aeb793eef10c5dc110ab821fc31fbd572a/store_stats.go#L22-L32

@steveyen any thoughts on the two approaches?

steveyen:
> @steveyen any thoughts on the two approaches?

Hi @abhinavdangeti  & @mschoch -- I like asking the filesystem for stuff like this, especially if you're expecting the stats call to be infrequent (like once a second?  every 5 secs?  not millions/sec)...

That's because the filesystem has the ground truth, and maintaining a separate # is more code.  For example, maybe you got some files building up while in the middle of compaction/merging. 
 Or you had files leftover that were half-written from some previous power loss.  Or, some sysadmin did some "mkdir backup; cp * backup" into the same directory at 3AM due to some crazy downtime firefight, but forgot about it the next day (this has happened).  Might as well count 'em all up.

mschoch:
Related, if we switch to this approach, let's add a stat for number of files as well (counting in the same loop).  This will be useful as we start to worry about too many files...

abhinavdangeti:
Thanks for that feedback @steveyen. And sounds good @mschoch. I'll have a patch up where the stats API will itself be responsible for putting together these stats (including the num_files).

PR #744, opened 2018-01-27
MB-26396: Handling docs with geopoints in slice format

abhinavdangeti:
This handles the missing case for a geopoint when the
coordinates are in a slice format.


abhinavdangeti:
I'll update the commit message.
I tested using a couchbase-server instance, added docs based on the description in the ticket, and it was pretty straight forward to reproduce the bug and test the fix.
I'll include a unit test like you mention for this.

PR #749, opened 2018-02-02
unblock the files for clean up, esp for merged new segment files

sreekanth-cb:
Making the unblock outside the loop to ensure that it covers all the cases like only merged segments and the usual mix of both mem & zap segments.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -266,9 +266,7 @@ func (s *Scorch) persistSnapshot(snapshot *IndexSnapshot) error {
 		for k, v := range s.root.internal {
 			newIndexSnapshot.internal[k] = v
 		}
-		for _, filename := range filenames {
-			delete(s.ineligibleForRemoval, filename)
-		}
+
 		rootPrev := s.root
 		s.root = newIndexSnapshot
 		s.rootLock.Unlock()
@@ -276,6 +274,12 @@ func (s *Scorch) persistSnapshot(snapshot *IndexSnapshot) error {
 			_ = rootPrev.DecRef()
 		}
 	}
+	// unlock the files for clean up
+	s.rootLock.Lock()
+	for _, filename := range filenames {
+		delete(s.ineligibleForRemoval, filename)
+	}
+	s.rootLock.Unlock()
 
 	return nil
 }
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -274,7 +274,7 @@ func (s *Scorch) persistSnapshot(snapshot *IndexSnapshot) error {
 			_ = rootPrev.DecRef()
 		}
 	}
-	// unlock the files for clean up
+	// allow files to become eligible for removal
 	s.rootLock.Lock()
 	for _, filename := range filenames {
 		delete(s.ineligibleForRemoval, filename)


index/scorch/persister.go
```diff
@@ -266,16 +266,20 @@ func (s *Scorch) persistSnapshot(snapshot *IndexSnapshot) error {
...
 		if rootPrev != nil {
 			_ = rootPrev.DecRef()
 		}
 	}
+	// unlock the files for clean up
+	s.rootLock.Lock()
+	for _, filename := range filenames {
+		delete(s.ineligibleForRemoval, filename)
+	}
+	s.rootLock.Unlock()
```
steveyen: I'm wondering about the case mentioned in the comment in line 222... what if only the internal keys were updated and needed to be persisted (line 266)?

wondering instead if the if-check on line 223 should just be removed?

sreekanth-cb: yea.. But removing the 223 if-check looked risky as I read the commit msg from @mschoch for it. Looked like this is the safest / easiest option.

mschoch: @steveyen your concern is valid, but perhaps we open separate issue to track that, as it appears to be a logical flaw with the way this works.   Specifically, the loop-breaking condition this relies on is incompatible with the case you point out.

index/scorch/persister.go
```diff
@@ -266,16 +266,20 @@ func (s *Scorch) persistSnapshot(snapshot *IndexSnapshot) error {
...
-		}
+
 		rootPrev := s.root
 		s.root = newIndexSnapshot
 		s.rootLock.Unlock()
 		if rootPrev != nil {
 			_ = rootPrev.DecRef()
 		}
 	}
+	// unlock the files for clean up
```
mschoch: can we not use the phrase "unlock" since lock has very specific meaning, perhaps say

allow files to become eligible for removal again

index/scorch/persister.go
steveyen: @mschoch - thx -- I'm good with the idea to separate that other feedback into a different issue / track.  (To be clear, that's the issue of "what if only the internal keys were updated and needed to be persisted" and whether we're handling that internal-keys-only correctly.)

mschoch:
Yes, the case where only internal keys were updated is NOT handled correctly.  But, the fix is not as simple as just removing the if clause, because that will introduce a loop.

sreekanth-cb:
@mschoch , As there is no other changes needed in this PR, ready for merge?

mschoch:
I've broken out new issue for "internal values only" case here: #763

But I've started to convince myself its not an issue after all.

index/scorch/persister.go
mschoch: So I did break this out as a separate issue #763.  However I now believe it is correct, and the comment was trying to explain why.  The persister DID persist the snapshot containing the internal values.  What it is NOT doing is introducing a new root.  And the reason is that it does not have to.  The only reason to introduce a new root is if we want to swap in some newly persisted zap segments, but we don't have any of those in this case.  The new internal values are already available to readers through the work done by the introducer, and the persister did it's job to persist them in the root bolt with this snapshot.  So, for the case of internal only changes, there is nothing more to be done.

PR #750, opened 2018-02-02
Fix for keeping the number of merge tasks within the calculated budget

sreekanth-cb:
Adjusting the merge task creation loop to accommodate the newly merged segments so that the eventual merge results/ number of segments stay within the calculated budget.

### DIFF
--- a/index/scorch/mergeplan/merge_plan.go
+++ b/index/scorch/mergeplan/merge_plan.go
@@ -186,8 +186,7 @@ func plan(segmentsIn []Segment, o *MergePlanOptions) (*MergePlan, error) {
 
 	// While we’re over budget, keep looping, which might produce
 	// another MergeTask.
-	newlyMerged := int(0)
-	for (len(eligibles) + newlyMerged) > budgetNumSegments {
+	for len(eligibles) > 0 && (len(eligibles)+len(rv.Tasks)) > budgetNumSegments {
 		// Track a current best roster as we examine and score
 		// potential rosters of merges.
 		var bestRoster []Segment
@@ -223,7 +222,6 @@ func plan(segmentsIn []Segment, o *MergePlanOptions) (*MergePlan, error) {
 		rv.Tasks = append(rv.Tasks, &MergeTask{Segments: bestRoster})
 
 		eligibles = removeSegments(eligibles, bestRoster)
-		newlyMerged++
 	}
 
 	return rv, nil
--- a/index/scorch/mergeplan/merge_plan.go
+++ b/index/scorch/mergeplan/merge_plan.go
@@ -186,13 +186,14 @@ func plan(segmentsIn []Segment, o *MergePlanOptions) (*MergePlan, error) {
 
 	// While we’re over budget, keep looping, which might produce
 	// another MergeTask.
-	for len(eligibles) > budgetNumSegments {
+	newlyMerged := int(0)
+	for (len(eligibles) + newlyMerged) > budgetNumSegments {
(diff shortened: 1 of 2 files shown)


steveyen:
Hi Sreekanth --

First off, some minor housekeeping things...

Please "squash" these two commits together into one (take a look at 'git rebase -i' command)?

Also, I've heard that commit titles should be < 70 chars?  Not sure of the exact preferred length, and that probably differs depending on different projects, but definitely want to keep the commit title on one line instead of clipped across many ("Adjusting the merge task creation loop to accomodate...").  How about something like "keep merge tasks within budget"?

But, bigger worry is from the updated unit tests, I wonder if it's doing the right thing.  Given the DefaultMergePlanOptions, and if there are only 2 or 3 segments, I'm wondering if it makes sense that the planner decides that they should be merged?  Hmmm.

sreekanth-cb:
Hi @steveyen , the DefaultMergePlanOptions recommends that merge, isn't it?
Or I got how the confusion comes here, 
SegmentsPerMergeTask = 10, I think we should ideally consider this as the "MaxSegmentsPerMergeTask" , else there will be 'N'  segments ( 1 <= N <=SegmentsPerMergeTask) left un merged. That is the main the driver for this change. 

The intention is that the merge plan should favour to form segments which are at least as big as the tierSize at each floor. Here in UT, these segments are well below the tierSize recommended for a segment and hence should be merged?

Had these 2 segments in UT are of a cumulative size between (1 x tierSize) and (2 x tierSize), then nothing should have changed, as the budget would recommend to have 2 segments which we already have. So no merge tasks.

Not sure whether I am missing something here?

sreekanth-cb:
Sure @steveyen , I shall change the commit msgs. Thanks for reminding me one more time.

But would like to keep these commits separate as they are kind of touching different area. 
For this merge_plan, I have few more change proposals in line with the merged segments aligning to a  staircase format in terms of eventual size and making the count of merged segments predictable. [right now we are not considering the tierSizes during the merge task creation]

So lets keep all these merge related things to a separate PR than the file deletion fix.?

mschoch:
@sreekanth-cb i'm not sure if you're looking at the same change @steveyen is talking about.  There are 2 commits, one is just called "cleanup" and seems to undo part of the original change, I agree with Steve here.  I think perhaps you're confusing it with another PR, but Steve is NOT suggesting you combine the PRs, rather just the 2 commits on this single PR.

steveyen:
On the math-y parts of the commentary thread...

> ... SegmentsPerMergeTask = 10, I think we should ideally consider this as the "MaxSegmentsPerMergeTask" ...

I also sometimes think of SegmentsPerMergeTask as meaning MaxSegmentsPerMergeTask.

> ... else there will be 'N' segments ( 1 <= N <=SegmentsPerMergeTask) left un merged. That is the main the driver for this change.

Some alternate point of view here.  Having some segments left un-merged is okay... like if there's only 1 or 2 (or 3) segments and they're all summing up to under the budget, why choose to merge them together?

> The intention is that the merge plan should favour to form segments which are at least as big as the tierSize at each floor. Here in UT, these segments are well below the tierSize recommended for a segment and hence should be merged?

I see -- that's the intention of this change, but I guess I'm uncomfortable with that intention or favoring... where I'd take the stance that undersized segments don't have to be merged.

sreekanth-cb:
yes..
But merging undersized segments saves FDs and possibly search perf as well by saving some extra file system calls?

sreekanth-cb:
closing this as it looks easier to  start one afresh!

PR #751, opened 2018-02-05
fix for merger persister handshake stalemate

sreekanth-cb:
The slow merger was lagging behind the fast persister
to a persister notify send-loop while the persister awaits
for any new introductions from introducer totally blocking
the merger

This fix along with the deleted files eligibilty flipping
makes the file count to around 6 to 11 files  per shard
for both travel and beer samples

Need to understand and fix the rollback test failure.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -39,15 +39,29 @@ type notificationChan chan struct{}
 func (s *Scorch) persisterLoop() {
 	defer s.asyncTasks.Done()
 
-	var notifyChs []notificationChan
+	var persistWatchers []*epochWatcher
 	var lastPersistedEpoch uint64
+
+	notifyWatchers := func() {
+		var watchersNext []*epochWatcher
+		for _, w := range persistWatchers {
+			if w.epoch < lastPersistedEpoch {
+				close(w.notifyCh)
+			} else {
+				watchersNext = append(watchersNext, w)
+			}
+		}
+		persistWatchers = watchersNext
+	}
+
 OUTER:
 	for {
 		select {
 		case <-s.closeCh:
 			break OUTER
-		case notifyCh := <-s.persisterNotifier:
-			notifyChs = append(notifyChs, notifyCh)
+		case ew := <-s.persisterNotifier:
+			persistWatchers = append(persistWatchers, ew)
+			notifyWatchers()
 		default:
 		}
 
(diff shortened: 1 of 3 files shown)


index/scorch/persister.go
```diff
@@ -113,13 +114,27 @@ OUTER:
 		case s.introducerNotifier <- w:
 		}
 
-		s.removeOldData() // might as well cleanup while waiting
+		s.removeOldData() // this looks enough as of now
```
mschoch: i understood the old comment, but i don't understand the new one

index/scorch/persister.go
```diff
@@ -39,15 +39,15 @@ type notificationChan chan struct{}
...
+	var persistWatchers []*epochWatcher
 	var lastPersistedEpoch uint64
 OUTER:
 	for {
 		select {
 		case <-s.closeCh:
 			break OUTER
 		case notifyCh := <-s.persisterNotifier:
-			notifyChs = append(notifyChs, notifyCh)
+			persistWatchers = append(persistWatchers, notifyCh)
```
mschoch: i'm a bit confused as to why here we unconditionally append to persistWatchers, but down at the end we apply logic to the same operation

sreekanth-cb: yea..we should ideally handle it the same way here too..

index/scorch/persister.go
sreekanth-cb: hm, I didn't fully understood the original comment. :) As there is no waiting there right...we just happen to do this in between the persister notification sending and receiving the response from it..Those two could just happen within a speck time and we are not doing anything in while wait loop, isn't it. 
And made the comment this way, in context of our the file count handling document. A thread there talks about 

while(condition) {

s.removeOldData() 
until few of the break conditions met...
}
So as of now, it looks like we could manage with a single call than in a loop until certain conditions are met.

mschoch: the "waiting" refers to the fact that we just notified the introducer that we're all caught up to a certain point, and we're about to hit a select statement that will "wait" for notification of more work to be done.  the duration of the wait may be 0 though.

i don't understand the while loop you posted here, if this actual code could you link to it?

finally, the words "this looks enough as of now" doesn't make any sense to me, like if you walked up to me on the street, pointed at something and said those words, i would have no idea...

sreekanth-cb:
@mschoch , @steveyen , ready for final review.

PR #754, opened 2018-02-06
tuning the edge for merge-task execution loop

sreekanth-cb:
Adjusting the merge task creation loop to accommodate
the newly merged segments so that the eventual merge
results/ number of segments stay within the calculated budget.

The inner loop fix handles the smaller left over segments that 
were left un merged earlier.

### DIFF
--- a/index/scorch/mergeplan/merge_plan.go
+++ b/index/scorch/mergeplan/merge_plan.go
@@ -186,13 +186,13 @@ func plan(segmentsIn []Segment, o *MergePlanOptions) (*MergePlan, error) {
 
 	// While we’re over budget, keep looping, which might produce
 	// another MergeTask.
-	for len(eligibles) > budgetNumSegments {
+	for len(eligibles) > 0 && (len(eligibles)+len(rv.Tasks)) > budgetNumSegments {
 		// Track a current best roster as we examine and score
 		// potential rosters of merges.
 		var bestRoster []Segment
 		var bestRosterScore float64 // Lower score is better.
 
-		for startIdx := 0; startIdx < len(eligibles)-o.SegmentsPerMergeTask; startIdx++ {
+		for startIdx := 0; startIdx < len(eligibles); startIdx++ {
 			var roster []Segment
 			var rosterLiveSize int64
 
--- a/index/scorch/mergeplan/merge_plan_test.go
+++ b/index/scorch/mergeplan/merge_plan_test.go
@@ -73,7 +73,16 @@ func TestSimplePlan(t *testing.T) {
 				segs[2],
 			},
 			nil,
-			&MergePlan{},
+			&MergePlan{
+				Tasks: []*MergeTask{
+					&MergeTask{
+						Segments: []Segment{
+							segs[2],
+							segs[1],
+						},
+					},
+				},
+			},
 			nil,
 		},
 		{"3 segments",
@@ -83,7 +92,17 @@ func TestSimplePlan(t *testing.T) {
 				segs[9],
 			},
 			nil,
-			&MergePlan{},
+			&MergePlan{
+				Tasks: []*MergeTask{
+					&MergeTask{
+						Segments: []Segment{
+							segs[9],
+							segs[2],
+							segs[1],
+						},
+					},
+				},
+			},
 			nil,
 		},
 		{"many segments",


sreekanth-cb:
@steveyen , finally we saw one snapshot rollback error surfaces in CI.
Think this one failure (seen only in 1.6.x go version) "TestIndexRollback (0.03s) snapshot_rollback_test.go:145: unexpected doc count: 5" failure will get addressed in the PR #753  ?

steveyen:
> Think this one failure (seen only in 1.6.x go version) "TestIndexRollback (0.03s) snapshot_rollback_test.go:145: unexpected doc count: 5" failure will get addressed in the PR #753 ?

Oh, that's a new one / never seen that one before on my machine.  It might be fixed by PR #753, but not sure, as I don't have an explanation on how raciness could lead to a doc count of 5, even thought that's clearly happening and it looks race-y (other runs have passed, albeit for different go versions).

mschoch:
Can someone link to the failure?  I tried to find it in the history and couldn't

steveyen:
> Can someone link to the failure? I tried to find it in the history and couldn't

Whoops, my apologies -- I had pressed the "Restart" build button on travis-CI to see if another run would be reproducible.  But, I too can no longer find the previous run log that had the failure for go 1.6.

The more I think about it, too the bigger the mystery...

> snapshot_rollback_test.go:145: unexpected doc count: 5

Indeed... a doc count of 5 is quite unexpected given the rollback test, where the 1st batch inserts 2 docs, then the 2nd batch inserts 2 new docs and deletes 1 old doc.  It's not popping to mind right now for me how various concurrent goroutine interleavings can somehow add up to a doc count of 5.

This also looks like a separate mystery not really related to Sreekanth's PR #754, except that #754 just happened to be lucky enough to catch it in its travis-CI run.

mschoch:
So, I spent some time trying to theorize how we could ever get 5...

I'm not sure I have it yet, but I think I have identified one issue.  The introduction of segments and the introduction of rollbacks both come in through the introducer.  If you remember, prior to the introduction of segments, some optimistic work is done.  Then during introduction we check that optimistic work and possibly do some fixup.  But, I think if a rollback is applied, the fixup phase for the optimistic work isn't enough (ie, the optimistic work is wrong, not just incomplete).  A similar problem may face the merger (though I haven't thought about it yet).

It almost seems like if we decide to rollback, it should flush the pipeline of not yet applied things (returning error in the case of batches, and just dropping the merges).

I still don't see how this could have happened to get us 5, because we're waiting for the Batch operations to return, so they should already be applied before Rollback starts...

mschoch:
Another observation, having NumSnapshotsToKeep set to 1 (default) is kind of silly with the persister introducing snapshots now.  Because in the steady state (idle) the current snapshot, and the one right before it are identical (and both are persisted).  Probably one fix is for the persister to be able to more concretely identify snapshots it introduced and not persist them, since we know they are redundant (and now interfere with snapshot cleanup logic)

mschoch:
Oh, I should have mentioned that is the source of one failure in this unit test (not the one with doc count 5 though).  On my system, on approximately one out of every 3 runs, the snapshot we're trying to rollback to is already cleaned up.  This frequency could be affected by the logging I've added.

steveyen:
> NumSnapshotsToKeep set to 1 (default) is kind of silly with the persister introducing snapshots now

Pre-coffee thought... maybe a low (or 1?) default for NumSnapshotsToKeep is an okay number of out of the box bleve users... basically, rollback-ability is effectively disabled by default.   But it's one of those advanced options that users (e.g., cbft) can choose to pay for.

steveyen:
On your thoughts into the mysterious 5 docs, yah, it feels like there's something there to the path of thought...

> It almost seems like if we decide to rollback, it should flush the pipeline of not yet applied things (returning error in the case of batches, and just dropping the merges).

Also, one other thing is Rollback() currently might not wait or block until it's done, if unsafe_batch is true.  I have the feeling that Rollback should instead always fully block until it's really done.  I don't think this is an answer to the 5 docs mystery, but just yet another to-do item that popped to mind while looking through the codepaths on this.

mschoch:
I'm not arguing against 1 being the correct value for NumSnapshotsToKeep, I agree it is a reasonable default value.  Rather, I'm arguing that our current implementation has rendered this value useless.   It literally keeps 1 extra snapshot which is of no value (in most cases).

mschoch:
Reading your comment a different way now, I think there is confusion around the behavior of NumSnapshotsToKeep.  If it in fact only ever keeps 1 snapshot, then our looping doesn't make it any worse.  Instead, the rollback test itself is flawed because with 1 snapshot you can never reliably rollback (you're racing the delete).

mschoch:
The Rollback test doesn't set unsafe_batch to true, so I'm assuming it's operating in the most safe wait for persistence mode (seems to be confirmed by my logging).

I think your comment is problematic because there isn't a well defined notion of "really done".  The unsafe_batch flag is specifically to try and accommodate these different perspectives.

steveyen:
> Oh, I should have mentioned that is the source of one failure in this unit test (not the one with doc count 5 though). On my system, on approximately one out of every 3 runs, the snapshot we're trying to rollback to is already cleaned up.

Yep, the other PR #753 is about trying to remove raciness from TestIndexRollback to address "the snapshot we're trying to rollback to is already cleaned up".

mschoch:
Looking at that closer though it doesn't feel like a good way to test rollback.  I don't think thats how a client would actually use rollback (by holding onto something with a slow reader).

steveyen:
> Looking at that closer though it doesn't feel like a good way to test rollback. I don't think thats how a client would actually use rollback (by holding onto something with a slow reader).

Good point.  I'd probably go with an additional test case, then, (a separate PR?) that doesn't have a slow reader.

My thinking with #753 was to trick out all the scenario conditions so that for sure Rollback was going to happen... e.g., no more raciness, like via a slow reader.

Also, even more test cases now pop to mind... like a negative test case where for sure the Rollback ought to fail as the old rollback point was already purged by removeOldData().

mschoch:
The reason this is most troubling to me is that in order to get to 5 docs, it can't just be that the deleted bits are flipped somehow.  You have to end up with an additional or duplicated segment in your snapshot.  And that would represent a huge violation of the way we understand the system.  Further, the rollback code is pretty lean, so it seems like there is a good chance this glitch isn't limited to rollback, but perhaps is on the regular code path.

steveyen:
Yep -- just to get this all straight in my head, as there's a lot of good stuff in this thread, some recap...

A - this PR #754 looks good to me by itself -- it's ready to merge, imho.

B - there's occasional TestIndexRollback failures, orthogonal to this PR #754 and seen elsewhere, due to raciness where the rollback test occasionally fails because the target snapshot no longer exists.  PR #753 has a proposed fix to reduce raciness in the test.

B.1 - there's feedback that PR #753 can be even better -- that there can be even more or improved rollback testing (e.g., don't hold a slow reader; more negative test cases) that's probably worth even more PR's.

C - NumSnapshotsToKeep -- there's an issue or nascent subthread on what to do with NumSnapshotsToKeep -- is it right given recent persister changes and what to do about it?

D - there's a proposal that Rollback() should always be blocking or fully synchronous --
 don't return until Rollback is either fully done or hits an error.  e.g., the Rollback() implementation shouldn't look at the unsafe_batch flag (i.e., app probably does not want EventuallyRollback() behavior).

E - there's a proposal that rollback should blow away inflight merges and error on inflight batches

F - "5 docs" -- the big one (imho), there's the very troubling mystery of "5 docs" (how the hell can there be 5 docs?).  This is currently seemingly rare (only seen once in the universe so far) but quite worrisome.  As @mschoch notes, "seems like there is a good chance this glitch isn't limited to rollback, but perhaps is on the regular code path".  Might also be or might not be related to E.

Finally, meta thing, I worry if I hit the big green "Merge pull request" button on this PR #754, will github close this down and prevent more commentary?  Or, perhaps it's time to split some of the B-thru-F into separate issues or other channels.

mschoch:
I still disagree with D.  The unsafe_batch flag does not mean eventually rollback.  The rollback has happened when it returns.  All it means is that you could still crash and wake up as if the rollback never happened.  It's not clear to me that this is bad, in the case of FTS it would simply have to rollback again..

mschoch:
Let's create separate issues for each of these, even if some of them are disputed.

mschoch:
I have broken out several new issues so this one can proceed:

A - this PR
B - already in #753 
B.1 - #762
C - #761
D - #760 
E  - #760 
F - #759 

Based on this list I'll do final review and this PR can proceed.

PR #757, opened 2018-02-07
change indexMeta to public

mosuka:
I'd like to use indexMeta from outside of bleve package for controlling index_mapping.json and index_meta.json.

For example:

```
indexPath := "./data"

indexMappingFile, err := os.Open("index_mapping.json")
indexMappingBytes, err := ioutil.ReadAll(indexMappingFile)
var indexMapping *mapping.IndexMappingImpl
err = json.Unmarshal(indexMappingBytes, indexMapping)
if err != nil {
	fmt.Println(err)
	return
}

indexMetaFile, err := os.Open("index_meta.json")
indexMetaBytes, err := ioutil.ReadAll(indexMetaFile)
var indexMeta *bleve.IndexMeta
err = json.Unmarshal(indexMetaBytes, indexMeta)
if err != nil {
	fmt.Println(err)
	return
}

index, err = bleve.NewUsing(indexPath, indexMapping, indexMeta.IndexType, indexMeta.Storage, indexMeta.Config)
```

I propose to change indexMeta to public. How about this PR?

### DIFF
--- a/index_impl.go
+++ b/index_impl.go
@@ -39,7 +39,7 @@ import (
 type indexImpl struct {
 	path  string
 	name  string
-	meta  *indexMeta
+	meta  *IndexMeta
 	i     index.Index
 	m     mapping.IndexMapping
 	mutex sync.RWMutex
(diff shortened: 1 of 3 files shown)


mschoch:
IndexMeta is intended to be completely opaque to the applications using bleve.  It's just supposed to record the bare minimum info needed to properly reopen the index later.

Your request to make it public includes no justification for the change.  Can you elaborate on why you want this change made?

mosuka:
Hi @mschoch ,
Thank you for explaining IndexMeta.

When opening the index you can specify IndexMapping, indexType, storage, and config. So, I thought it better to have indexMeta treated as public like IndexMapping.
Currently, I'm working on deploy applications using bleve with IndexMapping in JSON file. Similarly, so when trying to manage indexType, storage and config in JSON file myself, I noticed that there is indexMeta that holds them.
If it already have indexMeta in bleve package, I thought that having it public made it better than preparing a similar struct myself.

PR #771, opened 2018-02-12
Fix for empty segment merge handling 

sreekanth-cb:
Avoid creating new segment files with empty segments
tasks during the merge operation, avoid the
incorrect appending of a newer segment to snapshot
during such merge introductions.

Issue can be consistently reproduced upon loading 
travel sample with 80K docs. Budget Planning and 
persistence is falling into an infinite loop with
files keep created all the time.

UT update pending..

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -193,6 +193,12 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 	// prepare new index snapshot
 	currSize := len(s.root.segment)
 	newSize := currSize + 1 - len(nextMerge.old)
+
+	// empty segments deletion
+	if nextMerge.new == nil {
+		newSize--
+	}
+
 	newSnapshot := &IndexSnapshot{
 		parent:   s,
 		segment:  make([]*SegmentSnapshot, 0, newSize),
@@ -210,7 +216,7 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 		segmentID := s.root.segment[i].id
 		if segSnapAtMerge, ok := nextMerge.old[segmentID]; ok {
 			// this segment is going away, see if anything else was deleted since we started the merge
-			if s.root.segment[i].deleted != nil {
+			if segSnapAtMerge != nil && s.root.segment[i].deleted != nil {
 				// assume all these deletes are new
 				deletedSince := s.root.segment[i].deleted
 				// if we already knew about some of them, remove
@@ -238,14 +244,16 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 		}
 	}
 
-	// put new segment at end
-	newSnapshot.segment = append(newSnapshot.segment, &SegmentSnapshot{
-		id:         nextMerge.id,
-		segment:    nextMerge.new, // take ownership for nextMerge.new's ref-count
-		deleted:    newSegmentDeleted,
-		cachedDocs: &cachedDocs{cache: nil},
-	})
-	newSnapshot.offsets = append(newSnapshot.offsets, running)
+	if nextMerge.new != nil {
+		// put new segment at end
+		newSnapshot.segment = append(newSnapshot.segment, &SegmentSnapshot{
+			id:         nextMerge.id,
+			segment:    nextMerge.new, // take ownership for nextMerge.new's ref-count
+			deleted:    newSegmentDeleted,
+			cachedDocs: &cachedDocs{cache: nil},
+		})
+		newSnapshot.offsets = append(newSnapshot.offsets, running)
+	}
 
 	// swap in new segment
 	rootPrev := s.root
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -100,8 +100,8 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 	// prepare new index snapshot
 	newSnapshot := &IndexSnapshot{
 		parent:   s,
-		segment:  make([]*SegmentSnapshot, nsegs, nsegs+1),
-		offsets:  make([]uint64, nsegs, nsegs+1),
+		segment:  make([]*SegmentSnapshot, 0, nsegs+1),
+		offsets:  make([]uint64, 0, nsegs+1),
 		internal: make(map[string][]byte, len(s.root.internal)),
 		epoch:    s.nextSnapshotEpoch,
 		refs:     1,
@@ -124,24 +124,29 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 				return err
 			}
 		}
-		newSnapshot.segment[i] = &SegmentSnapshot{
+
+		newss := &SegmentSnapshot{
 			id:         s.root.segment[i].id,
 			segment:    s.root.segment[i].segment,
 			cachedDocs: s.root.segment[i].cachedDocs,
 		}
-		s.root.segment[i].segment.AddRef()
-
+		
 		// apply new obsoletions
 		if s.root.segment[i].deleted == nil {
-			newSnapshot.segment[i].deleted = delta
+			newss.deleted = delta
 		} else {
-			newSnapshot.segment[i].deleted = roaring.Or(s.root.segment[i].deleted, delta)
+			newss.deleted = roaring.Or(s.root.segment[i].deleted, delta)
 		}
-
-		newSnapshot.offsets[i] = running
-		running += s.root.segment[i].Count()
-
+		
+		// check for live size before copying
+		if newss.LiveSize() > 0 {
+			newSnapshot.segment = append(newSnapshot.segment, newss)
+			s.root.segment[i].segment.AddRef()
+			newSnapshot.offsets = append(newSnapshot.offsets, running)
+			running += s.root.segment[i].Count()
+		}
 	}
+
 	// append new segment, if any, to end of the new index snapshot
 	if next.data != nil {
 		newSegmentSnapshot := &SegmentSnapshot{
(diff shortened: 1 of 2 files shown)


index/scorch/merge.go
```diff
@@ -127,42 +127,61 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot) error {
...
+				oldNewDocNums: make(map[uint64][]uint64),
+				new:           segment,
+				notify:        make(notificationChan),
+			}
+
+			for i, segNewDocNums := range newDocNums {
+				sm.oldNewDocNums[task.Segments[i].Id()] = segNewDocNums
+			}
+
+		} else if emptySegmentTask {
```
steveyen: wondering if you need the emptySegmentTask?  Or, would comparing len(task.Segments) > 0 be enough or the same info?

index/scorch/introducer.go
```diff
@@ -209,19 +215,21 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
...
-				// if we already knew about some of them, remove
-				if segSnapAtMerge.deleted != nil {
-					deletedSince = roaring.AndNot(s.root.segment[i].deleted, segSnapAtMerge.deleted)
-				}
-				deletedSinceItr := deletedSince.Iterator()
-				for deletedSinceItr.HasNext() {
-					oldDocNum := deletedSinceItr.Next()
-					newDocNum := nextMerge.oldNewDocNums[segmentID][oldDocNum]
-					newSegmentDeleted.Add(uint32(newDocNum))
+			if segSnapAtMerge != nil {
```
steveyen: maybe fold this into the previous line so there's some less "indentation" of deep blocks, like...

    if segSnapAtMerge, ok := ...; ok && segSnapAtMerge != nil {

sreekanth-cb: one reason for doing it this way was to avoid the `else` part of the if check.
if segSnapAtMerge, ok := ...; ok { 
}. Keeping it this way let us skip the else part of appending to the new snapshot created by getting into the `if` part and not doing anything. 
not sure whether there is a cleaner way to do this..Or I may be missing something...?

index/scorch/merge.go
sreekanth-cb: yep, looks enough..

index/scorch/introducer.go
steveyen: i see -- i read the nesting of if-else incorrectly -- you're right.

index/scorch/merge.go
```diff
@@ -132,37 +132,54 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot) error {
 			if segSnapshot, ok := planSegment.(*SegmentSnapshot); ok {
 				oldMap[segSnapshot.id] = segSnapshot
 				if zapSeg, ok := segSnapshot.segment.(*zap.Segment); ok {
-					segmentsToMerge = append(segmentsToMerge, zapSeg)
+					if segSnapshot.LiveSize() == 0 {
+						oldMap[segSnapshot.id] = nil
+					} else {
+						segmentsToMerge = append(segmentsToMerge, zapSeg)
+					}
 					docsToDrop = append(docsToDrop, segSnapshot.deleted)
```
steveyen: wondering if the maintaining the docsToDrop needs to be moved into line 138.5?

That is, segmentsToMerge and docsToDrop need to have the same lengths... and the same positional pairwise entries?

index/scorch/introducer.go
steveyen: instead, maybe smush the if-checks on line 218 and 220?

not a big thing, but deeper indentation or nesting of the IF blocks increases some cognitive load

index/scorch/merge.go
```diff
@@ -132,37 +132,54 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot) error {
 			if segSnapshot, ok := planSegment.(*SegmentSnapshot); ok {
 				oldMap[segSnapshot.id] = segSnapshot
 				if zapSeg, ok := segSnapshot.segment.(*zap.Segment); ok {
-					segmentsToMerge = append(segmentsToMerge, zapSeg)
+					if segSnapshot.LiveSize() == 0 {
```
mschoch: maybe i'm misunderstanding something, but this feels like an inconvenient way to drop segments that have live size of 0.  what i mean is, at this point in the code, we waited for the merger to plan to do something with a segment (combine it with some others) and now while walking through that list, we happen to notice that it's live size is 0.  we still don't do anything directly about it, but was pass it as nil in the old map, so that the introducer will end up dropping it when introducing this merge.

wouldn't it be easier if as a general rule, the introducer ALWAYS dropped segments with live size of 0.  not just for merges, like any operation it performs that copies over a root and creates a new one it also performs livesize=- dropping.  if we did that, this check here wouldn't be needed, as we would expect livesize=0 segments be dropped before the merger sees them.

index/scorch/merge.go
mschoch: agree with steve, if segmentsToMerge and docsToDrop are not the same lenth, merge will do the wrong thing.

index/scorch/merge.go
```diff
@@ -132,37 +132,54 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot) error {
...
+				oldNewDocNums: make(map[uint64][]uint64),
+				new:           segment,
+				notify:        make(notificationChan),
+			}
+
+			for i, segNewDocNums := range newDocNums {
+				sm.oldNewDocNums[task.Segments[i].Id()] = segNewDocNums
+			}
+
+		} else if len(task.Segments) > 0 {
```
mschoch: this else if case seems to append a very similar structure, seems like the code could be rearranged to do this a simpler way.

index/scorch/merge.go
```diff
@@ -132,37 +132,54 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot) error {
...
+			sm = &segmentMerge{
+				id:            newSegmentID,
+				old:           oldMap,
+				oldNewDocNums: nil,
+				new:           nil,
+				notify:        make(notificationChan),
+			}
 		}
 
+		notifications = append(notifications, sm.notify)
```
mschoch: related to the comment above, here we append to notifications, even if we didn't go through the if or else-if cases, which seems like a bug

index/scorch/merge.go
sreekanth-cb: thats a very valid point here... we should add that logic in introducer too.
But since merge planner has a logic to pass in a task with empty segments, the logic for handling of any given tasks/segments has to be present here as well..

index/scorch/introducer.go
```diff
@@ -124,23 +125,29 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
...
+			newSnapshot.segment[index].deleted = delta
 		} else {
-			newSnapshot.segment[i].deleted = roaring.Or(s.root.segment[i].deleted, delta)
+			newSnapshot.segment[index].deleted = roaring.Or(s.root.segment[i].deleted, delta)
 		}
 
-		newSnapshot.offsets[i] = running
-		running += s.root.segment[i].Count()
-
+		// check for live size before copying
```
mschoch: here the comment says "before copying", but you already did "copy" on line 128 right?
it seems like the else block then tries to undo this? but the pattern here

x = append(x[:offset],x[offset+1:]...)

is used in a general purpose remove element from slice operation, but i don't understand why we use that here, we're going through the old root, and building the new snapshot slices one element at a time, there shouldn't be anything after "index" in the new segment yet right?

index/scorch/introducer.go
```diff
@@ -238,14 +257,32 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
...
+			for obsoletedIter.HasNext() {
+				oldDocNum := obsoletedIter.Next()
+				newDocNum := nextMerge.oldNewDocNums[segID][oldDocNum]
+				newSegmentDeleted.Add(uint32(newDocNum))
+			}
+		}
+	}
+
+	if nextMerge.new != nil &&
+		nextMerge.new.Count() > newSegmentDeleted.GetCardinality() {
```
mschoch: i don't understand this condition, we're checking the size of the new segment, with the number of newly deleted items?  is this to catch the case that by the time the merge happened, all of the docs are actually now deleted (by doing these additional checks) and now there is no point in adding it?

if so, i'm wondering if there is a clearer way to express this, i had to sit and really think about the math to even imagine thats what it was doing

sreekanth-cb: Yes. The intent is to capture the case you mentioned. What if all the docs in the newly merged segment got deleted by the time we reach here. Then no point in adding. I agree that these may be confusing.. Shall we add some comments explaining the same at line:274?

index/scorch/introducer.go
sreekanth-cb: yea..copying completes by setting the right ref count and setting the remaining fields. But before this final step if the live size happen to be zero, then we need to unset the assignment made at line:128 at index position.  For eg: what if the index assignment  made  to the last element/segment which later turned to have a zero live size (line:142). In that case,  we have to remove the last zero sized segment at index position. 
I can see that its due to the way this is coded up.. May be we can assign that segment to a temp variable, apply deletes , check live size and insert into the new snapshot slice.

index/scorch/introducer.go
```diff
@@ -110,6 +110,7 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 
 	// iterate through current segments
 	var running uint64
+	var index int
```
mschoch: one last thought, i think if we initialized segments and offsets with size 0 and capacity nsegs+1,
then we could simply use append, instead of need a new variable 'index'
also, this then removes the block of code at the end which tries to trim down the size of the slice

sreekanth-cb: thanks @mschoch , this is a new go learning for me that size 0 helps in unnecessary allocation towards reaching capacity when we know the exact max capacity!

PR #776, opened 2018-02-13
Estimating RAM needed for a Search operation

abhinavdangeti:
The intention here is to estimate the amount of memory needed to execute a
search query over an index prior to actually running the search operation.

Sample estimates for certain queries:
{Size: 10, BenchmarkUpsidedownSearchOverhead}

    ESTIMATE     BENCHMEM       TEST
    4616          4796          TermQuery
    5210          5405          MatchQuery
    7700          8447          DisjunctionQuery (of match queries)
    6514          6591          DisjunctionQuery (of term queries)
    7524          8175          ConjunctionQuery (of match queries)
    10306         10708         DisjunctionQuery (disjunction of disjunctions)
    …

### DIFF
--- a/index_impl.go
+++ b/index_impl.go
@@ -362,8 +362,59 @@ func (i *indexImpl) Search(req *SearchRequest) (sr *SearchResult, err error) {
 	return i.SearchInContext(context.Background(), req)
 }
 
+// memNeededForSearch is a helper function that returns an estimate of RAM
+// needed to execute a search request.
+func memNeededForSearch(req *SearchRequest,
+	searcher search.Searcher,
+	topnCollector *collector.TopNCollector) uint64 {
+
+	backingSize := req.Size + req.From + 1
+	if req.Size+req.From > collector.PreAllocSizeSkipCap {
+		backingSize = collector.PreAllocSizeSkipCap + 1
+	}
+	numDocMatches := backingSize + searcher.DocumentMatchPoolSize()
+
+	estimate := 0
+
+	// overhead, size in bytes from collector
+	estimate += topnCollector.Size()
+
+	var dm search.DocumentMatch
+	sizeOfDocumentMatch := dm.Size()
+
+	// pre-allocing DocumentMatchPool
+	var sc search.SearchContext
+	estimate += sc.Size() + numDocMatches*sizeOfDocumentMatch
+
+	// searcher overhead
+	estimate += searcher.Size()
+
+	// overhead from results, lowestMatchOutsideResults
+	estimate += (numDocMatches + 1) * sizeOfDocumentMatch
+
+	// additional overhead from SearchResult
+	var sr SearchResult
+	estimate += sr.Size()
+
+	// overhead from facet results
+	if req.Facets != nil {
+		var fr search.FacetResult
+		estimate += len(req.Facets) * fr.Size()
+	}
+
+	// highlighting, store
+	var d document.Document
+	if len(req.Fields) > 0 || req.Highlight != nil {
+		for i := 0; i < (req.Size + req.From); i++ { // size + from => number of hits
+			estimate += (req.Size + req.From) * d.Size()
+		}
+	}
+
+	return uint64(estimate)
+}
+
 // SearchInContext executes a search request operation within the provided
-// Context.  Returns a SearchResult object or an error.
+// Context. Returns a SearchResult object or an error.
 func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr *SearchResult, err error) {
 	i.mutex.RLock()
 	defer i.mutex.RUnlock()
--- a/index_impl.go
+++ b/index_impl.go
@@ -50,6 +50,10 @@ const storePath = "store"
 
 var mappingInternalKey = []byte("_mapping")
 
+const SearchMemCheckCallbackKey = "_search_mem_callback_key"
+
+type SearchMemCheckCallbackFn func(size uint64) error
+
 func indexStorePath(path string) string {
 	return path + string(os.PathSeparator) + storePath
 }
@@ -477,6 +481,15 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 			}
 		}
 		collector.SetFacetsBuilder(facetsBuilder)
+	}
+
+	if memCb := ctx.Value(SearchMemCheckCallbackKey); memCb != nil {
+		if memCbFn, ok := memCb.(SearchMemCheckCallbackFn); ok {
+			err = memCbFn(memNeededForSearch(req, searcher, collector))
+		}
+	}
+	if err != nil {
+		return nil, err
 	}
 
 	err = collector.Collect(ctx, searcher, indexReader)
--- a/search/facets_builder.go
+++ b/search/facets_builder.go
@@ -15,10 +15,31 @@
 package search
 
 import (
+	"reflect"
 	"sort"
 
 	"github.com/blevesearch/bleve/index"
+	"github.com/blevesearch/bleve/size"
 )
+
+var reflectStaticSizeFacetsBuilder int
+var reflectStaticSizeFacetResult int
+var reflectStaticSizeTermFacet int
+var reflectStaticSizeNumericRangeFacet int
+var reflectStaticSizeDateRangeFacet int
+
+func init() {
+	var fb FacetsBuilder
+	reflectStaticSizeFacetsBuilder = int(reflect.TypeOf(fb).Size())
+	var fr FacetResult
+	reflectStaticSizeFacetResult = int(reflect.TypeOf(fr).Size())
+	var tf TermFacet
+	reflectStaticSizeTermFacet = int(reflect.TypeOf(tf).Size())
+	var nrf NumericRangeFacet
+	reflectStaticSizeNumericRangeFacet = int(reflect.TypeOf(nrf).Size())
+	var drf DateRangeFacet
+	reflectStaticSizeDateRangeFacet = int(reflect.TypeOf(drf).Size())
+}
 
 type FacetBuilder interface {
 	StartDoc()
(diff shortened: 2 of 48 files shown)


steveyen:
Quite a wide reaching change!

It has that first glance feel of brittleness, though, in that with any change to a struct, the developer needs to remember to manually change the SizeInBytes() math.

Have you considered using Sizeof(), either via the unsafe package or via the reflect package?

That might have the benefit of a) less brittleness to changes (less bugs for future developer), and b) automatic accounting for extra bytes that golang might add (ex: alignment padding, etc).

mschoch:
@steveyen I've actually advised the opposite in the past because relying on unsafe brings that code smell and inability to run in hosted environments like GAE.

My thinking with regards to runtime changing is that it's just an estimate, if lets say Go 1.11 increases the size, we just increase our size accordingly and we over-estimate in the old runtimes.

With respect to brittleness, I have no good solution.

steveyen:
> @steveyen I've actually advised the opposite in the past because relying on unsafe brings that code smell and inability to run in hosted environments like GAE.

Good point.  Instead of unsafe, perhaps the runtime.Type.Size() would do the trick, to avoid the unsafe smell?

mschoch:
@steveyen so my thought was that the performance penalty of using reflection was going to be too high.

steveyen:
one thought that might help reduce brittleness and also avoid a perf penalty for reflect is to have a init() func somewhere that uses reflect to populate an immutable lookup table of sizes for all the static stuff from the various structs... (e.g., add up the sizes for the the fields that are scalar int's and bool's and sync.Mutex'es, etc)

and, perhaps it might have enough reflect / metadata stuff in it that it can tell which fields are non-scalar (the strings, arrays, maps) that need a recursive maneuver

steveyen:
Also, that reminds me of a reflect story.  I had this struct for stats -- which had tons of fields of uint64's, which I knew was just going to keep on changing over time (adding new stats over time)...

https://github.com/couchbase/cbgt/blob/master/manager.go#L85

...and I needed a way to copy that struct atomically so that I could get a snapshot of it.

At first I had a big, brittle page of dst.foo1 = src.foo1; dst.foo2 = src.foo2; ... lines of code.

Then, later on, I figured if the stats call ain't happening too often, might as well use reflect, and ended up with a less performant, but much less brittle version...

https://github.com/couchbase/cbgt/blob/master/manager.go#L920

Since the stats poll was only every second or so, I thought that was the better tradeoff at that time.

abhinavdangeti:
Hey @steveyen , @mschoch, So I've reduced the brittleness factor by a bit by adopting steve's suggestion of using reflection during init(). Let me know what you think, I will take a look again to see if I can do better.

steveyen:
hi... your latest PR is indeed less brittle, so it's better.  But, still quite a wide-ranging change.  Hmmm.

One thought is that the reflection based approaches are often a lot more concise in comparison, like a page of code or so (*), that they're still really tempting if it was me.  I'd understand if you'd want to still take the direct, non-reflection approach, though, if it's a real performance issue.  Especially, if there was time, it might be helpful to find out "how slow really is reflection" for real rather than the rules of thumb of common wisdom (...says the person who doesn't have time to find out for himself on this).  Again, if we really want the direct, non-reflection route, don't think of this comment / feedback as a -1 or -2.

(*) - btw, googling pops up some sample reflection code that recursively walks through and copies data structures.  Instead of copying, summing up sizes seems viable too...  https://gist.github.com/hvoecking/10772475#file-translate-go-L191

mschoch:
Measuring it properly will be difficult, as reflection uses some synchronized maps internally (one of the reasons they introduce sync.Map).  So, if you confine yourself to standalone benchmarks, you might be misled that it isn't that expensive.  But with a high volume of queries coming in, computing the cost of each one might also face lock contention in the reflection paths.

steveyen:
> So, if you confine yourself to standalone benchmarks, you might be misled that it isn't that expensive.

good point -- i'd forgotten about the sync stuff in reflect

abhinavdangeti:
@steveyen Thanks for that feedback, and I do prefer your first suggestion of using reflection in init() to determine overhead from all attributes within a structure as opposed to my earlier non-reflection direct approach. However during query time to avoid any performance hit from reflection, we'd still need to fetch the sizes of the content within some of the data structures and this is what the SizeInBytes() apis are for.

mschoch:
So, I've been toying with this other idea, I guess I need to put together a prototype.  The idea is that in addition to an init function computing the size of the static portion of a structure, we use reflection to build a slice, the slice contains functions, which take a single argument which is a reflect.Value (reflection of the actual instance at runtime), and they return the memory used by that dynamic portion.

I went ahead and prototyped it here.  And it does work, though obviously would need to be extended to other types.  And if the value inside a slice or map is a pointer, it should see if it implements Sizer (interface defining Size() method) and if so call it as well.

https://play.golang.org/p/_QaNAL9at6J

But, it ended up uglier than I hoped, and I really don't know that it would be any faster than just using reflect on the whole damn thing.  So, maybe thats where we should start...

abhinavdangeti:
I'm working on a change that will use reflection on runtime, and we would be visiting the reflection idea above at a later time.

> Propose we shorten the method name to just Size() since we know the unit, it's the same everywhere and it will be easier to read. (not that important but starting with something simple)

Done.

> It seems like instead of a global map with entries we could just initialize per-type values (not exported) as it seems we only use them inside that structs own Size() method.

Done.

> It seems like fundamentally what we're saying is, the size of thing has some static portion (its structure) and some dynamic portion (how many elements in slice or key/value pairs in a map or runes in string) We've replaced the static portion with our pre-computed value, but we still have to know which fields have some dynamic portion, which could be a pointer to another item, in which case we have to call Size() again, or some variable sized value like a slice, map or string.

This sounds good, but I see one problem with invoking the Size() of a pointer to another item while estimating the dynamic portion .. reflect insists that the pointer be an "Exported field" which in most cases if not all - it isn't. So there's a panic here:
`panic: reflect: reflect.Value.Call using value obtained using unexported field`

Any thoughts on this?

abhinavdangeti:
@mschoch @steveyen This latest set of changes calculates the static portion of the structures within init() using Reflect. The dynamic part is calculated manually on demand.

size/sizes.go
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package size
+
+const SizeOfBool = 1
```
mschoch: could we initialize all of these using init/reflect?

search/search.go
```diff
@@ -148,3 +215,8 @@ type SearcherOptions struct {
 type SearchContext struct {
 	DocumentMatchPool *DocumentMatchPool
 }
+
+func (sc *SearchContext) SizeOverhead() int {
```
mschoch: is there a reason this is named SizeOverhead and not Size like the rest?

search/pool.go
```diff
@@ -42,11 +53,10 @@ func NewDocumentMatchPool(size, sortsize int) *DocumentMatchPool {
 	startBlock := make([]DocumentMatch, size)
 	startSorts := make([]string, size*sortsize)
 	// make these initial instances available
-	i, j := 0, 0
-	for i < size {
+	j := 0
+	for i := 0; i < size; i++ {
```
mschoch: what is the reason we're changing this in this PR?

search/facets_builder.go
```diff
@@ -213,6 +247,10 @@ type FacetResult struct {
 	DateRanges    DateRangeFacets    `json:"date_ranges,omitempty"`
 }
 
+func (fr *FacetResult) SizeOverhead() int {
```
mschoch: SizeOverhead not Size?

search/pool.go
abhinavdangeti: This shouldn't have been part of this change. Let me undo this.

size/sizes.go
abhinavdangeti: Yes we can, let me update these.

search/search.go
abhinavdangeti: Oh, let me update these, I missed a TODO I had to complete here.

abhinavdangeti:
Hey @mschoch, per our conversation from earlier today, the 2nd commit here uses context to set up a callback function to determine the search's memory estimate, and will return the decision made per the threshold. Let me know what you think, I could look into doing the same for the search result's estimate.

index_impl.go
```diff
@@ -362,15 +366,76 @@ func (i *indexImpl) Search(req *SearchRequest) (sr *SearchResult, err error) {
...
+				return memCbFn(size)
+			}
+		}
+		return nil
+	}
+
 	searchStart := time.Now()
 
 	if !i.open {
+		_ = memCheckCallback(0)
```
mschoch: what is the purpose of these memCheckCallback(0) calls? there are several of them

abhinavdangeti: The client (that sets up the callback function) will be responsible for aggregating the estimates from all the local pindexes before responding with a decision. This is just to make sure that every pindex invokes the callback that one time regardless of an error situation.

mschoch: OK, again I don't think that is quite the right design.  I don't think the estimate of the coordinating query should include the local queries.  Executing each individual query has some requirements, and merging them has requirements, we can (and for simplicity) treat these as independent.  The only relation they have is that estimating the size of the merge requirements is also based on the SearchRequest.  But it is a different estimate, and does not include the child queries memory requirements.

In order to get this merged, let's just simplify further.  SearcInContext invokes the callback it it's present in the context.  And we don't call it otherwise.  Callers have no guarantee that the callback will ever be invoked.

steveyen:
hi -- scrolled through this and nothing stood out for me here as overly worrisome

index_impl.go
abhinavdangeti: Ah understood, thanks for that. I'll have patch up in a sec

abhinavdangeti:
@mschoch The latest set of changes incorporate your feedback, thanks!

PR #777, opened 2018-02-14
pausing persister until merging catches up

sreekanth-cb:
adding a pause to persister until the merger catches up.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -164,16 +164,12 @@ func notifyMergeWatchers(lastPersistedEpoch uint64,
 func (s *Scorch) pausePersisterForMergerCatchUp(lastPersistedEpoch uint64, lastMergedEpoch *uint64,
 	persistWatchers []*epochWatcher) []*epochWatcher {
 
+	// first, let the watchers proceed if they lag behind
+	persistWatchers = notifyMergeWatchers(lastPersistedEpoch, persistWatchers)
+
 OUTER:
-	for {
-
-		// first, let the watchers proceed if they lag behind
-		persistWatchers = notifyMergeWatchers(lastPersistedEpoch, persistWatchers)
-
-		// check for slow merger and pause persister until merger catch up
-		if lastPersistedEpoch <= *lastMergedEpoch+epochDistance {
-			break OUTER
-		}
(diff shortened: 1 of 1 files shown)


index/scorch/persister.go
```diff
@@ -136,15 +140,57 @@ OUTER:
...
+
+OUTER:
+	for {
+
+		// first, let the watchers proceed if they lag behind
+		persistWatchers = notifyMergeWatchers(lastPersistedEpoch, persistWatchers)
+
+		// check for slow merger and pause persister until merger catch up
+		if lastPersistedEpoch > *lastMergedEpoch &&
+			lastPersistedEpoch-*lastMergedEpoch > epochDistance {
```
steveyen: since these are uint's, i become a wary about subtraction.  I see you've guarded the overflow/underflow case on line 174, but maybe it can be collapse into a single line, like...

    if lastPersistedEpoch > *lastMergedEpoch + epochDistance {

index/scorch/persister.go
```diff
@@ -136,15 +141,55 @@ OUTER:
...
+
+			select {
+			case <-s.closeCh:
+				break OUTER
+			case ew := <-s.persisterNotifier:
+				persistWatchers = append(persistWatchers, ew)
+				*lastMergedEpoch = ew.epoch
+				continue OUTER
+			}
+		} else {
```
mschoch: one thought here is that we have an unconditional for loop, but this else block is a case you check every loop iteration to bail on the loop.  so, wouldn't it be more clear to invert that (since it's an else case) and make that the condition of the for loop?

the only downside i see is that right now you always call notifyMergeWatchers regardless of this condition, you'd need it before the for loop, and again inside the loop (presumably now at the end of the loop).  but, even though that makes it feel longer, you can drop the else case, and the "continue OUTER" becomes unnecessary too as you'd want to proceed to notifyMergeWatchers again and let the loop naturally continue without need a continue

sreekanth-cb:
@mschoch , attempted the loop clean up!

sreekanth-cb:
I see what you are saying now, Earlier all my thought was behind/focused on making the notifyMergeWatchers call only once. But this now looks more readable.

PR #780, opened 2018-02-23
configurable mergePlanner options

sreekanth-cb:
mergePlanner options are parsed from the
scorch configs parameters

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -28,6 +28,7 @@ import (
 
 func (s *Scorch) mergerLoop() {
 	var lastEpochMergePlanned uint64
+	mergePlannerOptions := s.parseMergePlannerOptions()
 OUTER:
 	for {
 		select {
@@ -45,7 +46,7 @@ OUTER:
 				startTime := time.Now()
 
 				// lets get started
-				err := s.planMergeAtSnapshot(ourSnapshot)
+				err := s.planMergeAtSnapshot(ourSnapshot, mergePlannerOptions)
 				if err != nil {
 					s.fireAsyncError(fmt.Errorf("merging err: %v", err))
 					_ = ourSnapshot.DecRef()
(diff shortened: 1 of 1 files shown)


index/scorch/merge.go
```diff
@@ -82,7 +83,36 @@ OUTER:
 	s.asyncTasks.Done()
 }
 
-func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot) error {
+func (s *Scorch) parseMergePlannerOptions() *mergeplan.MergePlanOptions {
+	mergePlannerOptions := &mergeplan.DefaultMergePlanOptions
```
steveyen: oh, this seems it'll have the effect of inadvertently updating the default merge plan options, when instead it should be parsing into its own copy?

index/scorch/merge.go
```diff
@@ -82,7 +83,36 @@ OUTER:
...
+	scorchOptions := map[string]interface{}{}
+	if v, ok := s.config["scorchOptions"]; ok {
+		if scorchOptions, ok = v.(map[string]interface{}); ok {
+			if v, ok := scorchOptions["maxSegmentsPerTier"].(float64); ok {
```
steveyen: one thought is this approach as a "brittle" feel.  not that merge plan options are changing that often, but if we do add more options, we'll need to remember to update this code here.

one less brittle approach, but perhaps "uglier" is to turn the scorchOptions back into a string, and then just use the standard json parser.  Since this code hardly ever runs (0.0000001% of the time), I can convince myself the laziness tradeoff is worth it

index/scorch/merge.go
```diff
@@ -82,7 +83,36 @@ OUTER:
 	s.asyncTasks.Done()
 }
 
-func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot) error {
+func (s *Scorch) parseMergePlannerOptions() *mergeplan.MergePlanOptions {
+	mergePlannerOptions := &mergeplan.DefaultMergePlanOptions
+	scorchOptions := map[string]interface{}{}
+	if v, ok := s.config["scorchOptions"]; ok {
```
steveyen: ah, a naming thing!  Instead of "scorchOptions", perhaps this is just "scorchMergePlanOptions", so it's really specific.  If one future day we add even more "scorchOptions" that aren't related to merge planner options, they won't need to be mixed up together.

index/scorch/merge.go
steveyen: (aside: I vaguely recall another near-embarasssing trick (hack), where I needed a "deep copy" of some data structures somewhere, also in a codepath that "hardly ever runs".  Not sure if it was 3 AM or not, but the solution I think I ended up with was to marshal to json and parse it back.  :-) )

sreekanth-cb: looks like have to adopt the same approach, just that I do it @3 PM :)

index/scorch/merge.go
sreekanth-cb: Whoops..fixed!

PR #781, opened 2018-02-23
Handle case where store snapshot isn't closed in upsidedown's Batch() API

abhinavdangeti:


abhinavdangeti:
I don't think this is a suspect, sundar and I were browsing for potential problems, and this came up. There's some more changes in moss that sundar will be pushing up

PR #787, opened 2018-02-28
adding stats for scorch

sreekanth-cb:


sreekanth-cb:
need to discuss a bit on how to track snapshot begin and end over stats.

index/scorch/merge.go
```diff
@@ -129,6 +129,9 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 		return nil
 	}
 
+	mip := uint64(len(resultMergePlan.Tasks))
+	atomic.AddUint64(&s.stats.CurInProgressFileMerges, mip)
```
steveyen: One thing is I tend to prefer is upwards-only counters (that only grow, like TotFooBar) instead of gauges (numbers that go up & down like CurFooBar).  I figured a user of the stats can calculate their own "gauge" if they're given two counters  --one for function entry, one for function exit -- plus now they know more information with two counters.

A related issue is to remember to decrement the gauge (or increment an exit counter) when there's an early function return.

Related, the # of function enters ought to always be >= the # of function exits.  And, the # of function exits might be == # of good exits + # of error exits.  In moss this often has a naming pattern of three related counters like TotFooBarBeg >= TotFooBarDone + TotFooBarErr.

steveyen:
heads up - been working on this, so will have a change that'll come right after and based on sreekanth's changes with naming fixes, etc

steveyen:
my changes stacked up on top of this one are here at P$ #789 

if #789 goes in, then this one #787 is obsoleted

PR #790, opened 2018-03-01
update to use new vellum Reset API

mschoch:


index/scorch/segment/zap/build.go
```diff
@@ -471,15 +471,17 @@ func persistDictionary(memSegment *mem.Segment, w *CountHashWriter, postingsLocs
 	varintBuf := make([]byte, binary.MaxVarintLen64)
 
 	var buffer bytes.Buffer
+	builder, err := vellum.New(&buffer, nil)
+	if err != nil {
+		return nil, err
+	}
 	for fieldID, fieldTerms := range memSegment.DictKeys {
 		if fieldID != 0 {
```
sreekanth-cb: one doubt here, is there a reason why we explicitly make this check if fieldID != 0 {} and reset this than doing at the end of the for-loop.

sreekanth-cb: one doubt here, is there a reason why we explicitly make this check if fieldID != 0 {} and reset this than doing at the end of the for-loop.

mschoch:
addressed @sreekanth-cb suggestion

PR #796, opened 2018-03-05
adding Merge method to Batch

sreekanth-cb:
exposing this new method in case of applications like cbft may choose to merge smaller batches especially for scorch index formats.

### DIFF
--- a/index.go
+++ b/index.go
@@ -102,6 +102,18 @@ func (b *Batch) Reset() {
 	b.internal.Reset()
 }
 
+// Merge performs the merge of the internal Index.Batch
+// of the given batch to the self.
+func (b *Batch) Merge(other *Batch) {
+	for k, v := range other.internal.IndexOps {
+		b.internal.IndexOps[k] = v
+	}
+
+	for k, v := range other.internal.InternalOps {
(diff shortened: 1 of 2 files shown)


index.go
```diff
@@ -102,6 +102,18 @@ func (b *Batch) Reset() {
 	b.internal.Reset()
 }
 
+// Merge performs the merge of the internal Index.Batch
+// of the given batch to the self.
+func (b *Batch) Merge(other *Batch) {
```
mschoch: This method is unexpected.  Unfortunately, documents go through the mapping process when they are added to the Batch (initially).  With this method, I could create batches for 2 different indexes (which have difference mappings) then merge them, and add that batch to an index.  It's not wrong exactly, as you can do this directly with IndexAdvanced anyway, but it opens the door for more problems.

sreekanth-cb: I see, Does it make sense to take care of that conditional check inside this merge method?
like checking the index Mappings of the Batches for equality and merge only upon batches are of the same index Mapping()?

mschoch:
I'm thinking we close this for now, as @steveyen has persuaded me that we should not make higher-level changes that also affect moss/upside down in this phase.

PR #798, opened 2018-03-05
adding compaction_written_bytes/sec stats to scorch

sreekanth-cb:
I wasn't sure whether we need the written bytes to be tracked at the CountHashWriter write method level or not.

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -180,8 +180,9 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 			s.markIneligibleForRemoval(filename)
 			path := s.path + string(os.PathSeparator) + filename
 			atomic.AddUint64(&s.stats.TotFileMergeZapBeg, 1)
-			newDocNums, err := zap.Merge(segmentsToMerge, docsToDrop, path, 1024, &s.stats)
+			newDocNums, nBytes, err := zap.Merge(segmentsToMerge, docsToDrop, path, 1024)
 			atomic.AddUint64(&s.stats.TotFileMergeZapEnd, 1)
+			atomic.AddUint64(&s.stats.TotFileMergeWrittenBytes, nBytes)
 			if err != nil {
 				s.unmarkIneligibleForRemoval(filename)
 				atomic.AddUint64(&s.stats.TotFileMergePlanTasksErr, 1)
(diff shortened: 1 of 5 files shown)


index/scorch/stats.go
```diff
@@ -100,6 +100,8 @@ type Stats struct {
 	TotMemMergeZapBeg   uint64
 	TotMemMergeZapEnd   uint64
 	TotMemMergeSegments uint64
+
+	TotCompactionWrittenBytes uint64
```
steveyen: naming thin -- i think in scorch we don't usually use the word "compaction", even though that's a legit way to think of what the file merger is doing.

How about "TotFileMergeWrittenBytes", and move this stat up a few lines into the file merge related clumping of stats?  (but keep that backwards compat naming alias elsewhere that's "total_compaction_written_bytes"?

steveyen:
Hi @sreekanth-cb -- another thing (capturing something @mschoch mentioned)... not sure we need an additional StatsReporter interface for this?  That is, an interface or level-of-indirection might be warranted if there's different kinds of stats receivers, but in this case there's just that one stats struct?

index/scorch/merge.go
```diff
@@ -183,16 +183,17 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 			fileMergeZapStartTime := time.Now()
 
 			atomic.AddUint64(&s.stats.TotFileMergeZapBeg, 1)
-			newDocNums, err := zap.Merge(segmentsToMerge, docsToDrop, path, 1024)
+			newDocNums, nBytes, err := zap.Merge(segmentsToMerge, docsToDrop, path, 1024)
 			atomic.AddUint64(&s.stats.TotFileMergeZapEnd, 1)
-
+      atomic.AddUint64(&s.stats.TotFileMergeWrittenBytes, nBytes)
```
mschoch: wondering why the indentation doesn't match here, do we need to run go fmt?

index/scorch/merge.go
```diff
@@ -183,16 +183,17 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
...
+      
 			fileMergeZapTime := uint64(time.Since(fileMergeZapStartTime))
 			atomic.AddUint64(&s.stats.TotFileMergeZapTime, fileMergeZapTime)
 			if atomic.LoadUint64(&s.stats.MaxFileMergeZapTime) < fileMergeZapTime {
 				atomic.StoreUint64(&s.stats.MaxFileMergeZapTime, fileMergeZapTime)
 			}
-
-			if err != nil {
+
```
mschoch: seems like indentation changed here as well

PR #800, opened 2018-03-06
making NumSnapshotsToKeep configurable

sreekanth-cb:
NumSnapshotsToKeep read from index configs

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -633,14 +633,19 @@ func (s *Scorch) removeOldBoltSnapshots() (numRemoved int, err error) {
 		return 0, err
 	}
 
-	if len(persistedEpochs) <= NumSnapshotsToKeep {
+	numSnapshotsToKeep := NumSnapshotsToKeep
+	if val, ok := s.config["numSnapshotsToKeep"].(float64); ok && val > 0 {
+		numSnapshotsToKeep = int(val)
+	}
+
+	if len(persistedEpochs) <= numSnapshotsToKeep {
 		// we need to keep everything
 		return 0, nil
 	}
 
 	// make a map of epochs to protect from deletion
-	protectedEpochs := make(map[uint64]struct{}, NumSnapshotsToKeep)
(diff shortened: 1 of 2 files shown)


index/scorch/persister.go
```diff
@@ -633,14 +633,19 @@ func (s *Scorch) removeOldBoltSnapshots() (numRemoved int, err error) {
 		return 0, err
 	}
 
-	if len(persistedEpochs) <= NumSnapshotsToKeep {
+	numSnapshotsToKeep := NumSnapshotsToKeep
+	if val, ok := s.config["numSnapshotsToKeep"].(float64); ok && val > 0 {
```
mschoch: i think this keeps coming up from time to time.  from the Go library perspective this is kinda ugly as users who pass an integer (what you would expect) find that the feature doesn't work.  i think if we add these kind of options, they should make sense from the Go library perspective first, even if that means more work elsewhere to parse JSON differently...

index/scorch/scorch.go
```diff
@@ -191,6 +192,23 @@ func (s *Scorch) openBolt() error {
 		}
 	}
 
+	s.numSnapshotsToKeep = NumSnapshotsToKeep
```
mschoch: hmmn, i don't think round tripping the value through json is the right approach either

index/scorch/scorch.go
```diff
@@ -503,3 +515,15 @@ func (s *Scorch) unmarkIneligibleForRemoval(filename string) {
 func init() {
 	registry.RegisterIndexType(Name, NewScorch)
 }
+
+func parseToInteger(v interface{}) (int, error) {
+	switch v.(type) {
+	case float64:
+		return int(v.(float64)), nil
```
mschoch: Go has syntax to allow remembering the newly typed value in a type assertion, so you don't have to do the assertion a second time when returning the value here.  See this example: https://tour.golang.org/methods/16

sreekanth-cb: done!

PR #812, opened 2018-03-08
MB-28562: Support search query callbacks before and after execution

abhinavdangeti:
+ SearchQueryStartCallback
+ SearchQueryEndCallback

### DIFF
--- a/index_impl.go
+++ b/index_impl.go
@@ -50,9 +50,11 @@ const storePath = "store"
 
 var mappingInternalKey = []byte("_mapping")
 
-const SearchMemCheckCallbackKey = "_search_mem_callback_key"
+const SearchQueryStartCallbackKey = "_search_query_start_callback_key"
+const SearchQueryEndCallbackKey = "_search_query_end_callback_key"
 
-type SearchMemCheckCallbackFn func(size uint64) error
+type SearchQueryStartCallbackFn func(size uint64) error
+type SearchQueryEndCallbackFn func(size uint64) error
 
 func indexStorePath(path string) string {
 	return path + string(os.PathSeparator) + storePath
@@ -483,13 +485,22 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 		collector.SetFacetsBuilder(facetsBuilder)
 	}
 
-	if memCb := ctx.Value(SearchMemCheckCallbackKey); memCb != nil {
-		if memCbFn, ok := memCb.(SearchMemCheckCallbackFn); ok {
-			err = memCbFn(memNeededForSearch(req, searcher, collector))
+	memNeeded := memNeededForSearch(req, searcher, collector)
+	if cb := ctx.Value(SearchQueryStartCallbackKey); cb != nil {
+		if cbF, ok := cb.(SearchQueryStartCallbackFn); ok {
+			err = cbF(memNeeded)
 		}
 	}
 	if err != nil {
 		return nil, err
+	}
+
+	if cb := ctx.Value(SearchQueryEndCallbackKey); cb != nil {
+		if cbF, ok := cb.(SearchQueryEndCallbackFn); ok {
+			defer func() {
+				_ = cbF(memNeeded)
+			}()
+		}
 	}
 
 	err = collector.Collect(ctx, searcher, indexReader)
--- a/index_test.go
+++ b/index_test.go
@@ -1871,7 +1871,7 @@ func BenchmarkScorchSearchOverhead(b *testing.B) {
 	benchmarkSearchOverhead(scorch.Name, b)
 }
 
-func TestSearchMemCheckCallback(t *testing.T) {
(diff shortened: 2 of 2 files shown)


index_impl.go
```diff
@@ -483,15 +485,24 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
 		}
 	}
 	if err != nil {
 		return nil, err
 	}
 
+	defer func() {
+		if cb := ctx.Value(SearchQueryEndCallbackKey); cb != nil {
+			if cbF, ok := cb.(SearchQueryEndCallbackFn); ok {
+				_ = cbF(memNeeded)
```
mschoch: maybe we could reverse these.  meaning, only if the 2 conditions are satisfied do we bother defering anything at all.  then we can just defer cbF and not an anonymous function.

mschoch:
also ping @steveyen these test failures seem to be coming from roaring, possibly related to some of your changes?

index_impl.go
abhinavdangeti: Hmm - ya I could do that, but with defer cbF(memNeeded), errCheck complains that we're not handling the response of the function.

Now a return response is not needed for the QueryEndCallback, but if I were to change the definition of the SearchQueryEndCallbackFn, I cannot use this switch-case in cbft:

        func (a *appHerder) onQueryEvent(event cbft.QueryEvent, size uint64) error {
	        switch event.Kind {
	        case cbft.EventQueryStart:
		        return a.onQueryStart(size)

	        case cbft.EventQueryEnd:
		        return a.onQueryEnd(size)

	        default:
		        return nil
	        }
        }

Is there a reason why you'd rather I not defer an anonymous function, or perhaps is this something that's ok too:

        if cb := ctx.Value(SearchQueryEndCallbackKey); cb != nil {
                if cbF, ok := ok.(SearchQueryEndCallbackFn); ok {
                        defer func() {
                                _ = cbF(memNeeded)
                        }
                }
        }

mschoch: Ya this is fine, there is no problem with the anonymous func, just wanted to avoid it if possible.

mschoch: My first instinct is that the query event stuff is over-engineered.  But, I guess I can kinda see it being used for query stats in the future, so let's go with it.

abhinavdangeti: Yep, my thoughts too.

abhinavdangeti:
@mschoch the latest patch-set incorporates your feedback, thanks!

abhinavdangeti:
Sure, here's the link to the previously seen (sporadic) test failures: https://travis-ci.org/blevesearch/bleve/builds/350979910

steveyen:
@mschoch @abhinavdangeti -- assuming that travis-CI is pulling the latest roaring (it's seemingly not vendored by bleve) looks like the latest roaring up on github has the right line number-to-code alignment...

A cut-&-paste of the top of the panic stack dump...
`unexpected fault address 0x7f001c9df071 fatal error: fault [signal SIGSEGV: segmentation violation code=0x2 addr=0x7f001c9df071 pc=0x45ba12] goroutine 62 [running]: runtime.throw(0x66c7fb, 0x5) /home/travis/.gimme/versions/go1.9.4.linux.amd64/src/runtime/panic.go:605 +0x95 fp=0xc42004fbc8 sp=0xc42004fba8 pc=0x42b2a5 runtime.sigpanic() /home/travis/.gimme/versions/go1.9.4.linux.amd64/src/runtime/signal_unix.go:374 +0x227 fp=0xc42004fc18 sp=0xc42004fbc8 pc=0x4415c7 runtime.memmove(0x7f001c9df071, 0x7f001c9df06f, 0x2) /home/travis/.gimme/versions/go1.9.4.linux.amd64/src/runtime/memmove_amd64.s:143 +0x102 fp=0xc42004fc20 sp=0xc42004fc18 pc=0x45ba12 github.com/RoaringBitmap/roaring.(*arrayContainer).iorArray(0xc420332660, 0xc420332680, 0x7f001c9df0b4, 0x12) /home/travis/gopath/src/github.com/RoaringBitmap/roaring/arraycontainer.go:377 +0x285 fp=0xc42004fcb8 `

Looks like segfault is amongst some new, day old code in roaring arraycontainer.go / iorArray()...
https://github.com/RoaringBitmap/roaring/commit/d6540aab65a17321470b1661bfc52da1823871e9
...related to a fix yesterday for roaring issue https://github.com/RoaringBitmap/roaring/issues/152 ...

abhinavdangeti:
Ah, thanks @steveyen.

mschoch:
suggest we add a working SHA to the manifest?

abhinavdangeti:
the SHA in the manifest(s) is pretty old, I could bump it up to the latest?

mschoch:
@steveyen implied that roaring is not in the *bleve* manifest

abhinavdangeti:
ah yes, will do.

PR #815, opened 2018-03-09
minor optimisation to loadChunk method

sreekanth-cb:
moving to a starting offset based direct chunk loading


index/scorch/segment/zap/intcoder.go
```diff
@@ -134,3 +139,42 @@ func (c *chunkedIntCoder) Write(w io.Writer) (int, error) {
...
 	return len(c.final)
 }
+
+// chunkLengthsToOffsets converts the chunk length array
+// to a chunk starting offset array. The readChunkBoundary
+// will figure out the start and end of every chunk from
+// these offsets. The starting offset of the first/single
+// array element will always be zero and this position is
+// used for storing the size of the current last item in
+// the array at any given point.
```
steveyen: i'm wondering instead of using the 0'th position to store the last item's info, how about logically shifting everything leftwards by one position?

That is, if you have a bunch of offsets for N item items, then actually only hold onto offsets[1:] of N-1 items?

mschoch: great point @steveyen, that is i believe the same situation and solution i had used for the "running" field in the index snapshot, which similarly remembers where certain global doc num ranges start.  there was no need to "remember" that it starts with 0, but there is a need to know how many are in the last segment, so it seems like the same case.

sreekanth-cb: thanks @steveyen , done!

mschoch:
One higher level question I have.  We have intcoder and contentcoder.  I remember I refactored one of them out.  And then (possibly concurrently with that) @sreekanth-cb introduced the other for the doc value code he added.  At that time I thought they looked similar but different enough I didn't give it much thought.

So... fast-forward today, and I noticed that we now have a common method used by both of these coders, which converts length slices to offset slices.  I guess my question is are these two both really needed?  Is there another round of refactoring on the horizon?

sreekanth-cb:
hi @mschoch , the common function still operates on different types as of now. At the moment, I don't have any refactoring idea in mind. But eventually we may generalise either one of it and make the other one fit on top of it.  Let me extent the TODO list.

index/scorch/segment/zap/intcoder.go
```diff
@@ -134,3 +139,35 @@ func (c *chunkedIntCoder) Write(w io.Writer) (int, error) {
...
+// Lens ->  0 5 0 5 => 0 5 5 10
+// Lens ->  0 0 0 5 => 0 0 0 5
+// Lens ->  5 0 0 0 => 5 5 5 5
+// Lens ->  0 5 0 0 => 0 5 5 5
+// Lens ->  0 0 5 0 => 0 0 5 5
+func chunkLengthsToOffsets(lengths []uint64) {
+	var runningOffset uint64
+	var index, i int
+	for i = 1; i <= len(lengths); i++ {
+		runningOffset += lengths[i-1]
```
steveyen: wondering if the same approach from PR https://github.com/blevesearch/bleve/pull/821 should be applied here -- instead of needing this additional fixup step?

PR #821, opened 2018-03-12
docValue space savings

sreekanth-cb:
merging the meta doc value length and loc
slices into a single offset slice  as that
is enough to compute the starting offset and
length of the the doc values data for a given
document inside a docValue chunk.


cmd/bleve/cmd/zap/docvalue.go
```diff
@@ -255,7 +253,19 @@ func getDocValueLocs(docNum uint64, metaHeader []zap.MetaData) (uint64, uint64)
 		return metaHeader[i].DocNum >= docNum
 	})
 	if i < len(metaHeader) && metaHeader[i].DocNum == docNum {
-		return metaHeader[i].DocDvLoc, metaHeader[i].DocDvLen
+		var start, end uint64
+		if i > 0 {
+			start = metaHeader[i].DocDvOffset
+		}
+		// single element case
```
steveyen: not that this is incorrect, but it feels like this should be doable or simplify'able into two cases instead of three?  that is, we're at the end, or we're not at the end?

cmd/bleve/cmd/zap/docvalue.go
```diff
@@ -255,7 +253,19 @@ func getDocValueLocs(docNum uint64, metaHeader []zap.MetaData) (uint64, uint64)
...
+		if i == 0 && len(metaHeader) == 1 {
+			end = metaHeader[i].DocDvOffset
+		} else if i < len(metaHeader)-1 {
+			end = metaHeader[i+1].DocDvOffset
+		} else { // for last element
+			end = start + metaHeader[0].DocDvOffset
```
steveyen: the last element case of using metaHeader[0] feels wrong?

index/scorch/segment/zap/docvalues.go
```diff
@@ -163,7 +161,19 @@ func (di *docValueIterator) getDocValueLocs(docNum uint64) (uint64, uint64) {
...
+		}
+		// single element case
+		if i == 0 && len(di.curChunkHeader) == 1 {
+			end = di.curChunkHeader[i].DocDvOffset
+		} else if i < len(di.curChunkHeader)-1 {
+			end = di.curChunkHeader[i+1].DocDvOffset
+		} else { // for last element
+			end = start + di.curChunkHeader[0].DocDvOffset
+		}
+		return start, end
```
steveyen: this code roughly looks like repeated, duplicated code?  maybe pull it out into a helper func?

cmd/bleve/cmd/zap/docvalue.go
mschoch: if i recall from the meeting last week, this was the solution sreekanth described of having the last offset stored in the first position (which isn't needed because it's implicitly zero.  you had advised against this because of the complexity

cmd/bleve/cmd/zap/docvalue.go
sreekanth-cb: not sure @steveyen , can you please check the loadChunk PR: #815 . It tries to explain the single offset slice logic.

steveyen: i see - thx... will leave commentary on the other PR #815

cmd/bleve/cmd/zap/docvalue.go
steveyen: oh, right -- will leave comment on the other PR #815

index/scorch/segment/zap/contentcoder.go
```diff
@@ -94,9 +93,21 @@ func (c *chunkedContentCoder) flushContents() error {
...
+	// convert the document data lens to data offsets
+	if len(c.chunkMeta) > 1 {
+		var runningOffset uint64
+		var index, i int
+		for i = 1; i < len(c.chunkMeta); i++ {
+			runningOffset += c.chunkMeta[i-1].DocDvOffset
+			c.chunkMeta[index].DocDvOffset = runningOffset
+			index++
+		}
+		c.chunkMeta[index].DocDvOffset = c.chunkMeta[i-1].DocDvOffset
```
steveyen: I'm a little confused about the serialized encoding -- perhaps some doc commentary would help?

I suspect the chunkMeta.DocDvOffset field is "switching gears" or switching meanings somewhere along the way?  I suspect it starts off as a length but is getting transformed?

sreekanth-cb: we are following the same offset slice pattern as that used with loadChunk here, Let me add some comments there.. 
Yes, you are right, the length slice is getting transformed to a starting offset slice with the length of the last item getting stored in last position.

index/scorch/segment/zap/contentcoder.go
```diff
@@ -94,9 +93,27 @@ func (c *chunkedContentCoder) flushContents() error {
 		return err
 	}
 
+	// convert the document data lens to data offsets.
+	// This enables the direct look up for the start,end
+	// offsets for a given document entry inside a chunk.
+	// Starting offset of i'th index is stored in i-1'th
+	// position except for 0'th index and ending offset is
+	// stored at i'th index position.
```
mschoch: I find this sentence confusing.  It seems to say that the offset of the 1st index is in position 0 (i-1) but that the offset of the 0th index is also in position 0 (i).

sreekanth-cb: done!

index/scorch/segment/zap/contentcoder.go
```diff
@@ -130,17 +147,15 @@ func (c *chunkedContentCoder) Add(docNum uint64, vals []byte) error {
...
 	if err != nil {
 		return err
 	}
 
 	c.chunkMeta = append(c.chunkMeta, MetaData{
-		DocNum:   docNum,
-		DocDvLoc: uint64(dvOffset),
-		DocDvLen: uint64(dvSize),
+		DocNum:      docNum,
+		DocDvOffset: uint64(dvSize),
```
steveyen: Still find it a little disconcerting that a size is being stored in an offset.

One of the things I look for is to see if the "units" are the same.  Loc & Offset are synonyms for the same kind of units.  Len & Size are synonyms.  Kilograms & Weight are the same kind of thing.  At least in my head.

So, this has this weird feel to me of temporarily storing a kilogram-weight number into a telephone field.

Since the dvOffset is seemingly conveniently available (old/pink line number 134), wondering if storing that would be clearer, rather than remembering the length in the offset field?

index/scorch/segment/zap/contentcoder.go
```diff
@@ -173,3 +188,18 @@ func (c *chunkedContentCoder) Write(w io.Writer) (int, error) {
 	}
 	return tw, nil
 }
+
+// ReadDocValueBoundary elicits the start, end offsets from a
+// starting offset based metaData header slice
+func ReadDocValueBoundary(chunk int, metaHeaders []MetaData) (uint64, uint64) {
+	var start uint64
+	if chunk > len(metaHeaders) {
```
steveyen: wondering if you need this case?  The callers seem to always check that chunk is < len(metaHeaders)?

sreekanth-cb: since it is an exported method, I felt we need to check this..?

index/scorch/segment/zap/contentcoder.go
sreekanth-cb: dvOffset+dvSize [loc type and size type]  are breaking the similar stuff analogy you mentioned here know? :) Something in that line could be possible. But I am not sure whether its entirely free from the flaw mentioned here?

steveyen: > dvOffset+dvSize [loc type and size type] are breaking the similar stuff analogy you mentioned here know? :) Something in that line could be possible. But I am not sure whether its entirely free from the flaw mentioned here?

oh, to my mind, an offset (or location) + a size (or a delta) leads to yet another offset.  Starting to think my mind is strange!  :-)

sreekanth-cb: Renaming "DocDvOffset" to "DocDvRunLen" might fix this , I think?

PR #823, opened 2018-03-13
adding maxsegment size limit checks

sreekanth-cb:


steveyen:
> looks good

oh, but now I belatedly see your "do not merge" label.

PR #825, opened 2018-03-13
MB-27385: De-duplicate the list of requested fields

abhinavdangeti:
De-duplicate the list of fields provided by the client as part
of the search request, so as to not inadvertantly load the same
stored field more than once.

### DIFF
--- a/index_impl.go
+++ b/index_impl.go
@@ -534,7 +534,8 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 			doc, err := indexReader.Document(hit.ID)
 			if err == nil && doc != nil {
 				if len(req.Fields) > 0 {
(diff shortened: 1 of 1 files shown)


index_impl.go
```diff
@@ -830,3 +831,16 @@ func (f *indexImplFieldDict) Close() error {
 	}
 	return f.indexReader.Close()
 }
+
+// helper function to remove duplicate entries from slice of strings
+func deDuplicate(fields []string) []string {
+	entries := make(map[string]bool)
```
mschoch: can use empty struct as value here instead of bool

abhinavdangeti: will do.

abhinavdangeti:
@mschoch the latest patch incorporates your feedback.

PR #830, opened 2018-03-15
memoize the size of an entire index snapshot

mschoch:
by memoizing the size of index snapshots and their
constituent parts, we significantly reduce the amount
of time that the lock is held in the app_herder, when
calculating the total memory used


mschoch:
I kind of dislike this PR, partly because I've been looking at it so long.

It does seem to do what it says on the box, the app_herder lock no longer shows up in the synchronization blocking profile.

However, we still take > 60s to process the updates on my laptop (by about 15s).  Fortunately the changes to testrunner mean the test passes now.  However, it still means we're significantly slower in this regard than upsidedown/moss.

Finally, we now see to WAY overshoot the quota, like I told it to use an indexing quota of 0.2 of the total quota (512MB), and I observe it going all the way up to 800MB.  This is new to me, and I have no idea why.

index/scorch/snapshot_segment.go
```diff
@@ -128,15 +128,26 @@ func (s *SegmentSnapshot) Fields() []string {
...
 	readyCh chan struct{}     // closed when the cachedFieldDocs.docs is ready to be used.
 	err     error             // Non-nil if there was an error when preparing this cachedFieldDocs.
 	docs    map[uint64][]byte // Keyed by localDocNum, value is a list of terms delimited by 0xFF.
+	size    uint64
 }
 
 func (cfd *cachedFieldDocs) prepareFields(field string, ss *SegmentSnapshot) {
 	defer close(cfd.readyCh)
 
+	cfd.size += 8 /* size of uint64, sizeInBytes field */
```
abhinavdangeti: You could use size.SizeOfUint64

index/scorch/snapshot_segment.go
```diff
@@ -152,12 +163,14 @@ func (cfd *cachedFieldDocs) prepareFields(field string, ss *SegmentSnapshot) {
 			return
 		}
 
+		cfd.size += 8 /* size of uint64, map key */
```
abhinavdangeti: ditto

index/scorch/snapshot_segment.go
mschoch: done

index/scorch/snapshot_segment.go
mschoch: done

PR #838, opened 2018-03-19
vellum adoption for regex and fuzzy queries

sreekanth-cb:
draft changes for any discussions.
Surprisingly, for the fuzzy queries vellum based implementation takes more time.
Need to analyse that..

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -96,6 +96,13 @@ type IndexReader interface {
 	Close() error
 }
 
+// IndexReaderAdv is an optional interface for advanced users
+// Hope to have a better name here...
+type IndexReaderAdv interface {
+	FieldDictRegex(field string, regex []byte) (FieldDict, error)
+	FieldDictFuzzy(field string, term []byte, fuzziness int) (FieldDict, error)
+}
+
 // FieldTerms contains the terms used by a document, keyed by field
 type FieldTerms map[string][]string
 
--- a/index/index.go
+++ b/index/index.go
@@ -96,10 +96,11 @@ type IndexReader interface {
 	Close() error
 }
 
-// IndexReaderAdv is an optional interface for advanced users
-// Hope to have a better name here...
-type IndexReaderAdv interface {
-	FieldDictRegex(field string, regex []byte) (FieldDict, error)
+type IndexReaderRegexp interface {
+	FieldDictRegexp(field string, regex []byte) (FieldDict, error)
+}
+
+type IndexReaderFuzzy interface {
 	FieldDictFuzzy(field string, term []byte, fuzziness int) (FieldDict, error)
 }
 
--- a/index/scorch/segment/mem/dict.go
+++ b/index/scorch/segment/mem/dict.go
@@ -15,7 +15,9 @@
 package mem
 
 import (
+	"math"
 	"reflect"
+	"regexp"
 	"sort"
 	"strings"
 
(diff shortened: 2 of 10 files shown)


search/searcher/search_fuzzy.go
```diff
@@ -31,9 +34,10 @@ func NewFuzzySearcher(indexReader index.IndexReader, term string,
 			break
 		}
 	}
-
+	t := time.Now()
 	candidateTerms, err := findFuzzyCandidateTerms(indexReader, term, fuzziness,
```
mschoch: i think this is why fuzzy is slower, you're finding the candidate terms the old way even when you later use the push-down capability

index/index.go
```diff
@@ -96,6 +96,13 @@ type IndexReader interface {
 	Close() error
 }
 
+// IndexReaderAdv is an optional interface for advanced users
+// Hope to have a better name here...
+type IndexReaderAdv interface {
```
mschoch: agree, i'll try to propose some names, and might recommend we use 2 separate interfaces

index/scorch/segment/mem/dict.go
```diff
@@ -98,6 +98,30 @@ func (d *Dictionary) RangeIterator(start, end string) segment.DictionaryIterator
 	}
 }
 
+// RegexIterator returns an iterator which only visits terms matching
+// the given regex expression.
+// TODO complete the implementation
```
mschoch: what is missing from the "complete implementation"?

index/scorch/segment/mem/dict.go
```diff
@@ -98,6 +98,30 @@ func (d *Dictionary) RangeIterator(start, end string) segment.DictionaryIterator
...
+	return &DictionaryIterator{
+		d:      d,
+		offset: offset,
+		prefix: regex,
+	}
+}
+
+// FuzzyIterator returns an iterator which only visits terms matching
+// the given edit distance.
+// TODO complete the implementation
```
mschoch: what is missing from the "complete implementation"?

search/searcher/search_fuzzy.go
mschoch: Ya as @sreekanth-cb pointed out, disregard this comment.  I misread the code due to the way github folded it in the diff.

mschoch:
What dataset is used when testing the relative perf?  I expect there may be some floor where building the automaton is slower than brute forcing all the terms.

mschoch:
Hey @sreekanth-cb I'm going to pull this into my queue an look at it closer tomorrow.  Can you provide some quick answer to my questions above regarding what dataset/queries you tested with when you observed fuzzy getting slower?

sreekanth-cb:
Hi @mschoch , iirc , it was the travel-sample index with  a query
"query":{"field":"country","regexp":"u.*d"},
 and other regex queries without a prefix , for eg: "regexp":"[r-s]e*[k]anth"..

mschoch:
So I just ran some tests on the 1M dataset, and interestingly, I see the opposite of @sreekanth-cb 

Wildcard query (which compiles to regex query internally, and chosen because that is what our perf tests use) for "my*h".  Before the change average 47ms (over 5000 queries), after the change it was 50ms (over same 5000 queryes)

Fuzzy query for ideas~1
Before the change 1923ms (average over 100 queries) and after the change 16ms (over 1000 queries)

If the fuzzy improvement looks too good to be true, I did verify it was returning the same result set as before.

This suggests the results may be highly data dependent.

mschoch:
OK, I did some more testing, and found a couple of interesting facts.

First, I confirmed that regexp can also be a lost faster, particularly in the case where there is no prefix.  To test this I went back to the old code and ran the wildcard "?an", this took over 2s on average.  In the new code, it too was in the 50-51ms range.

At first it seemed odd, it too was coming in with almost identical time (50ms) to the 'my*h' that I ran earlier.  I tried a few more patterns, and they all seem to hit a floor around 50ms.  I think what we're identifying is that building the regexp FSA has a cost which consumes around 50ms.

What does this mean?  It means that if you had a highly selective prefix wildcard or regexp, the old code might be faster.  The reason is, we already have prefix optimization in the regular code path, so if that selects a very small number of items, brute forcing the regex on the remaining items is faster.  Unfortunately, we don't have a way to know how selective the prefix is in advance.

I'm going to clean this patch up and proceed under the idea that it's better for a lot of cases.  However, it may be worse for some common cases.  And, it may actually be worse at some cases that our perf tests hit.  We'll see... 😀

mschoch:
I've cleaned up some things and finished the mem segment implementation.

I see there are still some rough edges.  First, we compile the regular expression in the Query, then when using this optimization have to turn it back to string to build the FSA.  Related to this, if there an errors along the way we don't have a clean way to return them.  Fixing this would require some API changes we find unpalatable right now.  Fortunately, we because we compile the regex early (too early) we at least don't expect it to fail to compile later on.

I can live with these issues for now, but will request review from Steve.

PR #851, opened 2018-03-21
 MB-28782: Error handling in merger/persister when index is closed

abhinavdangeti:
When the index is closed, do not fire an AsyncError (fatal) from either
the merger or the persister that is actively working. This is quite a
probable situation, so exit the loop within the goroutine.

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -17,7 +17,6 @@ package scorch
 import (
 	"bytes"
 	"encoding/json"
-
 	"fmt"
 	"os"
 	"sync/atomic"
@@ -59,6 +58,11 @@ OUTER:
 				// lets get started
 				err := s.planMergeAtSnapshot(ourSnapshot, mergePlannerOptions)
 				if err != nil {
+					if err == ErrClosed {
+						// index has been closed
+						_ = ourSnapshot.DecRef()
+						break OUTER
+					}
 					s.fireAsyncError(fmt.Errorf("merging err: %v", err))
 					_ = ourSnapshot.DecRef()
 					atomic.AddUint64(&s.stats.TotFileMergeLoopErr, 1)
@@ -231,7 +235,7 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 		select {
 		case <-s.closeCh:
 			_ = segment.Close()
-			return nil
+			return ErrClosed
 		case s.merges <- sm:
 			atomic.AddUint64(&s.stats.TotFileMergeIntroductions, 1)
 		}
@@ -242,7 +246,7 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 	for _, notification := range notifications {
 		select {
 		case <-s.closeCh:
-			return nil
+			return ErrClosed
 		case newSnapshot := <-notification:
 			atomic.AddUint64(&s.stats.TotFileMergeIntroductionsDone, 1)
 			if newSnapshot != nil {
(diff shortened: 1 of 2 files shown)


index/scorch/merge.go
```diff
@@ -59,7 +59,13 @@ OUTER:
 				// lets get started
 				err := s.planMergeAtSnapshot(ourSnapshot, mergePlannerOptions)
 				if err != nil {
-					s.fireAsyncError(fmt.Errorf("merging err: %v", err))
+					if err == ErrClosed {
+						// if the index was closed while in the process of merging,
+						// just log a message before closing
+						log.Printf("merging err: %v", err)
```
mschoch: ok, so had a long conversation with @steveyen 

1.  let's not log this, feels like it's only going to be confusing seeing an error that's not really an error
2.  but, let's add a break OUTER, this should ensure that we directly exit the goroutine, not go around the loop again and have to detect closing a second time (raising chances that somehow we end up not actually exiting this goroutine)

abhinavdangeti: Done.

abhinavdangeti:
@mschoch The latest patch incorporates your feedback.

mschoch:
I see you also expanded the case to cover merge.  One key difference here is that with persist, we actually need to return error in order to correctly rollback the bolt transaction, which I don't think applies to the merge case.

The handling of breaking out of the loop is cleaner though.  What do you think @steveyen?

index/scorch/merge.go
```diff
@@ -59,7 +58,12 @@ OUTER:
 				// lets get started
 				err := s.planMergeAtSnapshot(ourSnapshot, mergePlannerOptions)
 				if err != nil {
-					s.fireAsyncError(fmt.Errorf("merging err: %v", err))
+					if err == ErrClosed {
+						// index has been closed
+						break OUTER
```
steveyen: feels like this is skipping the ourSnapshot.DecRef() at line 67?

index/scorch/persister.go
```diff
@@ -94,7 +94,12 @@ OUTER:
 				close(ch)
 			}
 			if err != nil {
-				s.fireAsyncError(fmt.Errorf("got err persisting snapshot: %v", err))
+				if err == ErrClosed {
+					// index has been closed
+					break OUTER
```
steveyen: here too with the missing DecRef()?

abhinavdangeti:
Yes, I did extend it - mainly so it looks consistent. Also in case of merge - there is no change in behavior with returning an error as opposed to nil.

steveyen:
> The handling of breaking out of the loop is cleaner though. What do you think @steveyen?

Yah, looks cleaner.  Left some feedback on the ref-counting, though.

index/scorch/merge.go
abhinavdangeti: Ah, yes, good spot @steveyen.

abhinavdangeti:
@mschoch @steveyen Let me know if the latest patchset addresses all your concerns?

index/scorch/merge.go
```diff
@@ -59,8 +58,13 @@ OUTER:
 				// lets get started
 				err := s.planMergeAtSnapshot(ourSnapshot, mergePlannerOptions)
 				if err != nil {
-					s.fireAsyncError(fmt.Errorf("merging err: %v", err))
 					_ = ourSnapshot.DecRef()
```
steveyen: hmm, not sure if this is correct either...

in the old behavior, the refcount was held, the error callback is fired, and then the ref-count is released.

in the proposed change, 1) the refcount is released, 2) then the error callback is fired.  If the callback depends on the index still being alive, the refcount release at step #1 might inadvertently cause trouble.

abhinavdangeti: fireAsyncError just logs a fatal event right now, but I see you point if that were to change.

abhinavdangeti:
@steveyen latest patch incorporates your feedback.

index/scorch/merge.go
```diff
@@ -59,7 +58,13 @@ OUTER:
 				// lets get started
 				err := s.planMergeAtSnapshot(ourSnapshot, mergePlannerOptions)
 				if err != nil {
-					s.fireAsyncError(fmt.Errorf("merging err: %v", err))
+					if err == ErrClosed {
+						// index has been closed
+						_ = ourSnapshot.DecRef()
+						break OUTER
+					} else {
```
steveyen: might as well avoid the else clause here and fall through?

abhinavdangeti:
@steveyen done.

PR #855, opened 2018-03-23
TermFieldReader Advance optimisation

sreekanth-cb:
skips to the target segment and avoid
un necessary read of freq,loc,norm details

### DIFF
--- a/index/scorch/segment/mem/posting.go
+++ b/index/scorch/segment/mem/posting.go
@@ -155,6 +155,27 @@ func (i *PostingsIterator) Next() (segment.Posting, error) {
 	return &i.reuse, nil
 }
 
+func (i *PostingsIterator) Advance(docNumber uint64) (segment.Posting, error) {
+	if i.reuse.Number() == docNumber {
+		return &i.reuse, nil
+	}
+	next, err := i.Next()
+	if err != nil {
+		return next, err
+	}
+
+	nnum := next.Number()
+	for nnum < docNumber {
+		next, err = i.Next()
+		if err != nil || next == nil {
+			return next, err
+		}
+		nnum = next.Number()
+	}
+
+	return next, nil
+}
+
 // Posting is a single entry in a postings list
 type Posting struct {
 	iterator  *PostingsIterator
--- a/index/scorch/segment/mem/posting.go
+++ b/index/scorch/segment/mem/posting.go
@@ -160,7 +160,7 @@ func (i *PostingsIterator) Advance(docNumber uint64) (segment.Posting, error) {
 		return &i.reuse, nil
 	}
 	next, err := i.Next()
-	if err != nil {
+	if err != nil || next == nil {
 		return next, err
 	}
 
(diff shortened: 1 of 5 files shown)


index/scorch/snapshot_index_tfr.go
```diff
@@ -146,6 +162,27 @@ func (i *IndexSnapshotTermFieldReader) Advance(ID index.IndexInternalID, preAllo
 	return next, nil
 }
 
+// glean expects preAlloced TermFieldDoc
+func (i *IndexSnapshotTermFieldReader) glean(ID index.IndexInternalID,
```
steveyen: In the category of "naming is difficult", I think the method name "glean" probably needs a better name :-)

index/scorch/segment/mem/posting.go
```diff
@@ -155,6 +155,24 @@ func (i *PostingsIterator) Next() (segment.Posting, error) {
 	return &i.reuse, nil
 }
 
+func (i *PostingsIterator) Advance(localDocNumber uint64) (segment.Posting, error) {
```
steveyen: one thought that pops up is whether Advance() would ever be called to "advance" to the same localDocNumber that it's already on?  If so, then that might not work with this implementation?

index/scorch/segment/segment.go
```diff
@@ -75,6 +75,11 @@ type PostingsIterator interface {
 	Next() (Posting, error)
 
 	Size() int
+
+	// Advance will return the respective posting by loading
+	// the chunk in which the doc detail resides and calling Next() until
+	// it finds the given doc details.
```
steveyen: This doc-comment for Advance feels a bit too much tied to the proposed initial implementation, when it mentions "calling Next()" in a loop.

For example, I can imagine a future segment implementation (zap2?) that uses an idea like your docValues approach of having an index... so that it might jump more directly to the right place?

That is, I'm thinking the method doc comment should more about what-not-how?

index/scorch/segment/mem/posting.go
sreekanth-cb: whoops..I missed it..Done! thanks!

mschoch: One thought on the variable naming.  When looking at code at the segment level (api or implementation) we should just use "doc number" not "local doc number".  The reason is that in the context of a segment, there is no distinction, it's just the number of a document inside the segment.

index/scorch/segment/mem/posting.go
```diff
@@ -155,6 +155,27 @@ func (i *PostingsIterator) Next() (segment.Posting, error) {
 	return &i.reuse, nil
 }
 
+func (i *PostingsIterator) Advance(docNumber uint64) (segment.Posting, error) {
+	if i.reuse.Number() == docNumber {
+		return &i.reuse, nil
+	}
+	next, err := i.Next()
+	if err != nil {
```
steveyen: feels like there should be a next == nil as an OR check here too?

Issue #377, opened 2016-05-28
Match Phrase Prefix


Venommmm:
Is it possible to match query like 'build softwa" for index item "build software"?

mschoch:
Match Phrase Prefix search is possible, but it's currently not implemented.

There are two ways I can think of implementing it (there may be more).

The first is simple, but won't perform as well, that is to first expand the prefix.  So "softwa" would expand to a few terms, one of which is "software".  Then you combine that with the other phrase terms, and do N separate phrase searches.

The second solution is more complicated, but probably performs better, particularly in cases where the prefix matches a lot of terms.  You would first find documents that contain the start of the phrase, and then you would combine that with a query that expanded the prefix and also ensured the position offsets were 1.  Doing it this way is a bit more complicated but starting with a search on phrase probably cuts down on the candidate documents considerably in practice.

If you're interested in work on this I would suggest starting with the first approach.  It's relatively straightforward enhancement to get started with developing bleve.  Let us know if you try this and have questions.

Venommmm:
Here is my noob code (using this with twitter typeahead), seems it works nicely.
`var searchResults *bleve.SearchResult
	words := strings.Fields(q)

	if len(words) == 1 {

		query_list := make([]bleve.Query, 2)
		query_list[0] = bleve.NewPrefixQuery(q)
		query_list[1] = bleve.NewMatchQuery(q)
		query := bleve.NewDisjunctionQuery(query_list)

		search := bleve.NewSearchRequestOptions(query, 1000, 0, false)
		searchResults, _ = Search_index[Folder].Search(search)

	} else { // more than 1 word
		phrase := ""
		k := 0
		for k != len(words)-1 {
			phrase += words[k] + " "
			k++
		}

		phrase = phrase[0 : len(phrase)-1]
		prefix := words[len(words)-1]

		// query list
		query_list := make([]bleve.Query, 2)

		query1 := bleve.NewMatchPhraseQuery(phrase)

		query_list[0] = query1
		query_list[1] = bleve.NewPrefixQuery(prefix)

		query := bleve.NewConjunctionQuery(query_list)

		search := bleve.NewSearchRequestOptions(query, 5, 0, false)
		searchResults, _ = Search_index[Folder].Search(search)

		// if no results, try match prefix as match
		if len(searchResults.Hits) == 0 {
			query_list[0] = query1
			query_list[1] = bleve.NewMatchQuery(prefix)
[...]

PR #858, opened 2018-03-23
added MatchPhrasePrefixQuery (autocomplete functionality)

zippoxer:
In bleve's current state, it's impossible to perform a phrase prefix search. Yes, we have PrefixQuery, but it's impossible to search phrases with it against an analyzed field.

For instance, if I have a field named `title` which is analyzed with the english analyzer and split into terms, I can only prefix search for one of those terms ("quic"). I can't reliably prefix search `title` for a phrase like "quick fo" without resorting to tricks or duplicating the `title` field to a new one with an analyzer that won't split it.

Following @mschoch's instructions at [https://github.com/blevesearch/bleve/issues/377#issuecomment-300487250](url), I implemented a MatchPhrasePrefixQuery to solve this problem.

I tested it locally and it works for my use case.

I hope this helps anyone who looked for a autocomplete functionality in bleve.

### DIFF
--- a/query.go
+++ b/query.go
@@ -173,6 +173,22 @@ func NewPrefixQuery(prefix string) *query.PrefixQuery {
 	return query.NewPrefixQuery(prefix)
 }
 
+// NewMatchPhrasePrefixQuery creates a new Query which finds
+// documents containing phrases that start with the
+// specified phrase prefix.
+
+// NewMatchPhrasePrefixQuery creates a new Query
+// for matching phrase prefix in the index.
+// An Analyzer is chosen based on the field.
+// Input text is analyzed using this analyzer.
+// Token terms resulting from this analysis are
+// used to build a search phrase.  Result documents
+// must match this phrase prefix. Queried field must have been indexed with
+// IncludeTermVectors set to true.
+func NewMatchPhrasePrefixQuery(matchPhrasePrefix string) *query.MatchPhrasePrefixQuery {
+	return query.NewMatchPhrasePrefixQuery(matchPhrasePrefix)
+}
+
 // NewRegexpQuery creates a new Query which finds
 // documents containing terms that match the
 // specified regular expression.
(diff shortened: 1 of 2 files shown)


connorgorman:
There is a slight bug in this implementation. If there is no value emitted from the first fieldDict.Next() then `terms` is the `nil` slice. This is then added in the second segment below and has the value `[[]]` which causes NewMultiPhraseQuery to panic

```
tfd, err := fieldDict.Next()
for err == nil && tfd != nil {
  terms = append(terms, tfd.Term)
  tfd, err = fieldDict.Next()
}
```

```
phrase[len(phrase)-1] = terms // adds an empty slice to phrase which is then passed to NewMultiPhraseQuery
```

Reproducible panic:
```
type Example struct {
	Field1 string
	Field2 string
}

func TestPanic(t *testing.T) {
	index, _ := bleve.NewMemOnly(bleve.NewIndexMapping())
	example := &Example {
		Field1: "field1",
		Field2: "field2",
	}
	if err := index.Index("example", example); err != nil {
		t.Fatal(err)
	}
	autoReq := bleve.NewMatchPhrasePrefixQuery("field1")
	autoReq.FieldVal = "Field3"
	searchRequest := bleve.NewSearchRequest(autoReq)
	if _, err := index.Search(searchRequest); err != nil {
		t.Fatal(err)
	}
}
```

zippoxer:
@connorgorman Thanks for the feedback!

I fixed the nil bug, but I screwed up this pull request by mistake. I committed merges from the main repo to mine. I tried reverting them, but had conflicts for some reason.

Should I create a new pull request or can we rescue this one?

ffhenkes:
Is this one going forward? I would much appreciate this functionality for my use case :)

amsingh28:
I was going to take a crack at implementing this, but saw this is already done. Is there an ETA for this to  be merged to master?

tmm1:
Ping?

PR #863, opened 2018-03-29
removing dependency on govarint

sreekanth-cb:


index/scorch/segment/zap/merge.go
```diff
@@ -548,6 +547,8 @@ func writePostings(postings *roaring.Bitmap, tfEncoder, locEncoder *chunkedIntCo
 	return postingsOffset, nil
 }
 
+type VarintEncoder func(uint64) (int, error)
```
mschoch: this type probably doens't need to be exported, but we can do this in a follow-on change

sreekanth-cb: done!

PR #868, opened 2018-03-30
persisting the zap version to bolt

sreekanth-cb:

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -16,6 +16,7 @@ package scorch
 
 import (
 	"bytes"
+	"encoding/binary"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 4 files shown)


index/scorch/persister.go
```diff
@@ -317,6 +318,22 @@ func (s *Scorch) persistSnapshotDirect(snapshot *IndexSnapshot) (err error) {
 		return err
 	}
 
+	// persist meta values
+	metaBucket, err := snapshotBucket.CreateBucketIfNotExists(boltMetaDataKey)
+	if err != nil {
+		return err
+	}
+	err = metaBucket.Put([]byte("type"), []byte("zap"))
```
mschoch: feels like "zap" should be a constant in the zap package, just like zap.Version

sreekanth-cb: done!

mschoch:
@sreekanth-cb I just had another thought.  I think we should have bumped the Scorch version, because we changed the metadata we expect to find in the root bolt file.

PR #886, opened 2018-04-09
bumping the scorch version with meta persistence in root bolt

sreekanth-cb:


mschoch:
This is incorrect.  The zap format did not change.  It is the scorch format that changed.  The scorch version number should be bumped.

sreekanth-cb:
Got it. Your comment was very specific about scorch in the last PR. 
To my bad, I thought it was a typo for zap version!

PR #888, opened 2018-04-09
Numeric range pushdown

mschoch:


mschoch:
I set the do not merge label, because I need to do more perf testing first, but you can go ahead and review the changes.

index/scorch/segment/zap/dict.go
```diff
@@ -191,13 +192,55 @@ func (d *Dictionary) FuzzyIterator(term string,
...
+
+	onlyFST, err := vellum.Load(buf.Bytes())
+	if err != nil {
+		rv.err = err
+		return rv
+	}
+
+	itr, err := d.fst.Search(onlyFST, nil, nil)
+	if err == nil {
+		rv.itr = itr
```
steveyen: looks like you have an rv.err = err pattern that you're following, except for the fst.Search() call?

mschoch: good point, looks like that same bug applies to the regexp one above i copied from

mschoch: seems like that bug is in ALL of the dictionary iterators

mschoch: oh i see, presumably it was because they didn't want vellum.ErrIteratorDone to be treated as error further up the stack.  but we should explicitly check that...

mschoch: pushed another commit to try and at least make it more consistent

mschoch:
On my test cluster this brings numeric range search avg down from 63ms to 6ms.

PR #891, opened 2018-04-11
Optimization for reducing merger overhead from doc-values

abhinavdangeti:
The goal here is to reduce the memory overhead from the merger routine, by reducing the number of copies of doc-values. This is a 3 part change:

1. Change the way chunkedContentCoder writes (reversing the order)
2. Add a new flushingChunkedContentCoder which commits data as and when a certain buffer size is reached, follows the file format introduced by the first change as the chunkOffsets and the number of chunks will be written at the end.
3. Getting rid of docTermMap entirely (in persistMergedRest(..))


steveyen:
quick thing is the zap version needs an update, since another change went that took over version 10

index/scorch/segment/zap/contentcoder.go
```diff
@@ -34,10 +34,14 @@ var termSeparator byte = 0xff
...
-	chunkSize    uint64
-	currChunk    uint64
-	chunkLens    []uint64
+	final     []byte
+	chunkSize uint64
+	currChunk uint64
+	chunkLens []uint64
+
+	w                io.Writer
+	writeOutInChunks bool
```
steveyen: minor naming thing -- i have the feeling that this code is writing out chunks no matter what this flag says, so the naming is too close.

maybe it's more like a 'progressive write' kind of thing?

index/scorch/segment/zap/contentcoder.go
```diff
@@ -111,6 +117,15 @@ func (c *chunkedContentCoder) flushContents() error {
 	c.final = append(c.final, c.compressed...)
 
 	c.chunkLens[c.currChunk] = uint64(len(c.compressed) + len(metaData))
+
+	if c.writeOutInChunks {
+		_, err := c.w.Write(c.final)
+		if err != nil {
+			return err
+		}
+		c.final = nil
```
steveyen: this nil'ing will cause memory allocations the next time somebody appends() to final.

instead, should zero out the slice length via c.final = c.final[:0]

index/scorch/segment/zap/contentcoder.go
```diff
@@ -146,33 +161,55 @@ func (c *chunkedContentCoder) Add(docNum uint64, vals []byte) error {
...
 		if err != nil {
 			return tw, err
 		}
 	}
-	// write out the data
-	nw, err = w.Write(c.final)
+
+	chunkOffsetsLen := uint64(tw) - chunkOffsetsStart
+
+	buf = make([]byte, 8)
```
steveyen: should be able to reuse the buf allocated from line 182, resliced to 8 bytes length

index/scorch/segment/zap/contentcoder.go
```diff
@@ -146,33 +161,55 @@ func (c *chunkedContentCoder) Add(docNum uint64, vals []byte) error {
...
+		nw, err := c.w.Write(c.final)
+		tw += nw
+		if err != nil {
+			return tw, err
+		}
 	}
 
+	chunkOffsetsStart := uint64(tw)
+
+	buf := make([]byte, binary.MaxVarintLen64)
```
steveyen: instead of allocating new memory here, you ought to be done with c.final, so probably can reuse it and reslice it (but need to check its cap() first)

index/scorch/segment/zap/contentcoder_test.go
```diff
@@ -73,3 +78,49 @@ func TestChunkContentCoder(t *testing.T) {
...
+	_, err := cic1.Write()
+	if err != nil {
+		t.Fatalf("error writing: %v", err)
+	}
+	_, err = cic2.Write()
+	if err != nil {
+		t.Fatalf("error writing: %v", err)
+	}
+
+	if !reflect.DeepEqual(string(actual1.Bytes()), string(actual2.Bytes())) {
```
steveyen: can use bytes.Compare() instead of DeepEqual()

index/scorch/segment/zap/docvalues.go
```diff
@@ -62,37 +62,44 @@ func (di *docValueReader) curChunkNumber() uint64 {
...
-		return nil, fmt.Errorf("failed to read the field "+
-			"doc values for field %s", field)
+	// read the number of chunks, and chunk offsets position
+	var numChunks, chunkOffsetsPosition uint64
+
+	if fieldDvLocEnd-fieldDvLocStart > 16 {
+		numChunks = binary.BigEndian.Uint64(s.mem[fieldDvLocEnd-8 : fieldDvLocEnd])
+		// read the length of chunk offsets
+		chunkOffsetsLen := binary.BigEndian.Uint64(s.mem[fieldDvLocEnd-16 : fieldDvLocEnd-8])
+		// estimate position of chunk offsets
```
steveyen: the word "estimate" here in the comment feels wrong, as it ought to be a pretty exact calculation

abhinavdangeti:
Thanks @steveyen. I will get all the comments addressed by tomorrow.

index/scorch/segment/zap/contentcoder.go
abhinavdangeti: Done.

index/scorch/segment/zap/contentcoder.go
abhinavdangeti: Done.

index/scorch/segment/zap/docvalues.go
abhinavdangeti: Done.

index/scorch/segment/zap/contentcoder.go
abhinavdangeti: I'll change it to progressiveWrite then.

index/scorch/segment/zap/contentcoder.go
abhinavdangeti: Done.

index/scorch/segment/zap/contentcoder_test.go
abhinavdangeti: Done.

abhinavdangeti:
@steveyen the latest set of changes should incorporate all your feedback.

abhinavdangeti:
Merging this for now, will add the 3rd part as a separate PR.

PR #896, opened 2018-04-12
conditional snappy compression

sreekanth-cb:
Apply compression only if it helps in size reduction by at least 20%


mschoch:
Do we have any numbers on how this helps search speed?

mschoch:
@sreekanth-cb I think we decided against this approach for now, can we close?

sreekanth-cb:
yes @mschoch , closing this.

sreekanth-cb:
As of now, we always apply snappy compression for docValues and a better fix ought to try out newer compression techniques or applying compression on a lengthier slice of data to attain better compression.

You may unsubscribe to all those features which you don't use as  to reduce the index size.
for eg: term vectors, stored fields, include in _all, docValues etc.

PR #897, opened 2018-04-13
improve scorch Advance implementation

mschoch:
advance will jump directly to the expected segment
then it will skip directly to the expectec chunk

### DIFF
--- a/index/scorch/segment/zap/posting.go
+++ b/index/scorch/segment/zap/posting.go
@@ -504,7 +504,18 @@ func (i *PostingsIterator) readLocation(l *Location) error {
 
 // Next returns the next posting on the postings list, or nil at the end
 func (i *PostingsIterator) Next() (segment.Posting, error) {
-	docNum, exists, err := i.nextDocNum()
+	return i.nextAtOrAfter(0)
+}
+
+// Advance returns the posting at the specified docNum or it is not present
+// the next posting, or if the end is reached, nil
+func (i *PostingsIterator) Advance(docNum uint64) (segment.Posting, error) {
+	return i.nextAtOrAfter(docNum)
+}
+
+// Next returns the next posting on the postings list, or nil at the end
+func (i *PostingsIterator) nextAtOrAfter(atOrAfter uint64) (segment.Posting, error) {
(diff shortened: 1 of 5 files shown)


index/scorch/segment/zap/posting.go
```diff
@@ -617,6 +633,13 @@ func (i *PostingsIterator) nextDocNum() (uint64, bool, error) {
...
 
 	n := i.Actual.Next()
+	for uint64(n) < atOrAfter && i.Actual.HasNext() {
+		n = i.Actual.Next()
+	}
+	if uint64(n) < atOrAfter {
+		// couldn't find anything
+		return 0, false, nil
+	}
 	allN := i.all.Next()
```
steveyen: wondering the i.all iterator also needs the same treatment?

mschoch: hmmn, i think we have very different understandings of the way this works.  the purpose of the loop below is to safely allow "all" to catch up to where "n" is.  i say safely, because when we get to the right chunk we have to skip over freqs and locs, as dicated by the all iterator.  if we just jump it ahead here we're lost.

steveyen: ah, right

index/scorch/snapshot_index_tfr.go
```diff
@@ -134,24 +135,37 @@ func (i *IndexSnapshotTermFieldReader) Advance(ID index.IndexInternalID, preAllo
...
 		*i = *(i2.(*IndexSnapshotTermFieldReader))
 	}
-	// FIXME do something better
-	next, err := i.Next(preAlloced)
+	num, err := docInternalToNumber(ID)
+	if err != nil {
+		return nil, fmt.Errorf("error converting to doc number % x - %v", ID, err)
+	}
+	segIndex, ldocNum := i.snapshot.segmentIndexAndLocalDocNumFromGlobal(num)
+	if segIndex > len(i.snapshot.segment) {
```
steveyen: maybe should be >= here?

mschoch: yep, that is carryover from sreekanth's original patch, i noticed that too and forgot

steveyen:
yep, a belated: looks good

PR #900, opened 2018-04-17
Optimization for reducing merger overhead from doc-values

abhinavdangeti:
Part3: Getting rid of docTermMap during the merge operation,
       instead fetch all valid docNums and their corresponding
       values from the segmentBases via docValueReaders.

### DIFF
--- a/index/scorch/segment/zap/docvalues.go
+++ b/index/scorch/segment/zap/docvalues.go
@@ -34,10 +34,11 @@ func init() {
 	reflectStaticSizedocValueReader = int(reflect.TypeOf(dvi).Size())
 }
 
+type docNumTermsVisitor func(docNum uint64, terms []byte) error
+
 type docValueReader struct {
 	field          string
 	curChunkNum    uint64
-	numChunks      uint64
 	chunkOffsets   []uint64
 	dvDataLoc      uint64
 	curChunkHeader []MetaData
@@ -148,6 +149,34 @@ func (di *docValueReader) loadDvChunk(chunkNumber uint64, s *SegmentBase) error 
 	di.curChunkData = s.mem[compressedDataLoc : compressedDataLoc+dataLength]
 	di.curChunkNum = chunkNumber
 	di.uncompressed = di.uncompressed[:0]
+	return nil
+}
+
+func (di *docValueReader) iterateAllDocValues(s *SegmentBase, visitor docNumTermsVisitor) error {
+	for i := 0; i < len(di.chunkOffsets); i++ {
+		err := di.loadDvChunk(uint64(i), s)
+		if err != nil {
+			return err
+		}
+
+		// uncompress the already loaded data
+		uncompressed, err := snappy.Decode(di.uncompressed[:cap(di.uncompressed)], di.curChunkData)
+		if err != nil {
+			return err
+		}
+		di.uncompressed = uncompressed
+
+		start := uint64(0)
+		for _, entry := range di.curChunkHeader {
+			err = visitor(entry.DocNum, uncompressed[start:entry.DocDvOffset])
+			if err != nil {
+				return err
+			}
+
+			start = entry.DocDvOffset
+		}
+	}
+
 	return nil
 }
 
--- a/index/scorch/segment/zap/merge.go
+++ b/index/scorch/segment/zap/merge.go
@@ -175,12 +175,6 @@ func persistMergedRest(segments []*SegmentBase, dropsIn []*roaring.Bitmap,
 	tfEncoder := newChunkedIntCoder(uint64(chunkFactor), newSegDocCount-1)
 	locEncoder := newChunkedIntCoder(uint64(chunkFactor), newSegDocCount-1)
 
-	// docTermMap is keyed by docNum, where the array impl provides
-	// better memory usage behavior than a sparse-friendlier hashmap
-	// for when docs have much structural similarity (i.e., every doc
-	// has a given field)
-	var docTermMap [][]byte
-
 	var vellumBuf bytes.Buffer
 	newVellum, err := vellum.New(&vellumBuf, nil)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


index/scorch/segment/zap/docvalues.go
```diff
@@ -135,6 +137,40 @@ func (di *docValueReader) loadDvChunk(chunkNumber uint64, s *SegmentBase) error
...
+func (di *docValueReader) iterateDocTerms(s *SegmentBase, visitor docNumTermsVisitor) error {
+	for i := uint64(0); i < di.numChunks; i++ {
+		err := di.loadDvChunk(i, s)
+		if err != nil {
+			return err
+		}
+
+		var uncompressed []byte
+		// use the uncompressed copy if available
+		if len(di.uncompressed) > 0 {
```
steveyen: after calling loadDvChunk(), looks like len(di.uncompressed) will always be 0?  so, don't need the if-check here and the else-clause codepath always runs?

index/scorch/segment/zap/docvalues.go
```diff
@@ -135,6 +137,40 @@ func (di *docValueReader) loadDvChunk(chunkNumber uint64, s *SegmentBase) error
...
+			}
+			di.uncompressed = uncompressed
+		}
+
+		start := uint64(0)
+		for _, entry := range di.curChunkHeader {
+			err = visitor(entry.DocNum, uncompressed[start:entry.DocDvOffset])
+			start = entry.DocDvOffset
+
+			if err != nil {
```
steveyen: this err check should happen at line 162.5?

index/scorch/segment/zap/merge.go
```diff
@@ -361,14 +349,29 @@ func persistMergedRest(segments []*SegmentBase, dropsIn []*roaring.Bitmap,
...
+			}
+
+			dropI := drops[segmentI]
+
+			err = fdvReader.iterateDocTerms(segment, func(docNum uint64, terms []byte) error {
+				if dropI != nil && dropI.Contains(uint32(docNum)) {
+					return nil
 				}
+				err := fdvEncoder.Add(newDocNum, terms)
+				newDocNum++
```
steveyen: hmmm, this newDocNum++ thing feels wrong...

have to instead look it up in the newDocNums?  perhaps by doing something like... newDocNums[segmentI][docNum]

abhinavdangeti: Hmm, let me think a little bit more about this..

steveyen: >> ... perhaps by doing something like... newDocNums[segmentI][docNum]
> Hmm, let me think a little bit more about this..

Basically, the idea is that it has to remap from an old docNum to the right new doc num.  Just doing a "newDocNum++" would be assigning an incorrect different number.

abhinavdangeti: Yes I agree on that, but I think newDocNums[segmentI][docNum] wouldn't fetch us the right newDocNum. Also I'm working through a unit test failure seen with this change, where certain terms that should've been present in certain old segments aren't..

steveyen: > ... I think newDocNums[segmentI][docNum] wouldn't fetch us the right newDocNum. Also I'm working through a unit test failure seen with this change, where certain terms that should've been present in certain old segments aren't.

good point -- looks like the "segmentI" lookup int has a different "basis" w.r.t. newDocNums vs the segmentsInFocus

steveyen: > good point -- looks like the "segmentI" lookup int has a different "basis" w.r.t. newDocNums vs the segmentsInFocus

uh, nevermind -- i might have misunderstood that (where I was incorrectly looking at or thinking about "newDocsNumIn" when I wrote that)

abhinavdangeti: oh, so segmentsInFocus is intended to keep the segmentI lookup correct. Its the second 'docNum' lookup which doesn't seem right.

abhinavdangeti: A slight correction on my last comment here - the unit test failure is because certain segments don't have an initialized fieldDvReader for the fieldID.

index/scorch/segment/zap/merge.go
```diff
@@ -361,14 +349,30 @@ func persistMergedRest(segments []*SegmentBase, dropsIn []*roaring.Bitmap,
...
+			fieldIDPlus1 := uint16(segment.fieldsMap[fieldName])
+			if dvIter, exists := segment.fieldDvReaders[fieldIDPlus1-1]; exists &&
+				dvIter != nil {
+				i := 0
+				err = dvIter.iterateAllDocValues(segment, func(docNum uint64, terms []byte) error {
+					if newDocNums[segmentI][i] == docDropped {
+						i++
+						return nil
+					}
+					err := fdvEncoder.Add(newDocNums[segmentI][i], terms)
```
steveyen: using 'i' here feels wrong?  That is, using the callback param 'docNum' here instead of 'i' feels more right to me?

So, more like... newDocUms[segmentI][docNum] 

Thoughts?

abhinavdangeti: Yes agreed, docNums will align with index positions of newDocNums[seg..].

index/scorch/segment/zap/merge.go
```diff
@@ -361,14 +349,27 @@ func persistMergedRest(segments []*SegmentBase, dropsIn []*roaring.Bitmap,
...
+					}
+					return nil
+				})
 				if err != nil {
 					return nil, 0, err
 				}
 			}
 		}
+
 		err = fdvEncoder.Close()
```
steveyen: one thought perhaps to address the issue of zap.Segment.VisitableDocValueFields()...

is perhaps if the fdvEncoder is empty, then don't do an fdvEncoder.Close() and don't do a final fdvEncoder.Write()?

instead, an empty fdvEncoder should have a fieldDvLocsStart/fieldDvLocsEnd value that tells a future reader that there's no info here.

of note, there might a be a subtle difference between "there's no info here" versus fieldDocValues are disabled...  not sure.  The subtlety might tell the reader in the future whether to do the cached-doc-values in-memory uninversion maneuver or not.

PR #901, opened 2018-04-19
alternate disjunction searcher impl using heap

mschoch:
introduce an alternate disjunction searcher implementation that
uses a heap to track searchers.  when you have a disjunction
of many other searchers, this should be more efficient by
performing much fewer key comparisions.

currently a cut-over from slice to heap is set at 10,
this can be adjusted by setting the package-level variable
named DisjunctionHeapTakeover

### DIFF
--- a/search/searcher/search_disjunction_heap.go
+++ b/search/searcher/search_disjunction_heap.go
@@ -1,0 +1,342 @@
+//  Copyright (c) 2018 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package searcher
+
+import (
+	"bytes"
+	"container/heap"
+	"math"
+	"reflect"
+
+	"github.com/blevesearch/bleve/index"
+	"github.com/blevesearch/bleve/search"
+	"github.com/blevesearch/bleve/search/scorer"
+	"github.com/blevesearch/bleve/size"
+)
+
+var reflectStaticSizeDisjunctionHeapSearcher int
+var reflectStaticSizeSearcherCurr int
+
+func init() {
+	var dhs DisjunctionHeapSearcher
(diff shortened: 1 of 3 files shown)


search/searcher/search_disjunction_heap.go
```diff
@@ -0,0 +1,342 @@
...
+		for _, matchingCurr := range s.matchingCurrs {
+			if matchingCurr.curr != rv {
+				ctx.DocumentMatchPool.Put(matchingCurr.curr)
+			}
+			curr, err := matchingCurr.searcher.Next(ctx)
+			if err != nil {
+				return nil, err
+			}
+			if curr != nil {
+				matchingCurr.curr = curr
```
steveyen: wondering if updating the matchingCurr.curr is needed, as later on, i suspect the s.updateMatches() is going to blow away the matchingCurr's anyways?

mschoch: not sure i follow your question, the matchingCurrs slice will be reset later, here we're mutating the curr field of a single matchingCurr instance, which matters because we're add it back into the heap on the next line, if we didn't update it, we'd be placing it into the wrong place in the heap based on it's previous value, not it's current value.

search/searcher/search_disjunction_heap.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+// heap impl
+
+func (s *DisjunctionHeapSearcher) Len() int { return len(s.heap) }
+
+func (s *DisjunctionHeapSearcher) Less(i, j int) bool {
+	if s.heap[i].curr == nil {
+		return true
+	} else if s.heap[i].curr != nil && s.heap[j].curr == nil {
```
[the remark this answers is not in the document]
mschoch: yes, you're correct

mschoch:
I did a bit of further testing and did not come to any strong conclusion.

I tested disjunction queries of medium frequency terms (between 5k and 6k hits when we last analyzed the dictionary).  The exact frequency is likely not the same in my dataset, but presumably it's representative and these terms have similar frequency.

I then built disjunctions queries of 5, 10 and 15 of these (always same grouping, not mixing it up).
The latencies observed before were 6ms, 8ms and 13ms respectively.  The latencies after this change were 6ms, 9ms and 13ms.  Taken at face value this would suggest that the heap approach isn't significantly slower, even on smaller groupings.

However, my results with a fuzzy~1 query seemed to contradict this, so it may matter more the details of the terms themselves.

Based on this I'm going to propose we merge it as is, and we can revisit it if we find any perf regressions.

search/searcher/search_disjunction_slice.go
```diff
@@ -0,0 +1,298 @@
+//  Copyright (c) 2014 Couchbase, Inc.
```
abhinavdangeti: 2018 :)

PR #907, opened 2018-04-21
MB-28847: Account for bytes of an allocated buffer while building new segment

abhinavdangeti:
Account for these bytes in the memory used by the scorch index.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -76,10 +76,12 @@ type Scorch struct {
 }
 
 type internalStats struct {
-	persistEpoch        uint64
-	persistSnapshotSize uint64
-	mergeEpoch          uint64
-	mergeSnapshotSize   uint64
+	persistEpoch          uint64
+	persistSnapshotSize   uint64
+	mergeEpoch            uint64
+	mergeSnapshotSize     uint64
+	newSegBufBytesAdded   uint64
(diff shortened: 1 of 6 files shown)


index/scorch/scorch.go
```diff
@@ -308,11 +310,13 @@ func (s *Scorch) Batch(batch *index.Batch) (err error) {
...
 
 	var newSegment segment.Segment
+	var bufBytes uint64
 	if len(analysisResults) > 0 {
-		newSegment, err = zap.AnalysisResultsToSegmentBase(analysisResults, DefaultChunkFactor)
+		newSegment, bufBytes, err = zap.AnalysisResultsToSegmentBase(analysisResults, DefaultChunkFactor)
 		if err != nil {
 			return err
 		}
+		atomic.AddUint64(&s.iStats.newSegBufBytesAdded, bufBytes)
```
steveyen: I have the feeling that this change won't do the trick...

That is, by the time the bufBytes has been added to the stats here, it's too late and the show's already over.

mschoch: I'm no longer convinced any of our memory accounting "does the trick".

abhinavdangeti: @steveyen Let me plug in some timing stats here, to get that information.

PR #910, opened 2018-04-24
Include missing initialization for size entry of integers

abhinavdangeti:


size/sizes.go
```diff
@@ -19,28 +19,30 @@ import (
...
+	var b bool
+	SizeOfBool = int(reflect.TypeOf(b).Size())
+	var f32 float32
+	SizeOfFloat32 = int(reflect.TypeOf(f32).Size())
+	var f64 float64
+	SizeOfFloat64 = int(reflect.TypeOf(f64).Size())
+	var i int
+	SizeOfInt = int(reflect.TypeOf(i).Size())
+	var m map[int]int
+	SizeOfMap = int(reflect.TypeOf(m).Size())
```
mschoch: SizeOfMap implies something more general, but is it really the same regardless of type?

abhinavdangeti: Yes, the overhead is the same regardless of the types the map contains.

mschoch: thanks, in that case i can simplify my change as well and use this

PR #911, opened 2018-04-24
account for memory allocated during text analysis

mschoch:


analysis/freq.go
```diff
@@ -34,6 +59,14 @@ type TokenFreq struct {
 	frequency int
 }
 
+func (tf *TokenFreq) Size() int {
+	rv := reflectStaticSizeTokenFreq
```
abhinavdangeti: Shouldn't we also include len(tf.Term) here?

mschoch: ya good catch

index/scorch/scorch.go
```diff
@@ -295,12 +297,17 @@ func (s *Scorch) Batch(batch *index.Batch) (err error) {
 	// wait for analysis result
 	analysisResults := make([]*index.AnalysisResult, int(numUpdates))
 	var itemsDeQueued uint64
+	var totalAnalysisSize int
 	for itemsDeQueued < numUpdates {
 		result := <-resultChan
+		resultSize := result.Size()
+		atomic.AddUint64(&s.iStats.newSegBufBytesAdded, uint64(resultSize))
```
abhinavdangeti: Sorry for missing this before, but shouldn't this have been analysisBytesAdded, and below analysisBytesRemoved?

mschoch:
Did an index build with these changes, seemed better, but not still not great.

PR #913, opened 2018-04-25
minor memory savings during collection

sreekanth-cb:


mschoch:
can you elaborate how this saves memory?  does it have to do with the closure creating garbage inside a hot loop?

sreekanth-cb:
yes, thats the thinking..

mschoch:
thinking or measurement? seems like this one should be straightforward to verify with bleve-bench

mschoch:
@sreekanth-cb I believe this has already been rolled into one of the other PRs and merged.

sreekanth-cb:
thats right @mschoch , thx!

PR #916, opened 2018-04-27
dvReader improvement

sreekanth-cb:
-attempt to improve the reuse of dvReaders
-avoid the redundant dvCache prep and checks for a given segment

there is this ugly FieldDocValueState thing introduced to save the runtime dv cache preparations and checks, looking for better ways to handle this..  (thought of putting these methods to the original DocVisitState interface, but even that doesn't looked any better..) . another flip side is dvState management kind of spread across the index_snapshot and zap layer(docvalues.go)..

### DIFF
--- a/index/scorch/segment/segment.go
+++ b/index/scorch/segment/segment.go
@@ -120,19 +120,4 @@ type DocumentFieldTermVisitable interface {
 }
 
 type DocVisitState interface {
-	State() *FieldDocValueState
-	SetState(*FieldDocValueState)
-}
-
-// FieldDocValueState represents the state details,
-// which intents to save the redundant dvCache preparations
-type FieldDocValueState struct {
-	DvFieldsAllPersisted bool
-	DvFieldsPending      []string
-	DvCachePrepared      bool
-	DvSegment            DocumentFieldTermVisitable
-}
-
-func (fdvs *FieldDocValueState) CurrentSegment() DocumentFieldTermVisitable {
-	return fdvs.DvSegment
 }
--- a/index/scorch/segment/segment.go
+++ b/index/scorch/segment/segment.go
@@ -111,10 +111,28 @@ type Location interface {
 // postings or other indexed values.
 type DocumentFieldTermVisitable interface {
 	VisitDocumentFieldTerms(localDocNum uint64, fields []string,
-		visitor index.DocumentFieldTermVisitor) error
+		visitor index.DocumentFieldTermVisitor, optional DocVisitState) (DocVisitState, error)
 
 	// VisitableDocValueFields implementation should return
 	// the list of fields which are document value persisted and
 	// therefore visitable by the above VisitDocumentFieldTerms method.
 	VisitableDocValueFields() ([]string, error)
+}
+
+type DocVisitState interface {
+	State() *FieldDocValueState
+	SetState(*FieldDocValueState)
+}
+
+// FieldDocValueState represents the state details,
+// which intents to save the redundant dvCache preparations
+type FieldDocValueState struct {
+	DvFieldsAllPersisted bool
+	DvFieldsPending      []string
+	DvCachePrepared      bool
+	DvSegment            DocumentFieldTermVisitable
+}
+
+func (fdvs *FieldDocValueState) CurrentSegment() DocumentFieldTermVisitable {
+	return fdvs.DvSegment
 }
--- a/index/scorch/segment/zap/docvalues.go
+++ b/index/scorch/segment/zap/docvalues.go
@@ -38,16 +38,8 @@ func init() {
 type docNumTermsVisitor func(docNum uint64, terms []byte) error
 
 type docVisitState struct {
-	dvrs  map[uint16]*docValueReader
-	state *segment.FieldDocValueState
-}
-
-func (dvs *docVisitState) SetState(state *segment.FieldDocValueState) {
-	dvs.state = state
-}
-
-func (dvs *docVisitState) State() *segment.FieldDocValueState {
-	return dvs.state
+	dvrs    map[uint16]*docValueReader
+	segment *Segment
 }
 
 type docValueReader struct {
@@ -252,10 +244,6 @@ func (di *docValueReader) getDocValueLocs(docNum uint64) (uint64, uint64) {
 	return math.MaxUint64, math.MaxUint64
 }
 
-func (s *Segment) CurrentSegment() segment.DocumentFieldTermVisitable {
-	return s
-}
-
 // VisitDocumentFieldTerms is an implementation of the
 // DocumentFieldTermVisitable interface
 func (s *Segment) VisitDocumentFieldTerms(localDocNum uint64, fields []string,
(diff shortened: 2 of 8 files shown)


index/scorch/segment/zap/docvalues.go
```diff
@@ -238,13 +252,31 @@ func (di *docValueReader) getDocValueLocs(docNum uint64) (uint64, uint64) {
...
+	} else {
+		if dvs.state.DvSegment != s {
+			dvs.state = &segment.FieldDocValueState{DvSegment: s}
+			dvs.dvrs = make(map[uint16]*docValueReader, len(fields))
+		}
+	}
+
 	fieldIDPlus1 := uint16(0)
-	ok := true
+	ok = true
```
mschoch: why is this needed? it seems like in all subsequent places we check it, we have assigned to it again

sreekanth-cb:
Tried to benchmark these on perf tests.
http://perf.jenkins.couchbase.com/job/atlas-sdk/1688/console
latency: "value": 199.0
throughput:  "value": 50.89
looks better.

index/scorch/segment/segment.go
```diff
@@ -111,10 +111,28 @@ type Location interface {
...
+		visitor index.DocumentFieldTermVisitor, optional DocVisitState) (DocVisitState, error)
 
 	// VisitableDocValueFields implementation should return
 	// the list of fields which are document value persisted and
 	// therefore visitable by the above VisitDocumentFieldTerms method.
 	VisitableDocValueFields() ([]string, error)
 }
+
+type DocVisitState interface {
+	State() *FieldDocValueState
```
steveyen: exposing FieldDocValueState has a feeling of leaking a lot of internal info to the world?

index/scorch/segment/segment.go
```diff
@@ -111,10 +111,28 @@ type Location interface {
...
+	SetState(*FieldDocValueState)
+}
+
+// FieldDocValueState represents the state details,
+// which intents to save the redundant dvCache preparations
+type FieldDocValueState struct {
+	DvFieldsAllPersisted bool
+	DvFieldsPending      []string
+	DvCachePrepared      bool
```
steveyen: like, the idea that there's a cache or whether fields are "all persisted" is "internal baseball"?

index/scorch/segment/segment.go
```diff
@@ -111,10 +111,28 @@ type Location interface {
...
+	DvSegment            DocumentFieldTermVisitable
+}
+
+func (fdvs *FieldDocValueState) CurrentSegment() DocumentFieldTermVisitable {
+	return fdvs.DvSegment
+}
```
steveyen: The term "segment" kinda already has overloaded meaning in scorch, and this isn't returning a segment?  Perhaps the caller should access the DvSegment field directly?

steveyen:
overall, I agree with @marty on the complexity jump in this PR.  One thought (from intuition) is that I sense that the caching stuff shouldn't need to be changed as part of this optimization, and perhaps that's why the complexity has increased due to the linkage with caching as part of this PR

sreekanth-cb:
hm..agree to some extent but  at the same time  
I strongly felt we shouldn't be computing the delta fields to be dv computed for every hit on a segment.  Worst case it will trigger the cache preparation go routine a 1024 times per segment? felt that qualifies the cache part fix  as a correctness one than an optimisation?

steveyen:
> I strongly felt we shouldn't be computing the delta fields to be dv computed for every hit on a segment. Worst case it will trigger the cache preparation go routine a 1024 times per segment? felt that qualifies the cache part fix as a correctness one than an optimisation?

that might be true, but couple more reactions...

- is it possible to break this into smaller commits, where each one has smaller mental "surface area"?

- the caching might be important, but is there a way where the caching concept doesn't leak or break abstraction layers?  That is, from one point of view, before this change, the caching was just a concern of the bleve/index/scorch package, and other packages didn't care about caching?  Now, it appears with this PR that the various segments and segment implementations also now have knowledge about the caching concern.  So, there's a feel of a "abstractions layering or blackbox'ing" issue here.

A final thing is with time getting short, I realize getting this in sooner rather than perfect might be the right approach, but still having some mental blockage.

sreekanth-cb:
thanks @steveyen for bringing up the layering issue, that gave another thought to move these dvState/cacheState to the snapshot_segment along with the cachedDocs and updated the PR with those changes. You may please check it again..

index/scorch/segment/zap/docvalues.go
```diff
@@ -238,13 +244,28 @@ func (di *docValueReader) getDocValueLocs(docNum uint64) (uint64, uint64) {
 	return math.MaxUint64, math.MaxUint64
 }
 
+func (s *Segment) CurrentSegment() segment.DocumentFieldTermVisitable {
+	return s
```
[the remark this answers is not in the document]
sreekanth-cb: Oops..this was a left over function from the first commit. Now its no longer needed. Cleaning it.

index/scorch/segment/zap/docvalues.go
```diff
@@ -254,20 +275,22 @@ func (s *SegmentBase) VisitDocumentFieldTerms(localDocNum uint64, fields []strin
...
+				dvs.dvrs[fieldIDPlus1-1] = dvIter.cloneInto(dvs.dvrs[fieldIDPlus1-1])
+			}
 
 			// check if the chunk is already loaded
-			if docInChunk != dvIterClone.curChunkNumber() {
-				err := dvIterClone.loadDvChunk(docInChunk, s)
+			if docInChunk != dvs.dvrs[fieldIDPlus1-1].curChunkNumber() {
+				err := dvs.dvrs[fieldIDPlus1-1].loadDvChunk(docInChunk, &s.SegmentBase)
 				if err != nil {
 					continue
```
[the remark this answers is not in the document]
sreekanth-cb: even if the dvReader load fails for a Field, we may attempt to try it for other fields requested is the motive here.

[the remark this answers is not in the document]
mschoch: ok, that's fine, but what happens to that first field?  will fail to return correct data for it silently?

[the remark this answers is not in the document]
sreekanth-cb: hm..so far I thought that was fine as for some reason my impression was that we are not really propagating an err at the collector level for this. But now when I question myself about this I am not sure whether this is correct..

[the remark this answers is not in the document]
mschoch: one thought (which i can't confirm because i don't really understand the flow any more) is that it would be OK if we knew what we'd end up un-inverting the field at runtime to mask the problem.  but i don't know if thats the case or not.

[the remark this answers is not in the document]
sreekanth-cb: for non persisted doc value fields, the control won't come here as it is handled at a segment snapshot level cache and this is only for already persisted docValues. And then a read error could be treated like a major and bail out right there..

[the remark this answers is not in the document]
mschoch: I've taken a look at `loadDvChunk` and it seems the only case it returns an error is when we fail to decode a Uvarint at a location where we expect one.  I think this should be considered fatal as it suggests some sort of corruption.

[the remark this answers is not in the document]
sreekanth-cb: done!

index/scorch/segment/zap/docvalues.go
```diff
@@ -254,20 +271,22 @@ func (s *SegmentBase) VisitDocumentFieldTerms(localDocNum uint64, fields []strin
 
 		if dvIter, exists := s.fieldDvReaders[fieldIDPlus1-1]; exists &&
 			dvIter != nil {
-			dvIterClone = dvIter.cloneInto(dvIterClone)
+			if _, ok := dvs.dvrs[fieldIDPlus1-1]; !ok {
```
steveyen: looks fieldIDPlus1-1 is recomputed a few times?  might as well do it once with a helper var?

steveyen: also, there's a lot of map lookups (dvs.dvrs[fieldIDPlus1-1]) -- ought to be able to do it once and reuse?

index/scorch/segment/zap/docvalues.go
```diff
@@ -254,20 +271,22 @@ func (s *SegmentBase) VisitDocumentFieldTerms(localDocNum uint64, fields []strin
```
steveyen: looks like this docInChunk calculation can be moved to outside the loop?

index/scorch/snapshot_index.go
```diff
@@ -496,30 +496,55 @@ func docInternalToNumber(in index.IndexInternalID) (uint64, error) {
...
 	if segmentIndex >= len(i.segment) {
-		return nil
+		return nil, nil
 	}
 
 	ss := i.segment[segmentIndex]
 
 	if zaps, ok := ss.segment.(segment.DocumentFieldTermVisitable); ok {
+		if ss.dvState == nil {
+			ss.dvState = &fieldDocValueState{}
```
steveyen: the segment snapshot (ss) ought to immutable?  having a mutable dvState field here feels wrong?

index/scorch/snapshot_index.go
```diff
@@ -496,30 +496,55 @@ func docInternalToNumber(in index.IndexInternalID) (uint64, error) {
...
+		}
+
+		// if all fields are dv persisted
+		if ss.dvState.dvFieldsAllPersisted {
```
steveyen: still wondering about this dvFieldsAllPersisted bool?  that is, why there should be a special case for all fields getting persisted (as opposed to, say, some or none of the fields getting persisted)... that is all cases ought to perhaps just "fall out" for free?

mschoch:
I just want to echo one of @steveyen's comments.  I think the attempt to include that second level caching optimization is making this change harder to review.  Since we are under some time pressure to get this through, could you remove that and propose it as a separate change.  If there is time, we will try to include it as well.

sreekanth-cb:
@steveyen , the segment_snapshot was made mutable for the dvCache preparations originally? not sure whether I am overlooking something else here?

index/scorch/snapshot_index.go
sreekanth-cb: Added this bool to save further calls to VisitableDocValueFields and extractDvPendingFields on every invocation.

sreekanth-cb:
Sure, let me make it two separate PRs.

sreekanth-cb:
hey @mschoch , @steveyen , I removed the cache prep optimisations from this PR to target it with only the dvReader reuse optimisation for the perf regression observed.

search/collector/topn.go
```diff
@@ -140,6 +142,20 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
 		DocumentMatchPool: search.NewDocumentMatchPool(backingSize+searcher.DocumentMatchPoolSize(), len(hc.sort)),
 	}
 
+	hc.dvReader, err = reader.DocValueReader(hc.neededFields)
+	if err != nil {
+		return err
+	}
+
+	if hc.updateFieldVisitor == nil {
```
steveyen: since this is now initialized earlier in Collect(), i think hc.updateFieldVisitor at this point in time is always going to be nil?

index/scorch/segment/zap/docvalues.go
```diff
@@ -240,34 +246,49 @@ func (di *docValueReader) getDocValueLocs(docNum uint64) (uint64, uint64) {
...
-		// find the chunkNumber where the docValues are stored
-		docInChunk := localDocNum / uint64(s.chunkFactor)
-
-		if dvIter, exists := s.fieldDvReaders[fieldIDPlus1-1]; exists &&
+		fieldID := fieldIDPlus1 - 1
+		if dvIter, exists := s.fieldDvReaders[fieldID]; exists &&
 			dvIter != nil {
-			dvIterClone = dvIter.cloneInto(dvIterClone)
+			if dvr, ok = dvs.dvrs[fieldID]; !ok {
+				dvr = dvIter.cloneInto(dvr)
```
steveyen: one thought is to move the initialization of the dvs.dvrs to earlier, like up to line 260.5, where you know it needs to be initialized because dvrs is newly created.  Then, that would save an extra map lookup here at line 273 of s.fieldDvReaders[fieldID], where instead you'd only need to do a single lookup into the dvs.dvrs[fieldID] map.

steveyen:
looks much better -- also, the commit msg is now slightly wrong as you're no longer affecting the caching in this particular PR

search/collector/topn.go
sreekanth-cb: sreekanth acting like a review response bot..hence this is not seen earlier..

index/scorch/segment/zap/docvalues.go
```diff
@@ -240,34 +246,49 @@ func (di *docValueReader) getDocValueLocs(docNum uint64) (uint64, uint64) {
...
+		if dvs.segment != s {
+			dvs.segment = s
+			dvs.dvrs = make(map[uint16]*docValueReader, len(fields))
+		}
+	}
+
+	// find the chunkNumber where the docValues are stored
+	docInChunk := localDocNum / uint64(s.chunkFactor)
+	var dvr *docValueReader
+	var fieldIDPlus1 uint16
```
mschoch: this variable ins't used outside the for loop, perhaps declare it inside?

steveyen:
yep, looks good to me

PR #922, opened 2018-04-30
Micro optimizations while estimating mem needed for search

abhinavdangeti:


index_impl.go
```diff
@@ -385,35 +404,27 @@ func memNeededForSearch(req *SearchRequest,
...
 	}
 
 	// highlighting, store
-	var d document.Document
 	if len(req.Fields) > 0 || req.Highlight != nil {
-		for i := 0; i < (req.Size + req.From); i++ { // size + from => number of hits
-			estimate += (req.Size + req.From) * d.Size()
-		}
+		// Size + From => number of hits
+		estimate += (req.Size + req.From) * documentSizeEstimate
```
steveyen: the math here seems different than the original -- did the original code have a bug?

abhinavdangeti: Yes, it seems like we were double calculating it.

index_impl.go
```diff
@@ -368,6 +368,25 @@ func (i *indexImpl) Search(req *SearchRequest) (sr *SearchResult, err error) {
 	return i.SearchInContext(context.Background(), req)
 }
 
+var documentMatchSizeEstimate int
+var searchContextSizeEstimate int
+var facetResultSizeEstimate int
+var documentSizeEstimate int
```
steveyen: minor thing on naming -- these don't feel like estimates to me, but seem pretty deterministic

steveyen: that is, if you remove the suffix "Estimate", then it'd have more correct naming?

steveyen: or, perhaps replace "SizeEstimate" with "EmptySize"?

abhinavdangeti: Sounds good.

mschoch:
Ya, we should be careful tweaking this logic right now.  We *just* got our first indication from Alex that the memory quota was working acceptably in the DGM case.  So we run the risk that even if this is more correct, it could under-count in the big picture again.

PR #924, opened 2018-05-03
save redundant dvCache prep calls

sreekanth-cb:


sreekanth-cb:
not ready...up for discussion..

mschoch:
@sreekanth-cb since we'll be reviewing this asynchronously, and I see some bigger changes from what we saw before, would be possible to include like a paragraph here explaining what you changed or what the new approach is?

index/scorch/snapshot_index.go
```diff
@@ -491,6 +491,22 @@ func (i *IndexSnapshot) DocumentVisitFieldTerms(id index.IndexInternalID,
 	return err
 }
 
+func checkSubset(set, subset []string) bool {
+OUTER:
+	for _, f1 := range subset {
+		found := false
```
steveyen: the found variable is unneeded

index/scorch/snapshot_segment.go
```diff
@@ -215,9 +215,32 @@ func (cfd *cachedFieldDocs) prepareFields(field string, ss *SegmentSnapshot) {
...
+	c.m.Lock()
+	for k, _ := range c.cache {
+		rv = append(rv, k)
+	}
+	c.m.Unlock()
+	return rv
+}
+
+func (c *cachedDocs) getPersistedFields() []string {
+	c.m.Lock()
```
steveyen: the lock area on these methods is perhaps too small, where a caller can get the persisted or cached fields, does some work, but in the meantime another thread can sneak in an change the persisted / cached fields?

index/scorch/snapshot_index.go
```diff
@@ -507,11 +523,29 @@ func (i *IndexSnapshot) documentVisitFieldTerms(id index.IndexInternalID,
...
+		// if all fields are dv presisted
+		if len(pFields) > 0 && checkSubset(pFields, fields) {
+			return zaps.VisitDocumentFieldTerms(localDocNum, fields, visitor, dvs)
+		}
+
+		cFields := ss.cachedDocs.getCachedFields()
+		// if all fields are already in dvCache or persisted
+		if len(cFields) > 0 && checkSubset(append(pFields, cFields...), fields) {
+			dvPendingFields := extractDvPendingFields(fields, pFields)
+			visitDocumentFieldCacheTerms(localDocNum, dvPendingFields, ss, visitor)
```
steveyen: looks like visitDocumentFieldCacheTerms() is not multithread safe, as it accesses the shared cache without acquiring the lock.  It should likely instead be a method of the cachedDocs struct, as it hardly uses the ss param.

index/scorch/snapshot_segment.go
mschoch: so, i agree in principle, however here the thinking is this:

the persisted fields is actually immutable, it's just that someone has to do the work to load it and figure out which fields it is.  yes, if the first time in, multiple threads race to do it, and they both do, they always come to the same conclusion.

index/scorch/snapshot_index.go
```diff
@@ -507,11 +523,29 @@ func (i *IndexSnapshot) documentVisitFieldTerms(id index.IndexInternalID,
...
-			return nil, err
+		pFields := ss.cachedDocs.getPersistedFields()
+		// if all fields are dv presisted
+		if len(pFields) > 0 && checkSubset(pFields, fields) {
+			return zaps.VisitDocumentFieldTerms(localDocNum, fields, visitor, dvs)
+		}
+
+		cFields := ss.cachedDocs.getCachedFields()
+		// if all fields are already in dvCache or persisted
+		if len(cFields) > 0 && checkSubset(append(pFields, cFields...), fields) {
```
steveyen: this approaching to union'ing the pFields and cFields is unsafe, as the underlying array of the pFields slice is shared across multiple threads, even though it was probably intended to be immutable.  The append(), though, may mutate the input slice's underlying backing array.

index/scorch/snapshot_index.go
steveyen: The repetition of the visitDocumentFieldCacheTerms() and VisitDocumentFieldTerms() and multiple return exit conditions doesn't have a cleanliness to it.  Instead, I find it easier to think about if there were no early exits and "all the waters have one exit in the Nile delta".

index/scorch/snapshot_index.go
mschoch: i belive this comment was already made by sreekanth an in our morning review, not sure what happened to that

index/scorch/snapshot_index.go
mschoch: so again, this was already discussed, sreekanth did not make any attempt at that clean up in this change because he spent all day at the passport office.  we should sync up to discuss further.

index/scorch/snapshot_index.go
```diff
@@ -507,11 +523,29 @@ func (i *IndexSnapshot) documentVisitFieldTerms(id index.IndexInternalID,
 	ss := i.segment[segmentIndex]
 
 	if zaps, ok := ss.segment.(segment.DocumentFieldTermVisitable); ok {
-		// get the list of doc value persisted fields
-		pFields, err := zaps.VisitableDocValueFields()
-		if err != nil {
-			return nil, err
+		pFields := ss.cachedDocs.getPersistedFields()
+		// if all fields are dv presisted
+		if len(pFields) > 0 && checkSubset(pFields, fields) {
```
steveyen: The multiple special case checking here might be repetitious...

It has a bit of feel of a math library, where it logically handles special cases... like, "first, let me handle all the even integers", "next, we're going to handle any odd integers", "finally, let me handle 0 as the final case" -- such an approach is logically correct, but doesn't have a minimal, concise feel?

sreekanth-cb: I agree to your point, and that will definitely be elegant.
 But let me think of another perspective here.. these simpler procedure kind of logic has less cognitive load at a later point while debugging. You need to load less into your mind..
Else, lets say we need to jump into all those visitCache, zaps Visits function codes to figure out whether things work correctly if we pass the whole set of fields..where is it getting filled..or from where fields are getting served etc... need to load all that flow while debugging at a later point.. :)

index/scorch/snapshot_index.go
sreekanth-cb: one thought is,  "if there were no early exits " and if we end up moving the visitDocumentFieldCacheTerms() under cachedDocs within a mutex, then we also need ensure that we don't call it when not needed(when not served), otherwise it will add a lock in a hot path...

PR #933, opened 2018-05-11
Fix bug in bleve's zap stored command, formatting

abhinavdangeti:
+ Handle the _id field special case
+ Format the output for better readability

### DIFF
--- a/cmd/bleve/cmd/zap/dict.go
+++ b/cmd/bleve/cmd/zap/dict.go
@@ -45,8 +45,8 @@ var dictCmd = &cobra.Command{
 		vellumLen, read := binary.Uvarint(data[addr : addr+binary.MaxVarintLen64])
 		fmt.Printf("vellum length: %d\n", vellumLen)
 		fstBytes := data[addr+uint64(read) : addr+uint64(read)+vellumLen]
-		fmt.Printf("raw vellum data % x\n", fstBytes)
-		fmt.Printf("dictionary:\n\n")
+		fmt.Printf("raw vellum data:\n % x\n", fstBytes)
+		fmt.Printf("dictionary:\n")
 		if fstBytes != nil {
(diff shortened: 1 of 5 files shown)


cmd/bleve/cmd/zap/dict.go
```diff
@@ -45,8 +45,8 @@ var dictCmd = &cobra.Command{
 		vellumLen, read := binary.Uvarint(data[addr : addr+binary.MaxVarintLen64])
 		fmt.Printf("vellum length: %d\n", vellumLen)
 		fstBytes := data[addr+uint64(read) : addr+uint64(read)+vellumLen]
-		fmt.Printf("raw vellum data % x\n", fstBytes)
-		fmt.Printf("dictionary:\n\n")
+		fmt.Printf("raw vellum data:\n % x\n", fstBytes)
```
steveyen: this was there before, but feels like an extra space in the "% x" is wrong?

cmd/bleve/cmd/zap/explore.go
```diff
@@ -46,7 +46,7 @@ var exploreCmd = &cobra.Command{
 		vellumLen, read := binary.Uvarint(data[addr : addr+binary.MaxVarintLen64])
 		fmt.Printf("vellum length: %d\n", vellumLen)
 		fstBytes := data[addr+uint64(read) : addr+uint64(read)+vellumLen]
-		fmt.Printf("raw vellum data % x\n", fstBytes)
+		fmt.Printf("raw vellum data:\n % x\n", fstBytes)
```
steveyen: (looks like same here)

mschoch:
% x is by design to put spaces between each byte value, not a typo

steveyen:
> % x is by design to put spaces between each byte value, not a typo

cool - learned something new.

looks good to me

PR #935, opened 2018-05-16
MB-29654: Copy value, arrayPositions into the interim structure(s)

abhinavdangeti:


abhinavdangeti:
@mschoch Yep that's my understanding as well, to preserve the array positions beyond the scope of the callback method - we need to create a copy.

index/scorch/segment/zap/merge.go
```diff
@@ -649,7 +649,10 @@ func mergeStoredAndRemap(segments []*SegmentBase, drops []*roaring.Bitmap,
 				fieldID := int(fieldsMap[field]) - 1
 				vals[fieldID] = append(vals[fieldID], value)
 				typs[fieldID] = append(typs[fieldID], typ)
-				poss[fieldID] = append(poss[fieldID], pos)
+
+				poss[fieldID] = append(poss[fieldID], make([]uint64, len(pos)))
+				copy(poss[fieldID][len(poss[fieldID])-1], pos)
```
steveyen: this is a bit of a weird formulation of this that feels unnatural, but perhaps logically correct.

another thought that pops up is this might increase memory usage to O(numDocs), and O(numDocs) trips to allocator, so... hmmm.  (meaning, noodling about alternatives)

steveyen: more as a 'note to self' on this, it looks like the VisitDocument() might invoke the visitor callback with a reused pos array, so the callback should copy the values if it wants to hold onto those values...

https://github.com/blevesearch/bleve/blob/master/index/scorch/segment/zap/segment.go#L356

...so the fix here is logically correct, but perhaps more performance is available by copying instead to a giant preallocated buffer

steveyen: one edge case on this... if the pos in the callback has length 0, then we're always allocating something?

that is, make([]uint64, 0) seems to return something non-nil

abhinavdangeti:
Thanks @steveyen for the feedback, let me make some improvements to this change, I'll have another patch up soon.

index/scorch/segment/zap/merge.go
```diff
@@ -649,7 +649,10 @@ func mergeStoredAndRemap(segments []*SegmentBase, drops []*roaring.Bitmap,
 				fieldID := int(fieldsMap[field]) - 1
 				vals[fieldID] = append(vals[fieldID], value)
 				typs[fieldID] = append(typs[fieldID], typ)
-				poss[fieldID] = append(poss[fieldID], pos)
+
```
steveyen: Hi @abhinavdangeti - not quite the same issue, but I'm thinking the value param of the callback might have the same problem (where the callback needs to copy the value)?

mschoch: I mentioned the 'value' in my comment earlier.  It does not have the same problem, because it is shared from the uncompressed slice.  But it could have a different problem, if say you closed the index and were holding on to a search result from earlier...

steveyen: Hmm, one worry is perhaps there might be a problem due to sync.Pool reuse of the visitDocumentCtx structs, where the uncompressed struct might be inadvertently reused by a concurrent goroutine.  Hmm.

abhinavdangeti:
@steveyen This latest patch makes sure that the value remains in scope as well. It doesn't however carry the memory optimization we discussed for the array positions during merge, for which I will create a separate PR.

abhinavdangeti:
@mschoch @steveyen The 2 commits on this PR are ready for review.

steveyen:
probably should get @mschoch's eyes on this, too

index/scorch/segment/zap/merge.go
```diff
@@ -645,11 +648,14 @@ func mergeStoredAndRemap(segments []*SegmentBase, drops []*roaring.Bitmap,
...
 				poss[i] = poss[i][:0]
 			}
-			err := segment.VisitDocument(docNum, func(field string, typ byte, value []byte, pos []uint64) bool {
+			err := segment.visitDocument(vdc, docNum, func(field string, typ byte, value []byte, pos []uint64) bool {
 				fieldID := int(fieldsMap[field]) - 1
 				vals[fieldID] = append(vals[fieldID], value)
 				typs[fieldID] = append(typs[fieldID], typ)
-				poss[fieldID] = append(poss[fieldID], pos)
+
+				// MB-29654: copy array positions to preserve them beyond the scope of this callback
```
mschoch: In bleve we don't reference MB's in the source code.  Bleve users have no idea what they are.

index/scorch/segment/zap/segment.go
```diff
@@ -295,10 +295,17 @@ var visitDocumentCtxPool = sync.Pool{
...
 		vdc := visitDocumentCtxPool.Get().(*visitDocumentCtx)
+		defer visitDocumentCtxPool.Put(vdc)
+		return s.visitDocument(vdc, num, visitor)
+	}
+	return nil
+}
 
+func (s *SegmentBase) visitDocument(vdc *visitDocumentCtx, num uint64, visitor segment.DocumentFieldValueVisitor) error {
+	// first make sure this is a valid number in this segment
+	if num < s.numDocs {
```
mschoch: Do we need to check the docNum again?  Seems like the contract could be that it's always check one place or the other.

index/scorch/snapshot_index.go
```diff
@@ -313,21 +313,26 @@ func (i *IndexSnapshot) Document(id string) (rv *document.Document, err error) {
 	segmentIndex, localDocNum := i.segmentIndexAndLocalDocNumFromGlobal(docNum)
 
 	rv = document.NewDocument(id)
-	err = i.segment[segmentIndex].VisitDocument(localDocNum, func(name string, typ byte, value []byte, pos []uint64) bool {
+	err = i.segment[segmentIndex].VisitDocument(localDocNum, func(name string, typ byte, val []byte, pos []uint64) bool {
 		if name == "_id" {
 			return true
 		}
+
+		// MB-29654: copy value, array positions to preserve them beyond the scope of this callback
```
mschoch: same comment on referencing MBs

index/scorch/segment/zap/merge.go
abhinavdangeti: Done.

index/scorch/segment/zap/segment.go
abhinavdangeti: Yes, as this internal API can be invoked directly or via the public one above.

index/scorch/snapshot_index.go
abhinavdangeti: Done.

index/scorch/segment/zap/merge.go
```diff
@@ -654,7 +658,17 @@ func mergeStoredAndRemap(segments []*SegmentBase, drops []*roaring.Bitmap,
...
+				if len(pos) > 0 {
+					if cap(posTemp) < len(pos) {
+						posBuf = make([]uint64, len(pos)*len(fieldsInv))
+						posTemp = posBuf
+					}
+					curPos = posTemp[0:len(pos)]
+					copy(curPos, pos)
+					posTemp = posTemp[len(pos):]
+				}
+				poss[fieldID] = append(poss[fieldID], curPos)
```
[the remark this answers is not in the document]
abhinavdangeti: Yes, I've confirmed that nil array positions are valid.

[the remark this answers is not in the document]
abhinavdangeti: posTemp's cursor is right shifted over the course of iterating over the fields within the document. The idea here is to be able to reuse posBuf over subsequent iterations over multiple documents across multiple segments subject to posBuf not being reinitialized again.

index/scorch/segment/zap/segment.go
mschoch: chatted briefly with steve, since external method always calls internal one, we could ditch the extra check in the outer one.  yes, it might mean an extra call to get/return from the sync.pool, but an invalid doc num is an exceptional case anyway, who care if it is slightly slower.  keeping the regular case clearer and with one less branch seems more important.

abhinavdangeti: Done.

abhinavdangeti:
@mschoch My mistake, the latest patch corrects that.

mschoch:
errcheck is supposed to ignore the fmt package by default, because it's incredibly verbose to use it on the ?printf family of functions –  where did you see the error?

abhinavdangeti:
In the previous check: https://travis-ci.org/blevesearch/bleve/builds/380345807

mschoch:
ok, well something changed in the tool then, i'd rather remove errcheck then use this ridiculous form

abhinavdangeti:
Haha, alright then, let me update travis.yml as a separate change :)

mschoch:
looks like once upon a time they ignored the entire fmt package, now they've whitelisted things, but not Fprintf specifically

abhinavdangeti:
travis-ci update: https://github.com/blevesearch/bleve/pull/938

abhinavdangeti:
Just a rebase.

PR #938, opened 2018-05-17
errCheck to ignore fmt pkg during checks that travis-ci performs

abhinavdangeti:


mschoch:
Instead of removing it, lets change to `errcheck -ignorepkg fmt`

abhinavdangeti:
@mschoch done.

PR #939, opened 2018-05-18
MB-29576 - Indexing blocked on scorch

sreekanth-cb:
More stats to keep track of last merged, persisted and current root epoch.
Haven't added any extra in-memory merging events, thinking that the in-line persistence call would trigger another progress event during it's introduction.


mschoch:
@steveyen's concenr about when we update CurRootEpoch was the same point I tried to raised with him Friday in hipchat.  I would feel better if we updated this setting right after we do `s.root = ...`

sreekanth-cb:
indeed @steveyen , @mschoch,  that looks more consistent. Added stats update for rollback too.

PR #942, opened 2018-05-25
Scorch optimize via vellum.FST.Reader() API

abhinavdangeti:
+ This change sets up a vellum.Reader (which carries a prealloc'ed
      fstState) for the term dictionaries.
    + However the Reader cannot be used concurrently, so rather than
      re-using a fieldDicts initialized for the IndexSnapshot, re-use
      one that is set up for TermFieldReaders which are recycled.

### DIFF
--- a/index/scorch/segment/zap/dict.go
+++ b/index/scorch/segment/zap/dict.go
@@ -28,10 +28,11 @@ import (
 
 // Dictionary is the zap representation of the term dictionary
 type Dictionary struct {
-	sb      *SegmentBase
-	field   string
-	fieldID uint16
-	fst     *vellum.FST
+	sb        *SegmentBase
+	field     string
+	fieldID   uint16
+	fst       *vellum.FST
+	fstReader *vellum.Reader
 }
 
 // PostingsList returns the postings list for the specified term
@@ -46,14 +47,14 @@ func (d *Dictionary) PostingsList(term []byte, except *roaring.Bitmap,
 }
 
 func (d *Dictionary) postingsList(term []byte, except *roaring.Bitmap, rv *PostingsList) (*PostingsList, error) {
-	if d.fst == nil {
+	if d.fstReader == nil {
 		if rv == nil || rv == emptyPostingsList {
 			return emptyPostingsList, nil
 		}
 		return d.postingsListInit(rv, except), nil
 	}
 
-	postingsOffset, exists, err := d.fst.Get(term)
+	postingsOffset, exists, err := d.fstReader.Get(term)
 	if err != nil {
 		return nil, fmt.Errorf("vellum err: %v", err)
 	}
--- a/index/scorch/snapshot_index.go
+++ b/index/scorch/snapshot_index.go
@@ -59,9 +59,8 @@ type IndexSnapshot struct {
 	m    sync.Mutex // Protects the fields that follow.
 	refs int64
 
-	m2         sync.Mutex                                 // Protects the fields that follow.
-	fieldTFRs  map[string][]*IndexSnapshotTermFieldReader // keyed by field, recycled TFR's
-	fieldDicts map[string][]segment.TermDictionary        // keyed by field, recycled dicts
+	m2        sync.Mutex                                 // Protects the fields that follow.
+	fieldTFRs map[string][]*IndexSnapshotTermFieldReader // keyed by field, recycled TFR's
 }
 
 func (i *IndexSnapshot) Segments() []*SegmentSnapshot {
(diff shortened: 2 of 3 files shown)


abhinavdangeti:
Hey @steveyen, for scorch to be able to use the vellum Reader, we'd have maintain an array of field dictionaries for re-use and re-cycle, so multiple threads don't end up re-using the field dictionaries and there by the vellum Reader. Can you take a look to see if it's worth the overhead, or if we can do better?

index/scorch/snapshot_index.go
```diff
@@ -439,21 +439,26 @@ func (i *IndexSnapshot) TermFieldReader(term []byte, field string, includeFreq,
 func (i *IndexSnapshot) allocTermFieldReaderDicts(field string) (
 	tfr *IndexSnapshotTermFieldReader, dicts []segment.TermDictionary) {
 	i.m2.Lock()
+	defer i.m2.Unlock()
```
steveyen: btw, one thing is I have the impression (might be old/outdated) that defer will allocate memory resources underneath the hood for a closure, so if the code's simple enough, or if it's on the "hot million-ops/sec path", i tend to avoid defer and do it the old-school way of explicit unlock's.  Not sure if allocTermFieldReaderDicts() happens enough for this to matter, so just FYI

index/scorch/segment/zap/dict.go
```diff
@@ -46,14 +47,14 @@ func (d *Dictionary) PostingsList(term []byte, except *roaring.Bitmap,
...
-	if d.fst == nil {
+	if d.fstReader == nil {
 		if rv == nil || rv == emptyPostingsList {
 			return emptyPostingsList, nil
 		}
 		return d.postingsListInit(rv, except), nil
 	}
 
-	postingsOffset, exists, err := d.fst.Get(term)
+	postingsOffset, exists, err := d.fstReader.Get(term)
```
steveyen: Trying to load this back into my head...

And, looks like there are two separate, independent changes here...

1) add a reused fstReader to the Dictionary

2) fix to the IndexSnapshot.fieldDicts reuse

The first change of the Dictionary.fstReader looks pretty straightforward and looks to good to me.  The 2nd change, though, is taking a bit more mental effort on my part and still thinking about it.  So, one thought is perhaps splitting this PR into two independent PR's, if you want that 1st Dictionary.fstReader change to get in sooner.

abhinavdangeti: @steveyen The 1st change that you speak about is what caused the panic seen in Alex's tests, and we had to revert it with this commit: https://github.com/blevesearch/bleve/commit/477688c147c679e43f471d36286c1e462295554b

To avoid that panic, these set of changes would all need to remain together.

index/scorch/snapshot_index.go
abhinavdangeti: Noted.

index/scorch/snapshot_index.go
```diff
@@ -464,9 +469,9 @@ func (i *IndexSnapshot) recycleTermFieldReader(tfr *IndexSnapshotTermFieldReader
 	}
 	i.fieldTFRs[tfr.field] = append(i.fieldTFRs[tfr.field], tfr)
 	if i.fieldDicts == nil {
-		i.fieldDicts = map[string][]segment.TermDictionary{}
+		i.fieldDicts = map[string][][]segment.TermDictionary{}
 	}
-	i.fieldDicts[tfr.field] = tfr.dicts
+	i.fieldDicts[tfr.field] = append(i.fieldDicts[tfr.field], tfr.dicts)
```
steveyen: Looks like the tfr has a dicts field, so one "localized, peephole" thought is I'm wondering if tfr.dicts needs to be cleared?  That way, if a future user picks up the tfr from the i.fieldTFRs, the dicts ought to be empty?

A second, larger thought, is perhaps we only need reuse of i.fieldTFRs and don't need the extra fieldDicts?  That is, if a tfr has a dicts field, the perhaps reuse of tfr's is enough?  Not sure if that's right, but perhaps, @abhinavdangeti, can you look into this and see if that simplification might work?

abhinavdangeti: Hmm, it seems like that simplification might just work, let me look into the code a bit.

abhinavdangeti:
@steveyen the latest patchset incorporates the feedback.

abhinavdangeti:
Yes it should.

steveyen:
> Yes it should.

hehe... wondering if we can clarify that up to "the fix seems to work on my machine" territory? :-)

abhinavdangeti:
I'll work with Alex to get this tested then :)

abhinavdangeti:
Verified with alex's tests that caught the panic.

PR #955, opened 2018-07-09
Fix to boolean searcher's Advance()

abhinavdangeti:
+ Advance nested searchers within a boolean searcher only if
  the cursor is trailing the ID being looked up.
+ Unit test
+ Fixes blevesearch#954

Here's an example bug scenario with nested boolean searchers ..

    conjunctionSearcher [a]
        => booleanSearcher [b]
            Must => conjunctionSearcher [d]
                => disjunctionSearcher [f]
                    => disjunctionSearcher
                        => termSearcher
                    => disjunctionSearcher
                        => termSearcher
        => booleanSearcher [c]
            Must => conjunctionSearcher [e]
                => termSearcher

    Consider there to be docs ranging from 1 to 100 in the index
    and these are the expected matches, for the query:
    1, 11, 16, 21,..

    On the first Next() call by the collector ..
        [d]
            - initSearchers .. sets local cursor of [f] to 1
            - moves local cursor of disjunction cursor to 11

        [b]
            - init: local cursor of [d] set to 1
            - moves [d]’s local cursor of [f] to 12
            - moves local cursor of [d] to 11

        [c]
            - init: local cursor of [e] set to 1
            - moves local cursor of [e] to 11

        [a]
            - initSearchers: sets local cursors of [b], [c] to 1
            - moves [b]’s local cursor of [d] to 12
            - moves [d]’s local cursor of [f] to 16
            - moves [c]’s local cursor of [e] to 16
            - updates local cursor of [b] to 11
            - updates local cursor of [c] to 11

    (Note that at this point [b]’s local cursor to [d] and [c]’s
     local cursor to [e] are pointing to different documents,
     owing to the different nesting of searchers underneath)

    On the second Next() call by the collector ..
        [a] .. maxID: 11
            - moves [b]’s local cursor of [d] to 16
                - moves [d]’s local cursor of [f] to 17
            - moves [c]’s local cursor of [e] to 21
            - updates local cursor of [b] to 12
            - updates local cursor of [c] to 16

    On the third Next() call by the collector ..
        [a] .. maxID: 16
            - sees that maxID 16 > 12 (cursor position of [b])
            - invokes advanceChild([b]) ..
                --> this causes [b]’s local cursor to move from
                    16 to 17 and this prevents the match:16 from
                    being collected.

### DIFF
--- a/search/searcher/search_boolean.go
+++ b/search/searcher/search_boolean.go
@@ -332,31 +332,38 @@ func (s *BooleanSearcher) Advance(ctx *search.SearchContext, ID index.IndexInter
 	}
 
 	var err error
+	// Advance nested searcher(s) only if the cursor is trailing the lookup ID
 	if s.mustSearcher != nil {
-		if s.currMust != nil {
-			ctx.DocumentMatchPool.Put(s.currMust)
-		}
-		s.currMust, err = s.mustSearcher.Advance(ctx, ID)
-		if err != nil {
-			return nil, err
+		if s.currMust == nil || s.currMust.IndexInternalID.Compare(ID) < 0 {
+			if s.currMust != nil {
+				ctx.DocumentMatchPool.Put(s.currMust)
+			}
+			s.currMust, err = s.mustSearcher.Advance(ctx, ID)
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
 	if s.shouldSearcher != nil {
-		if s.currShould != nil {
-			ctx.DocumentMatchPool.Put(s.currShould)
-		}
-		s.currShould, err = s.shouldSearcher.Advance(ctx, ID)
-		if err != nil {
-			return nil, err
+		if s.currShould == nil || s.currShould.IndexInternalID.Compare(ID) < 0 {
+			if s.currShould != nil {
+				ctx.DocumentMatchPool.Put(s.currShould)
+			}
+			s.currShould, err = s.shouldSearcher.Advance(ctx, ID)
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
 	if s.mustNotSearcher != nil {
-		if s.currMustNot != nil {
-			ctx.DocumentMatchPool.Put(s.currMustNot)
-		}
-		s.currMustNot, err = s.mustNotSearcher.Advance(ctx, ID)
-		if err != nil {
-			return nil, err
+		if s.currMustNot == nil || s.currMustNot.IndexInternalID.Compare(ID) < 0 {
+			if s.currMustNot != nil {
+				ctx.DocumentMatchPool.Put(s.currMustNot)
+			}
+			s.currMustNot, err = s.mustNotSearcher.Advance(ctx, ID)
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
 
--- a/search_test.go
+++ b/search_test.go
@@ -17,12 +17,21 @@ package bleve
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"reflect"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/blevesearch/bleve/analysis"
+	"github.com/blevesearch/bleve/analysis/analyzer/custom"
+	"github.com/blevesearch/bleve/analysis/token/lowercase"
+	"github.com/blevesearch/bleve/analysis/tokenizer/single"
+	"github.com/blevesearch/bleve/analysis/tokenizer/whitespace"
+	"github.com/blevesearch/bleve/document"
 	"github.com/blevesearch/bleve/search"
+	"github.com/blevesearch/bleve/search/query"
 )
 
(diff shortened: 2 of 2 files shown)


abhinavdangeti:
Ah my mistake, thanks for pointing that out @mschoch . 
So that seems to be the case here .. where the searcher is already pointing to something past the key we want to advance to. Invoking advanceChild essentially resets the cursor of the searcher, and advances it to the position of the key we want.
Does it make sense to get rid of that bytes.Compare(..) of the id entirely, or is it perhaps that we should've never entered this situation in the first place?

mschoch:
@abhinavdangeti you haven't provided enough information about the problem for me to answer your questions.  If a searcher is already pointing at key D, and you ask it to advance to B, the intended behavior is to do nothing (or continue in this loop).  The reasoning for this is deeper, in that the way we walk all of these searchers simultaneously, should mean that this is safe (ie, there is no key B or C on the that searcher).

If what you're saying is that no we really do need to go backwards, because B or C is on that searcher, then there is a completely different problem, and you'd have to start by seeing how we skipped over B or C already in the first place.  Is it because any of the searchers are returning things out of order?

abhinavdangeti:
I believe that the problem lies with the searcher skipping over certain relevant keys.
I've added some debug logs and here's evidence that the searcher has indeed skipped relevant information (I've removed that conditional and always have the searcher's cursor advance regardless of it's position for this): 

```
[0xc4202a4a50] BOOL Advance to 16, mustSearcher: 0xc4202832c0
[0xc4202832c0] CONJ: Look up 16 
[0xc4202832c0] CONJ: Current cursor at [17-0.739195]
[0xc4202832c0] CONJ advanceChild, sub searcher: 0xc4203b2280
...
[0xc4202832c0] CONJ advanceChild aftermath, Cursor now at: [16-0.739195], sub searcher's p: 0xc4203b2280
```

**16** happens to be one of the keys that is expected in the result set but is absent.

From my logs I don't see any evidence that the searcher's returning things out of order however.

abhinavdangeti:
Hey @mschoch, think I've found the issue, the new patch addresses it. Let me know what you think.

abhinavdangeti:
Let me also add the unit test that I used to verify it along with this change.

mschoch:
Adding a unit test will be helpful, but I'd like to dig deeper into this solution.

Can you, in a few sentences explain why this works?  What scenario arises, and do these additional conditions prevent the wrong thing from happening?  This would be a useful addition to the commit message.

The other question is, why does this fix belong here?  Specifically, shouldn't it be valid for anyone to ask any searcher to Advance() anywhere?  Why does this searcher need to be "careful" about when to invoke Advance()?  I haven't heard your reasoning yet, but it seems to me that its better to keep this code clearer, and always call Advance() when its logically needed, and that it should be the down-stream searchers which either protect themselves from doing the wrong thing, or just doing nothing if it should in fact be a no-op.  Or maybe there is some other reason it is better done here?

mschoch:
Thanks, that helps some, but I still might be missing the big picture.  Would it be correct to say that the problem originates because a parent searcher tracks the "current" location of each of it's child searchers, but in fact the child searchers have already moved past that point?  So then, at some later time when it attempts to make decisions about "do I need to reposition this child?" it ends up making wrong decisions because the child searcher is actually further ahead?

That is, what I sort of get the feeling is happening here.

Why is that important?  Well it would seem that if this is the case, then the fix works because it does need to tell child searchers to go backwards sometimes.  We're advancing to "16", child searcher is already at "17", we need to make it go back to not miss something.

If that is the case, meaning that this fix works because it tells child searchers to go backwards, then we have to re-evaluate a larger issue.  We have a number of internals that are highly inefficient if searchers ever go backwards (they basically start over from 0 again).  And, so if this fix is introducing anything that causes us to go backwards (even if that is correct), we need to think about it more.

abhinavdangeti:
Hey @mschoch, all points you make sound valid. But I haven't found evidence of any searchers moving backwards yet.

For the moment at least, I believe the problem lies with invoking a boolean searcher's Advance() to an ID that it already points to, which in turn invokes the child conjunction searcher's Advance(). [Here](https://github.com/blevesearch/bleve/blob/master/search/searcher/search_conjunction.go#L237) the child conjunction searcher determines that it doesn't need to advance, and returns Next() - the cached value. The boolean searcher updates it's currMust with the cached value returned by the child searcher. This is where the boolean searcher moves past a match (there by skipping a result), as the child searcher was already ahead of the parent searcher. The change in this PR prohibits this update within the boolean searcher from happening.

We already do this for the [conjunction searcher](https://github.com/blevesearch/bleve/blob/master/search/searcher/search_conjunction.go#L237), [disjunction searcher](https://github.com/blevesearch/bleve/blob/master/search/searcher/search_disjunction_slice.go#L235) and [phrase searcher](https://github.com/blevesearch/bleve/blob/master/search/searcher/search_phrase.go#L400). I made this change as I felt the handling of this scenario was missing within the boolean searcher.

PR #956, opened 2018-07-11
MB-29923 - high memory consumption in scorch

sreekanth-cb:
The fix aims to loosen the persister-merger lock
stepping and also introduces a configurable
wait in the persister work loop to favour healthier
in-memory segment merges.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -17,6 +17,7 @@ package scorch
 import (
 	"bytes"
 	"encoding/binary"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -35,16 +36,22 @@ import (
 
 var DefaultChunkFactor uint32 = 1024
 
-// Arbitrary number, need to make it configurable.
-// Lower values like 10/making persister really slow
-// doesn't work well as it is creating more files to
-// persist for in next persist iteration and spikes the # FDs.
-// Ideal value should let persister also proceed at
-// an optimum pace so that the merger can skip
-// many intermediate snapshots.
-// This needs to be based on empirical data.
-// TODO - may need to revisit this approach/value.
-var epochDistance = uint64(5)
+var DefaultPersisterNapTimeMSec int = 2000 // ms
+
+var DefaultPersisterNapUnderNumFiles int = 1000
+
+type persisterOptions struct {
+	// PersisterNapTimeMSec controls the wait/delay injected into
+	// persistence workloop to improve the chances for
+	// a healthier and heavier in-memory merging
+	PersisterNapTimeMSec int
+
+	// PersisterNapTimeMSec > 0, and the number of files is less than
+	// PersisterNapUnderNumFiles, then the persister will sleep
+	// PersisterNapTimeMSec amount of time to improve the chances for
+	// a healthier and heavier in-memory merging
+	PersisterNapUnderNumFiles int
+}
 
 type notificationChan chan struct{}
 
@@ -54,6 +61,13 @@ func (s *Scorch) persisterLoop() {
 	var persistWatchers []*epochWatcher
 	var lastPersistedEpoch, lastMergedEpoch uint64
 	var ew *epochWatcher
+	po, err := s.parsePersisterOptions()
+	if err != nil {
+		s.fireAsyncError(fmt.Errorf("persisterOptions json parsing err: %v", err))
+		s.asyncTasks.Done()
+		return
+	}
+
 OUTER:
 	for {
 		atomic.AddUint64(&s.stats.TotPersistLoopBeg, 1)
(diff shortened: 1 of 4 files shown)


index/scorch/persister.go
```diff
@@ -35,16 +37,27 @@ import (
...
-// persist for in next persist iteration and spikes the # FDs.
-// Ideal value should let persister also proceed at
-// an optimum pace so that the merger can skip
-// many intermediate snapshots.
-// This needs to be based on empirical data.
-// TODO - may need to revisit this approach/value.
-var epochDistance = uint64(5)
+var DefaultWorkLoopWait int = 2000 // ms
+
+var MaxWorkLoopWait int = 5000 // ms
```
steveyen: wondering why have a MaxWorkLoopWait?

index/scorch/persister.go
```diff
@@ -35,16 +37,27 @@ import (
...
+type persisterOptions struct {
+	// WorkLoopWait controls the wait/delay injected into
+	// persistence workloop to improve the chances for
+	// a healthier and heavier in-memory merging
+	WorkLoopWait int
+
+	// PauseOnFileCount threshold mandates the persister to wait
+	// until the merger compacts the files on disk within this
+	// threshold
+	PauseOnFileCount int
```
steveyen: One thought was PauseOnFileCount might need a rename.

And, related, an edge case popped to mind... imagine there's a whole bunch of N files on disk (because there's a lot of data, perhaps due to the app wanting to ingest a lot of data)... but those files are already in "perfect staircase" formation.

In that case, the merger wakes up and thinks "cool, there's nothing to merge; I'm done".

But, if N is > this threshold option, then the persister would be incorrectly blocked.

So, I'm thinking the option that you'd really want to specify would be more like "pause the persister if the # of _mergable_ files goes above this threshold"?

index/scorch/persister.go
```diff
@@ -180,14 +200,25 @@ func notifyMergeWatchers(lastPersistedEpoch uint64,
...
 	persistWatchers = notifyMergeWatchers(lastPersistedEpoch, persistWatchers)
 
+	// check the merger lag by counting the segment files on disk,
+	// On finding fewer files on disk, persister takes a short pause
+	// for sufficient in-memory segments to pile up for the next
+	// memory merge cum persist loop.
+	// On finding too many files on disk, persister pause until the merger
+	// catches up to reduce the segment file count under the threshold.
+	numFilesOnDisk, _ := s.diskFileStats()
+	if numFilesOnDisk < uint64(po.PauseOnFileCount) {
```
steveyen: Ah, I see, I had misinterpreted its meaning.  Now that I'm seeing the comparison here, it seems to mean "temporarily pause persistence as long as the number of files is less than X".

Might want to have a check here, too, to only do this sleep+return maneuver only when po.WorkLoopWait > 0

index/scorch/persister.go
sreekanth-cb: hm, thats a interesting angle.
But as a measure to limit the FD count, irrespective of the potential for further merge operations - we may still want to block persister as the number of FD's/files are over threshold at this point and by blocking persister the file count eventually comes under threshold either due to completion of concurrent queries or due to merger progress.?

steveyen: right -- i think it's perhaps because i misunderstood what PauseOnFileCount meant.

I initially took it to mean (when I wrote the above comment)... "pause the persister if the file count is greater than X"... but it actually means... "slow down or sleep the persister for a finite amount of time as long as the file count is < X"

index/scorch/persister.go
sreekanth-cb: as it is configurable, thought of making a max check as well. 
But this may also end up being an advanced, rarely used configuration setting too.,

sreekanth-cb:
@steveyen , addressed the suggestions you mentioned earlier. please have a look. thanks!

index/scorch/persister.go
sreekanth-cb: hi @steveyen , 
MaxNumberOfFiles / MaxFilesOnDisk or any other thoughts on the naming?

index/scorch/persister.go
```diff
@@ -35,16 +36,20 @@ import (
...
-var epochDistance = uint64(5)
+var DefaultWorkLoopWait int = 2000 // ms
+
+var DefaultPersisterNapUnderNumFiles int = 1000
+
+type persisterOptions struct {
+	// WorkLoopWait controls the wait/delay injected into
+	// persistence workloop to improve the chances for
+	// a healthier and heavier in-memory merging
+	WorkLoopWait int
```
steveyen: oh, i guess this needs a rename too, now... like PersisterNapMSecs or something (or whatever naming pattern we have already for duration based config options)

index/scorch/persister.go
```diff
@@ -35,16 +36,20 @@ import (
...
+
+	// PersisterNapUnderNumFiles threshold mandates the persister to wait
+	// until the merger compacts the files on disk within this threshold
```
steveyen: this doc comment looks wrong, as is more something like, when PersisterNapTimeMSec > 0, and the number of files is less than PersisterNapUnderNumFiles, then the persister will sleep PersisterNapTimeMSec amount of time, which can help blah blah blah

PR #960, opened 2018-07-16
MB-30342 - ref_count leak from FieldDict

sreekanth-cb:
Attempt to fix the ref_count leak by closing the
field dictionary for term_prefix and term_range
searches

### DIFF
--- a/search/searcher/search_term_prefix.go
+++ b/search/searcher/search_term_prefix.go
@@ -27,12 +27,20 @@ func NewTermPrefixSearcher(indexReader index.IndexReader, prefix string,
 	if err != nil {
 		return nil, err
 	}
+	defer func() {
+		if cerr := fieldDict.Close(); cerr != nil && err == nil {
+			err = cerr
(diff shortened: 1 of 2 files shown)


search/searcher/search_term_prefix.go
```diff
@@ -35,5 +35,10 @@ func NewTermPrefixSearcher(indexReader index.IndexReader, prefix string,
 		tfd, err = fieldDict.Next()
 	}
```
mschoch: there is another error unchecked here, using this pattern, we must check for non nil err upon exiting the loop, as we may have broken out of the for loop due to an error.

sreekanth-cb: Done!

steveyen:
> So, calling Close() per API contract is better, but oddly (as far as i can figure out right now) ought not to do anything.

Oh, nevermind!  This is a leak with upsidedown -- my confusion solved.

PR #965, opened 2018-07-28
MB-30694: reflect.Value.Type on zero Value panics

sreekanth-cb:
Adding checks to validate the geo point values
while parsing


sreekanth-cb:
thanks @mschoch , Yes, UT update is pending and hence didn't explicitly start review request by adding reviewers. :)

geo/parse.go
```diff
@@ -36,10 +36,14 @@ func ExtractGeoPoint(thing interface{}) (lon, lat float64, success bool) {
 	var foundLon, foundLat bool
 
 	thingVal := reflect.ValueOf(thing)
+	if !thingVal.IsValid() {
+		return lon, lat, false
```
abhinavdangeti: Just return would do here :)

PR #969, opened 2018-07-31
replacing maps with slice for faster lookups

sreekanth-cb:
This change shows a throughput improvement from 63 -> 88 (~25%)

http://perf.jenkins.couchbase.com/job/atlas-sdk/2799
Vs
http://perf.jenkins.couchbase.com/job/atlas-sdk/2813

### DIFF
--- a/search/facets_builder.go
+++ b/search/facets_builder.go
@@ -54,14 +54,14 @@ type FacetBuilder interface {
 
 type FacetsBuilder struct {
 	indexReader index.IndexReader
-	facets      map[string]FacetBuilder
+	facetNames  []string
+	facets      []FacetBuilder
 	fields      []string
 }
 
 func NewFacetsBuilder(indexReader index.IndexReader) *FacetsBuilder {
 	return &FacetsBuilder{
 		indexReader: indexReader,
-		facets:      make(map[string]FacetBuilder, 0),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


search/facets_builder.go
```diff
@@ -54,23 +54,22 @@ type FacetBuilder interface {
...
 	}
 }
 
 func (fb *FacetsBuilder) Size() int {
 	sizeInBytes := reflectStaticSizeFacetsBuilder + size.SizeOfPtr
 
 	for k, v := range fb.facets {
-		sizeInBytes += size.SizeOfString + len(k) +
-			v.Size()
+		sizeInBytes += size.SizeOfString + v.Size() + len(fb.facetNames[k])
```
abhinavdangeti: Actually you dont need this extra SizeOfString anymore.

sreekanth-cb: My thinking was that, size.SizeOfString needs to be counted for the new facetNames[] string slice and just reused the same existing facets size count loop for doing the same. 
size.SizeOfString + len(fb.facetNames[k] => for the new facetNames slice 

But think I missed something here,
Now for accounting the fb.facets correctly,  need to add a 
size.SizeOfUint64 along with this, right?

abhinavdangeti: Ah you're right, my mistake.

PR #976, opened 2018-08-13
MB-29923 - high scorch mem usage with items to persist

sreekanth-cb:
This change includes,

-persister nap to favour in-memory merging
-skip in-memory merging on memory pressure
-skip persister nap on memory pressure
-checks during persister-merger wait to guard against an already advanced merger

toy build passes the tests here: http://qa.sc.couchbase.com/job/test_suite_executor/85255/console

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -73,6 +73,10 @@ type Scorch struct {
 	onAsyncError func(err error)
 
 	iStats internalStats
+
+	pauseLock sync.RWMutex
+
+	pauseCount uint64
 }
 
 type internalStats struct {
@@ -117,9 +121,30 @@ func NewScorch(storeName string,
 	return rv, nil
 }
 
+func (s *Scorch) paused() uint64 {
+	s.pauseLock.Lock()
+	pc := s.pauseCount
+	s.pauseLock.Unlock()
+	return pc
+}
+
+func (s *Scorch) incrPause() {
+	s.pauseLock.Lock()
+	s.pauseCount++
+	s.pauseLock.Unlock()
+}
+
+func (s *Scorch) decrPause() {
+	s.pauseLock.Lock()
+	s.pauseCount--
+	s.pauseLock.Unlock()
+}
+
 func (s *Scorch) fireEvent(kind EventKind, dur time.Duration) {
 	if s.onEvent != nil {
+		s.incrPause()
 		s.onEvent(Event{Kind: kind, Scorch: s, Duration: dur})
+		s.decrPause()
 	}
 }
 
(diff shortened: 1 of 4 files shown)


sreekanth-cb:
@abhinavdangeti , Can you please cross whether the above jenkin job run the same job you tried earlier?

index/scorch/scorch.go
```diff
@@ -117,9 +121,30 @@ func NewScorch(storeName string,
...
+
+func (s *Scorch) incrWaterMark() {
+	s.memLock.Lock()
+	s.highWaterMark++
+	s.memLock.Unlock()
+}
+
+func (s *Scorch) decrWaterMark() {
+	s.memLock.Lock()
+	s.highWaterMark--
```
steveyen: quick thing, on naming, a "high water mark" i think usually means some level or number that doesn't go down

abhinavdangeti:
@sreekanth-cb ya it does seem to be the same test suite. I'm not sure I completely understand how the watermark() check before persistSnapshotMaybeMerge made the difference however; per your comment in persistSnapshot(..) - "perform in-memory merging only when there is no memory pressure" - note that fireEvent is not just invoked by the merger/persister upon completion but also invoked by ExecuteBatch to record memory consumption.

sreekanth-cb:
@abhinavdangeti, the idea is, in-memory merging isn't an advisable thing to perform when you are running out of memory. It would worsen the situation.  Similarly Persister nap during a memory hit situation is also a bad idea. We should avoid those two manoeuvres when we are hit by app_herder.
Yes, we mainly wanted to capture the Execute batch  memory consumption add up calls, as that acts as the knob for bumping the memory usage/water mark.

index/scorch/scorch.go
```diff
@@ -434,24 +459,31 @@ func (s *Scorch) currentSnapshot() *IndexSnapshot {
...
 		}
 	}
+	return numFilesOnDisk, numBytesUsedDisk
+}
+
+func (s *Scorch) StatsMap() map[string]interface{} {
+	m := s.stats.ToMap()
+
+	numFilesOnDisk, numBytesUsedDisk := s.diskFileStats()
+	if numFilesOnDisk > 0 || numBytesUsedDisk > 0 {
```
steveyen: just noticed the safety checks here?  Wondering why the > 0 checks are needed?

index/scorch/scorch.go
```diff
@@ -117,9 +121,30 @@ func NewScorch(storeName string,
...
+	s.memLock.Lock()
+	s.highWaterMark--
+	s.memLock.Unlock()
+}
+
 func (s *Scorch) fireEvent(kind EventKind, dur time.Duration) {
 	if s.onEvent != nil {
+		s.incrWaterMark()
 		s.onEvent(Event{Kind: kind, Scorch: s, Duration: dur})
+		s.decrWaterMark()
```
steveyen: naming wise, this doesn't look like a water mark?

perhaps other thoughts of that this looks like "inMiddleOfEventCallback"?

or, actually, the reason why this is here is the even handler (the app-herder) might end up pausing during the event callback (e.g., it's stepping on the breaks), so maybe the "water mark" counter is more like "pauseCount" or something like that?

index/scorch/scorch.go
sreekanth-cb: hm... don't remember why we need it..removing them..

PR #985, opened 2018-08-17
Use fmt.Sprintf to generate kv pairs in test

bt:
I'm in the middle of re-implementing the latest version of BadgerDB into Bleve and I came across an issue in the this specific test in which, by the time the batch is executed, the key/value buffers have already been garbage collected and therefore, is causing the test to fail.

The solution here is to make a new buffer with `make([]byte, 5)` on each for-loop, however, I figured it's cleaner to use `fmt.Sprintf` to convert a string into the corresponding byte array.


steveyen:
Hi @bt,
One worry about this is I think the interface contract for KVBatch.Set() method has the implied rule that the backend store implementation is supposed to make its own copy of the incoming key/val bytes.  For example, see: https://github.com/blevesearch/bleve/blob/master/index/store/kvstore.go#L147

PR #988, opened 2018-08-19
Dynamic document mapping should take precedence over index dynamic sp…

connorgorman:
…ecification

Explanation:
 - See discussion and issue here: https://groups.google.com/forum/#!topic/bleve/7z-Zod2ssyU
 - Currently, the check verifies that if ClosestDocumentMapping.Dynamic, then use stub for dynamic mapping field. However, we adhere to the index mappings dynamic specification. AFAIK, we should instead use the dynamic specification from the document mapping (which is always true in this case)


connorgorman:
@mschoch what do you think about this one?

connorgorman:
Hey folks, any thoughts on this PR? Seems like just a simple bug fix where the dynamic mapping will always be indexed. Please let me know if you have any other suggestions. Thanks!

PR #992, opened 2018-08-26
adding new batch merge api

sreekanth-cb:


index.go
```diff
@@ -119,6 +119,12 @@ func (b *Batch) Reset() {
 	b.internal.Reset()
 }
 
+func (b *Batch) Merge(o *Batch) {
+	b.internal.Merge(o.internal)
+	b.lastDocSize = o.LastDocSize()
+	b.totalSize += o.TotalDocsSize()
```
steveyen: From a quick first-gut reaction, not sure if it's right for total docs size to be simply summed this way.  For example, if the 2nd batch had updates or deletions of docs of the same key(s) as from the 1st batch, then the ultimate size might be different?

sreekanth-cb: you are right, thanks!

index.go
```diff
@@ -119,6 +119,12 @@ func (b *Batch) Reset() {
 	b.internal.Reset()
 }
 
+func (b *Batch) Merge(o *Batch) {
+	b.internal.Merge(o.internal)
+	b.lastDocSize = o.LastDocSize()
+	b.totalSize = uint64(b.internal.TotalDocSize())
```
steveyen: one edge case is if the incoming 'o' batch is empty?  In that case, taking o's lastDocSize and totalSize might be wrong?

sreekanth-cb: lastDocSize might be wrong, but totalSize is computed from updated b. So may not change.

steveyen: > but totalSize is computed from updated b.

ah, didn't catch that -- thx!

PR #999, opened 2018-09-06
Fix checks for empty PostingsList and PostingsIterator

abhinavdangeti:


steveyen:
Hi @abhinavdangeti -- wondering if there's some functional or correctness issue here?  That is, the emptyPostingsList, I think, was meant to be a singleton instance?

steveyen:
Oh, on reading your PR a little closer, I see emptyPostingsList is still a singleton, but there are changes w.r.t. comparisons.  But, still wondering if something functional was wrong or incorrect in the existing approach?

abhinavdangeti:
@steveyen On initial look, comparing pointers to check for empty structs seemed incorrect to me. But you're right - these are singleton instances, we don't need to worry about this then.

PR #1009, opened 2018-09-17
Add support for GeoSpatial points in string format

abhinavdangeti:
Following specification used in elastic search:
    https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-point.html

Also see:
    https://issues.couchbase.com/browse/MB-30542

### DIFF
--- a/geo/parse.go
+++ b/geo/parse.go
@@ -16,6 +16,7 @@ package geo
 
 import (
 	"reflect"
+	"strconv"
 	"strings"
 )
 
@@ -24,6 +25,8 @@ import (
 // Container:
 // slice length 2 (GeoJSON)
 //  first element lon, second element lat
+// string (coordinates separated by comma, or a geohash)
+//  first element lat, second element lon
 // map[string]interface{}
 //  exact keys lat and lon or lng
 // struct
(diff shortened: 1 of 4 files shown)


geo/parse.go
```diff
@@ -16,14 +16,19 @@ package geo
...
+
+	"github.com/mmcloughlin/geohash"
 )
 
 // ExtractGeoPoint takes an arbitrary interface{} and tries it's best to
 // interpret it is as geo point.  Supported formats:
 // Container:
 // slice length 2 (GeoJSON)
 //  first element lon, second element lat
+// string (coordinates seperated by comma, or a geohash)
```
sreekanth-cb: typo on seperated?

geo/parse.go
```diff
@@ -59,6 +64,35 @@ func ExtractGeoPoint(thing interface{}) (lon, lat float64, success bool) {
...
+					foundLat = true
+				}
+				lon, err = strconv.ParseFloat(points[1], 64)
+				if err == nil {
+					foundLon = true
+				}
+			}
+		} else {
+			// geo hash
+			lat, lon = geohash.Decode(geoStr)
```
sreekanth-cb: worth having a few UTs to test the negative cases for this Decode() fn?,
For eg: like pass couple of invalid hashes,  
Another thing to see is that how that lat/lon values returned on an invalid hash percolates to later sections of the logic.

geo/parse.go
```diff
@@ -16,14 +16,19 @@ package geo
...
+	"github.com/mmcloughlin/geohash"
 )
 
 // ExtractGeoPoint takes an arbitrary interface{} and tries it's best to
 // interpret it is as geo point.  Supported formats:
 // Container:
 // slice length 2 (GeoJSON)
 //  first element lon, second element lat
+// string (coordinates seperated by comma, or a geohash)
+//  first element lat, second element lon
```
sreekanth-cb: minor extra space? :)

abhinavdangeti: That extra space before first, is following the format from before.

Per elasticsearch spec here: https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-point.html

With slices, its lon first and lat after, but with strings its lat first and lon after - that's what the comment(s)^ above talk about.

geo/parse.go
abhinavdangeti: Well..  geohash is a library, that already has it's own suite of unit tests.

sreekanth-cb: geohash is a library -> its always a debatable thing about the robustness or maturity of the 3rd party libs.  
Main wrinkle points for me here are,
1. How confident are you that, the values returned from that lib for an invalid hash are valid Or that it won't break any of the remaining code in bleve? 

2. Having few unit tests to set/validate the response expectations from a library helps in confirming minimum sanity at least during a SHA bump for it at a later point in time for accommodating a genuine feature or bug fix on their side.

abhinavdangeti:
@steveyen @sreekanth-cb Ping for re-review. The latest patch borrows only the necessary functionality from https://github.com/mmcloughlin/geohash for decoding geohashes.

abhinavdangeti:
Ping for re-review @sreekanth-cb .

PR #1015, opened 2018-10-01
add IndexAdvanced to the index interface

tzneal:
It seemed odd that this was on the Batch and the indexImpl, but not part of the Index interface itself.

### DIFF
--- a/index.go
+++ b/index.go
@@ -199,6 +199,10 @@ type Index interface {
 	// requests. See Index interface documentation for details about mapping
 	// rules.
 	Index(id string, data interface{}) error
+	// IndexAdvanced takes a document.Document object skips the mapping and
+	// indexes it without further analysis.
+	IndexAdvanced(doc *document.Document) error
+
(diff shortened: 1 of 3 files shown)


tzneal:
Ping @sreekanth-cb

abhinavdangeti:
I am OK with this change, but curious if you actually need this API to be part of the interface?

tzneal:
@abhinavdangeti I got around it by using NewBatch(), which does provide IndexAdvanced.  My use cases was indexing new documents as user's submitted them, where I needed a lot of control over the indexing.  I thought it might be helpful to someone else.

abhinavdangeti:
@tzneal I see. So we here at couchbase have another project that deploys this interface. As this isn't super pressing at the moment, I shall merge this when I get around to making the necessary changes in the other project as well.

tzneal:
@abhinavdangeti No problem, thanks for considering it!

PR #1017, opened 2018-10-10
add support for query string searches for boolean fields

### LINKED ISSUE #626: Query String Query for BoolFieldQuery
Hello,
I am wondering what's the proper syntax for bool field query, like this one:
```
	q := query.NewBoolFieldQuery(true)
	q.SetField("Married")
```
I was trying "Married:true", 'Married:"t"' but these queries do not return the expected result.
So, what's the right way to do that?

tzneal:
The format is "field_name:true" or "field_name:false", where case must match.  In the absence of a field name (e.g. just "true" or "false") it reverts back to a standard string match.

Fixes #626

### DIFF
--- a/search/query/query_string.y
+++ b/search/query/query_string.y
@@ -19,10 +19,11 @@ s string
 n int
 f float64
 q Query
+b bool
 pf *float64}
 
 %token tSTRING tPHRASE tPLUS tMINUS tCOLON tBOOST tNUMBER tSTRING tGREATER tLESS
-tEQUAL tTILDE
+tEQUAL tTILDE tBOOLEAN
 
 %type <s>                tSTRING
 %type <s>                tPHRASE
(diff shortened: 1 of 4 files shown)


tzneal:
Ping @sreekanth-cb

activeshadow:
Any idea when this pull request might be reviewed/merged/released?

mschoch:
So my concern with this PR is that it goes really deep into the query string, add tokens to the lexing phase all just to support a simple boolean search.  And yet, we don't index these terms in any special way, we just use 'T' and 'F', so it shouldn't be so complicated.

In the linked bug I had proposed trying to query with a particular syntax, and it was reported that that syntax didn't work.  But, I'd like to first really understand why that syntax doesn't work.  I speculated it might have to do with the text being analyzed, but we should be choosing the analyzer based on the field, so that explanation is at best incomplete.

PR #1019, opened 2018-10-13
update to use only supported versions of golang

techknowlogick:
per: https://golang.org/doc/devel/release.html#policy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,8 @@ sudo: false
 language: go
 
 go:
- - 1.7.x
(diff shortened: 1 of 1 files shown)


abhinavdangeti:
Thanks for raising this @techknowlogick . We'll continue to support 1.9.x as well, I've pushed up a PR for it: https://github.com/blevesearch/bleve/pull/1068

techknowlogick:
Thanks @abhinavdangeti 😄

PR #1028, opened 2018-10-23
Update golang version used for CI

mavimo:
Added go 1.11 and head as version that we use on test phase


mavimo:
@mschoch build failure seems related to wrong permission to publish coverage report on coverall since I'm pushing PR from a fork some secret variable is not exposed and script fail.

techknowlogick:
Related #1019

Issue #1031, opened 2018-10-26
Lack of examples


oskarwojciski:
I am seeing lack of examples how to use bleve, existing or new analyzers, etc.
I am going to prepare PR with proposed ones

oskarwojciski:
I have committed my proposition - please take a look and say what do you think. If you accepted this I can prepare more examples or even some list of examples needed in the Readme file

gedw99:
Nice !

You should probably look into what database type though. It seems Scorch is the current one used and not boltdb.

nvma:
Any example on how to use RegExp and Wildcard query would be really nice.

unisqu:
Working demo examples of every small minor thing instead of a beersearch which u still need to go through the code will be best.

hubyhuby:
It is definitely very complicated to find example on bleve.
In beer search I get stuck not even with bleve itself, but trying to understand the wrap around code for interfaces etc... and it gets hard to get to the actual point of understanding bleve itself...
For example it uses bleveHttp.NewSearchHandler("beer") to handle the search , but when you get there there is no explanation on how to use it here : 
https://github.com/blevesearch/bleve/tree/master/http

If you want to try bleve-explorer then you stumble on Angular :
https://github.com/blevesearch/bleve-explorer

nvma:
When I was trying to go-by-example with a beer search I've made very slow progress since as @unisqu said small chunks of examples are much more easily digested than digging into a working application. Besides it feel like these examples are quite dated by now.

PR #1032, opened 2018-10-26
Initial commit for examples of using Bleve

oskarwojciski:
Related to #1031

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,9 @@
+# Examples
+
+Each example is described inside it directory in separate README file.
+
+To run example, clone repository and run:
+
+```
+go run examples/{exampleDirectory}/main.go
+```
--- a/examples/analyzers_usage/README.md
+++ b/examples/analyzers_usage/README.md
@@ -1,0 +1,5 @@
+# Analyzers usage example
+
+This example shows how to use Bleve defined analyzers.
+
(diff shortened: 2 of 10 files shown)


mschoch:
Seems like there is some overlap with the existing examples.  https://github.com/blevesearch/bleve/blob/master/examples_test.go

oskarwojciski:
> Seems like there is some overlap with the existing examples. https://github.com/blevesearch/bleve/blob/master/examples_test.go

you're right but I like to have things more structured so maybe unify to one version?

hubyhuby:
@oskarwojciski ,
This is wonderfull to have those examples now. Thanks.
An interesting one would be Geo localised search ;which doesn t seem to be in your PR.

paivagustavo:
I'm just trying out Bleve and I have to say that this type of examples are more effective for me.

Sure `examples_test.go` have some overlapping but having a more structured and friendly examples helps for newcomers like myself as I find that the [documentation](http://blevesearch.com/docs/Home/) seems to be a lot more for terminology than a documentation.

I think that these type of examples should be extended and the documentation be improved, with some examples of types of indexing mapping and so on. As I said, I only discovered bleve today, so take this with a little grain of salt. And thanks @oskarwojciski  for this examples it helped me.

PR #1033, opened 2018-10-27
Return FieldArrayPositions along with Fields in DocumentMatch

viswajithiii:
We use Bleve internally. For our use case, when we received fields from a search result, we also needed to know exactly which element had which field value (for documents with nested arrays in them). 

This patch returns `FieldArrayPositions` in the DocumentMatch, which basically is a companion array to `Fields` which returns the array positions of the corresponding element. We did it this way so that we didn't break any clients of the existing struct. We've been using this patch internally for a while now, and it's worked for us, so we're sending this PR out in case you folks upstream think it would be useful.

### DIFF
--- a/search/search.go
+++ b/search/search.go
@@ -97,7 +97,7 @@ type DocumentMatch struct {
 
 	// Fields contains the values for document fields listed in
 	// SearchRequest.Fields. Text fields are returned as strings, numeric
-	// fields as float64s, boolean fiels as bools, and date fields as
+	// fields as float64s, boolean fields as bools, and date fields as
 	// time.RFC3339 formatted strings.
 	// If there is more than one value for a field,
 	// Fields will return a []interface{}, and the client will have to cast each
@@ -106,7 +106,7 @@ type DocumentMatch struct {
 
 	// FieldArrayPositions is a companion map to Fields which returns the array positions
 	// corresponding to each element in Fields.
-	FieldArrayPositions map[string][]ArrayPositions
+	FieldArrayPositions map[string][]ArrayPositions `json:"field_array_positions,omitempty"`
 
 	// if we load the document for this hit, remember it so we dont load again
 	Document *document.Document `json:"-"`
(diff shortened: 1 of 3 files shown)


viswajithiii:
@abhinavdangeti What do you think of this change?

search/search.go
```diff
@@ -97,9 +97,17 @@ type DocumentMatch struct {
 
 	// Fields contains the values for document fields listed in
 	// SearchRequest.Fields. Text fields are returned as strings, numeric
-	// fields as float64s and date fields as time.RFC3339 formatted strings.
+	// fields as float64s, boolean fiels as bools, and date fields as
```
abhinavdangeti: typo: fields

search/search.go
```diff
@@ -97,9 +97,17 @@ type DocumentMatch struct {
...
+	// time.RFC3339 formatted strings.
+	// If there is more than one value for a field,
+	// Fields will return a []interface{}, and the client will have to cast each
+	// value in that slice individually to string, bool or float64 individually.
 	Fields map[string]interface{} `json:"fields,omitempty"`
 
+	// FieldArrayPositions is a companion map to Fields which returns the array positions
+	// corresponding to each element in Fields.
+	FieldArrayPositions map[string][]ArrayPositions
```
abhinavdangeti: \`json:"fields,omitempty"\`

viswajithiii:
Yeah, it wasn't possible to get the actual field values from the `FieldTermLocationMap`. (The term was a lowercase string for string fields, and an opaque byte sequence for non-string fields.)

I've addressed your comments.

abhinavdangeti:
@viswajithiii Before we get this merged, can you to sign the CLA, as described here: https://github.com/blevesearch/bleve/blob/master/CONTRIBUTING.md

viswajithiii:
@abhinavdangeti I think I already did that a couple days ago. Can you confirm that I did things correctly?
image

mschoch:
The original intention was that the returned sliced should recreate the structure of the original slice.  I can see that the current implementation probably doesn't always do that. But that would be considerably more elegant than this solution.

Right now people already complain that the returned stored fields don't resemble the original values (numbers and dates are returned in an internal representation).  Returning the slice position info as another field seems like its taking us further down this path.  If this patch returns a slice of array positions so that the original can be recreated, why not just do this recreation inside of bleve?

viswajithiii:
Closing since stale.

PR #1038, opened 2018-11-01
Fix data race

typeless:
An alternative fix of #1036


typeless:
I have encountered the failed test `TestRegexpSearchScorch` in the CI a couple of times, but I never reproduce it on my local machine. Adding `-race` to the CI builds doesn't show any DATA RACE either.

PR #1045, opened 2018-11-14
Back porting fixes to Alice/6.0.1 branch

sreekanth-cb:
MB-31405 - Fix for higher disk usage
MB-31660 - Fix for faster index deletions

Back porting above fixes to Alice/6.0.1 branch


abhinavdangeti:
@sreekanth-cb The message above indicates there are 2 commit messages, but you seem to have squashed them into a single commit, could you re-do this backport - where there should be 2 commits going back.

PR #1051, opened 2018-11-19
fixing `go vet` errs

sreekanth-cb:


index/scorch/segment/zap/merge.go
```diff
@@ -329,7 +329,7 @@ func persistMergedRest(segments []*SegmentBase, dropsIn []*roaring.Bitmap,
 
 			err = enumerator.Next()
 		}
-		if err != nil && err != vellum.ErrIteratorDone {
+		if err != vellum.ErrIteratorDone {
```
mschoch: i see that go vet reported this one too, but i don't follow it.  how do we know that err is not nil here? the for doesn't guarantee that (if there had been a break).

steveyen: hmm.  Your comment also made me do a double take & head scratching.  I think the idea is that to leave the "for err == nil" loop, the err definitely has to be non-nil, as there also don't seem to be any break's in the loop, just 'return's.

sreekanth-cb: yes @mschoch , on one hand, I had the same feeling..like if someone change the above logic to add a break stmt then it would be a problem.  
But then, thought that such a future break-ing out of the `nil err loop` is worthy of/demands a thorough  inspection of the return logic that follows the loop.

mschoch: Hmmn, if that's the case it seems questionable to me.  To me it seems that breaking out of a loop is a first class programming construct, the loop condition isn't considered some sort of invariant for all subsequent code.

mschoch: I guess to me it feels that `go vet` is being too smart here.  The code can be streamlined due to an incidental property of the code as it exists today, not due to some logical guarantee of the way the code is organized.

mschoch: I guess I'll try to create an independent reproduction and raise the issue upstream.

dgryski: Please tag me on the vet issue. I'd be interested to follow it.

PR #1053, opened 2018-11-22
Add callbacks to Batch

aviadl:
So we can know when a batch was persisted properly to disk

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -248,15 +248,19 @@ type DocIDReader interface {
 	Close() error
 }
 
+type BatchCallbackFunction func(error)
+
 type Batch struct {
 	IndexOps    map[string]*document.Document
 	InternalOps map[string][]byte
+	callBack    []BatchCallbackFunction
 }
 
 func NewBatch() *Batch {
 	return &Batch{
 		IndexOps:    make(map[string]*document.Document),
 		InternalOps: make(map[string][]byte),
+		callBack:    nil,
 	}
 }
 
@@ -274,6 +278,14 @@ func (b *Batch) SetInternal(key, val []byte) {
 
 func (b *Batch) DeleteInternal(key []byte) {
 	b.InternalOps[string(key)] = nil
+}
+
+func (b *Batch) AddCallback(f BatchCallbackFunction) {
+	b.callBack = append(b.callBack, f)
+}
+
+func (b *Batch) Callback() []BatchCallbackFunction {
+	return b.callBack
 }
 
 func (b *Batch) String() string {
@@ -298,6 +310,7 @@ func (b *Batch) String() string {
 func (b *Batch) Reset() {
 	b.IndexOps = make(map[string]*document.Document)
 	b.InternalOps = make(map[string][]byte)
+	b.callBack = nil
 }
 
 func (b *Batch) Merge(o *Batch) {
--- a/index/index.go
+++ b/index/index.go
@@ -248,19 +248,19 @@ type DocIDReader interface {
 	Close() error
 }
 
-type BatchCallbackFunction func(error)
+type BatchCallback func(error)
 
 type Batch struct {
 	IndexOps    map[string]*document.Document
 	InternalOps map[string][]byte
-	callBack    []BatchCallbackFunction
+	callbacks   []BatchCallback
 }
 
 func NewBatch() *Batch {
 	return &Batch{
 		IndexOps:    make(map[string]*document.Document),
 		InternalOps: make(map[string][]byte),
-		callBack:    nil,
+		callbacks:   nil,
 	}
 }
 
@@ -280,12 +280,12 @@ func (b *Batch) DeleteInternal(key []byte) {
 	b.InternalOps[string(key)] = nil
 }
 
-func (b *Batch) AddCallback(f BatchCallbackFunction) {
-	b.callBack = append(b.callBack, f)
+func (b *Batch) AddCallback(f BatchCallback) {
+	b.callbacks = append(b.callbacks, f)
 }
 
-func (b *Batch) Callback() []BatchCallbackFunction {
-	return b.callBack
+func (b *Batch) Callbacks() []BatchCallback {
+	return b.callbacks
 }
 
 func (b *Batch) String() string {
@@ -310,7 +310,7 @@ func (b *Batch) String() string {
 func (b *Batch) Reset() {
 	b.IndexOps = make(map[string]*document.Document)
 	b.InternalOps = make(map[string][]byte)
-	b.callBack = nil
+	b.callbacks = nil
 }
 
 func (b *Batch) Merge(o *Batch) {
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -216,9 +216,6 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 	if next.persisted != nil {
 		s.rootPersisted = append(s.rootPersisted, next.persisted)
 	}
-	if next.callbacks != nil {
-		s.callbacks = append(s.callbacks, next.callbacks...)
-	}
 	s.callbacks = append(s.callbacks, next.callbacks...)
 	// swap in new index snapshot
 	newSnapshot.epoch = s.nextSnapshotEpoch
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -33,7 +33,7 @@ type segmentIntroduction struct {
 
 	applied   chan error
 	persisted chan error
-	callbacks []index.BatchCallbackFunction
+	callbacks []index.BatchCallback
 }
 
 type persistIntroduction struct {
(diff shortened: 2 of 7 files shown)


aviadl:
@steveyen @mschoch this is my offered solution to the callback problem we have discussed

index/index.go
```diff
@@ -248,15 +248,19 @@ type DocIDReader interface {
 	Close() error
 }
 
+type BatchCallbackFunction func(error)
```
steveyen: glad this takes an error param -- makes sense!

aviadl:
@steveyen I have addressed your comments

index/scorch/introducer.go
```diff
@@ -213,6 +216,10 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 	if next.persisted != nil {
 		s.rootPersisted = append(s.rootPersisted, next.persisted)
 	}
+	if next.callbacks != nil {
+		s.callbacks = append(s.callbacks, next.callbacks...)
+	}
```
steveyen: Now it looks like lines 219-thru-221 need to be removed, as line 222 takes care of things -- otherwise the callback is appended twice()?

aviadl: you are right, this was a bad merge ...
Fixed

mschoch:
I gave this some high-level thought over the long weekend.  It seems that one observation here is that advanced use of bleve now requires the of several different callbacks.  Perhaps, before we've introduced too many, we should review what we have so far, and see if there is some consistency we can follow for developer friendliness.

In particular there are 2 aspects I would like to consider (feel free to raise others as well):

1.  Instance callbacks vs Index callbacks vs Global callbacks

I think as proposed, this is the first of what I would call an instance callback.  Each invocation of executing a batch provides its own set of callbacks.  The advantage is the caller knows exactly which batch was persisted when it fires, no need to correlate with other data.  Possible downside is garbage associated with these callbacks (if I remember correctly, if those callbacks are closures over some other state, we have observed this penalty in the past).

I feel like instance level callback does make sense here, because there is no obvious way for callers to identify which batch got executed otherwise.  Relying on the "internal" values would be consistent with other approaches, but that would mean retaining all the data of the batch for the duration of the callback.  Not impossible, but maybe other side-effects here.

2.  Implicit vs Explicit

As proposed, this is what I would call a more explicit callback, by which I mean it's presence is clear from looking at the API on the batch object.  Whereas some of our other callbacks are more implicit or hidden.  Specifically, we already have callbacks during the index operations which are hidden inside the context object.

Would placing this object on the context object be something people would consider?  Would moving the other callback out of the context object be better?  Does anyone else think these should be handled in the same way?

index/scorch/scorch.go
```diff
@@ -54,6 +54,7 @@ type Scorch struct {
 	rootLock             sync.RWMutex
 	root                 *IndexSnapshot // holds 1 ref-count on the root
 	rootPersisted        []chan error   // closed when root is persisted
+	callbacks            []index.BatchCallback
```
steveyen: Hmmm, still thinking after reading marty's comments on global versus local callbacks...

but, this stands out to me as worrisome now, and I should have seen it earlier (my apologies).  I'm thinking this tracking of callbacks ought to be on the snapshot, not on the Scorch instance, as putting it on the Scorch instance feels more 'global' to me now.

steveyen:
> Would placing this object on the context object be something people would consider? Would moving the other callback out of the context object be better? Does anyone else think these should be handled in the same way?

Hi @mschoch -- I don't have a super strong feel either way.  These callbacks has a feel to me of "power user" API's, so I figure a context based approach wouldn't be too unfamiliar to advanced or  long-time go folks.  I guess the explicit instance callback API as @aviadl proposes here, though, is pretty clear.

A related thought... towards "future readiness" is whether there might be even more future callbacks?  For Batch related lifecycle, not much is popping to mind other than "it persisted (or not with some error)".  Other future callbacks feel like they might be in the "global" space as you've outlined... but activities in the global space doesn't seem to me to have a Context instance nearby.

These ideas of larger API design & "taste" might be worth spawning off to some separate place (not sure where) as it might not be just limited to this PR?

index/scorch/scorch.go
aviadl: I think we should separate between 2 things:
1. the APIs to be exposed
2. the implementation, and where to hold each struct/func

Regarding the APIs, i think that a more explicit approach makes it easier for a developer to use.
Regarding the implementation of `this` PR i used the same mechanism of `persistent` channel
For simplicity i would leave the implementation this way, and see how it works
The API will should stay this way anyhow ...

mschoch:
Yes, to clarify all of my comments are regarding the API exposed not the implementation.  My understanding was that this implementation closely followed the way that the existing "batch persisted" channels were closed, thus was "correct" in that regard.

To continue discussion around explicit vs implicit, one thing I find problematic with explicit (as in this PR) is that of naming.  The batch object now has AddCallbacks(), but which callback is that?  The lifecycle of a batch has at least 2 callbacks now, the first fires before a batch is processed (used for controlling memory use by blocking) and now this new one which fires after the batch is persisted.

I would propose that we use the same mechanism for both, and that however it is exposed the naming make it clear which is which.

Also, another thought experiment, consider this code in unsafe mode:

```
err = idx.Batch(b)
doSomethingElse()
```

Couldn't one accomplish the same in safe mode, and gain your own callback with:

```
go func() {
  err = idx.Batch(b)
  myBatchPersistedCallback(err)
}()
doSomethingElse()
```

I'm not arguing against the callback, just seeing if we all agree that should be logically equivalent.  The only slight difference I see is that an actual bleve callback could block bleve internals (sometimes desirable) whereas this external approach cannot.

aviadl:
@mschoch I guess hat regarding the function names, we can probably change it to:
`PersistCallback` - if we agree on this, i will do it quickly, shouldn't be a problem ...

Regarding the equivalence, you are right, but the main problem is that there is a completely different performance between safe and unsafe, so while logically they are equivalent, one would probably rather use the `callback` approach

mschoch:
@aviadl I think that was the point I was trying to illustrate though.  If there is any performance difference, it has to do with the way calling code interacts with bleve.  Bleve is not faster or doing anything else different on the inside when unsafe batches are enabled.  The indexing methods simply return earlier, allowing a higher effective throughput (trade-off being more memory used).

In practice, it might actually be faster at persisting as well because in-memory segment merging allows for more efficient (larger) segments to be written to disk.  But this too is a function of there being multiple segments in memory at the time, not a code-path that is specifically enabled or disabled when using unsafe batches.

mschoch:
OK, I've done a bit more review of current callbacks, and existing APIs to consider, and I'd like to frame the discussion slightly differently.

1.  My idea about using the Context object doesn't make sense, because there is currently only SearchInContext, there is no corresponding context versions of Index() or Batch().  So, short of introducing that (big API change) this idea was a dead end.

2.  The Batch object is the only place we could add a per-invocation callback without changing/breaking the API.  NOTE: this also means that the Index() methods would have no analogous capability...

3.  We have an already somewhat rich set of events/callbacks in scorch, but these are all scorch instance level callbacks, which have no way to pass along exactly which batch just got persisted.  The segment ID seems like the only appropriate option, but the caller of Index() or Batch() would have no idea about these, so again without deeper API changes that seems like a dead end.

So, that leaves us with I think the realization that if we're going to add support for this callback, it's going to look a lot like this PR right now.  (I would propose we change the name, and perhaps switch to a single callback on the batch not multiple...)

The question I'd like everyone to consider though, is do we want to add code into bleve to support this, when we all seem to agree it's possible to get equivalent functionality outside of bleve without these changes (see my comments above).  The argument against this, as I understand it, is that we have a perception that unsafe mode is faster.  But I'm questioning that perception, it seems like if you simply use safe mode, but yourself invoke Index/Batch asynchronously, followed by your callback, you should be able to achieve the same performance (there are no special unsafe mode code paths internal to bleve).

I would appreciate feedback from @aviadl @steveyen @sreekanth-cb @abhinavdangeti @rvncerr and if we think a short meeting to discuss this would help, I can try to arrange it.

aviadl:
I am in favor of having a short meeting on this.
My gut feeling tells me that `unsafe` would be faster, and i think we saw it over some tests.
As you @mschoch said above, mainly because of better memory utilization.
If this `feeling` is wrong, then i think we can drop this PR

mschoch:
@aviadl cool so how about as a next step I produce a small benchmark, and then those results can drive the meeting.  I think I can tweak the beer-search example to illustrate some loading patterns and measure the corresponding perf.

mschoch:
I wanted to circle back and update based on the introduction of #1076

The alternate code I proposed above can have a correctness issue.  One of the reasons one chooses unsafe mode, is that one would like to have multiple batches in flight, even though they are not persisted yet.  However, one still has to be careful though if there is a chance those batches contain the same keys.  To do this safely, one must wait for the batches to be "introduced" to the index before proceeding (as this ensures the correct ordering, if a subsequent batch uses the same key).  Unsafe mode makes this easy, once the Index/Batch method has returned, you know that the batch has been introduced, though not yet persisted.

The alternate I proposed above has no way to know when this "introduction" phase is complete.  Therefore it is a poor choice, and thus I've proceeded with the persisted callback approach.

PR #1062, opened 2018-12-04
DocNumbers - skip segments, that don't include requested ids

rvncerr:
Small performance improvement in SegmentBase.DocNumbers function.

Check, if current segment maximum docId smaller then smallest requested id.
If so, just skip this segment. This code uses recent vellum update (`GetMaxKey()`).

### DIFF
--- a/index/scorch/segment/zap/segment.go
+++ b/index/scorch/segment/zap/segment.go
@@ -418,38 +418,29 @@ func (s *SegmentBase) DocNumbers(ids []string) (*roaring.Bitmap, error) {
 		}
 
 		postingsList := emptyPostingsList
-
-		skipCheck := false
+		filteredIds := ids[:0]
 		sMax := ""
-		iMin := ""
 
-		if len(ids) > 0 {
-			sMaxB, err := idDict.fst.GetMaxKey()
-			if err != nil {
-				skipCheck = true
-			} else {
-				sMax = string(sMaxB)
-				iMin = ids[0]
-				for i := 1; i < len(ids); i++ {
-					if ids[i] < iMin {
-						iMin = ids[i]
-					}
+		sMaxB, err := idDict.fst.GetMaxKey()
+		if err == nil {
+			sMax = string(sMaxB)
+			for _, id := range ids {
+				if id <= sMax {
+					filteredIds = append(filteredIds, id)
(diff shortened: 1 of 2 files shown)


index/scorch/segment/zap/segment.go
```diff
@@ -418,12 +418,37 @@ func (s *SegmentBase) DocNumbers(ids []string) (*roaring.Bitmap, error) {
 		}
```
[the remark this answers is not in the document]
rvncerr: I was thinking in caching min/max in vellum FTS object.

index/scorch/segment/zap/segment.go
```diff
@@ -418,38 +418,29 @@ func (s *SegmentBase) DocNumbers(ids []string) (*roaring.Bitmap, error) {
...
-			if err != nil {
-				skipCheck = true
-			} else {
-				sMax = string(sMaxB)
-				iMin = ids[0]
-				for i := 1; i < len(ids); i++ {
-					if ids[i] < iMin {
-						iMin = ids[i]
-					}
+		sMaxB, err := idDict.fst.GetMaxKey()
```
steveyen: would it be simpler to return the err right here if err != nil?

if the fst is empty, what does GetMaxKey() return?  "", nil?

If so, then sMaxB ought to be lower than any real id, so that might be an edge case that "falls out for free"?

index/scorch/segment/zap/segment.go
```diff
@@ -418,13 +418,29 @@ func (s *SegmentBase) DocNumbers(ids []string) (*roaring.Bitmap, error) {
 		}
 
 		postingsList := emptyPostingsList
-		for _, id := range ids {
+		filteredIds := ids[:0]
```
steveyen: oh, this is nice to reuse the incoming slice backing array.

but, should double check that all the callers of DocNumbers() are ok with their input ids[] getting modified?

rvncerr: I got rid of this. Even if current callers are ok, we cannot be sure about future callers.

index/scorch/segment/zap/segment.go
rvncerr: Yes, now fixed.

Issue #1059, opened 2018-11-30
Boolean search problem with upside-down


abhinavdangeti:
From @aviadl, see https://github.com/blevesearch/bleve/issues/1029#issuecomment-443153832

@abhinavdangeti  @mschoch i I got the problem again using latest version and upside-down indexing
you can see in the bellow the query that is sent, and the results as they are received from the the bleve.hit:

```
--------------- bleve query -----------------------------
{
  "must": {
    "conjuncts": [
      {
        "must": {
          "conjuncts": [
            {
              "start": "2018-11-23T11:51:59+02:00",
              "end": "0001-01-01T00:00:00Z",
              "inclusive_start": true,
              "field": "created"
            }
          ]
        },
        "must_not": {
          "disjuncts": [
            {
              "should": {
                "disjuncts": [
                  {
                    "term": "179d299f-a38a-4a93-89d2-54eed8857ae5",
                    "field": "investigationId"
                  },
                  {
                    "term": "205a2c16-9797-4731-8d17-769e8b168d3d",
                    "field": "investigationId"
                  },
                  {
                    "term": "2451b2dc-7ccb-4905-8264-c0880b15c7ec",
                    "field": "investigationId"
                  },
                  {
                    "term": "512d2a33-7621-48f9-8b21-076ef159555d",
                    "field": "investigationId"
                  },
                  {
                    "term": "55014a53-c563-4291-815d-d1f9fe699c45",
                    "field": "investigationId"
                  },
                  {
                    "term": "5f526439-147a-4743-842e-828dbcf055b0",
                    "field": "investigationId"
                  },
                  {
                    "term": "612f5559-d77c-43f1-89d2-d139c53a4cba",
                    "field": "investigationId"
                  },
                  {
                    "term": "7e1cc1c7-8ed0-41f7-869d-f440da595e97",
                    "field": "investigationId"
                  },
                  {
                    "term": "ae8513ed-84e6-4cea-89ea-9bdb0d52a0b6",
                    "field": "investigationId"
                  },
                  {
                    "term": "d1a523a0-b90e-45be-8b33-fd0ee430372d",
                    "field": "investigationId"
                  },
                  {
                    "term": "e222e1b2-ef5e-4d20-8d7f-1e1e23383252",
                    "field": "investigationId"
                  },
                  {
                    "term": "e5c881e8-1e2e-496e-8ac0-b9de09d3635e",
                    "field": "investigationId"
                  },
                  {
[...]

PR #1065, opened 2018-12-05
Fix Boolean Searcher's MustNotSearcher from advancing incorrectly

abhinavdangeti:
+ Noted incorrect advancing of the boolean's currMustNot position:
    [0xc4200b82c0] BOOL Advance to ID: 13965, currID: 13943, currMust: 13943, currMustNot: 13966
    [0xc4200b82c0] BOOL Advanced, now currID: 13965, currMust: 13965, currMustNot: 13967
+ Fix the incorrect comparison for Boolean searcher's MustNot cached entry
    - If the current position is already where it should be do not advance
    - If the position is only less than (and not just inequal) to the requested ID, advance
+ Unit tests (thanks @mschoch)
+ Addresses https://github.com/blevesearch/bleve/issues/1059

### DIFF
--- a/search/searcher/search_boolean.go
+++ b/search/searcher/search_boolean.go
@@ -356,10 +356,10 @@ func (s *BooleanSearcher) Advance(ctx *search.SearchContext, ID index.IndexInter
 		}
 
 		if s.mustNotSearcher != nil {
-			// Additional check for mustNotSearcher whose cursor isn't tracked by
-			// currentID to prevent it from moving when the searcher's already
-			// where it should be.
-			if s.currMustNot == nil || !s.currMustNot.IndexInternalID.Equals(ID) {
+			// Additional check for mustNotSearcher, whose cursor isn't tracked by
+			// currentID to prevent it from moving when the searcher's tracked
+			// position is already ahead of or at the requested ID.
+			if s.currMustNot == nil || s.currMustNot.IndexInternalID.Compare(ID) < 0 {
 				if s.currMustNot != nil {
 					ctx.DocumentMatchPool.Put(s.currMustNot)
 				}
--- a/search/searcher/search_boolean.go
+++ b/search/searcher/search_boolean.go
@@ -332,8 +332,14 @@ func (s *BooleanSearcher) Advance(ctx *search.SearchContext, ID index.IndexInter
 		}
 	}
 
-	// Advance the searcher only if the cursor is trailing the lookup ID
-	if s.currentID == nil || s.currentID.Compare(ID) < 0 {
+	// Advance the searchers only if the currentID cursor is trailing the lookup ID,
+	// additionally if the mustNotSearcher has been initialized, ensure that the
+	// cursor used to track the mustNotSearcher (currMustNot, which isn't tracked by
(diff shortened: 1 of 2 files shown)


abhinavdangeti:
Just a rebase.

mschoch:
Ya, so my reading of it now is that yes, this particular dataset sort of masked or didn't reveal the problem in scorch, but since the bug was in the searcher, it is likely that a different query/dataset would have also revealed the problem in scorch (and maybe not shown it in upsidedown).

abhinavdangeti:
@mschoch Found yet another corner case, pushed up another commit along with the original. Let me know of any concerns with it.

abhinavdangeti:
Oh no, we don't go back to how we did this in the beginning with the latest change :)
Here're my notes on why this latest change is correct ..
- One thing about the boolean's mustNotSearcher is that - it always works hand-in-hand with either the mustSearcher or the shouldSearcher.
- Moving the check of lookup ID vs currMustNot up and AND'ing it with the check of lookup ID vs current ID ensures that we don't move the currentID to any intermediate position that the currMustNot is already ahead of which will generate spurious results. 
- This also ensures that all the test cases from @aviadl and from you pass as well.
- I've highlighted the test case that fails without the latest change, i've renamed it within the latest change to TestNestedBooleanMustNotSearcherUpsidedown.
- Here're some logs to corroborate my theory ..
    - The case under consideration has:
        ```
                                           DISJ
                                         /      \
                                     TERM        BOOL
                                                 (must:MatchAll + mustNot:Disj)
        ```
    - In the following logs,
        - Both the mustSearcher and the mustNotSearcher of the boolean searcher have reached the end, however the mustSearcher being a matchAll searcher returns nil (currentID) while the mustNotSearcher being a disjunction searcher - currMustNot remains at the last entry in the list.
        - The enclosing disjunction searcher however is a bit behind owing to the position of the term searcher, so when the disjunction tries to advance the boolean searcher to a position that the boolean searcher is already ahead of - _we must deny it_.
    - Before the latest change ..
        ```
        [0xc4200ca580] BOOL Next Ends, currID: , currMustNot: 7@1
        ...
        [0xc4200ca580] BOOL Advance to ID: 4@1, currID: , currMustNot: 7@1
        [0xc4200ca580] BOOL Next Ends, currID: 5@1, currMust: 5@1, currMustNot: 7@1, returns doc: 4@1
        [0xc4200ca580] BOOL Advanced, now currID: 4@1, currMust: 4@1, currMustNot: 7@1
        ```
    - 	After the latest change ..
         ```
         [0xc4202480b0] BOOL Next Ends, currID: , currMustNot: 7@1
         ...
         [0xc4202480b0] BOOL Advance to ID: 4@1, currID: , currMustNot: 7@1
         [0xc4202480b0] BOOL Advanced, now currID: , currMustNot: 7@1
         [0xc4202480b0] BOOL Next Ends, currID: , currMustNot: 7@1
         ```
    - As you can see in the first set of logs above, the boolean after being incorrectly advanced, returns a spurious entry (4@1).

abhinavdangeti:
^^Rebased

PR #1067, opened 2018-12-06
Defer closing the index in tests to avoid cascade failures

abhinavdangeti:
If a test fails, to avoid a subsequent test from failing because
of the earlier test not closing the index, defer the closing part.

### DIFF
--- a/search_test.go
+++ b/search_test.go
@@ -430,13 +430,6 @@ func TestMemoryNeededForSearchResult(t *testing.T) {
 
 // https://github.com/blevesearch/bleve/issues/954
 func TestNestedBooleanSearchers(t *testing.T) {
-	defer func() {
-		err := os.RemoveAll("testidx")
-		if err != nil {
-			t.Fatal(err)
-		}
-	}()
-
 	// create an index with a custom analyzer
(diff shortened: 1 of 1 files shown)


search_test.go
```diff
@@ -453,6 +446,18 @@ func TestNestedBooleanSearchers(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	defer func() {
+		err = idx.Close()
```
sreekanth-cb: idx will be nil if test failed at 440..

abhinavdangeti: If the test fails at line 440, this defer won't be invoked :)

abhinavdangeti:
> Another issue I see here is that, the removeAll () call failed with errors like directory not empty earlier.
And if we place the index Close right before the removeAll call (without a sleep or delay), then the Close triggered async operations in the background may not have completed while we try to removeAll or cleanup the directory.

Isn't Close() synchronous? Pretty sure we wait for all the async tasks to complete within it. 

> A better solution would be to have some some randomness as well in the index names or have a setup function at every test to clean any pending things from left overs?

Yea this may do good, when tests are run in parallel, let me push up a patch here.

PR #1072, opened 2018-12-10
Index store writer

maudetes:
In order to allow for replication, which is a common use case, I added the ()WriteTo function to the KVReader interface and implemented it for BoltDB.

Using the .Advanced() function, it is possible to access the Reader and use WriteTo(io.Writer) to replicate the store.

For example:
```golang
_, store, err := index.Advanced()
if err != nil {
   return err
}

r, err := store.Reader()
if err != nil {
   return err
}
defer r.Close()

tmpFile, err := ioutil.TempFile(path, "store.replicate.")
if err != nil {
   return err
}
defer tmpFile.Close()

err = r.WriteTo(tmpFile)
if err != nil {
   return err
}
```

I tested it locally and it works for my use case. Please let me know if there's anything I should add or change for this pull request to go through.

Thanks.

### DIFF
--- a/index/store/boltdb/reader.go
+++ b/index/store/boltdb/reader.go
@@ -15,6 +15,8 @@
 package boltdb
 
 import (
+	"io"
+
 	"github.com/blevesearch/bleve/index/store"
 	"github.com/boltdb/bolt"
 )
@@ -66,6 +68,11 @@ func (r *Reader) RangeIterator(start, end []byte) store.KVIterator {
 
 	rv.Seek(start)
 	return rv
+}
+
+func (r *Reader) WriteTo(w io.Writer) error {
+	_, err := r.tx.WriteTo(w)
+	return err
 }
 
 func (r *Reader) Close() error {
--- a/index/store/boltdb/reader.go
+++ b/index/store/boltdb/reader.go
@@ -70,6 +70,12 @@ func (r *Reader) RangeIterator(start, end []byte) store.KVIterator {
 	return rv
 }
 
+// As said in the BoltDB documentation:
+// You can use the Tx.WriteTo() function to write
+// a consistent view of the database to a writer.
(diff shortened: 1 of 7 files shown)


maudetes:
Thank you a lot for your time and feedback.

If I understood what you said correctly, a solution could be to keep the WriteTo() method for the boltdb store only (and not expose it on the KVReader interface), then use dynamical type assertion to a boltdb.Reader instance and finally use the WriteTo() method from it.

The code to use it would look like:
```golang
_, store, err := index.Advanced()
if err != nil {
   return err
}

r, err := store.Reader()
if err != nil {
   return err
}
defer r.Close()

br, ok := r.(*boltdb.Reader)
if !ok {
   return  fmt.Errorf("KVReader not a boltdb.Reader")
}

tmpFile, err := ioutil.TempFile(path, "store.replicate.")
if err != nil {
   return err
}
defer tmpFile.Close()

_, err = br.WriteTo(tmpFile)
if err != nil {
   return err
}
```

Moreover, I understand how the WriteTo function for a reader interface might seem "not quite right", but I think it is mostly a semantic question as we don't want to write *to* the store but only replicate what's read *from* the store to a different file. I could change the function name to ReplicateTo() if it seems better to you.

If this solution seems correct for you (and it seems to apply to my case), the only file to modify on the bleve side would be index/store/boltdb/reader.go to add the WriteTo() (or ReplicateTo()) method.

Thank you again for your time.

maudetes:
Hi!

I was wondering if you'd had the time to consider the suggestion I made ? Any feedback on this PR is welcome.

Thank you in advance.

hubyhuby:
Dear @maudetes ,
Not sure if it might help you, but there is a project here using bleve and providing:
Distributed search/indexing
Index replication
Bringing up cluster
Cluster Federation

https://github.com/mosuka/blast

PR #1073, opened 2018-12-10
MB-32295: Optimization for conjunction/disjunction searchers

abhinavdangeti:
If the number of nested searchers within a conjunction/disjunction
searcher is one, just return the nested searcher unwrapped as is, to
avoid an unnecessary level of indirection.

### DIFF
--- a/search/query/conjunction.go
+++ b/search/query/conjunction.go
@@ -74,7 +74,7 @@ func (q *ConjunctionQuery) Searcher(i index.IndexReader, m mapping.IndexMapping,
 	if len(ss) < 1 {
 		return searcher.NewMatchNoneSearcher(i)
 	} else if len(ss) == 1 {
-        // return single nested searcher as is
+		// return single nested searcher as is
 		return ss[0], nil
 	}
 
--- a/search/query/disjunction.go
+++ b/search/query/disjunction.go
@@ -79,8 +79,8 @@ func (q *DisjunctionQuery) Searcher(i index.IndexReader, m mapping.IndexMapping,
 
 	if len(ss) < 1 {
 		return searcher.NewMatchNoneSearcher(i)
-	} else if len(ss) == 1 {
-		// return the single nested searcher as is
+	} else if len(ss) == 1 && q.Min <= 1 {
+		// return the single nested searcher as is; only if min clauses is not greater than 1;
 		return ss[0], nil
 	}
 
--- a/search_test.go
+++ b/search_test.go
@@ -1012,5 +1012,74 @@ func TestQueryStringEmptyConjunctionSearcher(t *testing.T) {
 	query := NewQueryStringQuery("foo:bar +baz:\"\"")
 	searchReq := NewSearchRequest(query)
 
-    _, _ = index.Search(searchReq)
+	_, _ = index.Search(searchReq)
+}
+
+func TestDisjunctionQueryIncorrectMin(t *testing.T) {
+	// create an index with default settings
+	idxMapping := NewIndexMapping()
+	idx, err := New("testidx", idxMapping)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		err = idx.Close()
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = os.RemoveAll("testidx")
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	// create and insert documents as a batch
+	batch := idx.NewBatch()
+	docs := []struct {
+		field1 string
+		field2 int
+	}{
+		{
+			field1: "one",
+			field2: 1,
+		},
+		{
+			field1: "two",
+			field2: 2,
+		},
+	}
+
+	for i := 0; i < len(docs); i++ {
+		doc := document.NewDocument(strconv.Itoa(docs[i].field2))
+		doc.Fields = []document.Field{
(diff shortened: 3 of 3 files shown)


steveyen:
indeed -- back in the day, i vaguely recall I couldn't get my attempt to pass all the tests (unit tests?).  I see @abhinavdangeti's PR here does pass the travis-CI gauntlet, though.

On what @mschoch is referring to -- maybe the queryStringMode flags that are part of the conjunction and disjunction queries are related...

https://github.com/blevesearch/bleve/blob/master/search/query/conjunction.go#L29

Here's the commit that talks a bit about queryStringMode for convenience (and perhaps to tickle more memory neurons): https://github.com/blevesearch/bleve/commit/f391b991c20f02681bacd197afc6d8aed444e132

mschoch:
One other recollection was that @steveyen's change was similar but different.  I seem to recall it had to do with phrase searches that after analysis had just a single term, rewrite that to term search.  Or something like that.  But, even though it seemed right, Steve said some tests failed.  I'm not sure if there is any history of what Steve tried.

search/query/conjunction.go
```diff
@@ -70,9 +70,14 @@ func (q *ConjunctionQuery) Searcher(i index.IndexReader, m mapping.IndexMapping,
 		}
 		ss = append(ss, sr)
 	}
+
 	if len(ss) < 1 {
 		return searcher.NewMatchNoneSearcher(i)
+	} else if len(ss) == 1 {
+        // return single nested searcher as is
```
steveyen: oh, maybe need a go fmt on this.

mschoch:
I had thought about min and concluded that it didn't matter.  My thinking was that a disjunction searcher can only return items with at least one matching clause anyway.  So, on it's own replacing a disjunction containing one clause with just that clause seemed OK.

But, there are at least 2 other cases to consider.  The first is that boolean searches also support "min", but in a way that is a bit beyond just passing it down to the inner "should" disjunction.  A boolean searcher can meaningfully support min=0, because something satisfying the must clause, but none of the should clauses is still a match.  I think this is done by extra code in the boolean searcher, and so would be safe even with the rewrite, but this should be checked.

Another case that I just thought of though is the (hopefully not very important) corner case where you construct a disjunction with a single clause, but also set min to something > 1.  Logically this should return nothing.  But again, there may be additional concern around rewriting that as such, since in the query string case in particular, ES has difference behavior for something which "happens to return nothing" vs something which "logically cannot return anything".  See the link @steveyen gave for more info on that.

Then when it comes to boosting.  I think the rewrite is safe, but there might be some extra work to make sure the correct boost in used.  Remember that a boost on the disjunction itself would only matter if it was part of some larger query, and the boost on the single clause inside the disjunction is actually meaningless because it would weight that clause relative to it's siblings (but there are none).  So, I think if you rewrite to eliminate the disjunction, inheriting the parent boost would be correct (since you're basically pushing that clause up the tree).  But, this would be another case where if you're trying to basically score like ES, more testing would be required to make this match.

abhinavdangeti:
@mschoch Thanks for the description, let me look into it and push up another commit here addressing the correct propagation of the boost value and if need the min value.

abhinavdangeti:
I'm pretty sure I have changed the current behavior with the 2nd patch above^.

Here's an example .. lets consider this index with 2 docs, with one of them having "a" (doc1), and the other having "b" (doc2) for a certain field. With the following query ..
```
                                DISJ
                             /        \
                       TERM(a)        TERM(b)
                       boost: 15      boost: 10
```
I expect to see doc1 with field value "a" as the first hit (higher score and all), which I do.
Now consider the following query here:
```
                                DISJ
                             /        \
                       TERM(a)        CONJ
                       boost: 15      boost: 20
                                           \
                                           TERM(b)
                                           boost: 10
```
With this query, with the current code base - I still however see doc1 as the first hit with the higher score (same scores as the 1st query). Is this correct? Because my latest patch changes this behavior and returns doc2 as the first hit as it's scored higher now.

mschoch:
Honestly, I don't understand the change at all.  The implementation of SetBoost on things like conjunction/disjunction is calling SetBoost on all the children, but that's not how boost works at all.  Boosting is valid at any level in the hierarchy, and only serves to boost things at that level relative to other things at that same level.  I thought the intention was to simply set the boost on the child element when promoting/replacing something else, but this change seems to go way beyond that.

abhinavdangeti:
Yikes, you're right, thanks for the spot @mschoch !
One question following that though, if the conjunction/disjunction searcher move the immediate child up by one level, and update just the child's boost with their's, does that mean that - if the child were to further have more child searchers under it, we'd have to move the boost setting of them up by one level as well?

mschoch:
Hmmn, well I think it shows this isn't quite as simple as we thought.  Hypothetically a user could set a boost on each node in the hierarchy.  Rewriting the query seems to mean doing some math if we're going to try and preserve the users intention and not just throw away part of it.  Worse, I'm not sure exactly what that math looks like...

abhinavdangeti:
The latest patchset(s), only check for the "min" setting for the optimization to go through.

The boost setting from the parent is not propagated to the child while removing the parent (conjunction/disjunction) from the equation => this is to retain the boost settings by the user in the hierarchy.

Also this means that with the introduction of this change - the behavior doesn't change when it comes to how boost affects scoring with conjunction/disjunction searchers.

PR #1077, opened 2018-12-12
Fix query string number in query string query

gilliek:
Query string query lexer reads IP addresses as tNUMBER instead of tSTRING.

The error is in the **inNumOrStrState** (_search/query/query_string_lex.go_) function where the **seenDot** field of the lexer is never set to true:

```go
// see where to go
if !l.seenDot && next == '.' {
    // stay in this state
    l.buf += string(next)
    return inNumOrStrState, true
}
```

This PR fixes this issue.

### DIFF
--- a/search/query/query_string_lex.go
+++ b/search/query/query_string_lex.go
@@ -273,6 +273,7 @@ func inNumOrStrState(l *queryStringLex, next rune, eof bool) (lexState, bool) {
 	// see where to go
 	if !l.seenDot && next == '.' {
 		// stay in this state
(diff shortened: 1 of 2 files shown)


abhinavdangeti:
@gilliek This being your first contribution, can you follow the instructions here and sign the CLA, if you haven't already .. https://github.com/blevesearch/bleve/blob/master/CONTRIBUTING.md

gilliek:
@abhinavdangeti I already signed the CLA ;-)

PR #1089, opened 2018-12-17
Fix #1079 - Incorrect scorch score

### LINKED ISSUE #1079: Upside_down and Scorch showing different scores
Here's a sample program, where upsidedown and scorch seem to be returning different scores for the same query, although the documents are in the same order ..
```
package main

import (
	"fmt"
	"os"
	"strconv"

	"github.com/blevesearch/bleve/document"
	"github.com/blevesearch/bleve/index/scorch"
	"github.com/blevesearch/bleve/index/upsidedown"
)

func index2DocsQueryReturnScores(indexName string) (id1 string, score1 float64,
	id2 string, score2 float64) {
	// create an index
	idxMapping := bleve.NewIndexMapping()
	idx, err := bleve.NewUsing("testidx", idxMapping, indexName, bleve.Config.DefaultKVStore, nil)
	if err != nil {
		panic(err)
	}
	defer func() {
		err = idx.Close()
		if err != nil {
			panic(err)
		}
		err = os.RemoveAll("testidx")
		if err != nil {
			panic(err)
		}
	}()

	// create and insert documents as a batch
	batch := idx.NewBatch()
	docs := []struct {
		field1 string
		field2 int
	}{
		{
			field1: "one",
			field2: 1,
		},
		{
			field1: "two",
			field2: 2,
		},
	}
	for i := 0; i < len(docs); i++ {
		doc := document.NewDocument(strconv.Itoa(docs[i].field2))
		doc.Fields = []document.Field{
			document.NewTextField("field1", []uint64{}, []byte(docs[i].field1)),
			document.NewNumericField("field2", []uint64{}, float64(docs[i].field2)),
		}
		doc.CompositeFields = []*document.CompositeField{
			document.NewCompositeFieldWithIndexingOptions(
				"_all", true, []string{"text"}, []string{},
				document.IndexField|document.IncludeTermVectors),
		}
		if err = batch.IndexAdvanced(doc); err != nil {
			panic(err)
		}
	}
	if err = idx.Batch(batch); err != nil {
		panic(err)
	}

	/*
		Query:
				 DISJ
			        /    \
			     CONJ    TERM(two)
			     /
		          TERM(one)
	*/
	tq1 := bleve.NewTermQuery("one")
	tq1.SetBoost(2)
	tq2 := bleve.NewTermQuery("two")
	tq2.SetBoost(3)

	cq := bleve.NewConjunctionQuery(tq1)
	cq.SetBoost(4)

	q := bleve.NewDisjunctionQuery(tq1, tq2)
	sr := bleve.NewSearchRequestOptions(q, 2, 0, false)
	res, err := idx.Search(sr)
	if err != nil {
		panic(err)
	}
	if len(res.Hits) != 2 {
		panic(fmt.Sprintf("Expected 2 hits, but got: %v", len(res.Hits)))
	}

	return res.Hits[0].ID, res.Hits[0].Score, res.Hits[1].ID, res.Hits[1].Score
}

func main() {
	up_id1, up_score1, up_id2, up_score2 := index2DocsQueryReturnScores(upsidedown.Name)
	sc_id1, sc_score1, sc_id2, sc_score2 := index2DocsQueryReturnScores(scorch.Name)

	if up_id1 != sc_id1 || up_id2 != sc_id2 {
		fmt.Printf("upsidedown, scorch returned different docs;\nupsidedown: (%s, %s), scorch: (%s, %s)\n",
			up_id1, up_id2, sc_id1, sc_id2)
	}

	if up_score1 != sc_score1 || up_score2 != sc_score2 {
		fmt.Printf("upsidedown, scorch showing different scores;\nupsidedown: (%v, %v), scorch: (%v, %v)\n",
			up_score1, up_score2, sc_score1, sc_score2)
	}
}
```
This is the output on my computer for the above program:
```
upsidedown, scorch showing different scores;
upsidedown: (0.10090091739019828, 0.06726727826013218), scorch: (0.09805806795731893, 0.06537204530487929)
```

sreekanth-cb:
The composite field norm/scoring is broken with
scorch, as it incorrectly considers the "_id"
field during the composite field analysis.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -526,7 +526,7 @@ func (s *Scorch) Analyze(d *document.Document) *index.AnalysisResult {
 			rv.Analyzed[i] = tokenFreqs
 			rv.Length[i] = fieldLength
 
-			if len(d.CompositeFields) > 0 {
+			if len(d.CompositeFields) > 0 && field.Name() != "_id" {
 				// see if any of the composite fields need this
 				for _, compositeField := range d.CompositeFields {
 					compositeField.Compose(field.Name(), fieldLength, tokenFreqs)
--- a/test/versus_score_test.go
+++ b/test/versus_score_test.go
@@ -1,0 +1,135 @@
+//  Copyright (c) 2018 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package test
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"testing"
+
+	"github.com/blevesearch/bleve"
+	"github.com/blevesearch/bleve/document"
+	"github.com/blevesearch/bleve/index/scorch"
+	"github.com/blevesearch/bleve/index/upsidedown"
+	"github.com/blevesearch/bleve/mapping"
+	"github.com/blevesearch/bleve/search"
+)
+
+func TestDisjunctionSearchScoreIndexWithCompositeFields(t *testing.T) {
+	upHits := disjunctionQueryiOnIndexWithCompositeFields(upsidedown.Name, t)
+	scHits := disjunctionQueryiOnIndexWithCompositeFields(scorch.Name, t)
+
+	if upHits[0].ID != scHits[0].ID || upHits[1].ID != scHits[1].ID {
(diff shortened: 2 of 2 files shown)


index/scorch/scorch.go
```diff
@@ -526,7 +526,7 @@ func (s *Scorch) Analyze(d *document.Document) *index.AnalysisResult {
 			rv.Analyzed[i] = tokenFreqs
 			rv.Length[i] = fieldLength
 
-			if len(d.CompositeFields) > 0 {
+			if len(d.CompositeFields) > 0 && field.Name() != "_id" {
```
steveyen: great catch -- looks good

(separately, another day... looks like "_id" ought to be refactored into a global var or const somewhere for scorch, if not already)

mschoch:
Seems like this goes beyond just scoring, but also the question of whether the doc id should be indexed in the _all field.  It's surely inconsistent with upsidedown, but would be worth comparing with what ES does.

index/scorch/scorch.go
sreekanth-cb: Sure @steveyen , I shall make the "_id" const in a separate PR soon.

thanks @mschoch for bringing that point,
Let me create a new ticket to track that down?

PR #1091, opened 2018-12-19
Add a test for data race

typeless:
```
go test -race -v ./index/upsidedown
```


mschoch:
Thanks for contributing this, a modified version of the test been committed as a part of the fix.  See #1121

PR #1098, opened 2019-01-04
Reduce the number of docs in versus_test: TestScorchVersusUpsideDownBoltAll

abhinavdangeti:
This one test seems to take an exorbitant amount of time to run
on my computer, not sure if it has to do with the new macOS - Mojave.

Scaling the number of docs down in the test, lets the test finish
rather quickly.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -33,7 +33,7 @@ import (
 	"github.com/blevesearch/bleve/index"
 	"github.com/blevesearch/bleve/index/scorch/segment"
 	"github.com/blevesearch/bleve/index/scorch/segment/zap"
-	bolt "github.com/etcd-io/bbolt"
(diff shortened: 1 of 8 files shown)


mschoch:
How long does it actually take?

The other option is to honor the short flag, and configure your IDE (or when you're at the command-line) to run with the short flag:

```
  if testing.Short() {
    t.Skip("skipping versus test in short mode")
  }
```

abhinavdangeti:
It never finishes actually, just busy dumping a bunch of stuff to command line, and I lost patience after 300s. It's super weird that I started noticing this only after an update that I did this morning.

steveyen:
hmmm.  you're on Mojave OSX?  Perhaps they have new things like anti-virus scanning, or disk encryption / compression -- and the "versus" tests happen to put its files in the disk areas with those features?  I'm not a mojave person though.

Issue #1096, opened 2019-01-04
duplicates observed when using scorch


mschoch:
The following program illustrates the problem.  I believe this program follows the rules for correctly using bleve indexes and reusing batches.

```
package main

import (
	"fmt"
	"log"
	"math/rand"
	"os"
	"time"

	"github.com/blevesearch/bleve"
)

func main() {

	// cleanup from previous runs
	os.RemoveAll("idx.bleve")

	// use default mapping
	mapping := bleve.NewIndexMapping()

	// create a scorch index with default SAFE batches
	idx, err := bleve.NewUsing("idx.bleve", mapping, "scorch", "scorch", nil)
	if err != nil {
		log.Fatal(err)
	}

	// create a single batch instance that we will reuse
	// this should be safe because we have single goroutine
	// and we always wait for batch execution to finish
	batch := idx.NewBatch()

	// number of batches to execute
	for i := 0; i < 10; i++ {

		// track which id's we've already used in this batch
		usedAlready := map[string]struct{}{}

		// number of documents to put into the batch
		for j := 0; j < 50; j++ {

			// create a doc id 0-99
			// this could duplicate something already in the index
			//   this too should be OK and update the item in the index
			id := fmt.Sprintf("%d", rand.Intn(100))

			// if we've already used this id in this batch, skip it and keep going
			if _, ok := usedAlready[id]; ok {
				continue
			}
			usedAlready[id] = struct{}{}
			err = batch.Index(id, map[string]interface{}{
				"name":  id,
				"batch": fmt.Sprintf("%d", i),
			})
			if err != nil {
				log.Fatal(err)
			}
		}

		// execute the batch
		err = idx.Batch(batch)
		if err != nil {
			log.Fatal(err)
		}

		// reset the batch before reusing it
		batch.Reset()
	}

	// wait for any merging to finish
	time.Sleep(10 * time.Second)

	// search for docs having name starting with the number 9
	q := bleve.NewWildcardQuery("9*")
	q.SetField("name")
	req := bleve.NewSearchRequestOptions(q, 1000, 0, false)
	req.Fields = []string{"*"}
	res, err := idx.Search(req)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(res)

	// observe that the same doc id comes back multiple times
}

```
When I run the program I observe:
```
16 matches, showing 1 through 16, took 1.803845ms
    1. 90 (0.140728)
	batch
		5
	name
		90
    2. 91 (0.140728)
	batch
[...]

PR #1099, opened 2019-01-05
add unit test for #1096

mschoch:


mschoch:
this is a slightly tweaked version of the code shared in #1096 that hopefully makes it play nicer as a unit test.

confirmed to fail reliably before the fix and passes now

PR #1103, opened 2019-01-09
MB-31974 - Streaming search results

sreekanth-cb:
Adding optional builder/maker callbacks for the
document match handlers from applications.
These builder method gives back a threadsafe
document match handler callback for bleve which
it will invoke for every document hit during
a search.
The application should make its own copy of
the hit/documentMatch in the callback if it
wish to hold on to data.

### DIFF
--- a/search/collector.go
+++ b/search/collector.go
@@ -30,3 +30,23 @@ type Collector interface {
 	SetFacetsBuilder(facetsBuilder *FacetsBuilder)
 	FacetResults() FacetResults
 }
+
+// DocumentMatchHandler is the type of document match callback
+// bleve will invoke during the search.
+// Eventually, bleve will indicate the completion of an ongoing search,
+// by passing a nil value for the document match callback.
+// The application should take a copy of the hit/documentMatch
+// if it wish to own it or need prolonged access to it.
+type DocumentMatchHandler func(hit *DocumentMatch) error
+
+type MakeDocumentMatchHandlerKeyType string
+
+var MakeDocumentMatchHandlerKey = MakeDocumentMatchHandlerKeyType(
+	"MakeDocumentMatchHandlerKey")
+
+// MakeDocumentMatchHandler is an optional DocumentMatchHandler
+// builder function which the applications can pass to bleve.
+// These builder methods gives a DocumentMatchHandler function
+// to bleve, which it will invoke on every document matches.
+type MakeDocumentMatchHandler  func(ctx *SearchContext) (
+	callback DocumentMatchHandler, loadID bool, err error)
--- a/search/collector/topn.go
+++ b/search/collector/topn.go
@@ -140,6 +140,7 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher, 
 	}
 	searchContext := &search.SearchContext{
 		DocumentMatchPool: search.NewDocumentMatchPool(backingSize+searcher.DocumentMatchPoolSize(), len(hc.sort)),
+		Collector: hc,
 	}
 
 	hc.dvReader, err = reader.DocValueReader(hc.neededFields)
@@ -153,6 +154,19 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher, 
 		}
 		hc.sort.UpdateVisitor(field, term)
 	}
+
+	dmHandlerMaker := MakeTopNDocumentMatchHandler
+	if cv := ctx.Value(search.MakeDocumentMatchHandlerKey); cv != nil {
+		dmHandlerMaker = cv.(search.MakeDocumentMatchHandler)
+	}
+	// use the application given builder for making the custom document match
+	// handler and perform callbacks/invocations on the newly made handler.
+	dmHandler, loadID, err := dmHandlerMaker(searchContext)
+	if err != nil {
+		return err
+	}
+
+ 	hc.needDocIds = hc.needDocIds||loadID
 
 	select {
 	case <-ctx.Done():
@@ -169,13 +183,26 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher, 
 			}
 		}
 
-		err = hc.collectSingle(searchContext, reader, next)
+		err = hc.prepareDocumentMatch(searchContext, reader, next)
+		if err != nil {
+			break
+		}
+
+ 		err = dmHandler(next)
 		if err != nil {
 			break
 		}
 
 		next, err = searcher.Next(searchContext)
 	}
+
+ 	// help finalize/flush the results in case
+	// of custom document match handlers.
+	err = dmHandler(nil)
+	if err != nil {
+		return err
+	}
+
 	// compute search duration
 	hc.took = time.Since(startTime)
 	if err != nil {
@@ -191,8 +218,8 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher, 
 
 var sortByScoreOpt = []string{"_score"}
 
-func (hc *TopNCollector) collectSingle(ctx *search.SearchContext, reader index.IndexReader, d *search.DocumentMatch) error {
-	var err error
+func (hc *TopNCollector) prepareDocumentMatch(ctx *search.SearchContext,
+	reader index.IndexReader, d *search.DocumentMatch ) (err error) {
 
 	// visit field terms for features that require it (sort, facets)
 	if len(hc.neededFields) > 0 {
@@ -226,34 +253,51 @@ func (hc *TopNCollector) collectSingle(ctx *search.SearchContext, reader index.I
 		hc.sort.Value(d)
 	}
 
-	// optimization, we track lowest sorting hit already removed from heap
-	// with this one comparison, we can avoid all heap operations if
-	// this hit would have been added and then immediately removed
-	if hc.lowestMatchOutsideResults != nil {
-		cmp := hc.sort.Compare(hc.cachedScoring, hc.cachedDesc, d, hc.lowestMatchOutsideResults)
-		if cmp >= 0 {
-			// this hit can't possibly be in the result set, so avoid heap ops
-			ctx.DocumentMatchPool.Put(d)
-			return nil
-		}
-	}
-
-	removed := hc.store.AddNotExceedingSize(d, hc.size+hc.skip)
-	if removed != nil {
-		if hc.lowestMatchOutsideResults == nil {
-			hc.lowestMatchOutsideResults = removed
-		} else {
-			cmp := hc.sort.Compare(hc.cachedScoring, hc.cachedDesc, removed, hc.lowestMatchOutsideResults)
-			if cmp < 0 {
-				tmp := hc.lowestMatchOutsideResults
-				hc.lowestMatchOutsideResults = removed
-				ctx.DocumentMatchPool.Put(tmp)
-			}
-		}
-	}
-
 	return nil
 }
+
+func MakeTopNDocumentMatchHandler(
+	ctx *search.SearchContext) (search.DocumentMatchHandler, bool, error) {
+	var hc *TopNCollector
+	var ok bool
+	if hc, ok = ctx.Collector.(*TopNCollector); ok {
+		return func(d *search.DocumentMatch) error {
+			if d == nil {
+				return nil
+			}
+			// optimization, we track lowest sorting hit already removed from heap
+			// with this one comparison, we can avoid all heap operations if
+			// this hit would have been added and then immediately removed
+			if hc.lowestMatchOutsideResults != nil {
+				cmp := hc.sort.Compare(hc.cachedScoring, hc.cachedDesc, d,
+					hc.lowestMatchOutsideResults)
+				if cmp >= 0 {
+					// this hit can't possibly be in the result set, so avoid heap ops
+					ctx.DocumentMatchPool.Put(d)
+					return nil
+				}
+			}
+
+			removed := hc.store.AddNotExceedingSize(d, hc.size+hc.skip)
+			if removed != nil {
+				if hc.lowestMatchOutsideResults == nil {
+					hc.lowestMatchOutsideResults = removed
+				} else {
+					cmp := hc.sort.Compare(hc.cachedScoring, hc.cachedDesc,
+						removed, hc.lowestMatchOutsideResults)
+					if cmp < 0 {
+						tmp := hc.lowestMatchOutsideResults
+						hc.lowestMatchOutsideResults = removed
+						ctx.DocumentMatchPool.Put(tmp)
+					}
+				}
+			}
+			return nil
+		}, false, nil
+	}
+	return nil, false, nil
+}
+
 
 // visitFieldTerms is responsible for visiting the field terms of the
 // search hit, and passing visited terms to the sort and facet builder
--- a/search/search.go
+++ b/search/search.go
@@ -285,6 +285,7 @@ type SearcherOptions struct {
 // SearchContext represents the context around a single search
 type SearchContext struct {
 	DocumentMatchPool *DocumentMatchPool
+	Collector Collector
 }
 
 func (sc *SearchContext) Size() int {


mschoch:
Hey @sreekanth-cb can you share more about how you came up with this design?  

I would have expected Bleve to support streaming results directly (Go objects, over channel or callback), however this seems to stream bytes using JSON encoding?

Also, the use of context to use this feature makes it feel like a backdoor hacked into the existing implementation, instead of asking, what would a new proper streaming interface look like (new API).

sreekanth-cb:
Hi @mschoch ,

The current design is based on the thinking that,

1. Support streaming with no/minimal changes to bleve, with cbft implementing the new streamCollector interface.

2. Wanted to avoid channel based streaming due to inherent lock overheads, and context is used to pass the streamCollector implementation to avoid any API breaks.

3. Streaming bytes was an optimisation to save the serialisation efforts of Go structs to protobuf/IDL struct types. In this bytes streamed mode, the unmarshal of all the document hits will happen only once at the receiving client. (either SDK or FTS client for N1QL)

mschoch:
> 1. Support streaming with no/minimal changes to bleve, with cbft implementing the new streamCollector interface.

This seems like a missed opportunity to make Bleve better.  Streaming results seems like a really useful feature, worthy of a new API (and all the headaches/discussion that come with it).  Instead you've opted for something that works for cbft but skips the harder API issues.

> 2. Wanted to avoid channel based streaming due to inherent lock overheads, and context is used to pass the streamCollector implementation to avoid any API breaks.

Regarding channels/locks, it seems like ultimately there has to be synchronization somewhere.  You've started a search, but are processing results before the Search method has returned.  Callbacks vs channel just trade off where the synchronization lives.  I'm probably OK with the callback approach here, but not 100% sure yet.

Again, streaming the results seems like completely different behavior, worthy of a new method or updated signature to convey these very different semantics to the API user.

> 3. Streaming bytes was an optimisation to save the serialisation efforts of Go structs to protobuf/IDL struct types. In this bytes streamed mode, the unmarshal of all the document hits will happen only once at the receiving client. (either SDK or FTS client for N1QL)

Again, it seems to me like the mistake is in making this decision *inside* of bleve.  JSON happens to work for FTS because you've already got code at the other end that can process JSON versions of them.  What about the bleve user that has everything in Go, and it's embedded into a process.  They *never* serialize anything because it never leaves the process.  They still might want their results streamed.

Did you consider a design where the callback gives you a slice of search result hits?  And then FTS could still do the JSON in their callback implementation if they choose.  This feels more appropriate for the general purpose bleve user.

steveyen:
> > with no/minimal changes to bleve
> Again, it seems to me like the mistake is in making this decision inside of bleve...
> Did you consider a design where the callback gives you a slice of search result hits? And then FTS could still do the JSON in their callback implementation if they choose. This feels more appropriate for the general purpose bleve user.

Indeed -- agree very much with @mschoch here that the proposed change right now is too "big" or not orthogonal enough and that decisions like JSON'ification and batching should be instead on the outside of this -- so, that becomes the responsibility of whoever is implementing the callback.

That is, this should have more of a feel of... "this is an optional callback that's invoked by bleve on every DocumentMatch that's considered".  No streaming or serialization here, that is.  The implementer of the callback function, instead, would decide what to do (log it, count it, serialize it, batch it, push it into memcached or S3, etc -- it's up to the callback).

steveyen:
One more consideration that pops to mind is whether the callback would be able to return an error -- e.g., perhaps there's some issue that means the search processing should stop.  Or, whether something like that should be handled via the Context.

sreekanth-cb:
> Again, streaming the results seems like completely different behavior, worthy of a new method or updated signature to convey these very different semantics to the API user.

Thinking of creating a ticket/issue for this, as that is a deeper change.

> Again, it seems to me like the mistake is in making this decision _inside_ of bleve. JSON happens to work for FTS because you've already got code at the other end that can process JSON versions of them. What about the bleve user that has everything in Go, and it's embedded into a process. They _never_ serialize anything because it never leaves the process. They still might want their results streamed.
> 
> Did you consider a design where the callback gives you a slice of search result hits? And then FTS could still do the JSON in their callback implementation if they choose. This feels more appropriate for the general purpose bleve user.

Indeed. Its my bad with some last moment cleanups. My comment and the PR code not matching at all when I read it now. Lost sight of forest for trees.. 

Original callback was taking a [] of hits/DocumentMatch.  And all specifics should be handled at the callback level. 
Batching of hits was intended to save the synchronisation effort than it is with every individual hit.

mschoch:
Some ideas to think about.

1.  Create a small program to build a small index and run a search.  Think about how you'd want to write this program to handle see every hit as it is found as opposed to the end.  What I'm suggesting is first think about how a *bleve* user would want this program to look (don't think about how bleve should implement the details yet).

Specifically, what would the search function be called?  What arguments would it take?  Is one of them a callback?  Or should the callback be part of the search request options?  What should the function return?  And when does it return (at the completion of the search or right away?)  These are all things you can think of from the perspective of a bleve consumer, not a bleve implementor.  It'd be interesting to hear your ideas on these topics and discuss them here.

2.  This may be more clear after answering question 1 above, but I think one of two things might happen.

a.  The current functionality (no streaming) ends up being able to be reimplemented internally to use the streaming version, only bleve supplies the callback instead of the user.  The callback would find the top-N, compute the rolled up summary info, etc.  In this way its like Collect() vs collectSingle(), only collectSingle because the callback handling each hit.  I know there are more complexities here, just a rough idea.

b.  The current functionality (no streaming) ends up being like having 1 large batch.

"a" and "b" here may actually end up being the same, I don't know.  But these are just some ideas to explore.

sreekanth-cb:
let me attempt the callback approach (a) mentioned above.

search/collector.go
```diff
@@ -30,3 +30,20 @@ type Collector interface {
 	SetFacetsBuilder(facetsBuilder *FacetsBuilder)
 	FacetResults() FacetResults
 }
+
+
+type DocumentMatchHandler func(hit *DocumentMatch) error
+
+type MakeDocMatchHandlerKey string
```
steveyen: sometimes is called DocumentMatch, sometimes DocMatch?  Recommend be consistent on DocumentMatch?

search/collector/topn.go
```diff
@@ -154,6 +154,23 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
...
+		dmHandler =  NewTopNDocumentMatchHandler(searchContext, hc, nil)
+	} else if dmHandlerMaker, ok = cv.(search.MakeDocumentMatchHandler); !ok {
+		// on issues with the given doc match handler maker, 
+		// initialize the dmHandler with the default TopN handler
+		dmHandler =  NewTopNDocumentMatchHandler(searchContext, hc, nil)
+	} else {
+		// use the application given builder for making the custom document match
+		// handler and perform callbacks/invocations on the newly made handler.
+		dmHandler = dmHandlerMaker(searchContext)
+		loadID = true
```
steveyen: the loadID here looks a little like an odd, special case?  Instead, what if the dmHandlerMaker() returned an additional bool loadID return value, so that the callback can decide whether it wants ID's loaded or not?

search/collector/topn.go
```diff
@@ -169,13 +186,26 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
...
+
+		err = dmHandler(next)
 		if err != nil {
 			break
 		}
 
 		next, err = searcher.Next(searchContext)
 	}
+
+	// help finalize/flush the results in case
```
steveyen: should add to the public doc / API comments up above that the handler callback will be invoked with a nil hit, to signal that there aren't any more hits coming?

search/collector/topn.go
```diff
@@ -226,33 +257,46 @@ func (hc *TopNCollector) collectSingle(ctx *search.SearchContext, reader index.I
...
-	if hc.lowestMatchOutsideResults != nil {
-		cmp := hc.sort.Compare(hc.cachedScoring, hc.cachedDesc, d, hc.lowestMatchOutsideResults)
-		if cmp >= 0 {
-			// this hit can't possibly be in the result set, so avoid heap ops
-			ctx.DocumentMatchPool.Put(d)
+	return nil
+}
+
+func NewTopNDocumentMatchHandler(ctx *search.SearchContext, 
+	hc *TopNCollector, d *search.DocumentMatch) search.DocumentMatchHandler {
```
steveyen: looks like d param is not needed here

search/collector/topn.go
```diff
@@ -154,6 +154,23 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
 		hc.sort.UpdateVisitor(field, term)
 	}
 
+	var dmHandlerMaker search.MakeDocumentMatchHandler
+	var dmHandler search.DocumentMatchHandler
+	var loadID, ok bool
+	if cv := ctx.Value(search.MakeDocumentMatchHandlerKey); cv == nil {
+		// initialize the dmHandler with the default TopN handler
+		dmHandler =  NewTopNDocumentMatchHandler(searchContext, hc, nil)
```
steveyen: There's something not quite right about this, in that the NewTopNDocumentMatchHandler() func signature takes this extra hc param, so it doesn't have the same exact signature as the MakeDocumentMatchHandler func type.

That means, I think, these things are not chainable or wrappable.  Like, my application should be able to have its own handler, which might does some extra logging or stats or streaming or something, but then should be able to pass along the hit to a top-N handler.  But, my app can't construct a top-N handler if it doesn't have the hc instance?

search/collector/topn.go
sreekanth-cb: hm, thinking was that IndexInternalID doesn't make sense for applications. We can make it optional too..

search/collector/topn.go
```diff
@@ -154,6 +154,23 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
...
+	} else if dmHandlerMaker, ok = cv.(search.MakeDocumentMatchHandler); !ok {
+		// on issues with the given doc match handler maker, 
+		// initialize the dmHandler with the default TopN handler
+		dmHandler =  NewTopNDocumentMatchHandler(searchContext, hc, nil)
```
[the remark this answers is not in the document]
sreekanth-cb: indeed!

search/collector/topn.go
sreekanth-cb: should the second return value for dmHandlerMaker () should be just the loadID bool or a n options struct?

search/collector/topn.go
sreekanth-cb: done.

search/collector.go
```diff
@@ -30,3 +30,23 @@ type Collector interface {
...
+	"MakeDocumentMatchHandlerKey")
+
+// MakeDocumentMatchHandler is an optional DocumentMatchHandler 
+// builder function which the applications can pass to bleve.
+// These builder methods gives a DocumentMatchHandler function
+// to bleve, which it will invoke on every document matches in the index.
+// For better memory usage, the DocumentMatchHandler callback implementation
+// should put/give back the hit/DocumentMatch to the ctx.DocumentMatchPool
+// once it is done with the processing of each individual hit.
+type MakeDocumentMatchHandler  func(ctx *SearchContext) (callback DocumentMatchHandler, lookID bool)
```
steveyen: sort of a knee-jerk golanger reaction here... if there's no error in the return vals for stuff like this, it feels like it'll be a delayed future change where you'll be adding a error to the return vals, so might as well add it today :-)

search/collector.go
```diff
@@ -30,3 +30,24 @@ type Collector interface {
...
 }
+
+// DocumentMatchHandler is the type of document match callback
+// bleve will invoke during the search.
+// Bleve will indicate the completion of ongoing search, by passing
+// nil value for the document match callback eventually.
+// It is recommended that the application takes a copy of the
+// hit/documentMatch if they wish to own it or need prolonged
+// access to it.
+type DocumentMatchHandler func(hit *DocumentMatch) error
```
mschoch: Should document what happens when the handler returns an error (looks like it will stop the search early, which would be useful in some cases)

steveyen:
Update on this -- I should've taken notes from the recent discussion with @mschoch and @sreekanth-cb on this, so this is a bit from (poor, post-weekend) memory...

Marty shared feedback that the factory callback approach had the feel of internal implementation details bubbling upstairs (it's due to how bleve's MultiSearch leverages goroutines underneath-the-hood) and that felt like a complexity to application level dev awareness -- where it'd be more ideal if that wasn't the case.  But, we didn't have a better idea at the time or midst of the conversation.

Also, the thought was that app dev's might want to have a "search that sees all the document matches" as a pretty common case, so the feedback was that hiding the callback API (or callback maker) as a proposed, optional key in the Context didn't feel right if it's going to be a common situation.

Instead, a new top-level "advanced" search method felt better, where the existing SearchInContext() ought to turn into a thin wrapper method that invokes the new, advanced method with default behavior.

One discussed proposal on the new, advanced method was something like SearchInContextWithCallback() -- (I don't remember exactly what I proposed) -- but the observation by Marty was that many API's in golang world are moving towards Context-friendliness or assuming a Context approach, so there might not be any need to mention Context in the method name.

So, something like the following was what popped to mind (maybe it was only in my mind)...
`
SearchWithHandler(context.Context, *SearchRequest, MakeDocumentMatchHandler) (
    *SearchResult, error)
`
...where we should document why it's a factory/maker rather than a DocumentMatchHandler.

mschoch:
So, I don't quite get what adding the SearchProcessor gets us.  In our conversation last week I had proposed having some new interface, so maybe that is where this came from.  But my idea was that it'd be great to have both the "new thing for each shard" and the "process hits" on the same interface.  I had no idea what that would look like or if it was even possible.  But, we can agree this new interface doesn't accomplish that.

I think the current design is basically doomed.  Because of the way MultiSearch uses goroutines, there is simply no obvious way for this to work without users understanding the internals.  As ugly as the name is, we should probably just go back to MakeMakeDocumentMatchHander, because the one thing that has going for it, is that it does what it says.  And it also makes it really obvious that we're doing something weird and developers should pay attention.

If it sounds like I've come full circle on this it's because I have.  I still have the exact same concerns as when we started (basically no one would expect it to work this way as a user).  But, unless we make major changes to the internals, there is no other choice.

mschoch:
OK, here is another idea (for discussion, please don't rush to implement)

```
type DocumentMatchHandler interface {
  Handle(hit *DocumentMatch) error
  Copy() DocumentMatchHandler
}
```

Next, we add an optional DocumentMatchHandler to the SearchRequest struct.

Finally, we update the createChildRequest method, to copy the DocumentMatchHandler (using Copy()) just like we do for the Sort.  See: https://github.com/blevesearch/bleve/blob/master/index_alias_impl.go#L425-L438

In the copy method, one can ensure that each newly created DocumentMatchHandler gets its own buffer, to avoid any data races.

Would this work?

sreekanth-cb:
not sure whether I am following this.
As I understood, DocumentMatchHandler interface still needs to be implemented by the application.

So,  how does the Copy() method makes it more user intuitive than a handler builder approach?

mschoch:
It is subjective, and I want to be clear that I don't think `Copy()` is the right design either.  The connection to me is that we've already encountered a similar problem with MultiSearch in the past.  Specifically, parts of the request were being shared between goroutines incorrectly.  The fix we chose at the time was to simply create copies of the things.

It seemed to me that I could view this problem as the same, and apply the same solution.  So, one thing this has going for it, is that it matches how we already address something similar.

Regarding the notion of "more user intuitive" I believe that many beginners are a bit afraid of higher-order functions.  That doesn't mean we can't use them, but I am cautious to make them part of the public API.  This is subjective, and I suspect there won't be agreement on this.

sreekanth-cb:
We all agree at least on one point that, neither of these options are user intuitive. And don't have a cleaner solution popping to mind at the moment.

Copy is analogous to the way we handle Sort, but then Copy feels like a lesser logical name as against a name like a Maker/Builder for a callback method?  

Response/Result callbacks being part of SearchRequest itself felt a bit lesser attractive, hence moved to the newer API version.

One thing, I liked about the SearchProcessor interface is that it keeps options open for future, for example one can pass a newer Collector implementation to the newer API to manipulate/perform search in what ever way they like.

May be, once we explore more into the custom scoring territory we may have a better way to refactor things and ditch this API/interface all together.. 
Does it make sense to add comments like "Experimental/Interim/Depreciable" to these new API / interface documentation?

May be closures aren't that foreign to programmers as it used to be(with JS popularity), 
May be we can split the interface into two,

```
type SearchProcessor interface {
MakeDocumentMatchHandler(ctx *SearchContext) (
			impl DocumentMatchHandler, loadID bool, err error)
}

instead of being a function, DocumentMatchHandler is an interface

type DocumentMatchHandler interface {
       Handle(hit *DocumentMatch) error
}
```

mschoch:
I don't see the problem with the name Copy.  It does exactly what it says, it creates a copy of a DocumentMatchHandler.  It is called internally by the search implementation once for each shard the request is being forwarded to.

On the topic of 'Response/Result callbacks being part of SearchRequest itself felt a bit lesser attractive', can you explain why?  To me the SearchRequest holds all the information about how we want the search results delivered to us.  And to me, a callback which is going to be fired for each hit, seems like something I would configure as a part of the request.  How do you see it?

On, the topic of 'One thing, I liked about the SearchProcessor interface is that it keeps options open for future'.  I'm kind of against this, as it seems like we're adding a level of indirection with no concrete plans for how it will be used.  I don't understand the comment about a new collector implementation being able to do whatever it wants, that seems to be the case no matter what.

Given that we're so far apart on this, I'm going to suggest that we go back to the original proposal from you, and we just hide this capability behind the context and implement it as a Couchbase hack.  I had thought we could expose this to users as a more general "result streaming" solution, but it doesn't seem like we have consensus on how that should work.  So, perhaps we just abandon that for now.

search/collector.go
sreekanth-cb: done!

sreekanth-cb:
@mschoch , @steveyen , brought back the ctx based changes. ready for another round.

PR #1116, opened 2019-01-30
MB-32855: Propagate min setting when optimizing a single clause disjunction

abhinavdangeti:


search_test.go
```diff
@@ -1083,3 +1083,66 @@ func TestDisjunctionQueryIncorrectMin(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if err = idx.Batch(batch); err != nil {
+		t.Fatal(err)
+	}
+
+	mq1 := NewMatchQuery("dev-ops")
+	mq1.SetField("dept")
+	mq2 := NewMatchQuery("keelia@domain.com")
```
mschoch: i think this is a bad example, the match query will match both documents, because the way email addresses are analyzed.  you and i know that, but a user reading this test might be confused, or draw incorrect conclusions about why it really works.  it should be simple to change the data/query to avoid this confusion.

abhinavdangeti: Fair enough, let me use a better example :)

abhinavdangeti:
Yes the test case fails with the previous optimization.

steveyen:
looks like the associated test example was updated / simplified.  it looks good to me.  thinking @mschoch's weigh in on this, though, would be helpful in case @abhinavdangeti and i are not seeing something about the test/example or code approach

mschoch:
i just dont get it, this propogates down not up.  are you suggesting we have to propogate the min down, so that when the optimization pushes it up its right?

that seems to work in this case but be broken more generally.  like if i have disjunctions of disjunctions that are not part of a boolean searcher, then pushing the min values down the tree is just wrong.

mschoch:
ok, i at least now see how it works.

i think in hindsight the mistake was ever allowing min of 0 on a disjunction query.  it just doesn't make sense there and only really makes sense in the higher-level construct of the boolean searcher's should clause...

abhinavdangeti:
Yes - to pushing the min down, when the optimization pushes the child searcher up. 
I think we agree that this change would only affect the situation when the child searcher is a disjunction.

Note that this is the case when only min <= 1. And this min propagation (especially when zero) doesn't affect the case of disjunction of disjunction, with the checks in place within Next(). (If I'm overseeing something there let me know.). It's just to accommodate the boolean should case.

abhinavdangeti:
Missed your last comment :)

PR #1121, opened 2019-02-01
fix data race when indexing empty batches and using reset

mschoch:
When the batches passed to bleve are empty, this was a special
case that inadvertently avoided proper synchronization, and
led to the race detector to correctly point out that resetting
the batch was unsafe.  By simply avoiding an unnecessary step
when the batch is empty, we can avoid the problem.


steveyen:
very cool -- good find & fix!  looks good

PR #1123, opened 2019-02-01
enable race detector on travis

mschoch:
now that race detector passes for whole test suite, lets keep it that way


steveyen:
looks good 2 me

PR #1018, opened 2018-10-13
WIP: Switch to use bbolt

techknowlogick:
Fix #783

PR #1134, opened 2019-02-14
Techknowlogick switch to bbolt

mschoch:
Replace #1018 since I couldn't figure out how to add commits to it...


mschoch:
Also, recommend we use the squash and merge option, as the commit history has some missteps along the way that won't be helpful.

PR #1136, opened 2019-02-14
Re-enabling conjunction single clause optimization conditionally

abhinavdangeti:
If the conjunction query had a single clause, and this single clause
were to be a match-none searcher, do not apply the optimization
of returning the single searcher itself. This is because if the
conjunction query were a boolean query's must, then when the boolean
searcher is being initialized, it's must searcher would be reset to
nil - there by dropping what could have played a relevant role.

### DIFF
--- a/search/query/conjunction.go
+++ b/search/query/conjunction.go
@@ -73,6 +73,16 @@ func (q *ConjunctionQuery) Searcher(i index.IndexReader, m mapping.IndexMapping,
 
 	if len(ss) < 1 {
 		return searcher.NewMatchNoneSearcher(i)
+	} else if len(ss) == 1 {
+		// If the single searcher within this conjunction query happens to
+		// be a match none, do not apply this optimization - as in the case
+		// where this conjunction query is a boolean query's must, when the
+		// boolean searcher is being initialized, it's must searcher would
+		// be reset to nil - there by dropping what could have played a
+		// relevant role.
+		if _, ok := ss[0].(*searcher.MatchNoneSearcher); !ok {
+			return ss[0], nil
+		}
 	}
 
 	return searcher.NewConjunctionSearcher(i, ss, options)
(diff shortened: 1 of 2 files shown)


abhinavdangeti:
Not a very big fan of this change here, but I thought I'd push it up so I can get another opinion or two.

steveyen:
hmm, I'm a bit worried, from my past history on this, that I don't fully understand this area now (if ever!) well enough.  That is, the last few PR's here all seemed completely plausible to me, and they ended up being reverted.  This PR also seems pretty plausible to me too -- but I'm kinda feeling that I've seen this movie (and the followup movie sequels) before.  Clutching my $15 a bit more closely.

Apologies, as not sure how to express this in a way to give you more actionable feedback -- e.g., "please address X, Y and Z and then we ought to be good"

mschoch:
Please see this commit, to gain some understanding of what is going on...

https://github.com/blevesearch/bleve/commit/f391b991c20f02681bacd197afc6d8aed444e132

The issue had to do with compatibility with Elasticsearch for query strings (but ES showed different behavior when you manually built a boolean query to do the same thing).  The issue we observed was that these 2 queries would behave differently:

+stopword

+wordnotinthedictionary

You might expect these to behave the same, since in both cases you said MUST for something that isn't there.  But, what we observed was that +stopword is handled differently.  I guess that thinking was that +stopword means you asked for something logically not possible, whereas +wordnotinthedictionary is possible, but just happens to not be used.

My concern reading this change, is that it seems very similar, but operates on boolean queries regardless of if they originated as a query string, and my recollection was that we saw that behave differently.

So, I'd say I'm -1 on this for now until someone can clear all this up.

steveyen:
Thanks for the pointers & backstory

> My concern reading this change, is that it seems very similar, but operates on boolean queries regardless of if they originated as a query string

Indeed -- after pointing that out, I now can't "un-see" that observation too!

PR #1138, opened 2019-02-19
dict cmd additions

sreekanth-cb:
-exposing 1hit count and total term info
-exposing postings size and cardinality (should this be a separate cmd)

### DIFF
--- a/cmd/bleve/cmd/zap/dict.go
+++ b/cmd/bleve/cmd/zap/dict.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"math"
 
+	"github.com/RoaringBitmap/roaring"
 	"github.com/blevesearch/bleve/index/scorch/segment/zap"
 	"github.com/couchbase/vellum"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


mschoch:
The original idea was that the `dict` command just focused on the dictionary.  But, once we added 1-hit encodings, it was useful to identify them here (which @steveyen added a while back).  If users want to go beyond the dictionary into the postings, the explore command is useful.  But, I still think this is OK as the info you're adding is more summary in nature, about the postings, not about a specific posting.

One rule (I think I learned from @steveyen but don't follow myself often enough) is that it's a good idea to think of the output as eventually being consumed by another utility.  This doesn't mean that you always have to go to fully structured output like JSON, but something like "every line is an entry in the dictionary" (NOTE: not what this tool does today) would make it a lot easier for someone to use sed/awk/uniq/sort/etc to do interesting stuff downstream.  I'm venturing off-topic, but that would be one way to improve the tools long term in my opinion (make things like the hexdump of the dictionary go away or be hidden behind more options, maybe make extracting the dictionary in binary form to a .vellum file a separate tool and make the output of this simpler)

sreekanth-cb:
Thanks @mschoch , I shall keep these points in mind on next iteration on these cmds.

PR #1141, opened 2019-02-22
LoadAndHighlightFields API

sreekanth-cb:
Refactoring the code a bit to let
applications perform the field loading
and highlighting. Mainly helpful
with the streamed result hits.

### DIFF
--- a/index_impl.go
+++ b/index_impl.go
@@ -542,70 +542,12 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 	}
 
 	for _, hit := range hits {
-		if len(req.Fields) > 0 || highlighter != nil {
-			doc, err := indexReader.Document(hit.ID)
-			if err == nil && doc != nil {
-				if len(req.Fields) > 0 {
-					fieldsToLoad := deDuplicate(req.Fields)
-					for _, f := range fieldsToLoad {
-						for _, docF := range doc.Fields {
-							if f == "*" || docF.Name() == f {
-								var value interface{}
-								switch docF := docF.(type) {
-								case *document.TextField:
-									value = string(docF.Value())
-								case *document.NumericField:
-									num, err := docF.Number()
-									if err == nil {
-										value = num
-									}
-								case *document.DateTimeField:
-									datetime, err := docF.DateTime()
-									if err == nil {
-										value = datetime.Format(time.RFC3339)
-									}
-								case *document.BooleanField:
-									boolean, err := docF.Boolean()
-									if err == nil {
-										value = boolean
-									}
-								case *document.GeoPointField:
-									lon, err := docF.Lon()
-									if err == nil {
-										lat, err := docF.Lat()
-										if err == nil {
-											value = []float64{lon, lat}
-										}
-									}
-								}
-								if value != nil {
-									hit.AddFieldValue(docF.Name(), value)
-								}
-							}
-						}
-					}
-				}
-				if highlighter != nil {
-					highlightFields := req.Highlight.Fields
-					if highlightFields == nil {
-						// add all fields with matches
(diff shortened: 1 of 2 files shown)


index_impl.go
```diff
@@ -632,6 +574,75 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 	}, nil
 }
 
+func LoadAndHighLightFields(hit *search.DocumentMatch, req *SearchRequest,
```
steveyen: minor naming thing -- I see existing code calls it "Highlighter", instead of "HighLighter" -- english is weird.

sreekanth-cb: done!

steveyen:
looks ok to me.

one thing, perhaps for the future, is i'm wondering whether the loading of field and the highlighting are or can be orthogonal.

search/search.go
```diff
@@ -279,6 +279,7 @@ type SearcherOptions struct {
 type SearchContext struct {
 	DocumentMatchPool *DocumentMatchPool
 	Collector         Collector
+	IndexReader       index.IndexReader
```
mschoch: wondering why this is needed?  it seems like the method you refactored passes the IndexReader as an argument, and I don't see search context used elsewhere in this change

sreekanth-cb: The thinking is that, the "MakeDocumentMatchHandler"  ref: "https://github.com/blevesearch/bleve/blob/master/search/collector.go#L51 here takes the SearchContext as an argument. And when the application gets this builder callback, then it can use this SearchContext information to invoke these LoadFieldsAndHighlight()/similar methods in bleve.

sreekanth-cb: For eg: the Collector inside the context can be used by the application to get a Collector chaining effect. Like, the application can process the hit on https://github.com/blevesearch/bleve/blob/master/search/collector.go#L40 callback and later pass it back to the topN collector handler like here: https://github.com/blevesearch/bleve/blob/master/search/collector/topn_test.go#L588

mschoch: So, I'm not completely sure I follow your scenario yet, but my understanding from the commit message was that this was just a refactor of one really long method (SearchInContext) into two slightly less long methods.  That was an obvious good improvement.  However, this second change seems unrelated (unless I still haven't gotten the connection).

The reason it seemed confusing to me was that generally, everyone that needs the IndexReader, is directly passed a copy of it today.  The SearchContext was added later, and my thinking was that everyone that got the SearchContext already got the IndexReader (usually as a separate argument).

Regarding your scenario you describe, are you suggesting that loading stored fields and highlighting should be done inside a collector (or collector chain) and not done outside it?

sreekanth-cb:
-Regarding your scenario you describe, are you suggesting that loading stored fields and highlighting should be done inside a collector (or collector chain) and not done outside it?

In a hit streaming scenario, the collector is orchestrating the application's document match builders and handlers. And hence it is appending the IndexReader to the SearchContext so that, the application handlers can load the stored fields by calling this new refactored method upon such requirements in the application logic.

mschoch:
Got it, so in that I just think it would have been more clear if the commit message explained the "why" a little better.  I had assumed the motivation for the refactor was simply to make it clearer, but you had a very concrete purpose in mind that wasn't explained fully in the commit message.

Now that I understand the purpose, I would just point that to me it adds some element of confusion.  Searchers are given an IndexReader in their New method, but each call to Next/Advance provides a SearchContext, which happens to contain an IndexReader (which happens to always be the same now, but which it would never use even if it was different).

I'm not suggesting this is terrible, or the end of the world.  Rather it looks like we're just stuffing things where they happen to be convenient, even if it doesn't seem to make sense in the bigger picture any more.

sreekanth-cb:
Yes, Agree that the intention wasn't made clear in the commit message. Since I had this hacky feel for these streaming related changes, thought of not highlighting that part.

PR #1143, opened 2019-02-26
fix postings list count implementation

### LINKED ISSUE #1140: PostingList API Count() Error in Calculation
https://github.com/blevesearch/bleve/blob/cb59b33c8bb0e24675ce5f079307e67e11a0f72c/index/scorch/segment/zap/posting.go#L265-L281

When except is not nil, count = n-e, but the except include all delete document internalID, and the postingList bitmap only has docs which has the term, I'm trying to give you my way of doing it.

```go
func (p *PostingsList) Count() uint64 {
  var n, e uint64
  if p.normBits1Hit != 0 {
    n = 1
  } else if p.postings != nil {
    n = p.postings.GetCardinality()
    if p.except != nil {
      e = p.postings.AndCardinality(p.except)
    }
  }
  //var e uint64
  //if p.except != nil {
  //  e = p.except.GetCardinality()
  //}
  //if n <= e {
  //  return 0
  //}
  return n - e
}
```

mschoch:
previously the calculation subtracted the total number of
deleted documents, without regard for how the deleted documents
set might overlap with the postings list, this was incorrect.

fixes #1140

### DIFF
--- a/index/scorch/segment/zap/posting.go
+++ b/index/scorch/segment/zap/posting.go
@@ -264,18 +264,17 @@ func (p *PostingsList) iterator(includeFreq, includeNorm, includeLocs bool,
 
 // Count returns the number of items on this postings list
 func (p *PostingsList) Count() uint64 {
-	var n uint64
+	var n, e uint64
 	if p.normBits1Hit != 0 {
 		n = 1
+		if p.except != nil && p.except.Contains(uint32(p.docNum1Hit)) {
+			e = 1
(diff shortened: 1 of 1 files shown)


mschoch:
actually, think we need to check to return 0 rather than a negative, let me add that back

mschoch:
actually no, AndCardinatlity couldn't exceed `n`, so i think we're good, anyone confirm?

PR #1153, opened 2019-03-06
fix issue mapping documents containing *time.Time

mschoch:
If documents contained a field of type *time.Time AND they also
had an explicit mapping for the field of a type OTHER THAN text,
the field would not be mapped.

The issue was introduced when we added special support for
handling encoding.TextMarshaler because *time.Time satisfies
this interface.

This change fixes this issue, but there are some more complicated
cases exposed by this, specifically related to mapping the same
field multiple times, and handling cases where you try to mix
multiple types.

Instead, our solution is to further limit the way we support
using ecoding.TextMarshaler.  The new limitation is that we ONLY
use the output of MarshalText() if you have an explicit mapping
AND ALL the index fields are of type text.  All other cases
will ignore TextMarshaler and follow regular indexing rules.

### DIFF
--- a/mapping/document.go
+++ b/mapping/document.go
@@ -533,12 +533,14 @@ func (dm *DocumentMapping) processProperty(property interface{}, path []string, 
 					for _, fieldMapping := range subDocMapping.Fields {
 						if fieldMapping.Type != "text" {
 							allFieldsText = false
+							break
 						}
 					}
 					txt, err := property.MarshalText()
 					if err == nil && allFieldsText {
+						txtStr := string(txt)
 						for _, fieldMapping := range subDocMapping.Fields {
-							fieldMapping.processString(string(txt), pathString, path, indexes, context)
+							fieldMapping.processString(txtStr, pathString, path, indexes, context)
 						}
 						return
 					}
--- a/mapping/document.go
+++ b/mapping/document.go
@@ -525,19 +525,25 @@ func (dm *DocumentMapping) processProperty(property interface{}, path []string, 
 		if !propertyValue.IsNil() {
 			switch property := property.(type) {
 			case encoding.TextMarshaler:
-
-				txt, err := property.MarshalText()
-				if err == nil && subDocMapping != nil {
-					// index by explicit mapping
+				// ONLY process TextMarshaler if there is an explicit mapping
+				// AND all of the fiels are of type text
+				// OTHERWISE process field without TextMarshaler
+				if subDocMapping != nil {
+					allFieldsText := true
 					for _, fieldMapping := range subDocMapping.Fields {
-						if fieldMapping.Type == "text" {
-							fieldMapping.processString(string(txt), pathString, path, indexes, context)
+						if fieldMapping.Type != "text" {
+							allFieldsText = false
 						}
 					}
-				} else {
-					dm.walkDocument(property, path, indexes, context)
+					txt, err := property.MarshalText()
+					if err == nil && allFieldsText {
+						for _, fieldMapping := range subDocMapping.Fields {
+							fieldMapping.processString(string(txt), pathString, path, indexes, context)
+						}
+						return
+					}
 				}
-
+				dm.walkDocument(property, path, indexes, context)
 			default:
 				dm.walkDocument(property, path, indexes, context)
 			}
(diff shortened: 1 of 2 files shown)


mschoch:
As noted in the commit comment, in addition to fixing the issue, this commit changes behavior for some other corner cases, please review carefully.

mschoch:
Here is a more detailed explanation.

**What is the problem?**

If we get here (field value is pointer, pointer satisfies interface encoding.TextMarshaler, field has a mapping):

https://github.com/blevesearch/bleve/blob/master/mapping/document.go#L535

... we iterate over the field mappings (we support multiple) and if any of them are also of type `text` then we index the output of MarshalText().  But there is no else case, so if the type wasn't text, this field/mapping gets ignored completely.

If you didn't have an explicit mapping you were safe.  (and this was on purpose, we felt that we should only attempt using TextMarshaler if someone added an explicit mapping for the field)

**So, why isn't the fix just adding an else case?**

Well, initially that was what I thought too.  I made an else block that called:

```
dm.walkDocument(property, path, indexes, context)
```

Just like all the other else cases.  This will go back to the walkDocument function (the recursion is split into two methods, walkDocument tries to just navigate structure, and processProperty tries to cope with values, although they call each other as the structure unfolds).  The walkDocument function will traverse inside the pointer this time (as opposed to processing the pointer directly the last time through).

**So, what is the problem with this?**

Well, the simplest problem I could think of was if I had mapped it as, say a DateTime field, two times.  The else case would fire twice (once for each mapping), but because we're calling walkDocument which is still walking the source document, it means once it finally gets to the value, it will look at the mappings and create both fields, but it would be called twice.  So we'd end up with 4 fields, 2 of which duplicate names.  I don't even know what would happen beyond this point, but it seems wrong...

Instead of calling walkDocuemnt, maybe we could process the field right now.  But, this ends up having to duplicate a lot of the logic from earlier in the function, depending on what kind of type you want, etc...  The intended handling of pointers was to recurse through one more time.

**OK, by why did you rewrite this whole section, things seem to be done in a different order now?**

The order in which we check things has changed as well, and this was because I considered another corner case.

What would happen if someone had a field that implemented TextMarshaler, and they had one field mapped as text and another mapping of the same field as some other type?  Then there is a new question, should both field mappings used the output of MarshalText() or perhaps the other field was intended to work with the original value, not the text version.  Think this seems contrived?  I think with *time.Time you could actually imagine this making sense.

For now I've basically said let's not allow this.  If you want to use TextMarshaler, thats's fine, you can even have multiple fields, but they all must be text (could still be analyze differently).

If you don't have an explicit mapping, or if you map the field to anything other than text, we completely ignore that it happened to satisfy TextMarshaler, and the value is processed directly.

**Summary**

Based on everything above, the code in this section was rewritten.  First we check the very specify case in which the pointer satisfies TextMarshaler, the field has an explicit mapping, and ALL field mappings are type text.  In that case, we do what we did before, process each field as text with the result of MarshalText(), then return.

All other cases fall through and simply go back to walkDocument, which will eventually process the contents of the pointer, not the pointer itself.

mapping/document.go
```diff
@@ -525,19 +525,25 @@ func (dm *DocumentMapping) processProperty(property interface{}, path []string,
...
-					// index by explicit mapping
+				// ONLY process TextMarshaler if there is an explicit mapping
+				// AND all of the fiels are of type text
+				// OTHERWISE process field without TextMarshaler
+				if subDocMapping != nil {
+					allFieldsText := true
 					for _, fieldMapping := range subDocMapping.Fields {
-						if fieldMapping.Type == "text" {
+						if fieldMapping.Type != "text" {
+							allFieldsText = false
```
steveyen: minor optimization thought -- a break out of the loop here ought to be ok?

mapping/document.go
```diff
@@ -525,19 +525,25 @@ func (dm *DocumentMapping) processProperty(property interface{}, path []string,
...
+						}
+					}
+					txt, err := property.MarshalText()
+					if err == nil && allFieldsText {
```
steveyen: another non-important thought (as it was in the original code, too)... optimization where conversion to string(txt) can happen just once outside the loop

mapping/document.go
mschoch: done

mapping/document.go
mschoch: done

PR #1155, opened 2019-03-11
extract multisearch strategy into interface

mschoch:
this change would allow an IndexAlias or MultiSearch user
to control the amount and type of concurrency used while
executing searches across several indexes

this change also introduces a new pooled implementation which
limits a number of searches executed at one time

In one of my tests (doing MatchAll across 100 local bleve indexes)
I observe the following with default behavior today:

```
$ /usr/bin/time -v ./sirmemsalot -count 5
	Command being timed: "./sirmemsalot -count 5"
	User time (seconds): 11.56
	System time (seconds): 0.69
	Percent of CPU this job got: 287%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.27
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 860312
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 7769
	Minor (reclaiming a frame) page faults: 101423
	Voluntary context switches: 11912
	Involuntary context switches: 1579
	Swaps: 0
	File system inputs: 1341640
	File system outputs: 4000
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

And hten when I use the new pooled implementation with the pool
size capped at 4, I see:

```
$ /usr/bin/time -v ./sirmemsalot -count 5 -pool 4
2019/03/11 11:40:28 yeah pool is 4
	Command being timed: "./sirmemsalot -count 5 -pool 4"
	User time (seconds): 7.80
	System time (seconds): 0.34
	Percent of CPU this job got: 280%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:02.89
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 820284
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 65
	Minor (reclaiming a frame) page faults: 103393
	Voluntary context switches: 3937
	Involuntary context switches: 666
	Swaps: 0
	File system inputs: 1800
	File system outputs: 4000
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

### DIFF
--- a/index_alias_impl.go
+++ b/index_alias_impl.go
@@ -28,19 +28,27 @@ import (
 )
 
 type indexAliasImpl struct {
-	name    string
-	indexes []Index
-	mutex   sync.RWMutex
-	open    bool
+	name     string
+	indexes  []Index
+	mutex    sync.RWMutex
+	open     bool
+	strategy MultiSearchExecutionStrategy
 }
 
 // NewIndexAlias creates a new IndexAlias over the provided
 // Index objects.
 func NewIndexAlias(indexes ...Index) *indexAliasImpl {
+	return NewIndexAliasWithStrategy(nil, indexes...)
+}
+
+// NewIndexAlias creates a new IndexAlias over the provided
+// Index objects.
+func NewIndexAliasWithStrategy(strategy MultiSearchExecutionStrategy, indexes ...Index) *indexAliasImpl {
 	return &indexAliasImpl{
-		name:    "alias",
-		indexes: indexes,
-		open:    true,
+		name:     "alias",
+		indexes:  indexes,
+		open:     true,
+		strategy: strategy,
 	}
 }
 
@@ -159,7 +167,7 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
 		return i.indexes[0].SearchInContext(ctx, req)
 	}
 
-	return MultiSearch(ctx, req, i.indexes...)
+	return MultiSearchWithStrategy(ctx, req, i.strategy, i.indexes...)
 }
 
 func (i *indexAliasImpl) Fields() ([]string, error) {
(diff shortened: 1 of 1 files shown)


mschoch:
This commit is mainly to start the conversation.  Our goal was to reduce peak memory use, and this seems to only slightly improve that metric.  However, it did run considerably faster, so it may be beneficial anyway.

Also, I'm open to coming up with better ways to measure this.  Currently this was done by creating index alias over 100 copies of a beer-search index.  Then running MatchAll with size=math.MaxInt64.  All of the indexes have the same IDs, so the result merging behavior may not be typical, but that seemed outside the scope of the original measurement (focusing on memory usage).

mschoch:
OK, I think I have to walk back some of the original data I posted, and some of the conclusions that were drawn from them.  The major page faults were NOT related to the multi-search strategy, instead they indicate that during that run none of the indexes were in memory yet.  Even using the current default multi-search strategy, when run subsequently it is nearly as fast and also has 0 major page faults.

Here is some more complete data, including multiple runs for each setting, and throwing out the initial (cache-loading) runs.

sirmemsalot-conc-1

And, I added the ability to have multiple clients (goroutines) running their searches concurrently.  I tested with a modest setting of 4:

sirmemsalot-conc-4

It seems there are very few statistically significant variations...

sreekanth-cb:
thanks @mschoch for the detailed analysis. I tried the logic discussed above with CBFT (concurrent remote & controlled concurrent local partitions) and number variations were barely noticeable.

tylerkovacs:
Even without performance gains, this change would be extremely useful for me since it provides a mechanism to constrain query concurrency. Here's a little bit more detail about my use case:

I build my index on one machine and deploy (by copying the index) to other machines. The machine that builds the index has many more cores than the machines that handle queries. To make index creation as fast as possible, I build many index shards in parallel (where shard count == # of cores on the index build machine).

But since the machines handling queries have less cores, this means that each query maxes out all available cores (since, on master, bleve spawns a goroutine to query each shard in the `bleve.IndexAlias`). There are other apps running on my query machines, so it would be nice to be able to constrain concurrency on each query. `PooledMultiSearchExecutionStrategy.MaxConcurrent` provides an elegant hook to do so.

PR #1159, opened 2019-03-13
run integration tests with scorch

mschoch:
some minor changes have been made to a few of the queries to
ensure that the search results have a deterministic order

one test was removed as it was specifically checking the order
when no sort was specified, which is defined in a way that
does not lend itself to easy testing

we did not change all tests without sort order, as many of them
are correctly ordered by score already


mschoch:
An alternative would be force using `"sort": ["_score", "_id"],` everywhere, but I went for the more minimal change.

PR #1161, opened 2019-03-15
change sort strategy for integration tests

mschoch:
any test which is not explicitly setting a sort order
should instead specify:

"sort": ["-_score", "_id"]

this starts with the natural order of score descending,
but also imposes an additional tiebreaking rule to sort
by the documents identifier.  this comes with some
additional cost, but makes results more stable for
testing purposes.

### DIFF
--- a/test/tests/basic/searches.json
+++ b/test/tests/basic/searches.json
@@ -1,8 +1,10 @@
 [
 	{
+		"comment": "test term search, exact match",
 		"search": {
 			"from": 0,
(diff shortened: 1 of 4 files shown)


mschoch:
also, i started adding test comments describing what the tests were doing, this was a feature added later, which is why earlier tests did not have comments.  the comment is very useful when a test fails, because we print the comment, and you can search for it, as opposed to counting tests in the JSON to find the right one by using it's index.

PR #1174, opened 2019-03-25
Revert "MB-32846 - more aggressively removeOldData() in scorch persis…

sreekanth-cb:
…ter"

This reverts commit c8e737a945472f5f266c0799005216080443fbb8.

This recent introduction of removeOldData in
the persister main loop has slowed the persister
work flow to result in higher number of memory
based segments. This high number of in-memory
segments had resulted in perf regression across a few
query types like fuzzy, wildcard etc.

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -376,6 +376,7 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 				fileSegments++
 			}
 		}
+
 	}
 
 	// before the newMerge introduction, need to clean the newly
@@ -392,7 +393,6 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 			}
 		}
 	}
-
 	// In case where all the docs in the newly merged segment getting
 	// deleted by the time we reach here, can skip the introduction.
 	if nextMerge.new != nil &&
@@ -424,7 +424,6 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 	newSnapshot.AddRef() // 1 ref for the nextMerge.notify response
 
 	newSnapshot.updateSize()
-
 	s.rootLock.Lock()
 	// swap in new index snapshot
 	newSnapshot.epoch = s.nextSnapshotEpoch
@@ -502,7 +501,6 @@ func (s *Scorch) revertToSnapshot(revertTo *snapshotReversion) error {
 	}
 
 	newSnapshot.updateSize()
-
 	// swap in new snapshot
 	rootPrev := s.root
 	s.root = newSnapshot
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -111,7 +111,6 @@ OUTER:
 		if ew != nil && ew.epoch > lastMergedEpoch {
 			lastMergedEpoch = ew.epoch
 		}
-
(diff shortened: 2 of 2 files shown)


mschoch:
I guess this is where satisfying multiple stakeholders with bleve will be hard...

One could argue (as Couchbase did initially) that too many old (unnecessary) files were on disk.  The original commit addressed that.

However, that results in some slower runtime observations, which one could argue is a bug (as Couchbase is now).  This commit reverts that.

My gut reaction is to think that the number of files (and corresponding disk space wastage) is a more serious issue, and thus as a project we'd just have to eat the perf regression for now.  However, being asked to merge this suggests that Couchbase feels differently.

sreekanth-cb:
With 1M use case (more production similar than bleve bench), the number of files remain steady / un affected with this change.(Or the count of memory and/or file segments are both kept at a better range with this fix.)
http://cbmonitor.sc.couchbase.com/reports/html/?snapshot=fts_650-2715_run_test_eee7&snapshot=fts_650-19192019_run_test_0534

Its been tested and verified with many of the showfast results we have.

As it was mentioned in the original commit description,
removeOldData won't be skipped or too many files won't hit the disk due to the walk together /step at a distance logic between persister and the merger.  This original fix from @steveyen  is more of an experimental fix to target the index size bloating issue with the highly random low cardinality data set (auto generated) from Alex , for which we later identified the actual root cause as empty int coder metadata footprints and the space occupied by the sparse hits filled freq/norm/loc data.

mschoch:
@steveyen thanks for clarifying.  This then raised questions about the initial code review, so I went back to look closer at the commit, and you seemed to have evidence in the commit message about an actual improvement.  I'm still not clear if we're arguing that that was bad test in the first place, or if we think the real issue was fixed later in some other way, or both?

@sreekanth-cb I think the commit message should provide more explanation about why we think this is the right change.  On the surface it reads like, revert bug fix due to perf regression, users shouldn't have to click further to understand that isn't exactly the case.

steveyen:
> @steveyen thanks for clarifying. This then raised questions about the initial code review, so I went back to look closer at the commit, and you seemed to have evidence in the commit message about an actual improvement. I'm still not clear if we're arguing that that was bad test in the first place, or if we think the real issue was fixed later in some other way, or both?

Real issue (disk space usage can grow really large in some cases) isn't fixed yet, but I think we now know the root cause (the chunkMeta bytes from the intCoder & chunkedContentCoder's are way too big when the indexed terms have low cardinality.  Related: https://github.com/blevesearch/bleve/pull/1160  And, a fix for that means a scorch/zap file format change(s).)

The "bad test in the first place", I think, was that I overly relied on bleve-bench for that initial code review, along with self-blinding by my "beautiful-at-the-time" theorizing.

bleve-bench usually is an okay (imho) as a micro-benchmarking tool.  But, when Alex G. got his hands on the real build, it showed no improvements in his "showfast" perf tests w.r.t. disk usage.  Worse, it also showed a degradation in search performance.  @sreekanth-cb, to his credit, never gave up on this and was steadfast in his theory (and proof via larger 1M doc runs and experimental data!) that the extra removeOldData() call that I had added to the persister was to blame.  I would have lost $ if a bet had been placed on that!

sreekanth-cb:
@mschoch , its been consistently noted that my commit message isn't clear enough. I shall definitely try to improvise here.

PR #1175, opened 2019-03-25
Update to travis.yml to fix errcheck validation

abhinavdangeti:


abhinavdangeti:
Looks like the error is coming from go list ./...

Issue #1173, opened 2019-03-24
match phrase TermLocationMap got same locations


ansjsun:
![image](https://user-images.githubusercontent.com/1221947/54877891-b9592e80-4e5f-11e9-9501-c42251ac6952.png)

so , use findPhrasePaths function it return a large array 

![image](https://user-images.githubusercontent.com/1221947/54877927-0dfca980-4e60-11e9-82ac-d739569316ad.png)

mschoch:
If possible, please see if this PR addresses the issue: https://github.com/blevesearch/bleve/pull/1169

heidawei:
@mschoch  I try to bugfix is issue, the PR [#1177](https://github.com/blevesearch/bleve/pull/1177)

PR #1177, opened 2019-03-26
bugfix for duplicate term locations

heidawei:
bugfix for [issue #1173](https://github.com/blevesearch/bleve/issues/1173)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -105,6 +105,7 @@ import (
 	_ "github.com/blevesearch/bleve/index/store/goleveldb"
 	_ "github.com/blevesearch/bleve/index/store/gtreap"
 	_ "github.com/blevesearch/bleve/index/store/moss"
+	_ "github.com/blevesearch/bleve/index/store/rocksdb"
 
 	// index types
 	_ "github.com/blevesearch/bleve/index/upsidedown"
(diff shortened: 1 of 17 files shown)


mschoch:
I don't think adding a RocksDB adapater is an appropriate solution to the problem you've run in to.

The RocksDB wrapper was moved to the blevex repo due it reyling on CGO which is extremely difficult to support in practice.  We have no plans to bring it back, and in fact almost all our effort is currently on the new scorch indexing scheme.  That said, we are still interested in fixing bugs that affect the upsidedown index scheme.

We have recently merged a fix which should remove duplciate locations (however we've never seen as many locations in a response like you have shown).  If you could help us to try to reproduce the problem you're seeing, with the current master version of bleve we can try to help further.

heidawei:
I am sorry for add rocksdb, it is not for fix is issue. the test case is that: we test 384 documents, which has duplicate term in one field like is:"https://www.yy.com/yy/xx/yy/xx", for multi term search, the same term in diff searcher will has duplicate locations， I think the original intention of the design is that different terms correspond to different searchers，but it will has duplicate terms in a phrase.  than we use match phrase query, the array of locations will enlarge .

heidawei:
@mschoch  I will retry to submit new PR for this issue, I am sorry to rocksDB again

PR #1182, opened 2019-03-27
support search after style pagination

mschoch:
Typically deep pagination becomes quite expensive when using
the size+from mechanism on distributed indexes, due to each
child index having to build large result sets, and return
them over the network for merging at the coordinating
node.

Alternatively, the search after approach allows for the client
to maintain state, in this case, the sort key of the last
result for the current page.  By running the same search again,
but this time providing the last hit's sort key as this
new requests search after key, we allow for significantly
more efficient pagination.

TotalHits and Facets are unaffected by using this parameter,
because all hits are still seen by the collector.


mschoch:
I'd like to add a lot more tests, but wanted to get the concept up for early review/feedback.

search.go
```diff
@@ -311,6 +322,12 @@ func (r *SearchRequest) SortByCustom(order search.SortOrder) {
 	r.Sort = order
 }
 
+// SetSortAfter sets the request to skip over hits with a sort
+// value less than the provided sort after key
+func (r SearchRequest) SetSortAfter(after []string) {
```
steveyen: feels like this should be named SetSearchAfter?  not a strong feeling, but a naming-consistency tickle appeared in my mind

mschoch: oops that was typo, good catch

search.go
```diff
@@ -275,6 +276,7 @@ type SearchRequest struct {
 	Sort             search.SortOrder  `json:"sort"`
 	IncludeLocations bool              `json:"includeLocations"`
 	Score            string            `json:"score,omitempty"`
+	SearchAfter      []string          `json:"search_after"`
```
sreekanth-cb: snake_case -> search_after when all existing are camelCase?

steveyen: good_catch!

sreekanth-cb:
indeed...now it looks like very simple, but never thought it would be so.. :)

search.go
mschoch: Actually `snake_case` was what we intended to use for all JSON, and `camelCase` in the Go code.

Here are some longer-lived multi-word names:

https://github.com/blevesearch/bleve/blob/master/search.go#L447-L448
https://github.com/blevesearch/bleve/blob/master/search.go#L123-L124

I think we just missed the naming issue when we added `includeLocations` to the request.

mschoch:
@sreekanth-cb the implementation is simpler than I expected as well, but there are some limitations with this approach:

- No jump to page you haven't visited yet, just Next page, plus whichever page boundaries the client chooses to remember
- Not guaranteed to be stable (size+skip isn't stable either), but it means it still wouldn't be suitable for N1QL use cases
- Will need strong documentation to use correctly, perhaps even a reference implementation demonstrating it
-- specifically the sort order must be a total order (typically using `_id` as last sort will be sufficient)
-- more complexity on the client side to track pages

mschoch:
Additional feature request from @aviadl 

If in addition to `SearchAfter` we also supported `SearchBefore` then clients could access both Next and Prev pages without maintaining any state.

steveyen:
> If in addition to SearchAfter we also supported SearchBefore then clients could access both Next and Prev pages without maintaining any state.

That makes sense.  So, it moves towards an optional range that can be specified.

This SearchAfter change, btw, by itself, still looks good to me, even without the SearchBefore.

mschoch:
I've added an implementation for SearchBefore.

I still want to add tests to the alias versions of these, which isn't easily testable today.

Also, did someone else rename `coll` to `collector`?  I see that change in this PR now and I'm guessing someone else changed it and I reversed it when I merged master???

abhinavdangeti:
Hmm, it's always been collector, looks like your first commit here in this PR changes it to "coll" which I think is good, so there's no clash with "github.com/blevesearch/bleve/search/collector" that's imported in the file.

mschoch:
I've updated this again, the second to last commit refactors our integration tests in a way that allows us to now also create multiple indexes, have an alias span them, and then run the searches across the alias.  This is something that should be extended further, there are many more cases for aliases that have been under-tested in the past.

For now I did a single match-all query to verify that the alias even worked.  Then added one each for basic `search_before` and `search_after` functionality.

search.go
```diff
@@ -311,6 +336,12 @@ func (r *SearchRequest) SortByCustom(order search.SortOrder) {
 	r.Sort = order
 }
 
+// SetSearchAfter sets the request to skip over hits with a sort
+// value less than the provided sort after key
+func (r SearchRequest) SetSearchAfter(after []string) {
```
abhinavdangeti: func (r *SearchRequest)

tylerkovacs:
Thanks for this change! I've run into some problems trying to use it and I'm not sure if it is due to bugs or if I'm just using it incorrectly.

When setting req.SearchAfter, where does one get the sort key value? I've been using the `Sort` value from the last `search.DocumentMatchCollection` in the previous `SearchResult`. Using that value works in some of my tests but not others. I think this means that I'm using the correct value but there are bugs in some scenarios not covered by the tests (e.g., when sorting against multiple fields).

If you can confirm that I'm using it correctly, then I'll look into adding a test case that shows the problem.

mschoch:
@tylerkovacs Correct usage is to supply the sort key of the last item in a given page, to then access the next page.  The only other requirement is that for this to work correctly there cannot be any ties, the sort you use must impose a total order.  Usually, we would recommend adding "_id" to the end of the users sort to ensure there is a total order.

If you have an example that does not work, please open a new issue and include it.  Thanks.

PR #1189, opened 2019-04-02
MB-33600: Allow date ranges before epoch time

abhinavdangeti:
+ MinAllowedTime: 1677-12-01T00:00:00Z
    + MaxAllowedTime: 2262-04-11T11:59:59Z

    Any NonZero time.Time values that fall outside this range
    will cause UnixNano() to overflow (int64 cannot accommodate).

### DIFF
--- a/search/query/date_range.go
+++ b/search/query/date_range.go
@@ -41,12 +41,12 @@ type BleveQueryTime struct {
 	time.Time
 }
 
-var minValidTime time.Time
-var maxValidTime time.Time
+var MinRFC3339CompatibleTime time.Time
+var MaxRFC3339CompatibleTime time.Time
 
 func init() {
-	minValidTime, _ = time.Parse(time.RFC3339, "1677-12-01T00:00:00Z")
-	maxValidTime, _ = time.Parse(time.RFC3339, "2262-04-11T11:59:59Z")
+	MinRFC3339CompatibleTime, _ = time.Parse(time.RFC3339, "1677-12-01T00:00:00Z")
+	MaxRFC3339CompatibleTime, _ = time.Parse(time.RFC3339, "2262-04-11T11:59:59Z")
 }
 
 func queryTimeFromString(t string) (time.Time, error) {
@@ -151,7 +151,7 @@ func (q *DateRangeQuery) parseEndpoints() (*float64, *float64, error) {
 	min := math.Inf(-1)
 	max := math.Inf(1)
 	if !q.Start.IsZero() {
-		if q.Start.Before(minValidTime) || q.Start.After(maxValidTime) {
+		if !isDatetimeCompatible(q.Start) {
 			// overflow
 			return nil, nil, fmt.Errorf("invalid/unsupported date range, start: %v", q.Start)
 		}
@@ -159,7 +159,7 @@ func (q *DateRangeQuery) parseEndpoints() (*float64, *float64, error) {
 		min = numeric.Int64ToFloat64(startInt64)
 	}
 	if !q.End.IsZero() {
-		if q.End.Before(minValidTime) || q.End.After(maxValidTime) {
+		if !isDatetimeCompatible(q.End) {
 			// overflow
 			return nil, nil, fmt.Errorf("invalid/unsupported date range, end: %v", q.End)
 		}
@@ -179,4 +179,13 @@ func (q *DateRangeQuery) Validate() error {
 		return err
 	}
 	return nil
+}
+
+func isDatetimeCompatible(t BleveQueryTime) bool {
+	if QueryDateTimeFormat == time.RFC3339 &&
+		(t.Before(MinRFC3339CompatibleTime) || t.After(MaxRFC3339CompatibleTime)) {
+		return false
+	}
+
+	return true
 }
--- a/search/query/date_range.go
+++ b/search/query/date_range.go
@@ -41,6 +41,13 @@ type BleveQueryTime struct {
 	time.Time
 }
 
+var epochTime time.Time
+
+func init() {
+	unixEpochTime := "1970-01-01T00:00:00Z"
+	epochTime, _ = time.Parse(time.RFC3339, unixEpochTime)
+}
+
 func queryTimeFromString(t string) (time.Time, error) {
 	dateTimeParser, err := cache.DateTimeParserNamed(QueryDateTimeParser)
 	if err != nil {
@@ -144,7 +151,7 @@ func (q *DateRangeQuery) parseEndpoints() (*float64, *float64, error) {
 	max := math.Inf(1)
 	if !q.Start.IsZero() {
 		startInt64 := q.Start.UnixNano()
-		if startInt64 < 0 {
+		if startInt64 < 0 && q.Start.After(epochTime) {
 			// overflow
 			return nil, nil, fmt.Errorf("invalid/unsupported date range, start: %v", q.Start)
 		}
@@ -152,7 +159,7 @@ func (q *DateRangeQuery) parseEndpoints() (*float64, *float64, error) {
 	}
 	if !q.End.IsZero() {
 		endInt64 := q.End.UnixNano()
-		if endInt64 < 0 {
+		if endInt64 < 0 && q.End.After(epochTime) {
 			// overflow
 			return nil, nil, fmt.Errorf("invalid/unsupported date range, end: %v", q.End)
 		}
--- a/search/query/date_range.go
+++ b/search/query/date_range.go
@@ -41,11 +41,12 @@ type BleveQueryTime struct {
 	time.Time
 }
 
-var epochTime time.Time
+var minValidTime time.Time
+var maxValidTime time.Time
 
 func init() {
-	unixEpochTime := "1970-01-01T00:00:00Z"
-	epochTime, _ = time.Parse(time.RFC3339, unixEpochTime)
+	minValidTime, _ = time.Parse(time.RFC3339, "1677-12-01T00:00:00Z")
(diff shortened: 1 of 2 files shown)


search/query/date_range.go
```diff
@@ -144,15 +151,15 @@ func (q *DateRangeQuery) parseEndpoints() (*float64, *float64, error) {
 	max := math.Inf(1)
 	if !q.Start.IsZero() {
 		startInt64 := q.Start.UnixNano()
-		if startInt64 < 0 {
+		if startInt64 < 0 && q.Start.After(epochTime) {
```
sreekanth-cb: I think, this check only takes care of the end limit of year 2262, not the starting range year of 1678? What if any of the range values goes beyond that?

mschoch:
Curious on the thinking here.  

Is it just the realization that we're using an int64 representation, and thus currently wasting half the space we have available?  Or is that we did support them before, and the recent validation change accidentally broke that?

Either way it would seem like good info to include in the commit message.

mschoch:
Also, just noticed that this improved validation, while a good thing in general has implications for customers who worked-around this creatively.  Specifically, applications using `0001-01-01 00:01:00 +0000 UTC` as a start/end date end up broken by the original change.

mschoch:
Actually, I'd like to revisit the original change, unless you think this PR will fix it.

In Go it is accepted to use the zero-value of a time structure as a value with a particular meaning, in our case meaning an unspecified time boundary.  This is done because using pointers to time is generally discouraged, as the go docs state:

```
Programs using times should typically store and pass them as values, not pointers. That is, time variables and struct fields should be of type time.Time, not *time.Time.
```

But, unfortunately this creates an asymmetry then when you try and round-trip these values through JSON.  We detect it as a negative unix-nano value, and assume it wrapped.  When in fact the caller may simply be echoing back a value that Bleve gave them (due to the way a zero-value was formatted as part of a response)

mschoch:
Here is go playground to see this in action:

https://play.golang.org/p/HbmVOI6RqwP

mschoch:
OK, to be more constructive about what I think we should support.  I agree with Sreekanth, that we could do the same check for the lower bound.

But also, if the date parses and is recognized via the IsZero() method, we should treat it as not a validation error, and just as if the Go methods had been called with the zero-value of time.Time:

This snippet shows that this property should be identifiable after parsing:

https://play.golang.org/p/5AFDQLiRnyL

search/query/date_range.go
abhinavdangeti: Good point.

abhinavdangeti:
Yes @mschoch I think I agree with you. Let me put in another patch that captures the zero case and less than the year 1678 case. I'll add some test cases too, that checks for expected outcomes.

abhinavdangeti:
@mschoch I was just looking back into the case, looks like we handle the IsZero case somewhat correctly already.

- Both start and end cannot be zero, but one of them can be ..
https://github.com/blevesearch/bleve/blob/d8a53a56c039b7bb29fea965996e6829f4245232/search/query/date_range.go#L166

- Next I'll be doing the epoch check only if the time specified is non-zero ..
https://github.com/blevesearch/bleve/blob/d8a53a56c039b7bb29fea965996e6829f4245232/search/query/date_range.go#L145
https://github.com/blevesearch/bleve/blob/d8a53a56c039b7bb29fea965996e6829f4245232/search/query/date_range.go#L153

- I'll add some test cases that cover the above situations, and the 1678 check.

mschoch:
@abhinavdangeti on master right now we have tests that result in the following:

``
error="invalid/unsupported date range, start: 0001-01-01 00:01:00 +0000 UTC"
```

**Comment by abhinavdangeti** [comment]
Right, but that's not zero time. This is:

    0001-01-01 00:00:00 +0000 UTC

**Comment by mschoch** [comment]
Ah, then we're screwed.

**Comment by abhinavdangeti** [comment]
Let me push up another patchset here, and then take it from there :)

**Comment by mschoch** [comment]
OK, I guess I'll have to see about maintaining a branch without this validation, it's going to be too disruptive.

File: search/query/date_range.go [hunk 2]
```diff
@@ -41,6 +41,14 @@ type BleveQueryTime struct {
 	time.Time
 }
 
+var minValidTime time.Time
+var maxValidTime time.Time
+
+func init() {
+	minValidTime, _ = time.Parse(time.RFC3339, "1677-12-01T00:00:00Z")
+	maxValidTime, _ = time.Parse(time.RFC3339, "2262-04-11T11:59:59Z")
+}
```
mschoch: What do you think of this approach to finding the limits, it avoids parsing/errors and (I think) gets the exact end points we could support?

https://play.golang.org/p/Uz5m5i7CZZR

abhinavdangeti: Yea this looks good, but looks like the values that the playground code generates are a bit off. This is the output I see ...

```
epoch: 1970-01-01T00:00:00Z
max: 2262-04-11T23:47:16Z
min: 1677-09-21T00:12:43Z
```

When I plug in these values for minValidTime, maxValidTime, I see an overflow. 
https://play.golang.org/p/e7tAiuN8i7I

Am I doing something wrong here?

mschoch: No, it turns out that internally time is tracked in two variables, so limits are not quite as straightforward.  See this question for some of the details:

https://stackoverflow.com/questions/25065055/what-is-the-maximum-time-time-in-go

I guess I'll back off having some elegant way to compute these.

But I now have a new requirement, we should export those min/max values.  The reason is that we have examples of users try to express the query, "has a date/time indexed".  Previously this could be done with [zeroTime+1min, zeroTime).  Given that now there is some enforced restriction on the values, we should give users a symbolic way to express the same thing:

[minTime, zeroTime)

mschoch: Actually I now understand why the min overflowed in your example.  My way of initializing things went directly to the nanoseconds.  However, your version starts with the text represenation, without the nanosecond precision.  This rounds down to 0 nanoseconds, which is further in the negative direction from the epoch.

Here is a program (have to run locally since go playground is 32-bit) which attempted to binary search finding the first overflow situation:

```
package main

import (
	"fmt"
	"math"
	"sort"
	"time"
)

func main() {
	epoch := time.Unix(0, 0)
	epochUnixNano := epoch.UnixNano()
	fmt.Printf("epoch unix nano: %d\n", epochUnixNano)

	firstOverflow := sort.Search(math.MaxInt64, func(i int) bool {
		//fmt.Printf("trying %d\n", i)
		then := epoch.Add(time.Duration(i) * time.Nanosecond)
		fmt.Printf("trying: %s\n", then.Format(time.RFC3339Nano))
		thenUnixNano := then.UnixNano()
		if thenUnixNano != int64(i) {
			// overlflow
			fmt.Printf("overflowed: %d != %d", thenUnixNano, i)
			return true
		}
		return false
	})
	fmt.Printf("first overflow: %d\n", firstOverflow)

	firstUnderflow := sort.Search(math.MaxInt64, func(i int) bool {
		//fmt.Printf("trying %d\n", -i)
		then := epoch.Add(time.Duration(-i) * time.Nanosecond)
		fmt.Printf("trying: %s\n", then.Format(time.RFC3339Nano))
		thenUnixNano := then.UnixNano()
		if thenUnixNano != int64(-i) {
			// overlflow
			fmt.Printf("underflowed: %d != %d", thenUnixNano, -i)
			return true
		}
		return false
	})
	fmt.Printf("first underflow: %d\n", firstUnderflow)
}
```

But, it actually demonstrates full fidelity of int64 nanoseconds...

mschoch: But it does highlight a real issue, since we use RFC3339 as a our canonical format, we should prefer to use Min/Max values that can be round-tripped safely _in that format_

mschoch: Also, just found these:

https://github.com/blevesearch/bleve/blob/master/document/field_datetime.go#L38-L39

abhinavdangeti: Ah great. I could just reuse these settings then:
https://github.com/blevesearch/bleve/blob/d8a53a56c039b7bb29fea965996e6829f4245232/document/field_datetime.go#L38-L39

Let me make sure they work correctly with all the test cases.

mschoch: My concern is that from pure Go this would be fine, but someone mixing Go and JSON might have a problem.  For example, they might pass in MinTimeRepresentable, which is valid, but in a JSON response, we return the RFC3339 represenation of that time, which if later echo'd back in a search request via JSON, it unmarshals to a value that is now out of range (similar to how your example overflowed)...

abhinavdangeti: Very valid point, so I might just have to update these entries to the RFC3339 format that I've used in this change.

mschoch: I'm kind of torn on that too.  On the one hand, those existing constants accurately capture the restrictions of our data storage.

Only when querying does this new concern come up, and that happens because we chose RFC3339 as our format.  Another more precise format would not suffer the limitation.

It seems all of these choices might have some backwards compat issues if anyone has data indexed around these edges (unlikely but of course possible)

abhinavdangeti: Ugh yea, so I can't reuse those settings for that very reason. You're right - now we're end up turning down some queries that used to go through before, but these queries were quite likely returning some incorrect results due to the overflow.

That said, let me first push up a patch to export these settings and apply them iff QueryDateTimeFormat were RFC3339.

abhinavdangeti: On second thought they may not need exporting, as I'll be checking the QueryDateTimeFormat anyway, and this is something the user can choose to set to any other.

mschoch: I want them to be exported, because today users are hard-coding their own incorrect minimums

abhinavdangeti: Done.

Issue #1194, opened 2019-04-04
[scorch] locations not returned when "score":"none" is set in the search request along with "includeLocations":true


abhinavdangeti:
Check out this test code:

```
func TestSearchScoreNone(t *testing.T) {
	idx, err := NewUsing("testidx", NewIndexMapping(), scorch.Name, Config.DefaultKVStore, nil)
	if err != nil {
		t.Fatal(err)
	}

	defer func() {
		err := os.RemoveAll("testidx")
		if err != nil {
			t.Fatal(err)
		}
	}()

	doc := map[string]interface{}{
		"field1": "asd fgh jkl",
		"field2": "more content blah blah",
		"id":     "doc",
	}

	if err = idx.Index("doc", doc); err != nil {
		t.Fatal(err)
	}

	q := NewQueryStringQuery("content")
	sr := NewSearchRequest(q)
	sr.IncludeLocations = true
	sr.Score = "none"

	res, err := idx.Search(sr)
	if err != nil {
		t.Fatal(err)
	}

	fmt.Println("LOCATIONS:", res.Hits[0].Locations)
	fmt.Println("SCORE:", res.Hits[0].Score)
	fmt.Println("ID:", res.Hits[0].ID)
}
```

The test outputs this data:

```
LOCATIONS: map[]
SCORE: 0
ID: doc
```

However, when I remove this line:

    sr.Score = "none"

, I see:

```
LOCATIONS: map[field2:map[content:[0xc00041c180]]]
SCORE: 0.11597946228887497
ID: doc
```

cc @steveyen

PR #1195, opened 2019-04-04
Include freq/norm information when IncludeTermVectors:true

abhinavdangeti:
Fixes: https://github.com/blevesearch/bleve/issues/1194

### DIFF
--- a/search/scorer/scorer_term.go
+++ b/search/scorer/scorer_term.go
@@ -40,6 +40,7 @@ type TermQueryScorer struct {
 	idf                    float64
 	options                search.SearcherOptions
 	idfExplanation         *search.Explanation
+	includeScore           bool
 	queryNorm              float64
 	queryWeight            float64
 	queryWeightExplanation *search.Explanation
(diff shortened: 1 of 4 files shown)


search/scorer/scorer_term.go
```diff
@@ -159,7 +159,9 @@ func (s *TermQueryScorer) Score(ctx *search.SearchContext, termMatch *index.Term
 
 	rv := ctx.DocumentMatchPool.Get()
 	rv.IndexInternalID = append(rv.IndexInternalID, termMatch.ID...)
-	rv.Score = score
+	if s.options.Score != "none" {
```
steveyen: hmmm, thinking a string compare in the "hot loop" of scoring every term match might have a perf impact?

search/searcher/search_term.go
```diff
@@ -42,7 +42,7 @@ func NewTermSearcher(indexReader index.IndexReader, term string, field string, b
 }
 
 func NewTermSearcherBytes(indexReader index.IndexReader, term []byte, field string, boost float64, options search.SearcherOptions) (*TermSearcher, error) {
-	needFreqNorm := options.Score != "none"
+	needFreqNorm := options.IncludeTermVectors || options.Score != "none"
 	reader, err := indexReader.TermFieldReader(term, field, needFreqNorm, needFreqNorm, options.IncludeTermVectors)
```
steveyen: Instead of checking this at this level, I'm thinking the indexReader.TermFieldReader() has all the information that it needs to check this directly and more exactly?  That is, indexReader.TermFieldReader() also has the options.IncludeTermVectors bool? 

In contrast, the proposal here kinda overloads the `needFreqNorm` flag with "hidden" information about IncludeTermVectors?

search/scorer/scorer_term.go
abhinavdangeti: fair, let me add a bool flag to the scorer to avoid this string compare.

search/searcher/search_term.go
abhinavdangeti: Ya true, but with the code today, we initialize freqNormReader only when needFreqNorm is set to true, and this is needed to check and see if locations are actually available ..
https://github.com/blevesearch/bleve/blob/7f5d67a851143ce1ad4d8a0d493362300a814ccc/index/scorch/segment/zap/posting.go#L420-L425

And it became a somewhat invasive change to accommodate the situation where locations are required when score isn't - where now I'd need to check includeFreqNorm || includeLocs, and I felt more comfortable with this change at this stage.

PR #1198, opened 2019-04-08
reuse/cache the vellum fst instances at segment level 

sreekanth-cb:
This change aims to cache the vellum fst instances
per field at the SegmentBase level to help it’s reuse
across queries.
Its been noted from the cpu profiles that the
vellum.Load/new could be optimised.

Improvements observed for a 1M data set with 10K w/sec,

Wildcard queries -  147 – 184 [25%]
Fuzzy1 – 705 – 790 [12%]
Fuzzy2 - 71 - 79 [11%]
Prefix queries - 253 -> 270 [7%]

### DIFF
--- a/index/scorch/segment/zap/segment.go
+++ b/index/scorch/segment/zap/segment.go
@@ -20,8 +20,8 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"reflect"
 	"sync"
+	"unsafe"
 
 	"github.com/RoaringBitmap/roaring"
 	"github.com/blevesearch/bleve/index/scorch/segment"
@@ -35,7 +35,7 @@ var reflectStaticSizeSegmentBase int
 
 func init() {
 	var sb SegmentBase
-	reflectStaticSizeSegmentBase = int(reflect.TypeOf(sb).Size())
+	reflectStaticSizeSegmentBase = int(unsafe.Sizeof(sb))
 }
 
 // Open returns a zap impl of a segment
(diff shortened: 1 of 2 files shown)


index/scorch/segment/zap/segment.go
```diff
@@ -258,19 +262,28 @@ func (sb *SegmentBase) dictionary(field string) (rv *Dictionary, err error) {
...
-				rv.fstReader, err = rv.fst.Reader()
-				if err != nil {
-					return nil, fmt.Errorf("dictionary field %s vellum reader err: %v", field, err)
+			var ok bool
+			sb.m.Lock()
+			if rv.fst, ok = sb.fieldFSTs[rv.fieldID]; !ok {
+				// read the length of the vellum data
+				vellumLen, read := binary.Uvarint(sb.mem[dictStart : dictStart+binary.MaxVarintLen64])
+				fstBytes := sb.mem[dictStart+uint64(read) : dictStart+uint64(read)+vellumLen]
+				if fstBytes != nil {
```
steveyen: looks pretty good... but the fstBytes here looks like it can never be nil, so consider removing the check?  I see the nil check is also in the original code, though.

index/scorch/segment/zap/segment.go
```diff
@@ -35,7 +35,7 @@ var reflectStaticSizeSegmentBase int
 
 func init() {
 	var sb SegmentBase
-	reflectStaticSizeSegmentBase = int(reflect.TypeOf(sb).Size())
+	reflectStaticSizeSegmentBase = int(unsafe.Sizeof(sb))
```
abhinavdangeti: Any idea why this is showing up as a recent failure? This has been in the code base for some time now..

$ go vet $(go list ./... | grep -v vendor/)
index/scorch/segment/zap/segment.go:38: call of unsafe.Sizeof copies lock value: zap.SegmentBase contains sync.Mutex
exit status 1

sreekanth-cb: No idea..  reflect.TypeOf(sb).Size() was failing for all the Go versions in CI, With unsafe.Sizeof , it just fails for only one.( the oldest)

PR #1206, opened 2019-04-28
add optional validation hooks to scorch/zap

mschoch:
ValidateDocFields can be used to check each field
seen by zap

ValidateMerge can be used to check the results of
a merge operation

These hooks are intended to allow for application specific
checks inside the main bleve branch, so as to allow maximum
flexibility with minimal disruption.  These hooks are not
intended to be permanent, but that decision can be made
at a later time.

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -223,7 +223,7 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 			}
 			err = zap.ValidateMerge(segmentsToMerge, docsToDrop, seg.(*zap.Segment))
 			if err != nil {
-				return fmt.Errorf("merge validation failed: %v")
+				return fmt.Errorf("merge validation failed: %v", err)
 			}
 			oldNewDocNums = make(map[uint64][]uint64)
 			for i, segNewDocNums := range newDocNums {
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -316,7 +316,7 @@ func (s *Scorch) mergeSegmentBases(snapshot *IndexSnapshot,
 	}
 	err = zap.ValidateMerge(nil, sbs, sbsDrops, seg.(*zap.Segment))
 	if err != nil {
-		return nil, 0, fmt.Errorf("merge validation failed: %v", err)
+		return nil, 0, fmt.Errorf("in-memory merge validation failed: %v", err)
 	}
 
 	// update persisted stats
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -221,6 +221,10 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 				atomic.AddUint64(&s.stats.TotFileMergePlanTasksErr, 1)
 				return err
 			}
+			err = zap.ValidateMerge(segmentsToMerge, docsToDrop, seg.(*zap.Segment))
+			if err != nil {
+				return fmt.Errorf("merge validation failed: %v")
+			}
 			oldNewDocNums = make(map[uint64][]uint64)
 			for i, segNewDocNums := range newDocNums {
 				oldNewDocNums[task.Segments[i].Id()] = segNewDocNums
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -151,7 +151,6 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 		atomic.AddUint64(&s.stats.TotFileMergePlanNone, 1)
 		return nil
 	}
-
 	atomic.AddUint64(&s.stats.TotFileMergePlanOk, 1)
 
 	atomic.AddUint64(&s.stats.TotFileMergePlanTasks, uint64(len(resultMergePlan.Tasks)))
@@ -221,7 +220,7 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 				atomic.AddUint64(&s.stats.TotFileMergePlanTasksErr, 1)
 				return err
 			}
-			err = zap.ValidateMerge(segmentsToMerge, docsToDrop, seg.(*zap.Segment))
+			err = zap.ValidateMerge(segmentsToMerge, nil, docsToDrop, seg.(*zap.Segment))
 			if err != nil {
 				return fmt.Errorf("merge validation failed: %v", err)
 			}
@@ -314,6 +313,10 @@ func (s *Scorch) mergeSegmentBases(snapshot *IndexSnapshot,
 	if err != nil {
 		atomic.AddUint64(&s.stats.TotMemMergeErr, 1)
 		return nil, 0, err
+	}
+	err = zap.ValidateMerge(nil, sbs, sbsDrops, seg.(*zap.Segment))
+	if err != nil {
+		return nil, 0, fmt.Errorf("merge validation failed: %v", err)
 	}
 
 	// update persisted stats
--- a/index/scorch/segment/zap/merge.go
+++ b/index/scorch/segment/zap/merge.go
@@ -33,8 +33,9 @@ var DefaultFileMergerBufferSize = 1024 * 1024
 
 // ValidateMerge can be set by applications to perform additional checks
 // on a new segment produced by a merge, by default this does nothing.
+// Caller should provide EITHER segments or memSegments, but not both.
 // This API is experimental and may be removed at any time.
-var ValidateMerge = func(segments []*Segment, drops []*roaring.Bitmap, newSegment *Segment) error {
+var ValidateMerge = func(segments []*Segment, memSegments []*SegmentBase, drops []*roaring.Bitmap, newSegment *Segment) error {
 	return nil
 }
 
--- a/index/scorch/segment/zap/merge.go
+++ b/index/scorch/segment/zap/merge.go
@@ -31,6 +31,13 @@ import (
 
 var DefaultFileMergerBufferSize = 1024 * 1024
 
+// ValidateMerge can be set by applications to perform additional checks
+// on a new segment produced by a merge, by default this does nothing.
+// This API is experimental and may be removed at any time.
+var ValidateMerge = func(segments []*Segment, drops []*roaring.Bitmap, newSegment *Segment) error {
+	return nil
+}
+
 const docDropped = math.MaxUint64 // sentinel docNum to represent a deleted doc
 
 // Merge takes a slice of zap segments and bit masks describing which
--- a/index/scorch/segment/zap/new.go
+++ b/index/scorch/segment/zap/new.go
@@ -33,6 +33,14 @@ var NewSegmentBufferNumResultsBump int = 100
 var NewSegmentBufferNumResultsFactor float64 = 1.0
 var NewSegmentBufferAvgBytesPerDocFactor float64 = 1.0
 
+// ValidateDocFields can be set by applications to perform additional checks
+// on fields in a document being added to a new segment, by default it does
+// nothing.
+// This API is experimental and may be removed at any time.
+var ValidateDocFields = func(field document.Field) error {
+	return nil
+}
+
 // AnalysisResultsToSegmentBase produces an in-memory zap-encoded
 // SegmentBase from analysis results
 func AnalysisResultsToSegmentBase(results []*index.AnalysisResult,
@@ -520,6 +528,11 @@ func (s *interim) writeStoredFields() (
 
 			if opts.IncludeDocValues() {
 				s.IncludeDocValues[fieldID] = true
+			}
+
+			err := ValidateDocFields(field)
+			if err != nil {
+				return 0, err
 			}
 		}
 


mschoch:
I decided to see if you would be OK with an implementation that added these hooks to the master of bleve (as opposed to me maintaining a branch).  The idea is that some of the checks may be application specific, or have performance penalties, but with this implementation, only people that want to use them have to.

Here is a sample implementation I'm working with at the moment:

```
func init() {
	zap.ValidateMerge = func(segments []*zap.Segment, drops []*roaring.Bitmap, newSegment *zap.Segment) error {
		log.Printf("validating a merge")

		// iterate over each source segment
		for i, segment := range segments {
			fieldsInSegment := segment.Fields()
			// iterate over each local doc number
			for segmentLocalDocNum := uint64(0); segmentLocalDocNum < segment.Count(); segmentLocalDocNum++ {
				// if this doc is obsolete skip it
				if drops[i] != nil && drops[i].Contains(uint32(segmentLocalDocNum)) {
					continue
				}
				// otherwise, find all of that docs field terms
				docFieldTerms := make(map[string][]string)
				_, err := segment.VisitDocumentFieldTerms(segmentLocalDocNum, fieldsInSegment, func(field string, term []byte) {
					docFieldTerms[field] = append(docFieldTerms[field], string(term))
				}, nil)
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error visiting doc field terms: %v")
				}
				// now look up this docs external identifier
				externalDocID, err := segment.DocID(segmentLocalDocNum)
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error looking up id for doc number: %v",err)
				}
				// now go to the new segment, and find its doc number there
				newDocNums, err := newSegment.DocNumbers([]string{string(externalDocID)})
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error looking up new doc number: %v", err)
				}

				// we don't know that fields will be identical in the new segment,
				// BUT we do expect all fields we actually saw in the previous visit
				// method should STILL be there, so let's use those

				var docActuallyUsesFields []string
				for fieldName := range docFieldTerms {
					docActuallyUsesFields = append(docActuallyUsesFields, fieldName)
				}

				newDocNumsItr := newDocNums.Iterator()
				if newDocNumsItr == nil || !newDocNumsItr.HasNext() {
					return fmt.Errorf("VALIDATE MERGE unable to doc number in new segment")
				}
				newSegmentLocalDocNumber := uint64(newDocNumsItr.Next())
				newDocFieldTerms := make(map[string][]string)
				_, err = newSegment.VisitDocumentFieldTerms(newSegmentLocalDocNumber, docActuallyUsesFields, func(field string, term []byte) {
					newDocFieldTerms[field] = append(newDocFieldTerms[field], string(term))
				}, nil)
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error visiting doc field terms in new segment: %v")
				}

				// we expect that docFieldTerms and newDocFieldTerms should be the same
				if !reflect.DeepEqual(docFieldTerms, newDocFieldTerms) {
					return fmt.Errorf("VALIDATE MERGE doc field values do not match\n%v\n%v", docFieldTerms, newDocFieldTerms)
				}
			}
		}
		log.Printf("merge OK")
		return nil
	}

	zap.ValidateDocFields = func(field document.Field) error {
		if field.Name() == "_id" {
			return nil
		}
		if !field.Options().IncludeDocValues() {
			return fmt.Errorf("VALIDATE DOC FIELDS include doc values false")
		}
		return nil
	}
}
```

This does two things:

1.  The document field validation check ensures that all documents coming in to bleve have IncludeDocValues set to true.  In our application, that is the case, though it is not generally the case.  Documents with this value set to false *could* cause the behavior we're seeing, so this is a helpful check.

2.  The merge validation is a bit more complex.  The approach taken here is trying to be straightforward, and not use a lot of memory at any one time, though it trades off for a slower execution.  The idea is to go though each document in each input segment.  If it's deleted/obsolete, we skip it.  Otherwise, we look at all of it's DocValues and stuff them in a map.  Then, we find the corresponding document in the new segment, look at it's DocValues and stuff them in a map.  Then we compare the maps expecting them to be the same.  One wrinkle is that fields could have changed at the segment level (other docs with new fields), but we only care that every field in the source segments doc matches the new segment version, so we limit our map building to those fields.

sreekanth-cb:
@mschoch , forgot to verify this while reviewing. We may need to add such validations in the persister codepaths where in we do the in-memory segment merging. (we take a bunch of memory segments and merge and persist them and hence similar verification checks ought to be applicable there as well?)

mschoch:
@sreekanth-cb thanks that is a good point.  I'll review it today and see if the same callback would be appropriate there.

Also, i was wondering if you also reviewed the callback implementation I put in the comment, to see if you agree with how it works.  I was hoping to catch you online yesterday day, but we never overlapped.

sreekanth-cb:
I had a quick peek yesterday, but shall check it once again and respond here..

mschoch:
@sreekanth-cb it seems like extending this to in-memory merges is complicated.  Not because it should be, but because the methods we use to read doc values, are defined on *zap.Segment, and not SegmentBase.  It seems like logically they could be defined on SegmentBase, since the data they use is there.  I experimented moving the methods there, and got it compile, but it doesn't seem right to change the impl so drastically just to add debugging.

Which is unfortunate, because I'd love to believe the problem is actually related to in-memory merges, for one it would explain why we so rarely see this...

At the moment, the only 2 choices I see are:

1.  Move DocumentFieldTermVisitable method impls to SegmentBase, in this case our "checker" function can basically work as it does now.
2.  Have separate checker method, and be forced to manually muck with the doc values internals ourselves to validate it is correct.

Do you see any other options?

mschoch:
I did come up with a 3rd option.  I can create a fake segment wrapper around the segment base, and use that for validation purposes.  Not ideal, but I think this is less invasive than the alternatives...

mschoch:
@sreekanth-cb I've updated it to invoke the same callback (slightly different signature now) for both regular  and in-memory merges.

My callback has been updated as follows (and I'd still like your feedback on it as well):

```
func init() {
	zap.ValidateMerge = func(segments []*zap.Segment, memSegments []*zap.SegmentBase, drops []*roaring.Bitmap, newSegment *zap.Segment) error {

		checkSeg := func(i int, seg *zap.Segment) error {
			fieldsInSegment := seg.Fields()
			// iterate over each local doc number
			for segmentLocalDocNum := uint64(0); segmentLocalDocNum < seg.Count(); segmentLocalDocNum++ {
				// if this doc is obsolete skip it
				if drops[i] != nil && drops[i].Contains(uint32(segmentLocalDocNum)) {
					continue
				}
				// otherwise, find all of that docs field terms
				docFieldTerms := make(map[string][]string)
				_, err := seg.VisitDocumentFieldTerms(segmentLocalDocNum, fieldsInSegment, func(field string, term []byte) {
					docFieldTerms[field] = append(docFieldTerms[field], string(term))
				}, nil)
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error visiting doc field terms: %v")
				}
				// now look up this docs external identifier
				externalDocID, err := seg.DocID(segmentLocalDocNum)
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error looking up id for doc number: %v", err)
				}
				// now go to the new segment, and find its doc number there
				newDocNums, err := newSegment.DocNumbers([]string{string(externalDocID)})
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error looking up new doc number: %v", err)
				}

				// we don't know that fields will be identical in the new segment,
				// BUT we do expect all fields we actually saw in the previous visit
				// method should STILL be there, so let's use those

				var docActuallyUsesFields []string
				for fieldName := range docFieldTerms {
					docActuallyUsesFields = append(docActuallyUsesFields, fieldName)
				}

				newDocNumsItr := newDocNums.Iterator()
				if newDocNumsItr == nil || !newDocNumsItr.HasNext() {
					return fmt.Errorf("VALIDATE MERGE unable to doc number in new segment")
				}
				newSegmentLocalDocNumber := uint64(newDocNumsItr.Next())
				newDocFieldTerms := make(map[string][]string)
				_, err = newSegment.VisitDocumentFieldTerms(newSegmentLocalDocNumber, docActuallyUsesFields, func(field string, term []byte) {
					newDocFieldTerms[field] = append(newDocFieldTerms[field], string(term))
				}, nil)
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error visiting doc field terms in new segment: %v")
				}

				// we expect that docFieldTerms and newDocFieldTerms should be the same
				if !reflect.DeepEqual(docFieldTerms, newDocFieldTerms) {
					return fmt.Errorf("VALIDATE MERGE doc field values do not match\n%v\n%v", docFieldTerms, newDocFieldTerms)
				}
			}
			return nil
		}

		// iterate over each source segment
		if segments != nil {
			for i, seg := range segments {
				err := checkSeg(i, seg)
				if err != nil {
					return err
				}
			}
		} else if memSegments != nil {
			// or iterate over each in-memory segment
			for i, seg := range memSegments {
				fakeSeg := &zap.Segment{
					SegmentBase: *seg,
				}
				err := checkSeg(i, fakeSeg)
				if err != nil {
					return err
				}
			}
		}
		
		return nil
	}

	zap.ValidateDocFields = func(field document.Field) error {
		if field.Name() == "_id" {
			return nil
		}
		if !field.Options().IncludeDocValues() {
			return fmt.Errorf("VALIDATE DOC FIELDS include doc values false")
		}
		return nil
	}
}
```

index/scorch/merge.go
```diff
@@ -311,6 +314,10 @@ func (s *Scorch) mergeSegmentBases(snapshot *IndexSnapshot,
 		atomic.AddUint64(&s.stats.TotMemMergeErr, 1)
 		return nil, 0, err
 	}
+	err = zap.ValidateMerge(nil, sbs, sbsDrops, seg.(*zap.Segment))
+	if err != nil {
+		return nil, 0, fmt.Errorf("merge validation failed: %v", err)
```
sreekanth-cb: "merge validation failed:" -> looks exactly same across zap vs memory segments, Is it enough with the err to differentiate where it happened? meaning during zap or memory merging.. do we need to be slightly verbose to indicate the context as well

sreekanth-cb:
doubts in the callback impl are,
1. are we always spitting out the field name/details as well along with the error details 
2. what will happen if there is no docValue for a field in older segment as well as newer segment? Will it throw false errors ? Or given the always on/true docValue there will always be some value for the doc values in older segment as well as newer segment?

mschoch:
Thanks @sreekanth-cb 

1.  I was mostly interested in reviewing the logic, as the client will actually be using a different actual implementation.  All of the errors return except for the last one would be highly unexpected, as they represent the API just not working as expected.  It is still useful to log them properly, so below you'll see I've update the to include more relevant context about the error.  But, the expectation is that these will not happen.

2.  If I understand correctly, your question is what happens when there is no docValue for a field in older segment as well as newer segment?  My understanding/expectation is that when we invoke `VisitDocumentFieldTerms` on the old segment, the callback will never be fired for that field.  This means that the `docFieldTerms` map will not contain that field as a key.  This means that that field will NOT be in the slice `docActuallyUsesFields`.  And that means that we will not even request that field when invoking `VisitDocumentFieldTerms` on the new segment.  Which means that the map `newDocFieldTerms` should not contain that field as a key either.  So, in conclusion, I would expect that in that case, that field will not be in either map, thus as it relates to that field, the maps will be identical (both not containing it).

2a.  Regarding your follow-up question for 2, you ask if becuase DV is always on, would there always be a value.  And I think the answer to that is NO.  Consider a field which is analyzed using the "en" analyzer (will drop stop words), and consider the doc having the value "the".  The stored field would contain "the", but both the the inverted index, and the DV sections would contain no content, as we dropped all tokens during analysis, so there was nothing to index.  That is valid, and I don't see that it would cause any issues in this validation.

Finally, you asked in another comment about the error message not differentiating between which merge type encountered the error.  That is a good point, I'll update it.  But, again I want to emphasize that we will not be using this exact callback.  In fact, our implementation will not even be returning an error.  Instead, in addition to logging helpful information for us to understand the bug, we will be producing a list of document IDs that can be re-indexed to fix the problem.

mschoch:
Oops, here is the updated callback, I intended to include in the previous comment (this updates the log messages):

```
func init() {
	zap.ValidateMerge = func(segments []*zap.Segment, memSegments []*zap.SegmentBase, drops []*roaring.Bitmap, newSegment *zap.Segment) error {

		checkSeg := func(i int, seg *zap.Segment) error {
			fieldsInSegment := seg.Fields()
			// iterate over each local doc number
			for segmentLocalDocNum := uint64(0); segmentLocalDocNum < seg.Count(); segmentLocalDocNum++ {
				// if this doc is obsolete skip it
				if drops[i] != nil && drops[i].Contains(uint32(segmentLocalDocNum)) {
					continue
				}
				// otherwise, find all of that docs field terms
				docFieldTerms := make(map[string][]string)
				_, err := seg.VisitDocumentFieldTerms(segmentLocalDocNum, fieldsInSegment, func(field string, term []byte) {
					docFieldTerms[field] = append(docFieldTerms[field], string(term))
				}, nil)
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error visiting doc number %d in segment '%s' field terms: %v",
						segmentLocalDocNum, seg.Path(), err)
				}
				// now look up this docs external identifier
				externalDocID, err := seg.DocID(segmentLocalDocNum)
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error looking up id for doc number %d in segment '%s': %v",
						segmentLocalDocNum, seg.Path(), err)
				}

				// we don't know that fields will be identical in the new segment,
				// BUT we do expect all fields we actually saw in the previous visit
				// method should STILL be there, so let's use those
				var docActuallyUsesFields []string
				for fieldName := range docFieldTerms {
					docActuallyUsesFields = append(docActuallyUsesFields, fieldName)
				}

				// now go to the new segment, and find its doc number there
				newDocNums, err := newSegment.DocNumbers([]string{string(externalDocID)})
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error looking up doc number for '%s' in new segment '%s': %v",
						externalDocID, newSegment.Path(), err)
				}

				newDocNumsItr := newDocNums.Iterator()
				if newDocNumsItr == nil || !newDocNumsItr.HasNext() {
					return fmt.Errorf("VALIDATE MERGE unable to find doc number in new segment '%s'", newSegment.Path())
				}
				newSegmentLocalDocNumber := uint64(newDocNumsItr.Next())
				newDocFieldTerms := make(map[string][]string)
				_, err = newSegment.VisitDocumentFieldTerms(newSegmentLocalDocNumber, docActuallyUsesFields, func(field string, term []byte) {
					newDocFieldTerms[field] = append(newDocFieldTerms[field], string(term))
				}, nil)
				if err != nil {
					return fmt.Errorf("VALIDATE MERGE error visiting doc number %d in segment '%s' field terms: %v",
						newSegmentLocalDocNumber, newSegment.Path(), err)
				}

				// we expect that docFieldTerms and newDocFieldTerms should be the same
				if !reflect.DeepEqual(docFieldTerms, newDocFieldTerms) {
					return fmt.Errorf("VALIDATE MERGE doc field values do not match\n%v\n%v", docFieldTerms, newDocFieldTerms)
				}
			}
			return nil
		}

		// iterate over each source segment
		if segments != nil {
			for i, seg := range segments {
				err := checkSeg(i, seg)
				if err != nil {
					return err
				}
			}
		} else if memSegments != nil {
			// or iterate over each in-memory segment
			for i, seg := range memSegments {
				fakeSeg := &zap.Segment{
					SegmentBase: *seg,
				}
				err := checkSeg(i, fakeSeg)
				if err != nil {
					return err
				}
			}
		}

		return nil
	}

	zap.ValidateDocFields = func(field document.Field) error {
		if field.Name() == "_id" {
			return nil
		}
		if !field.Options().IncludeDocValues() {
			return fmt.Errorf("VALIDATE DOC FIELDS include doc values false for field %s", field.Name())
		}
		return nil
	}
}
```

Also, one additional commit will be pushed to clarify the error message when the merge callback returns an error.

sreekanth-cb:
LGTM 👍

PR #1208, opened 2019-05-06
move the impl of DocumentFieldTermVisitable

mschoch:
previously these methods were defined on Segment
but they only use fields in SegmentBase,
and having them defined on SegmentBase means you
can use them on in-memory segments

### DIFF
--- a/index/scorch/segment/zap/docvalues.go
+++ b/index/scorch/segment/zap/docvalues.go
@@ -39,7 +39,7 @@ type docNumTermsVisitor func(docNum uint64, terms []byte) error
 
 type docVisitState struct {
 	dvrs    map[uint16]*docValueReader
-	segment *Segment
+	segment *SegmentBase
 }
 
 type docValueReader struct {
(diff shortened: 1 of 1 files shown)


mschoch:
The reason for this proposal is that the recent validation callbacks that were added previously were using a "fake segment" approach to check in-memory segments, like this:

```
fakeSeg := &zap.Segment{
    SegmentBase: *seg,
}
err := checkSeg(i, fakeSeg)
```

We had to do this because the DocumentFieldTermVisitable methods only worked on a Segment, not a SegmentBase.  But, the problem is that Go vet flags this for copying a mutex inside of SegmentBase.

I think moving these methods is cleaner anyway.  All of the other Segment interface methods were implemented on *SegmentBase already.

mschoch:
NOTE: I had considered that this might have been the source of an issue all along.  The of thinking was, what if in-memory segments were being treated differently somewhere else, because we implemented these methods in the wrong place.

But, I don't think thats the case.  The reason is that merge impl doesn't use these methods, it operates directly on the underlying fields.

So, if there were any other side-effects, it would only affect bleve consumes, not bleve internals.  And with that in mind, it seems like possibly, searches with sorting/faceting that included in-memory segments, might have been doing a run-time uninversion, when they could have used docvalues directly.  But this is unconfirmed...

mschoch:
As a follow-up to my previous comment about how search including sort/facet might be affected, I believe in the past, in-memory segments would fail this check:

https://github.com/blevesearch/bleve/blob/master/index/scorch/snapshot_index.go#L566

And after this change, they would now pass the check.

As I understand it, this is OK, and probably performs better.  However, it is important that @steveyen and @sreekanth-cb review this...

mschoch:
OK, I modified bleve a little bit to allow me to disable persistence.  This ensured that all the segments in the index were in-memory, and I verified that when a query with facets was run, that check was not satisfied.  The following was added after: https://github.com/blevesearch/bleve/blob/master/index/scorch/snapshot_index.go#L566

```
 else {
		log.Printf("segment isnt dftv")
	}
```

And after applying this patch, that behavior changes, and the DocumentFieldTermVisitable interface is used.

In both cases I received correct results, so I'm going to proceed with this patch.

PR #1209, opened 2019-05-08
fix batch persisted callback for upsidedown

mschoch:
previous implementation would only execute the callback
if the control flow reached the end of the function

unfortunately there are other explicit returns earlier
in the function

users setting a callback, would reasonably expect it to
always be called (passing either nil or an err) and
due to this bug, certain error conditions resulted
in the callback not firing

this change checks for the callback, and if present
defers execution of the callback, ensuring it will
always be executed


mschoch:
@abhinavdangeti good catch, I opened #1210

Issue #1210, opened 2019-05-08
review batch persisted callback for scorch


mschoch:
See this comment from @abhinavdangeti https://github.com/blevesearch/bleve/pull/1209#pullrequestreview-234819057

abhinavdangeti:
Related: https://github.com/blevesearch/bleve/pull/1255

PR #1211, opened 2019-05-08
Invoking the persisted callbacks within scorch

abhinavdangeti:
The registered persisted callbacks are to be invoked
whether persistSnapshot(..) passes or not.

Should fix: https://github.com/blevesearch/bleve/issues/1210
More details on: https://github.com/blevesearch/bleve/pull/1209


abhinavdangeti:
Let me try and put together a unit test for this.

sreekanth-cb:
My understanding here is,

In this change, upon an ephemeral persist error, we throw the error back to user and on the next OUTER loop iteration, those segments may get persisted correctly and we won't be able to give any further/final intimations to the user.
So the user/app is left with an ephemeral/incorrect error status for the persistence of a segment even when it would have persisted by the time the app is going to take any action on it.
(For eg: an imaginary ephemeral error could space issue resolved out a free-ing of space due to a parallel compaction completion.)

In the original implementation, a persisted callback would have happened only when on a successful persistence happen on a segment. On errors, it continues with newer snapshots and eventually the user gets to know whether persistence happened or not, which make more sense to me.

Also, the application is not left without any clues when these error occurs. We give a s.fireAsyncError upon errors.

So I am not sure whether this change is needed or not?

abhinavdangeti:
> In this change, upon an ephemeral persist error, we throw the error back to user and on the next OUTER loop iteration, those segments may get persisted correctly and we won't be able to give any further/final intimations to the user.

Firstly, there remains the case when the segment were closed, we break out of OUTER.
Secondly, in the current code we won't be notifying the user on a retry attempt, as we've already purged the original persisted callbacks (could this be a problem?)

> In the original implementation, a persisted callback would have happened only on a successful persistence happens on a segment. On errors, it continues with newer snapshots and eventually the user gets to know whether persistence happened or not, which make more sense?.

That's implying that the persisted callbacks will always be invoked with err=nil. To remain consistent with upside-down's behavior now, the persisted callback will need to be invoked before we error out and lose the persisted callbacks.

mschoch:
So, I think it would be helpful to take a step back and clarify the use case that I think we want to support.

In the Couchbase model, the information to resume the source stream is persisted inside of bleve, along with the batch, using internal storage.  This means that Couchbase has no real need for a persisted callback.

However, there is another model, where an application uses a persistent queue to safely move data from some other primary storage into the index.  In this model, a persisted callback is used to remove items from the queue after the data is safely stored in the bleve index.  So, from this perspective, there are 2 very useful things to signal:

1) data persisted without error
2) data failed to persist (bleve has given up, and these updates should be considered lost by the app)

Knowing about any transient errors in the interim, might be nice, but doesn't let the app take any direct action.

Now, about the concept of knowing that the callback will always fire eventually.  The thinking here is that, although the queue can be used to re-index data later, sometimes we just want to cleanly shutdown the index.  In that case, we want to just have ref counts of how many batches we indexed, and know when they're all done.  

So, I bring all that up just to clarify what behavior would be useful for applications.  It sounds like, even with this PR there are some specific cases we need to rethink.

abhinavdangeti:
Thanks for that breakdown @mschoch :).

Right, so the case we'll need to handle is when persistSnapshot fails for a reason other than ErrClosed, and in this situation if fireAsyncError doesn't log a fatal error (couchbase logs a fatal error here) and no new snapshot was introduced in the interim - the persister attempts re-persisting the previous snapshot. At this point the persisted callbacks have however been cleaned out.
With the current change, we would have invoked them at least once with the error seen in the first attempt.
The unhandled bit is to preserve these persisted callbacks up till one of the following events ..
- a new snapshot has been introduced
- the persister attempts to persist the exact previous snapshot and succeeds

mschoch:
So, I would think that firing the callback on the first error, when scorch may continue to try to persist the data is not good (but perhaps better than never firing at all).

How hard do the additional changes appear to be?  The way it's worded it sounds hard, like you need to save something, and there may be no good place to attach that something today...

abhinavdangeti:
Oh, the latest commit here does accommodate the additional changes (rather straightforward). The persister now holds the persistedCallbacks until either of the events (mentioned above) occur.

abhinavdangeti:
But if you think that "firing the callback on the first error, when scorch may continue to try to persist the data is not good", the change to accommodate it is very simple.

abhinavdangeti:
Hey folks, I think that this is still something we'd need to consider .. any further thoughts on this?

index/scorch/persister.go
```diff
@@ -142,21 +146,38 @@ OUTER:
 				}
 				close(ch)
 			}
+
```
sreekanth-cb: To fire a callback,
What if we just move the reset of the s.persistedCallbacks from line no: 133 to 181 or so.
(Reset it only on a success case) and fire the callback wherever needed than having all these newer temp slices?

sreekanth-cb:
Yes, as was mentioned in the first comment - I still stand with the opinion of firing this callback only on a successful completion of persistence. 
And guess @mschoch is also suggesting the same?
But the current code I suspect - is doing it all the time?

abhinavdangeti:
I've created a new PR for preserve persisterCallbacks of a snapshot until needed here: https://github.com/blevesearch/bleve/pull/1255

PR #1213, opened 2019-05-08
MB-34021 - geo query not returning few docs

sreekanth-cb:
The geo coordinates location precision lost during
geohash decoding was resulting in this issue.

Introducing a new experimental geohash decoding method
with better numerical precision and performance. 
[Simpler approach of taking the centre of the bounding box
and that looks like fixing the issue reported.]
But this needs to be tested further to see how it fares against 
the ES in a broader use cases.

go test -benchmem -bench=.
goos: darwin
goarch: amd64
pkg: github.com/blevesearch/bleve/geo
BenchmarkGeoHashLen5Decode-8      	10000000	       119 ns/op	       0 B/op	       0 allocs/op
BenchmarkGeoHashLen5NewDecode-8   	20000000	        60.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkGeoHashLen6Decode-8      	10000000	       120 ns/op	       0 B/op	       0 allocs/op
BenchmarkGeoHashLen6NewDecode-8   	20000000	        70.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkGeoHashLen7Decode-8      	10000000	       119 ns/op	       0 B/op	       0 allocs/op
BenchmarkGeoHashLen7NewDecode-8   	20000000	        83.2 ns/op	       0 B/op	       0 allocs/op


sreekanth-cb:
As I mentioned earlier this approach gives numerically precise results, but not very sure to what extent that verify/confirm the practical relevance ? as I see the original algorithm tries many things like adjusting precision/rounding of numbers, though that changes proved lesser effective in our tests over these reported bugs.

sreekanth-cb:
> gotta admit, my understanding of the math on this isn't 100%, but looks like good work to me!
> 
> on the "more testing" side, one thought is perhaps check a lot of cases (perhaps even randomly or somewhat exhaustively) against the existing approach if there's a way to see if the new approach is "no worse" somehow w.r.t. correctness?

this approach gives numerically precise results, but not very sure to what extent that verify/confirm the practical relevance ? as I see the original algorithm tries many things like adjusting precision/rounding of numbers, though that changes proved lesser effective in our tests over these reported bugs.

Think its better to verify against ES to confirm the effectiveness/relevance of the precision in numbers. 
Good thing is we started testing geo more rigorously recently?

geo/parse.go
```diff
@@ -85,7 +85,7 @@ func ExtractGeoPoint(thing interface{}) (lon, lat float64, success bool) {
 			}
 		} else {
 			// geohash
-			lat, lon = GeoHashDecode(geoStr)
```
mschoch: what happened to this function, shouldn't it be removed?

sreekanth-cb: Yes, there is a cleanup pending here for the stale code. But thinking was that lets the new API stand the test of time (the system/longevity tests in CB which are ongoing now for MH) for some time and then I can clean it up..

mschoch: In that case, I would recommend we not do it this way.  Generally, we don't keep unused code in the repository (same reason we don't commit commented out code).  Everyone intends to come back and clean it up, but in reality, we all move on to other tasks and it gets forgotten.

If we're confident enough to use the new function, we should remove the old one.  It's in the git history if we ever need to switch it back.

PR #1215, opened 2019-05-10
when index is already closed, close returns ErrorIndexClosed

mschoch:
this change make Close() consistent with the other bleve
methods.  if an index is closed, the method should do
nothing, and return ErrorIndexClosed.

previously, the underlying index implementation would
be closed again, and in the case of scorch this can
lead to panics:

```
ERRO[0004] runtime.plainError close of closed channel
/usr/local/go/src/runtime/panic.go:513 (0x42e079)
	gopanic: reflectcall(nil, unsafe.Pointer(d.fn), deferArgs(d), uint32(d.siz), uint32(d.siz))
/usr/local/go/src/runtime/chan.go:335 (0x4074fe)
	closechan: panic(plainError("close of closed channel"))
```

### DIFF
--- a/index_impl.go
+++ b/index_impl.go
@@ -755,6 +755,10 @@ func (i *indexImpl) Close() error {
 	i.mutex.Lock()
 	defer i.mutex.Unlock()
 
+	if !i.open {
+		return ErrorIndexClosed
+	}
+
 	indexStats.UnRegister(i)
 
 	i.open = false


mschoch:
Previously asked this in slack, but maybe better to do it more asynchronously here:

@steveyen wondering if you had any thoughts on this PR?  specifically, if you'd consider returning an error in a new case that we didn't before an API change, and something we should avoid

Option 1: this PR

Option 2:  in this case we could reasonably improve the behavior by silently doing the right thing, but not necessarily returning the error, that would avoid breaking programs that might already rely on close not returning an error

Option 3: do nothing at all, and panic on double close is ok

mschoch:
Please do not merge this at this time.  I have some concerns about returning an error in cases we did not before.

I'm leaning towards a simpler fix that prevents the crash, but returns nil.

Mainly the concern is the possibility of breaking existing apps...

PR #1216, opened 2019-05-16
WIP - nestedArraySearcher draft

sreekanth-cb:
The intent is to have a new searcher which
helps to search inside slice of objects nested
with in a document.
ref: https://forums.couchbase.com/t/fts-inside-array/15476

The implementation takes the approach of having
a FilteringSearcher on top of the conjunction searcher.
The new filtering searcher implementation check whether
all the matched terms in the document are sharing an
array position.
Thinking is that a even single array position is shared
then the document match is a valid one.


mschoch:
The first question I have is why this is controlled with a Boolean?  How much of the array positions have to match in order for this to be satisfied?  I was expecting that users would want to specify the depth at which they want to require the array positions to match.  I was thinking that if you had more deeply nested objects and arrays, sometimes you don't want to require that the deepest level matches, just some higher portion matches.

sreekanth-cb:
@mschoch , thats right. I got tunnel sighted and overlooked the fact that array positions could indicate deeper nestedness. Current one only handles the a single depth. Will address the parameterised depth part..

sreekanth-cb:
pushed another patch set. 
@mschoch - getting a basic doubt here. While searching deeply nested fields, we need to specify the "field" in query as the entire nested path, right?

mschoch:
@sreekanth-cb I think the answer is, you must specify field path to properly constrain the original matches, which isn't unique to this type of search, it is always the case.  But that again makes me think that the qualifier for this should new query type should simply be integer depth of commonality.

I think it would be easier for us to discuss this with an example set of documents (1 is sufficient for now) an example query (what the user specifies) and the expected result.  I ask this because "deeply nested fields" could mean different things to us...

sreekanth-cb:
Few simple examples are added in the integration tests in this PR.

mschoch:
@sreekanth-cb in that case, can you ask your question again referencing one of the examples?  It seems like you think there is a problem with what is in there now, and I don't yet see it.

search/query/conjunction.go
```diff
@@ -26,6 +26,7 @@ import (
 type ConjunctionQuery struct {
 	Conjuncts       []Query `json:"conjuncts"`
 	BoostVal        *Boost  `json:"boost,omitempty"`
+	ArrayDepth      int     `json:"arraydepth,omitempty"`
```
abhinavdangeti: Per our JSON convention .. shouldn't this be array_depth?

sreekanth-cb: right. done!

sreekanth-cb:
@mschoch , @steveyen gentle reminder for bringing this into your radars.

search/searcher/search_nested.go
```diff
@@ -0,0 +1,154 @@
...
+		if totalMatchFound == len(apLists)-1 {
+			return true
+		}
+		baseIndex++
+	}
+
+	return false
+}
+
+func equalAPSlices(a, b search.ArrayPositions) int {
```
mschoch: typically comparison methods returning (-1, 0, 1) have "compare" in the name, not equals (typically bool)

search/searcher/search_nested.go
```diff
@@ -0,0 +1,154 @@
...
+	} else {
+		for k := range c.termAPMap {
+			list := c.termAPMap[k]
+			c.termAPMap[k] = list[:0]
+		}
+	}
+
+	termList := make(map[string]struct{}, 1)
+	for _, ftl := range d.FieldTermLocations {
+		if len(ftl.Location.ArrayPositions) == c.arrayDepth {
```
mschoch: What is special about the FTL having the exact depth requested as the query array depth?  I thought this was designed such that you could specify the minimum depth that you care about.  In the book example, depth 1 might mean same chapter, depth 2 might mean same page, etc.  I could query for depth 1, meaning I want to find matches that are in the same chapter, but the hits all actually have depth 3, because the matching terms occur inside pages, which are nested inside of chapters...

search/searcher/search_nested.go
```diff
@@ -0,0 +1,154 @@
...
+				ftl.Location.ArrayPositions)
+		}
+		termList[ftl.Term] = struct{}{}
+	}
+
+	// exit early if any of the search terms are not found
+	// at the given array depth.
+	for term := range termList {
+		if _, ok := c.termAPMap[term]; !ok {
+			return false
```
mschoch: Again, relating to my previous comment, I expected matches could be deeper than the requested depth, and that still makes sense.  Is there a use case for giving the user the choice of either "exact depth" vs "at least depth"?

search/searcher/search_nested.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// checkForIntersection check the existence of common
+// elements in a list of array position lists.
+func checkForIntersection(apLists [][]search.ArrayPositions) bool {
+	if len(apLists) == 1 {
+		return true
+	}
+
+	if len(apLists) == 2 && len(apLists[0]) == 1 && len(apLists[1]) == 1 {
```
mschoch: is this just a special case optimization? perhaps a comment could add justification

search/searcher/search_nested.go
```diff
@@ -0,0 +1,154 @@
...
+		if equalAPSlices(apLists[0][0], apLists[1][0]) == 0 {
+			return true
+		}
+	}
+
+	var baseIndex int
```
mschoch: from here down I don't really understand what the code is doing.  i don't have a concrete suggestion, just that the variable names, operations, and overall behavior don't fit.  Like here are some questions.  What does "smallestListOver" mean?  What is a baseIndex?  I expected we'd be ranging over the apLists, but instead we use baseIndex and some hard-coded 0's?

search/searcher/search_nested.go
sreekanth-cb: right.

search/searcher/search_nested.go
sreekanth-cb: hm. the intent was only consider the matches at the asked depth/exact depth and then check their array positions equality. 
Need to think more on the "exact depth" vs "at least depth" use cases.

at least depth - gives an impression that user isn't aware of the document structure or doesn't care about the match positions (except the exclusion factor of atleast) - but then he can stick to the blind search with out the depth parameter ?

search/searcher/search_nested.go
sreekanth-cb: hm. this code tries to find any single common element out of the given ordered array positions lists for the terms matched. 
smallestListOver -  stands for the arrayPosition lists for the term with the lowest number of matches/shorted array position list.   
for eg:  It tries to find a common element from a list of array position list of terms like below. (made up example)
[[0], [1], [0,2], [0,1,2]
[[0], [3], [0,2]
[[0], [4], [0,2], [1,2,0]

search/searcher/search_nested.go
sreekanth-cb: It was designed for exact depth. Now in book sample documents under tests, 
{book [chapters : [pages]]} . results in array Positions slices of depth 2, not depth 3.

search/searcher/search_nested.go
sreekanth-cb: @mschoch ,  to summarise my understanding of the at least array depth now is as below,
Please correct me,

In the book example, 
an array_depth of 1 in means - if matches occur within different/same pages of a particular chapter. Or the chapter level itself like a chapter title.

an array_depth of 2 in  means - if matches occur within a single page of any particular chapter.

If so then - after getting all array position lists of the matched terms in a document, we need to check whether any of those array position slices per matched term share a common prefix upto a length of given (array_depth - 1).   

Why (array_depth - 1) -> As I see a depth of only 2 when there is a 2 level hierarchy of of chapters and pages as in the samples d.json under integration tests.

steveyen:
Hi -- just to keep folks updated -- @sreekanth-cb and I ended up having a separate conversation about this a day or so ago, talking about @mschoch's feedback and sense that we ought to be able to get to the "=" and "<=" cases here ("either "exact depth" vs "at least depth") -- and that it's worth noodling more about that as there's user facing API impact here, too.

search/searcher/search_nested.go
```diff
@@ -0,0 +1,154 @@
...
+	return c.checkArrayDepth
+}
+
+func (c *depthFilterContext) checkArrayDepth(d *search.DocumentMatch) bool {
+	if d == nil || len(d.FieldTermLocations) == 0 {
+		return false
+	}
+
+	// reset and resuse
+	if len(c.termAPMap) == 0 {
```
steveyen: maybe this should check for nil rather than 0 len?

search/query/conjunction.go
```diff
@@ -92,8 +102,9 @@ func (q *ConjunctionQuery) Validate() error {
 
 func (q *ConjunctionQuery) UnmarshalJSON(data []byte) error {
 	tmp := struct {
-		Conjuncts []json.RawMessage `json:"conjuncts"`
-		Boost     *Boost            `json:"boost,omitempty"`
+		Conjuncts  []json.RawMessage `json:"conjuncts"`
+		Boost      *Boost            `json:"boost,omitempty"`
+		ArrayDepth int               `json:"array_depth,omitempty"`
```
steveyen: One of the thoughts... is that array-depth feels like a number that goes "downwards"...

Like, given a candidate document, are the candidate terms within some allowed depth, as measured from the root or top of the document?  (That is, if I understand arrayDepth correctly -- which I might not!)

I was also playing around with thoughts of going the opposite direction, or using a relative param.

Like, given a candidate document, and you see a matching term -- how many other terms are nearby "relatives" -- kind of pretending that the document hierarchy represents a family tree.

For example, perhaps I only want to find immediate siblings (orderline product == "t-shirt" AND same orderline's size is "x-large").  But, please ignore those first cousins and second cousins.

Or, maybe siblings & first cousins are (find "Bill" and "Clinton" within the same chapter), but 2nd cousins are not good enough ("Bill" and "Clinton" mentioned by books on the same shelf isn't good enough).

Something like that.  That is, wondering if that conceptual direction would handle more generic cases (trying to hit the == and <= intuitions mentioned earlier by @sreekanth-cb and @mschoch before).

And, of course, if that works, g'luck naming such a thing!  :-)

sreekanth-cb: @steveyen , on further thinking - i feel some impedance mismatch here with this new hierarchical relation among the matched terms in a document Vs the current PR changes.

1. Though it looked alike on first read, Its slightly different from the original issue we intend to address through this PR. Here we are specifically focused on/processing term matches using their array positions as a yard stick. Or the approach won't work if we have an empty array position lists. But even in that case, your proposed hierarchical match criteria is valid and feel that is a more generic or broader use case of this approach. Current approach now feels very narrow or tailor made for a particular problem at hand. 
Since we don't have any other hierarchical context information than array position lists,  not sure how to approach this in such a generic way as don't have the full document with us. (also for documents matches where array positions aren't applicable).

2. When we think of figuring sibling relations from a bottom up manner - it feels like that mandates more/exact know how from the user on the whole document structure so that he can specify a level (int) information from the deepest level of his interest to the highest level.?
Where in the current approach, leaves more leeway for user to just convey things from a top down way. For eg: at least this deep and at more depth (could be unknown doc structure ).

3. Though initially, I had qualms over the arrayDepth name, now upon re visiting this PR after a few days - it looks more reasonable as the whole checks/verification revolves around the array position lists. This PR/approach targets only the documents that have sub arrays within it..

mschoch: The problem with "array depth" is that it's all about how we think about the implementation, and neither the name, nor the exact behavior will really make sense to a user.

I like @steveyen's proposal from that perspective.

The big concern I have is that "relative" has to be rooted somewhere, and as we've described it so far it's a just a bag of hits we're filtering, so there is no logical starting place.  The only variation I've come up with is that we could compute like a max distance.  Like you could find the deepest and shallowest hits and consider it a hit if maxD - minD was within that range (or 0 for exact same depth).

Any thoughts on our thoughts @steveyen?

sreekanth-cb: @mschoch , what if there is no array positions available in the document match? I mean the document just had a hierarchy. But there were no [] of objects. Not clear on how do we get the sense of depth for a term's occurrence?

Issue #1222, opened 2019-05-24
indexing options with `docvalues` disabled is meaningless for geopoint fields


sreekanth-cb:
As the current implementation of geo queries are powered by doc values, there is no need to make docValue indexing option configurable for geopoint mapping. docvalue can be always enabled for geopoints.
Else, since we have exposed docValue indexing options over the type mapping, there is a chance that user may ignorantly disable the docValue option for geo fields during the indexing process.

[docvalue option is made configurable to better control the index size when geo/sort/facet queries aren't used]

PR #1223, opened 2019-05-24
 Enable docvalues option for geopoints always - #1222

sreekanth-cb:
Always enable docvalues option for geopoints


abhinavdangeti:
The commit header should carry a brief description for what the change implements, rather than "Fix for issue .." :)

sreekanth-cb:
@mschoch , I made this change thinking that the geo queries can't be served without docvalues. Hence thought of overriding the docvalue option to true always.
You have any better thoughts of handling this?

sreekanth-cb:
Indeed. Thats the right way to do this.

PR #1227, opened 2019-05-28
MB-33364 - perf wrinkles with geo queries

sreekanth-cb:
Geo queries are reported to have high memory consumption
and higher latencies.
One plausible root cause for this could be,
-Very high number of candidate terms computed during
 the bounded box ranges. (unlike other queries, looks like
 these terms are just mathematically created without actual
 existence in the bleve index)
-Too many candidate terms keeps more TermFieldReaders and
 term searcher objects live and manage over the disjunction heap.

This fix tries to filter out the onBoundary/notOnBoundary
terms early in the search path so that it creates
lesser garbage from TFRs, term byte searchers and lighter
disjunction heap management.

Fix have shown good improvement with numbers in local/mac
environment against the functional tests with
1000 documents and num_queries=100

Average FTS latency in ms for 100 queries
63.20892059(new) Vs 189.9222353(existing)

~3X improvement in average latency is observed with
latency reduction upto 6X for certain queries.

So the latency spikes got better normalised.
eg:
New ms             Vs          Old ms

80.392502	 Vs    427.297942
306.575176	 Vs    1555.539272
435.612816	 Vs    2431.383857
307.829132	 Vs    1402.138768
148.749478	 Vs     936.793944
293.825909	 Vs    944.349282
390.313823	 Vs    469.521313
620.221195	 Vs    2414.491882
455.174285      Vs   1655.026496
185.31835         Vs .  746.681948

### DIFF
--- a/search/searcher/search_geoboundingbox.go
+++ b/search/searcher/search_geoboundingbox.go
@@ -40,7 +40,7 @@ func NewGeoBoundingBoxSearcher(indexReader index.IndexReader, minLon, minLat,
 
 	// do math to produce list of terms needed for this search
 	onBoundaryTerms, notOnBoundaryTerms, err := ComputeGeoRange(0, GeoBitsShift1Minus1,
-		minLon, minLat, maxLon, maxLat, checkBoundaries)
+		minLon, minLat, maxLon, maxLat, checkBoundaries, indexReader, field)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


sreekanth-cb:
hm, my thinking was that the TFRs are getting re-used here:
https://github.com/blevesearch/bleve/blob/master/index/scorch/snapshot_index.go#L439
https://github.com/blevesearch/bleve/blob/master/index/scorch/snapshot_index_tfr.go#L182
So may be not that bad?

mschoch:
True, I guess it's better than I thought, but we do go through a mutex every single time (something that had external reuse would also be able to avoid that).

sreekanth-cb:
Indeed @mschoch, good point! 
Let me try checking newer look up apis for the dictionary and that can be a separate change.

PR #1228, opened 2019-05-29
WIP - dictionary random Lookup api

sreekanth-cb:
draft version of having a new
Exist iterator api implementation
for the FieldDictionary

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -121,6 +121,11 @@ type IndexReaderOnly interface {
 	FieldDictOnly(field string, onlyTerms [][]byte, includeCount bool) (FieldDict, error)
 }
 
+// better naming everywhere
+type IndexReaderRandom interface {
+	FieldDictRandom(field string) (AdvFieldDict, error)
+}
+
 // FieldTerms contains the terms used by a document, keyed by field
 type FieldTerms map[string][]string
 
@@ -228,6 +233,11 @@ type DictEntry struct {
 type FieldDict interface {
 	Next() (*DictEntry, error)
 	Close() error
+}
+
+type AdvFieldDict interface {
+	FieldDict
+	Exists(key []byte) (error, bool)
 }
 
 // DocIDReader is the interface exposing enumeration of documents identifiers.
--- a/index/scorch/segment/empty.go
+++ b/index/scorch/segment/empty.go
@@ -91,10 +91,24 @@ func (e *EmptyDictionary) OnlyIterator(onlyTerms [][]byte,
 	return &EmptyDictionaryIterator{}
 }
 
+func (e *EmptyDictionary) ExistsIterator() AdvDictionaryIterator {
+	return &EmptyDictionaryLookupIterator{}
+}
+
 type EmptyDictionaryIterator struct{}
 
 func (e *EmptyDictionaryIterator) Next() (*index.DictEntry, error) {
 	return nil, nil
+}
+
+type EmptyDictionaryLookupIterator struct{}
+
+func (e *EmptyDictionaryLookupIterator) Next() (*index.DictEntry, error) {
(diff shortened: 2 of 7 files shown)


mschoch:
Here are some initial thoughts on this:

1.  This PR focuses very specifically on exactly what you need (an Exists() method) and not a more general random access interface to the dictionary.  You could argue this is good or bad, just an observation.

2.  You end up with new methods inside scorch to create and AdvFieldDict, but it seems to me like its cleaner to just say in Scorch any method that creates a FieldDict returns one that is also AdvFieldDict.  I think streamlining this would reduce the amount of code.

3.  Going back to the previous PR that led to me asking about this.  In hindsight, I'm wondering why we couldn't just sort the full list of terms, and then use the existing FieldDict interface to do the filtering?  I'm assuming that the perf penalty was from actually looking for all those terms (even the ones that didn't exist), and not just from returning them in the list.  And while the list of terms is large, I don't think it is so large that sorting it is problematic.  Any thoughts on this?

4.  Have you measured perf with this to see if it even makes significant difference?

sreekanth-cb:
> Here are some initial thoughts on this:
> 
> 1. This PR focuses very specifically on exactly what you need (an Exists() method) and not a more general random access interface to the dictionary.  You could argue this is good or bad, just an observation.

Indeed, this change was no-brainer quick hack to get the numbers. 

> 2. You end up with new methods inside scorch to create and AdvFieldDict, but it seems to me like its cleaner to just say in Scorch any method that creates a FieldDict returns one that is also AdvFieldDict.  I think streamlining this would reduce the amount of code.

It does make sense. Will do this once we are going to take this change.

> 3. Going back to the previous PR that led to me asking about this.  In hindsight, I'm wondering why we couldn't just sort the full list of terms, and then use the existing FieldDict interface to do the filtering?  I'm assuming that the perf penalty was from actually looking for all those terms (even the ones that didn't exist), and not just from returning them in the list.  And while the list of terms is large, I don't think it is so large that sorting it is problematic.  Any thoughts on this?

This approach indeed works and tried this. I liked this approach as,
-we filter the terms by going over the FST using the automaton iterator and it all happens in a single go Or perhaps this result in better disk access patterns.

But worry point would be the memory footprint,
-Sorting huge lists. It could be in order of 100K items or more.
-not sure of the memory friendliness of the FieldDictOnly interface. It creates an in memory FST with that many terms and use that automaton to filter the terms. 
 So in memory footprint of  sorted terms + custom FST?

> 4. Have you measured perf with this to see if it even makes significant difference?

Now these next level optimisation decision needs a better stand alone perf tests like that of showfast.
I can't trust or decide the minute details from my local.
Did run the measurements.

Current master avg latency(with last PR) - 63.20892059ms
FieldDictOnly filter method latency - 63.61678323/61.33661654
This PR latency - 56.00787743ms 

As these are not differing much - I am thinking that let's polish this change and let me benchmark these once I have a showfast/high scale perf run to get the eventual numbers. 
We can decide that time?

sreekanth-cb:
Another more convincing data point derived for Exists api or the new FieldDictRandom interface in this PR.
I did a quick hack of the same new Exist method into the existing numeric filters and did a toy build run on numeric range queries in Showfast.
ref: https://github.com/sreekanth-cb/bleve/commit/765cb3fca3635c5c51acfe4d38b2204bbd960106

The latency remained unaffected. But throughput improved 44%. 
From 55 - 85 QPS.
This gives confidence to move ahead with this approach.
Also, attaching the cpu profile for comparison.

Older
FieldDictOnly

Newer CPU
FieldDictRandom_New

sreekanth-cb:
Opened a new PR #1233 for taking this change further.

PR #1236, opened 2019-06-04
Geo adopts the new field dictionary Exists (random look up) API

sreekanth-cb:
New Exists iterator api implementation for the FieldDictionary

geo queries adoption for the Exists api to filter the candidate terms.
This method performs better than the existing RangeIterator and OnlyIterator methods on memory and cpu usage for random dictionary lookups.

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -121,6 +121,11 @@ type IndexReaderOnly interface {
 	FieldDictOnly(field string, onlyTerms [][]byte, includeCount bool) (FieldDict, error)
 }
 
+// better naming everywhere
+type IndexReaderRandom interface {
+	FieldDictRandom(field string) (AdvFieldDict, error)
+}
+
 // FieldTerms contains the terms used by a document, keyed by field
 type FieldTerms map[string][]string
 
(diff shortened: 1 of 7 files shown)


index/index.go
```diff
@@ -230,6 +234,11 @@ type FieldDict interface {
 	Close() error
 }
 
+type AdvFieldDict interface {
+	FieldDict
+	Exists(key []byte) (bool, error)
+}
+
```
steveyen: One thought... instead of having AdvFieldDict extend the FieldDict interface, maybe this new interface can just have the Exists() method all by itself?  Perhaps that might help clean up the code, if the interface-casting check can then happen in simpler ways?  e.g., wondering if upsidedown's dictionary iterators just don't implement the Exists() method, and only scorch dictionary iterators have that new method?

Related, there seem to be a few synonyms for this new fast-path approach... "Random" access lookup of a term, "Adv"anced lookup, or whether a key or term "Exists" in the dictionary.  Would recommend instead just going with just a single concept name.  If you've no strong preference, I think I'd prefer Exists or (perhaps even KeyExists or TermExists if "Exists" by itself is too generic).

sreekanth-cb: good catch, thanks. done!

PR #1238, opened 2019-06-05
add test cases for multi-sort with search before/after

tylerkovacs:
This PR adds new test cases for sorts with tied (ie. identical) values.

The tests pass using the default index and keyvalue types:

```
$ go test ./test/... -count=1 -v -run TestIntegration
=== RUN   TestIntegration
--- PASS: TestIntegration (0.67s)
    integration_test.go:49: using index type upside_down and kv type boltdb
    integration_test.go:71: Running test: alias
    integration_test.go:71: Running test: basic
    integration_test.go:71: Running test: employee
    integration_test.go:71: Running test: facet
    integration_test.go:71: Running test: fosdem
    integration_test.go:71: Running test: geo
    integration_test.go:71: Running test: phrase
    integration_test.go:71: Running test: sort
PASS
ok  	_/home/tyler/src/bleve/test	0.685s
```

But the same tests fail when using `scorch`:

```
$ go test ./test/... -count=1 -v -run TestIntegration -indexType=scorch -kvType=scorch
=== RUN   TestIntegration
--- FAIL: TestIntegration (1.11s)
    integration_test.go:49: using index type scorch and kv type scorch
    integration_test.go:71: Running test: alias
    integration_test.go:71: Running test: basic
    integration_test.go:71: Running test: employee
    integration_test.go:71: Running test: facet
    integration_test.go:71: Running test: fosdem
    integration_test.go:71: Running test: geo
    integration_test.go:71: Running test: phrase
    integration_test.go:71: Running test: sort
    integration_test.go:150: test error - sort by name, ascending
    integration_test.go:151: test 0 - expected hit 2 to have ID a got g
    integration_test.go:150: test error - sort by name, ascending
    integration_test.go:151: test 0 - expected hit 3 to have ID g got a
    integration_test.go:150: test error - sort by name, descending
    integration_test.go:151: test 1 - expected hit 2 to have ID a got g
    integration_test.go:150: test error - sort by name, descending
    integration_test.go:151: test 1 - expected hit 3 to have ID g got a
    integration_test.go:150: test error - sort by name, descending, missing first
    integration_test.go:151: test 2 - expected hit 3 to have ID a got g
    integration_test.go:150: test error - sort by name, descending, missing first
    integration_test.go:151: test 2 - expected hit 4 to have ID g got a
    integration_test.go:150: test error - sort by name, ascending, before nancy
    integration_test.go:151: test 14 - expected hit 2 to have ID a got g
    integration_test.go:150: test error - sort by name, ascending, before nancy
    integration_test.go:151: test 14 - expected hit 3 to have ID g got a
FAIL
FAIL	_/home/tyler/src/bleve/test	1.130s
```

The failures occur on the "tied" values. Scorch appears to simply returned tied values in a different order than the default index/kv.

Results are returned in the expected order when using a multi-valued sort so that the results no longer tie.

This doesn't impact my usage of bleve (since I've always used scorch) but I'm opening this PR in case it is news/noteworthy for you. Feel free to simply close if the issue is known and/or not considered a problem.

### DIFF
--- a/test/tests/sort/data/g.json
+++ b/test/tests/sort/data/g.json
@@ -1,0 +1,8 @@
+{
+	"id": "g",
+	"name": "marty",
+	"age": 27,
+  "born": "2012-09-21",
+	"title": "mista",
+	"tags": ["foo", "bar"]
+}
--- a/test/tests/sort/searches.json
+++ b/test/tests/sort/searches.json
@@ -10,7 +10,7 @@
       "sort": ["name"]
 		},
 		"result": {
-			"total_hits": 6,
+			"total_hits": 7,
 			"hits": [
 				{
 					"id": "c"
@@ -20,6 +20,9 @@
         },
         {
           "id": "a"
+        },
+        {
+          "id": "g"
         },
         {
           "id": "e"
@@ -44,7 +47,7 @@
       "sort": ["-name"]
 		},
 		"result": {
-			"total_hits": 6,
+			"total_hits": 7,
 			"hits": [
         {
           "id": "b"
@@ -54,6 +57,9 @@
         },
         {
           "id": "a"
+        },
+        {
+          "id": "g"
         },
         {
           "id": "f"
@@ -78,7 +84,7 @@
       "sort": [{"by":"field","field":"name","missing":"first","desc":true}]
 		},
 		"result": {
-			"total_hits": 6,
+			"total_hits": 7,
 			"hits": [
         {
           "id": "d"
@@ -91,6 +97,9 @@
         },
         {
           "id": "a"
+        },
+        {
+          "id": "g"
         },
         {
           "id": "f"
@@ -112,7 +121,7 @@
       "sort": ["age"]
 		},
 		"result": {
-			"total_hits": 6,
+			"total_hits": 7,
 			"hits": [
 				{
 					"id": "f"
@@ -125,6 +134,9 @@
         },
         {
           "id": "c"
+        },
+        {
+          "id": "g"
         },
         {
           "id": "d"
@@ -146,11 +158,14 @@
       "sort": ["-age"]
 		},
 		"result": {
-			"total_hits": 6,
+			"total_hits": 7,
 			"hits": [
 				{
 					"id": "d"
 				},
+        {
+          "id": "g"
+        },
         {
           "id": "b"
         },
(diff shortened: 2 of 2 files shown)


mschoch:
OK, I've gone back to review this closer, there a few issues that all come together here:

1. Upsidedown and Scorch have different natural index order (the ordering of results when nothing was specified, or when 2 or more items are tied after comparing with the specified order)  This alone makes tests like this which do not impose a total ordering prone to fail.  This makes our job testing harder, but we accepted it as it's actually a useful feature.  Upsidedown essentially does more work to impose order, even when you don't care, and scorch improves this by letting you impose full order, or not, it's up to you.

2.  Scorch's natural index order isn't even stable.  Not only is it different from upsidedown, it can change from request to request.  If you care about the order, you must impose the order by sorting, you should not make any assumptions about the natural index order.

3.  Using the `SearchAfter` and `SearchBefore` capabilities **REQUIRES** that you impose a total order.  The tests in this PR actually fail due to 1 & 2 above, but there is a deeper problem.  Let's say you sort results by `name` as some of these tests do.  And lets say the last result on a page showing 10 hits has sort value ["marty"].  To access the next page, you specify `SearchAfter: ["marty"]`, but in fact hits 11 and 12 also had sort value ["marty"].  When you go to page 2 you will have skipped over those hits as well.  The only way to avoid this is to ensure that a hit's sort value uniquely identifies a single document.  For most applications, it is sufficient to simply add "_id" as the last sort column, changing the above example to `["marty", "a"]` for the 10th hit, ensuring that page 2 then starts with and correctly shows ["marty", "g"], etc.

I'm open to adding more documentation around this if you have any suggestions.

tylerkovacs:
Thanks @mschoch, that all makes sense. I discovered this "issue" while trying to create tests to reproduce my actual problem (more on that below). I'll close the PR since you're already aware of the behavior and have good reasons for keeping this way.

In my application, I'm using a total order with `_id` as a tiebreaker. Here's an example:

`searchRequest.SortBy([]string{"-_score", "rank", "_id"})`

Where:
* `-_score` orders by relevance to a text query.
* `rank` is a `NumericFieldMapping` set by my application and is unique per document.
* `_id` as a fallback tiebreaker (just in case a bug causes `rank` duplicate `rank` values)

I've yet to reproduce the issues I'm seeing with `SearchAfter/SearchBefore` in test cases. I tried to recreate it using the declarative test cases (ie. the ones I changed in this PR) but they don't work well for this issue for a few reasons:
* `_score` is a dynamic value and doesn't fit well in the declarative tests.
* `rank` is mapped to an odd (binary?) representation of the number and also cannot be expressed easily in a declarative test.

I'll work on an imperative test case soon but one question comes to mind with respect to `_score` and `MultiSearch` via `IndexAlias`. I suspect that `_score` might be dependent on the documents within each `Index` making it unsuitable for use within `SearchBefore/SearchAfter`. I'm not sure I'm explaining that clearly so perhaps it's best to just write the imperative test case...

PR #1241, opened 2019-06-06
Numeric range query adopts Exists Iterator

sreekanth-cb:
Numeric range query to use the new Exists iterator
from Vellum for filtering the candidate terms for
scorch index types.
Empirical evidence shows that this brought about
44% throughput improvements for qps.

### DIFF
--- a/search/searcher/search_numeric_range.go
+++ b/search/searcher/search_numeric_range.go
@@ -53,20 +53,49 @@ func NewNumericRangeSearcher(indexReader index.IndexReader,
 	if !*inclusiveMax && maxInt64 != math.MinInt64 {
 		maxInt64--
 	}
+
+	var fieldDict index.FieldDictExists
+	var isIndexed filterFunc
+	var err error
+	if irr, ok := indexReader.(index.IndexReaderExists); ok {
+		fieldDict, err = irr.FieldDictExists(field)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


sreekanth-cb:
Existing CPU Profile
image

Newer Profile
image

search/searcher/search_numeric_range.go
```diff
@@ -53,20 +53,51 @@ func NewNumericRangeSearcher(indexReader index.IndexReader,
...
+		}
+	}
+
+	var isIndexed filterFunc
+	if fieldDict != nil {
+		isIndexed = func(term []byte) bool {
+			if found, err := fieldDict.Exists(term); found && err == nil {
+				return true
+			}
+			return false
```
steveyen: minor thing -- (change not needed unless if there are other changes, imho)...

this might be done via a more compact... return err == nil && found

abhinavdangeti:
Hey @sreekanth-cb, looks like you're bleve change did not pick up your vellum change. Rebasing would make the travis-ci checks pass.

abhinavdangeti:
On second look - I take that back. FieldDictExists is introduced with https://github.com/blevesearch/bleve/pull/1236.

And as for filterFunc which is also introduced with the above PR, it needs to put in a better location so both searchers can use the function type.

Issue #1249, opened 2019-06-14
searchRequest.Fields  not works 


orange-jacky:
type GlobalAirport struct {
		Id           string `bson:"_id" json:"id"`
		Cnname       string `bson:"cnname" json:"cnname"`
		Citycode     string `bson:"citycode" json:"citycode"`
		Citycn       string `bson:"citycn" json:"citycn"`
		Enname       string `bson:"enname" json:"enname"`
		Country      string `bson:"country" json:"country"`
		Province     string `bson:"province" json:"province"`
		Cityen       string `bson:"cityen" json:"cityen"`
		Country_code string `bson:"country_code" json:"country_code"`
		Region_code  string `bson:"region_code" json:"region_code"`
		Region_name  string `bson:"region_name" json:"region_name"`
		City_pinyin  string `bson:"city_pinyin" json:"city_pinyin"`
		City_jianpin string `bson:"city_jianpin" json:"city_jianpin"`
	}

func buildGlobalAirportIndexMapping() (mapping.IndexMapping, error) {
	// a generic reusable mapping for english text
	englishTextFieldMapping := bleve.NewTextFieldMapping()
	englishTextFieldMapping.Analyzer = en.AnalyzerName

	// a generic reusable mapping for keyword text
	keywordFieldMapping := bleve.NewTextFieldMapping()
	keywordFieldMapping.Analyzer = keyword.Name

	aircompanyMapping := bleve.NewDocumentMapping()

	aircompanyMapping.AddFieldMappingsAt("id", englishTextFieldMapping)
	aircompanyMapping.AddFieldMappingsAt("cnname", englishTextFieldMapping)
	aircompanyMapping.AddFieldMappingsAt("citycode", englishTextFieldMapping)
	aircompanyMapping.AddFieldMappingsAt("citycn", englishTextFieldMapping)
	aircompanyMapping.AddFieldMappingsAt("enname", englishTextFieldMapping)
	aircompanyMapping.AddFieldMappingsAt("country", englishTextFieldMapping)
	aircompanyMapping.AddFieldMappingsAt("province", englishTextFieldMapping)
	aircompanyMapping.AddFieldMappingsAt("cityen", englishTextFieldMapping)
	aircompanyMapping.AddFieldMappingsAt("city_pinyin", englishTextFieldMapping)
	aircompanyMapping.AddFieldMappingsAt("city_jianpin", englishTextFieldMapping)

	indexMapping := bleve.NewIndexMapping()
[...]

PR #1254, opened 2019-07-01
MB-34817: Prototype for multi_match query

abhinavdangeti:
Directed towards addressing: https://github.com/blevesearch/bleve/issues/1249


abhinavdangeti:
Hey @mschoch Is this worth considering?

mschoch:
Two thoughts:

1.  There seems to be some complex corners to the code, like you can specify fields or analyzer or both, or neither, and there is some behavior defined for all of these.  Can we have a short call where you walk through these?

2.  ES has a multi-match query that looks similar, how similar is it?

abhinavdangeti:
Sounds good. Let me do a quick recap on this myself. I had looked into ES's approach on this, but I'll need a refresh on that too since it's been a while, so I'll do that before we chat.

abhinavdangeti:
Hey @mschoch , I've updated this commit a bit, and updated the description identifying what can be achieved using the multi_match query.
It is similar to ES in some ways. Let me know if we can chat on this on how we can do better?

PR #1255, opened 2019-07-01
Preserve persistedCallbacks of an unpersisted snapshot until needed

abhinavdangeti:
The persistedCallbacks of a snapshot that failed persistance need
to be preserved until one of the two events have occurred:
- a new snapshot was introduced
- the retry-attempt of the exact same snapshot succeeded

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -90,6 +90,10 @@ func (s *Scorch) persisterLoop() {
 	var persistWatchers []*epochWatcher
 	var lastPersistedEpoch, lastMergedEpoch uint64
 	var ew *epochWatcher
+
+	var lastUnpersistedEpoch uint64
+	var unpersistedCallbacks []index.BatchCallback
+
 	po, err := s.parsePersisterOptions()
 	if err != nil {
 		s.fireAsyncError(fmt.Errorf("persisterOptions json parsing err: %v", err))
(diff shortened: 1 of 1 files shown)


abhinavdangeti:
Good point @sreekanth-cb. Let me get a patch up here.

index/scorch/persister.go
```diff
@@ -149,11 +152,25 @@ OUTER:
...
+				// the retry attempt
+				unpersistedCallbacks = append(unpersistedCallbacks, ourPersistedCallbacks...)
+
 				s.fireAsyncError(fmt.Errorf("got err persisting snapshot: %v", err))
 				_ = ourSnapshot.DecRef()
 				atomic.AddUint64(&s.stats.TotPersistLoopErr, 1)
 				continue OUTER
 			}
+
+			if unpersistedCallbacks != nil {
```
sreekanth-cb: shouldn't we need an  " unpersistedEpoch <= ourSnapshot.Epoch " check as well to guard against cases of rollbacks where the cached callbacks (unpersistedCallbacks) becomes irrelevant?

abhinavdangeti: Hmm, I think we can do without, as along as there are unpersisted callbacks we should invoke. For even in case of a rollback, a new snapshot of segments will be introduced with a monotonically increasing epoch.

sreekanth-cb: you are right on the epoch numbers, But is it the right thing to fire a callback when things rollbacked and becomes irrelevant?
May be thats a CB specific concern, not a bleve specific one.

abhinavdangeti: I guess I'm not seeing what the concern is, could you elaborate a bit on it?
Also, if it's the couchbase case only - a fireAsyncError is treated as Fatal :)

sreekanth-cb: The concern was that - in case of situations like a rollback - the most recent/last snapshots become irrelevant and thus the associated batch contents/callbacks. So in such cases, we don't have to care about appending these cached callbacks and is there a way to identify or handle such cases.

PR #1270, opened 2019-07-31
Fix for potential removeOldData call starve issue

sreekanth-cb:
If the persister is busy doing the catch up with
the ongoing mutations then there is a chance for
the older snapshot clean up call/removeOldData to
wait for a longer time to get invoked. This can
result in unbounded growth in the number of scorch
files.

This change attempt to invoke removeOldData whenever
the number of files grow beyond the thresholds.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -245,20 +245,19 @@ func notifyMergeWatchers(lastPersistedEpoch uint64,
 	return watchersNext
 }
 
-func (s *Scorch) pausePersisterForMergerCatchUp(lastPersistedEpoch uint64, lastMergedEpoch uint64,
-	persistWatchers []*epochWatcher, po *persisterOptions) (uint64, []*epochWatcher) {
+func (s *Scorch) pausePersisterForMergerCatchUp(lastPersistedEpoch uint64,
+	lastMergedEpoch uint64, persistWatchers []*epochWatcher,
+	po *persisterOptions) (uint64, []*epochWatcher) {
 
-	// first, let the watchers proceed if they lag behind
+	// First, let the watchers proceed if they lag behind
 	persistWatchers = notifyMergeWatchers(lastPersistedEpoch, persistWatchers)
 
-	// check the merger lag by counting the segment files on disk,
+	// Check the merger lag by counting the segment files on disk,
+	numFilesOnDisk, _ := s.diskFileStats()
+
 	// On finding fewer files on disk, persister takes a short pause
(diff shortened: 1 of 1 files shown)


sreekanth-cb:
@steveyen , given the steady increase in the number of scorch index files on disk (in the use case under consideration), later I had this new theory that the removeOldData call starving could be a more probable reason than the corner case of a potential leak resulting from the un-flip on s.ineligibleForRemoval.

steveyen:
Wondering with the other PR on the ineligible-file-removals... https://github.com/blevesearch/bleve/pull/1271
...whether this PR is still needed.

The extra head-scratcher for me is that it introduces yet another goroutine, and something in the back of my mind is wondering an extra goroutine is avoidable.  Related, if the other PR actually solves the problem, hoping this one isn't needed.

sreekanth-cb:
The other PR won't fix the starve issue or the attempted fixes are for unrelated issues.. 

RemoveOldData starving can still happen  if the persister gets into a busy work loop is the thinking here.  This starve issue is easily reproducible with an artificial Sleep in the persister work loop with steady incoming mutations.

The other PR is more for handling / tightening any potential raciness in setting the removal eligibility flags.

sreekanth-cb:
@steveyen , you are right and thats the reason I initially tried to have that go-routine to make it a non blocking call.
 
Now, I have tried the toy runs for these changes(5 test runs) and the throughput looked intact for CB. No regressions observed.

http://perf.jenkins.couchbase.com/job/atlas-sdk/6109/console 
to
http://perf.jenkins.couchbase.com/job/atlas-sdk/6113/console.

index/scorch/persister.go
```diff
@@ -245,20 +245,19 @@ func notifyMergeWatchers(lastPersistedEpoch uint64,
 	return watchersNext
 }
 
-func (s *Scorch) pausePersisterForMergerCatchUp(lastPersistedEpoch uint64, lastMergedEpoch uint64,
-	persistWatchers []*epochWatcher, po *persisterOptions) (uint64, []*epochWatcher) {
+func (s *Scorch) pausePersisterForMergerCatchUp(lastPersistedEpoch uint64,
+	lastMergedEpoch uint64, persistWatchers []*epochWatcher,
+	po *persisterOptions) (uint64, []*epochWatcher) {
```
steveyen: btw, an extra newline popped up here?

steveyen: or, perhaps it was always there

Issue #1269, opened 2019-07-31
Potential file leaks with scorch indexes


sreekanth-cb:
Its seen with one of the under resourced CB cluster that - after a few days of live run, the number of scorch files on disk started growing. But the num of files at root always remained under limit. ~20 all the time.

Once we continue to track the number of scorch file growth over a period of time for these scorch indexes, the number of files started to grow steadily once the merger started treading ahead of the persister.  Once this happened - the file count is consistently climbing.

```
CurOnDiskFiles : 7612
--
CurRootEpoch : 12502409
LastMergedEpoch : 12502405
LastPersistedEpoch : 12502396
MaxBatchIntroTime : 2946750904
 
CurOnDiskFiles : 13297
CurRootEpoch : 12547973
LastMergedEpoch : 12547971
LastPersistedEpoch : 12547966
 
CurOnDiskFiles : 20182
CurRootEpoch : 12635253
LastMergedEpoch : 12635236
LastPersistedEpoch : 12635228
 
continues to climb like below.
 
CurOnDiskFiles : 48619
CurRootEpoch : 13072459
LastMergedEpoch : 13072440
LastPersistedEpoch : 13072435
 ```

Eventually, the persister do catches up with the merger, but that won't help fixing the suspected file leak here.

A non-proven theory pops up in mind is that - 

Merger during the merge process - flips the removal eligibility of the newly created bigger segment files here      https://github.com/blevesearch/bleve/blob/master/index/scorch/merge.go#L192
Persister makes these file based segments back to removal eligibility here- https://github.com/blevesearch/bleve/blob/master/index/scorch/persister.go#L584
Now since the Merger is ahead of the persister and nature of work load is more of data updates, there is a chance that the newly introduced merged file segments by the merger in step1 would have removed/(not a part of latest root) from the latest root by another subsequent file merger round before the persister gets a chance to un flip them. When the persister grabs a recent root - it's radar would have skipped another set of file segments which were a part of the few older skipped root versions.  This results in files leaking as ineligible for removal and thus
s.ineligibleForRemoval slice gets bigger with stale entries and this prevents the clean up of files from disk here:
https://github.com/blevesearch/bleve/blob/master/index/scorch/persister.go#L215
https://github.com/blevesearch/bleve/blob/master/index/scorch/persister.go#L892
 
[...]

PR #1271, opened 2019-07-31
Tightening the scorch stale file clean ups

sreekanth-cb:
Fix for #1269.
Its observed in an under resourced situation that
the merger was treading ahead of the persister and the
scorch files were piling up. It was reaching upto 70K
with an update heavy workload.

The intention of this fix is to unflip any
ineligible files which are genuinely eligible for
clean up.

Also, adding stats to track this for future.

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -157,6 +158,7 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 
 	// process tasks in serial for now
 	var notifications []chan *IndexSnapshot
+	var filenames []string
 	for _, task := range resultMergePlan.Tasks {
 		if len(task.Segments) == 0 {
 			atomic.AddUint64(&s.stats.TotFileMergePlanTasksSegmentsEmpty, 1)
@@ -181,6 +183,12 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 						segmentsToMerge = append(segmentsToMerge, zapSeg)
 						docsToDrop = append(docsToDrop, segSnapshot.deleted)
 					}
+					// track the files getting merged for unsetting the
+					// removal ineligibility. This helps to unflip files
+					// even with fast merger, slow persister work flows.
+					path := zapSeg.Path()
+					filenames = append(filenames,
+						strings.TrimPrefix(path, s.path+string(os.PathSeparator)))
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


index/scorch/persister.go
```diff
@@ -583,6 +583,7 @@ func (s *Scorch) persistSnapshotDirect(snapshot *IndexSnapshot) (err error) {
 	for _, filename := range filenames {
 		delete(s.ineligibleForRemoval, filename)
 	}
+	s.stats.CurFilesIneligibleForRemoval = uint64(len(s.ineligibleForRemoval))
```
steveyen: maybe instead of carefully maintaining CurFilesIneligibleForRemoval every time s.ineligibleForRemoval is updated...

instead during the stats call read the latest len()?

index/scorch/persister.go
```diff
@@ -765,12 +766,15 @@ func (s *Scorch) removeOldData() {
...
-	if removed > 0 {
+	// it is safer to attempt the cleaning of stale files even if there were no
+	// snapShot purges.
+	if removed >= 0 {
 		err = s.removeOldZapFiles()
 		if err != nil {
 			s.fireAsyncError(fmt.Errorf("got err removing old zap files: %v", err))
 		}
 	}
+	atomic.AddUint64(&s.stats.TotRemoveOldDataCompleted, 1)
```
steveyen: maybe stats for # of actual files that were removed versus skipped due to ineligibility would give a better picture?  so, some stat that's incremented inside removeOldZapFiles()?

now that I see the removed >= 0 check, wondering if the check is needed at all?  it's always true?

index/scorch/merge.go
```diff
@@ -265,7 +267,9 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 			}
 		}
 	}
-
+	// once the newly merged segment introduction is done,
+	// its safe to unflip the removal ineligibility.
+	s.unmarkIneligibleForRemoval(filename)
```
steveyen: this seems to be the key change.

One thought, though, is sometimes filename might be ""?

And, another thought is there's a for-loop before this that ranges over several filenames, but this is unmarking a single filename?

Also, handing the segment off to the introducer channel might not be enough?  E.g., it's in the channel, but the introducer hasn't actually gotten around to doing its work yet?

sreekanth-cb: yikes... very true.. I went blind with this change..

sreekanth-cb: Once we receive the notifications back from the introducer, the introductions are actually done?, But need to take care of all the files.

index/scorch/persister.go
sreekanth-cb: yep. make sense!

index/scorch/persister.go
sreekanth-cb: I suspect another wrinkle here with the file clean ups. 
ie whenever we retain a couple of older snapshots, we aren't really taking that fact into consideration while deleting the older files. 
s.loadZapFileNames() - as of now it returns the current live files at the root, instead it should give files from all snapshots(older+root) kept for rollback purpose and purge the rest of files?

PR #1273, opened 2019-08-05
Replace for-loop that calls HasNext/Next() with roaring.IntPeekable

alldroll:
Roaring bitmap library now has IntPeekable iterator implementation, that allows skipping values instead of using a naive sequential iteration with `HasNext/Next()`


abhinavdangeti:
Thanks for this contribution @alldroll. While we review it, would you please go ahead and sign the CLA (instructions here: https://github.com/blevesearch/bleve/blob/master/CONTRIBUTING.md).
Once that's set up we can go ahead and get this change in soon.

PR #1276, opened 2019-08-12
add method to access analyzer of a text field

mschoch:
this change allows applications to see the analyzer set for
a text field, useful if applications want to modify the document
after mapping, but prior to indexing.


mschoch:
cc @aviadl

PR #1277, opened 2019-08-13
Improve large segment waste

mschoch:
Do not merge to master, this is attempt for me to rebase @sreekanth-cb's PR onto latest master.  In the meantime @steveyen had introduce the MemUvarIntReader thing, so I attempted to blend the two.

Please do not laugh too hard, and review carefully.

### DIFF
--- a/index/scorch/segment/zap/chunk.go
+++ b/index/scorch/segment/zap/chunk.go
@@ -1,0 +1,30 @@
+package zap
+
+import "fmt"
+
+func getChunkSize(chunkMode uint32, cardinality uint64, maxDocs uint64) (uint64, error) {
+	switch chunkMode {
+	case 1024:
+		// legacy chunk size
+		return 1024, nil
+	case 1:
+		return 1, nil
+	case 0:
+		// attempt at simple improvement
+		// theory - the point of chunking is to put a bound on the maximum number of
+		// calls to Next() needed to find a random document.  ie, you should be able
+		// to do one jump to the correct chunk, and then walk through at most
+		// chunk-size items
+		// previously 1024 was chosen as the chunk size, but this is particularly
+		// wasteful for low cardinality terms.  the observation is that if there
+		// are less than 1024 items, why not put them all in one chunk,
+		// this way you'll still achieve the same goal of visiting at most
+		// chunk-size items.
+		// no attempt is made to tweak any other case
+		if cardinality < 1024 {
+			return maxDocs, nil
+		}
+		return 1024, nil
+	}
+	return 0, fmt.Errorf("unknown chunk mode %d", chunkMode)
+}
--- a/index/scorch/segment/zap/chunk.go
+++ b/index/scorch/segment/zap/chunk.go
@@ -1,15 +1,25 @@
 package zap
 
-import "fmt"
+import (
+	"fmt"
+)
+
+// LegacyChunkMode was the original chunk mode (always chunk size 1024)
+// this mode is still used for chunking doc values.
+var LegacyChunkMode uint32 = 1024
+
+// DefaultChunkMode is the most recent improvement to chunking and should
+// be used by default.
+var DefaultChunkMode uint32 = 1025
 
 func getChunkSize(chunkMode uint32, cardinality uint64, maxDocs uint64) (uint64, error) {
-	switch chunkMode {
-	case 1024:
+	switch {
+	// any chunkMode <= 1024 will always chunk with chunkSize=chunkMode
+	case chunkMode <= 1024:
 		// legacy chunk size
-		return 1024, nil
-	case 1:
-		return 1, nil
-	case 0:
+		return uint64(chunkMode), nil
+
+	case chunkMode == 1025:
 		// attempt at simple improvement
 		// theory - the point of chunking is to put a bound on the maximum number of
 		// calls to Next() needed to find a random document.  ie, you should be able
@@ -21,7 +31,7 @@ func getChunkSize(chunkMode uint32, cardinality uint64, maxDocs uint64) (uint64,
 		// this way you'll still achieve the same goal of visiting at most
 		// chunk-size items.
 		// no attempt is made to tweak any other case
-		if cardinality < 1024 {
+		if cardinality <= 1024 {
 			return maxDocs, nil
 		}
 		return 1024, nil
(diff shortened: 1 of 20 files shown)


mschoch:
I will not be surprised if this is completely broken.

mschoch:
Just merged latest master back into this, which I had thought I started with, but missed by mistake.  At least it was clean this time.

mschoch:
@sreekanth-cb yes, the plan is to merge it to an experimental branch, but I wanted to get you (and hopefully @steveyen) to review it, since the merge was confusing.

We still do not plan to merge this to master, until we have a better story about versioning.

mschoch:
@sreekanth-cb @steveyen @abhinavdangeti 

This PR now also includes my attempt to make chunking better (only for the very sparse <1024 docs out of the dataset).

DocValues continue to use the LegacyMode (always chunk 1024 no matter what) which is one area that this change is a bit awkward.

For one example dataset (which is already tuned to turn off a lot of things), it was observed that the total index size went from 195MB down to 137MB (almost 30% reduction).  More specifically looking at the largest segment inside that index, it's space efficiency (segment size / num docs) was significantly better 116.8bytes/doc and is now better than a smaller segment in the same index 131.9bytes/doc.  This is just a small sample, but indicates we do see improvements in the areas we expected

index/scorch/segment/zap/chunk.go
```diff
@@ -0,0 +1,40 @@
...
+// DefaultChunkMode is the most recent improvement to chunking and should
+// be used by default.
+var DefaultChunkMode uint32 = 1025
+
+func getChunkSize(chunkMode uint32, cardinality uint64, maxDocs uint64) (uint64, error) {
+	switch {
+	// any chunkMode <= 1024 will always chunk with chunkSize=chunkMode
+	case chunkMode <= 1024:
+		// legacy chunk size
+		return uint64(chunkMode), nil
```
sreekanth-cb: curious to know - why chunkMode(default modes too) is a uint32 when we need a uint64 back?

index/scorch/segment/zap/chunk.go
```diff
@@ -0,0 +1,40 @@
...
+		// calls to Next() needed to find a random document.  ie, you should be able
+		// to do one jump to the correct chunk, and then walk through at most
+		// chunk-size items
+		// previously 1024 was chosen as the chunk size, but this is particularly
+		// wasteful for low cardinality terms.  the observation is that if there
+		// are less than 1024 items, why not put them all in one chunk,
+		// this way you'll still achieve the same goal of visiting at most
+		// chunk-size items.
+		// no attempt is made to tweak any other case
+		if cardinality <= 1024 {
```
sreekanth-cb: In early trials, one thought i had here is that - though the default CF was 1024, we ever know/measured - whether there were really 1024 entries are going inside a chunk.  So there could be perf regressions resulting from a more number of Next() calls inside a chunk with these 1024 values getting coalesced into a single chunk now.  (more document values going inside a chunk than early days)
Didn't had a better cheap fix for that, so in my toy trials - I had tried 256/512 as the actual cardinality limit for coalescing to a single chunk, as for that use case terms involved were of high uniqueness.

index/scorch/segment/zap/contentcoder.go
```diff
@@ -86,6 +86,19 @@ func (c *chunkedContentCoder) Reset() {
...
+	total := int(maxDocNum/chunkSize + 1)
+	c.chunkSize = chunkSize
+	if cap(c.chunkLens) < total {
+		c.chunkLens = make([]uint64, total)
+	} else {
+		c.chunkLens = c.chunkLens[:total]
+	}
+	if cap(c.chunkMeta) < total {
+		c.chunkMeta = make([]MetaData, 0, total)
+	}
```
sreekanth-cb: blind comment here - why isn't an else part here for `c.chunkMeta = c.chunkMeta[:total]`

index/scorch/segment/zap/chunk.go
mschoch: In the file format (footer) we only use 32 bits to record chunk mode.  So, chunk mode really is only 32-bits, but i think the IntCoder takes chunkMode as uint64, probably because we do math with it against other uint64s.  So, basically someone has to do some converting somewhere.  I no longer know what is best.

index/scorch/segment/zap/chunk.go
mschoch: Agreed, it's always been arbitrary, but sticking 1024 seemed more consistent with the original design (no one should have to call Next more than 1024 times).  But you're right there was never any reason to think that was a good choice in the first place.

Also, this current version only attempts to help with the sparsest of field-terms.  If there are 1025 docs, you're back to lots of chunks again.  And if you have 100 million docs, even 1025 is sparse...

My thought is that this helps address the immediate problem, but we can define new chunk modes in the future that do better.

index/scorch/segment/zap/contentcoder.go
mschoch: The reason is that chunkMeta is used differently than chunkLens.  With chunkLens, the length of the slice must be correct, because we directly index into it, like `c.chunkLens[something] = blah`.  However, chunkMeta is always appended to, so length should be 0 initially (and that is already taken care of by Reset in the case where there is already sufficient capacity).  In the case where there is insufficient capacity, we use the 3-arg form of make to set len to 0.

It is confusing, but I believe correct.

mschoch:
Thanks for the review @sreekanth-cb I replied to your comments.

bt:
Hey guys -- first and foremost, wish you all a great year ahead!

I wanted to see if there's anything that I could do to push this PR forward as I think this is something a lot of us have been waiting for.

As a workaround, I have been using Zstd to compress and decompress segments but it's quite tedious and I'm hitting a concurrency issue somewhere so rather than fixing it over in my fork, this PR would solve it the proper way.

If there's something I can help out with, whether it's testing or development, please let me know!

gilliek:
Any plan to merge this PR? Like @bt, I am also interested :-)

mschoch:
@gilliek it is part of zap v12, which will be come available as a part of this larger plan: https://github.com/blevesearch/bleve/issues/1350

gilliek:
@mschoch great, thanks for the update :)

mschoch:
Closing this, as the functionality is now available.  In order to take advantage of it, you must pass additional configuration options:

https://github.com/blevesearch/bleve/blob/master/index/scorch/scorch.go#L107-L121

For example:

`forceSegmentType` set to `zap`
and
`forceSegmentVersion` set to `12`

PR #1280, opened 2019-08-17
Update RoaringBitmap/roaring package

alldroll:
This updates the version of `RoaringBitmap/roaring` package. Unfortunately, there was a really tricky [bug](https://github.com/RoaringBitmap/roaring/pull/229) with `Iterator.AdvanceIfNeeded`, so it is better to update the dependency.


alldroll:
ping!

abhinavdangeti:
Thanks for pushing this in @alldroll. We've already update the vendor/manifest SHA for RoaringBitmap/roaring to release 0.4.19 (4208ad825dda03a6a3d2197df8ec57948aebcc12) which includes the fix.

PR #1286, opened 2019-09-16
Support for bounded polygon query

sreekanth-cb:
Attempt to introduce GeoBoundingPolygonQuery
to enable the bounded polygon queries.
Polygon searcher internally finds the bounded
rectangle for the given polygon points and
apply the filter which performs the point-in-polygon
checks.
In this iteration, as simplicity is preferred over
precision(on boundary points), the point-in-polygon check
is based on ray-casting algorithm.

geo points can be in any of the following forms as with other geo queries
{"query": {"field": "geo", "polygon_points":[[-80.881,35.282],[-80.858,35.281],[-80.864,35.270]]}}
"query": {"field": "geo", "polygon_points":[{"lon" :-80.881,"lat":35.282},{"lon":-80.858,"lat":35.281},{"lon":-80.864,"lat":35.270}]}}
{"query": {"field": "geo", "polygon_points":["35.282,-80.881","35.281, -80.858","35.270,-80.864"]}}
{"query": {"field": "geo", "polygon_points":["dnq2xwbk8","dnq88n9n8","dnq2xusws"]}}

### DIFF
--- a/geo/geo.go
+++ b/geo/geo.go
@@ -37,6 +37,12 @@ var geoTolerance = 1E-6
 var lonScale = float64((uint64(0x1)<<GeoBits)-1) / 360.0
 var latScale = float64((uint64(0x1)<<GeoBits)-1) / 180.0
 
+// Point represents a geo point.
+type Point struct {
+	Lon float64
+	Lat float64
+}
+
 // MortonHash computes the morton hash value for the provided geo point
 // This point is ordered as lon, lat.
 func MortonHash(lon, lat float64) uint64 {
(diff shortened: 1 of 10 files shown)


sreekanth-cb:
> side thoughts (but not blockers, imho), assuming the query API with "points" is inspired from elasticsearch? also, is lon-lat the usual order of those attributes, or is it lat-lon (aiming for least surprise)?

lat-lon order follows the standard format we already follows and thats inline with that ES.
The query format is slightly different as ES has these queries applies like a filter.
I just followed the "points" field as it is. But now re-thinking on "points"  Vs "polygon_points"?

PR #1288, opened 2019-09-19
compute any score only when it is really needed

sreekanth-cb:


search/scorer/scorer_term.go
```diff
@@ -117,56 +117,60 @@ func (s *TermQueryScorer) SetQueryNorm(qnorm float64) {
 func (s *TermQueryScorer) Score(ctx *search.SearchContext, termMatch *index.TermFieldDoc) *search.DocumentMatch {
 	var scoreExplanation *search.Explanation
```
abhinavdangeti: This can be moved to inside the if s.includeScore || s.options.Explain conditional

sreekanth-cb: done! and merging it.

PR #1292, opened 2019-10-04
adding a disk size stats for the index

sreekanth-cb:
The new "num_bytes_used_disk_by_root" stats shows
the disk size used by the latest root index.
The idea is that this should help in suggesting
a sizing recommendation for RAM quota driven by
an index size resident ratio.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -466,8 +466,9 @@ func (s *Scorch) Stats() json.Marshaler {
 	return &s.stats
 }
 
-func (s *Scorch) diskFileStats() (uint64, uint64) {
-	var numFilesOnDisk, numBytesUsedDisk uint64
+func (s *Scorch) diskFileStats(rootSegmentPaths map[string]struct{}) (uint64,
+	uint64, uint64) {
+	var numFilesOnDisk, numBytesUsedDisk, numBytesOnDiskByRoot uint64
 	if s.path != "" {
 		finfos, err := ioutil.ReadDir(s.path)
 		if err == nil {
@@ -475,24 +476,48 @@ func (s *Scorch) diskFileStats() (uint64, uint64) {
 				if !finfo.IsDir() {
 					numBytesUsedDisk += uint64(finfo.Size())
 					numFilesOnDisk++
+					if rootSegmentPaths != nil {
+						fname := s.path + string(os.PathSeparator) + finfo.Name()
+						if _, fileAtRoot := rootSegmentPaths[fname]; fileAtRoot {
+							numBytesOnDiskByRoot += uint64(finfo.Size())
+						}
+					}
 				}
 			}
 		}
 	}
-	return numFilesOnDisk, numBytesUsedDisk
+	// if no root files path given, then consider all disk files.
+	if rootSegmentPaths == nil {
+		return numFilesOnDisk, numBytesUsedDisk, numBytesUsedDisk
+	}
+
+	return numFilesOnDisk, numBytesUsedDisk, numBytesOnDiskByRoot
+}
+
+func (s *Scorch) rootDiskSegmentsPaths() map[string]struct{} {
+	rv := make(map[string]struct{}, len(s.root.segment))
+	for _, segmentSnapshot := range s.root.segment {
+		switch seg := segmentSnapshot.segment.(type) {
+		case *zap.Segment:
+			rv[seg.Path()] = struct{}{}
+		}
+	}
+	return rv
 }
 
 func (s *Scorch) StatsMap() map[string]interface{} {
 	m := s.stats.ToMap()
 
-	numFilesOnDisk, numBytesUsedDisk := s.diskFileStats()
+	s.rootLock.RLock()
+	rootSegPaths := s.rootDiskSegmentsPaths()
+	m["CurFilesIneligibleForRemoval"] = uint64(len(s.ineligibleForRemoval))
+	s.rootLock.RUnlock()
+
+	numFilesOnDisk, numBytesUsedDisk, numBytesOnDiskByRoot := s.diskFileStats(rootSegPaths)
 
 	m["CurOnDiskBytes"] = numBytesUsedDisk
 	m["CurOnDiskFiles"] = numFilesOnDisk
 
-	s.rootLock.RLock()
-	m["CurFilesIneligibleForRemoval"] = uint64(len(s.ineligibleForRemoval))
-	s.rootLock.RUnlock()
 	// TODO: consider one day removing these backwards compatible
 	// names for apps using the old names
 	m["updates"] = m["TotUpdates"]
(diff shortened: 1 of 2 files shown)


index/scorch/scorch.go
```diff
@@ -507,8 +533,9 @@ func (s *Scorch) StatsMap() map[string]interface{} {
 	m["num_items_introduced"] = m["TotIntroducedItems"]
 	m["num_items_persisted"] = m["TotPersistedItems"]
 	m["num_recs_to_persist"] = m["TotItemsToPersist"]
-	m["num_bytes_used_disk"] = m["CurOnDiskBytes"]
-	m["num_files_on_disk"] = m["CurOnDiskFiles"]
+	m["num_bytes_used_disk"] = numBytesUsedDisk
```
mschoch: According to the comments in this section, some of the stats have 2 names for backwards compatibility.  But if that is the case, I don't think we should keep introducing multiple names for each stat.  Or is there more to this than I understand?

index/scorch/scorch.go
```diff
@@ -466,33 +466,59 @@ func (s *Scorch) Stats() json.Marshaler {
...
+	s.rootLock.RLock()
+	rootSegPaths := s.rootDiskSegmentsPaths()
+	m["CurFilesIneligibleForRemoval"] = uint64(len(s.ineligibleForRemoval))
+	s.rootLock.RUnlock()
+
+	numFilesOnDisk, numBytesUsedDisk, numBytesOnDiskByRoot := s.diskFileStats(rootSegPaths)
 
 	m["CurOnDiskBytes"] = numBytesUsedDisk
 	m["CurOnDiskFiles"] = numFilesOnDisk
+	m["CurOnDiskBytesByRoot"] = numBytesOnDiskByRoot
```
mschoch: Can you add a description for each of these new stats to the commit message?  Just like a sentence is enough.  The problem is that the stat name uses "root" which we use inside of scorch to mean a particular thing, but that isn't obvious to people outside of the scorch internals.

index/scorch/scorch.go
```diff
@@ -466,33 +466,59 @@ func (s *Scorch) Stats() json.Marshaler {
...
+						if _, fileAtRoot := rootSegmentPaths[fname]; fileAtRoot {
+							numBytesOnDiskByRoot += uint64(finfo.Size())
+						}
+					}
 				}
 			}
 		}
 	}
-	return numFilesOnDisk, numBytesUsedDisk
+	// if no root files path given, then consider all disk files.
```
mschoch: is this actually needed?  the persister pause logic passes nil, but only looks at the number of files on disk.

it just seems like it's adding complexity, defining special behavior for nil, but that behavior isn't used presently.

index/scorch/scorch.go
sreekanth-cb: agree..i was just following the pattern there. So m["CurOnDiskBytesByRoot"]  isn't needed here.

index/scorch/scorch.go
sreekanth-cb: CurOnDiskBytesByIndex or CurOnDiskIndexBytes ?  
Currently we don't have a stat which explicitly gives the index size on disk.  CurOnDiskBytes computes the whole files size in the index directory which might include the older snapshots as well.
Exact index disk size helps to suggest a resident ratio/ram quota for the index to the customer - is the context for this change,

index/scorch/scorch.go
sreekanth-cb: currently its not used, but felt this helps to have a more consistent behaviour for that function.
like - anyone calling with nil/users not bothered about the exact index disk size, should get the whole file size in directory than just receiving zero as a return value for the 3rd/new return value.

index/scorch/scorch.go
```diff
@@ -507,8 +532,11 @@ func (s *Scorch) StatsMap() map[string]interface{} {
 	m["num_items_introduced"] = m["TotIntroducedItems"]
 	m["num_items_persisted"] = m["TotPersistedItems"]
 	m["num_recs_to_persist"] = m["TotItemsToPersist"]
-	m["num_bytes_used_disk"] = m["CurOnDiskBytes"]
-	m["num_files_on_disk"] = m["CurOnDiskFiles"]
+	// total disk bytes found in index directory inclusive of older snapshots
+	m["num_bytes_used_disk"] = numBytesUsedDisk
+	// total disk bytes by index exclusive of older snapshots
+	m["num_bytes_used_disk_index"] = numBytesOnDiskByIndex
```
abhinavdangeti: This stat seems more like "num_bytes_used_disk" to me.
The older num_bytes_used_disk can be "num_bytes_on_disk" - what do you think?

sreekanth-cb: older stats name could be something like a "total_num_bytes_on_disk" etc, but we can't change stats due to backward compatibility as we don't know whether anyone actually started using it or not ?

sreekanth-cb:
@mschoch ,  can you please review this.

index/scorch/scorch.go
mschoch: So, my request was for you to add it to the commit message, as I think that people are more likely to consult that if they have questions about it.

Second, I disagree with the statement 'we don't have a stat which explicitly gives the index size on disk'.  If you have chosen to keep more than 1 snapshot, or you have queries using older snapshots, then your index does contain multiple snapshots.  So, the size of your index is the size of all of them.

That doesn't mean there isn't value in knowing the size of just the current root.  We just disagree on the definitions.

index/scorch/scorch.go
mschoch: OK, so it will be removed?

index/scorch/scorch.go
mschoch: OK, I don't feel too strongly about this, just dislike writing code that isn't used.

index/scorch/scorch.go
mschoch: Yes, we certainly cannot rename stat keys as they are just strings passed through to users.

mschoch:
@sreekanth-cb I reviewed and replied to your comments, but I don't see any changes since the last review.

index/scorch/scorch.go
sreekanth-cb: which one to be removed now? m["CurOnDiskBytesByRoot"]  is removed already if you thats what you were referring to.

index/scorch/scorch.go
sreekanth-cb: do you have a better naming suggestion as we disagree here on the definitions?

sreekanth-cb:
@mschoch I did make changes, but did force amend on the commit.

index/scorch/scorch.go
sreekanth-cb: ok, let me go back to the older CurOnDiskBytesByRoot name itself your request was about just changing the commit msg.

mschoch: i sense that you don't agree about something, but i don't ow what that is

sreekanth-cb: i just misread your original comment about renaming the stat from root -> index, but then on a second read I realised that its more about the commit description than about the exact naming stuff..

PR #1298, opened 2019-10-25
Unit tests for a single segment merge policy

sreekanth-cb:
Using an aggressive merge policy, scorch is able to create a single 
segment index.


sreekanth-cb:
thanks @steveyen , done!

Issue #1301, opened 2019-10-25
Inconsistent results when doing Geo search


mkishere:
When doing a `bleve.NewGeoDistanceQuery()` based on documents with Geohash string field, inconsistent results returned. I got 0/1/2/3 hits all the time (3 hits expected). Am I missing something or did I hit with race condition?
Test done on AWS instance (Ubuntu 18.04) and my local PC (Surface i5 Windows 10)

Sample code:
```go
package main

import (
	"fmt"

	"github.com/blevesearch/bleve"
	"github.com/blevesearch/bleve/analysis/analyzer/keyword"
	"github.com/blevesearch/bleve/mapping"
)

type Shop struct {
	ID       int      //Internal ID
	Name     string   //Shop name
	Address  string   //Shop address
	Geohash  string   //Geohash code for lat/long coordinates
	Type     string   //Shop type
	District string   //Where is the shop?
	URL      string   //URL, currently unused
	Tags     []string //Tags used
}

func (s Shop) String() string {
	return fmt.Sprintf("%s (%s)\n%s", s.Name, s.Type, s.Address)
}

//BleveType fulfills bleveType interface
func (s Shop) BleveType() string {
	return "Shop"
}

func prepareDataset() (bleve.Index, error) {
	idx, err := bleve.NewMemOnly(newShopIndexMapping())
	if err != nil {
		return nil, fmt.Errorf("Cannot create store file %w", err)
	}

	idx.Index("1", Shop{
		ID:       1,
		Name:     "A",
		Address:  "XXX",
		Type:     "XXX",
		District: "上環",
		Geohash:  "wecpjc2b27ev",
	})
	idx.Index("2", Shop{
		ID:       2,
		Name:     "B",
		Address:  "XXX",
		Type:     "XXX",
		District: "TW",
		Geohash:  "wecpkbeddsmf",
	})
	idx.Index("3", Shop{
		ID:       3,
		Name:     "C",
		Address:  "XXX",
		Type:     "XXX",
		District: "KT",
		Geohash:  "wecnzm94b80h",
	})
	idx.Index("4", Shop{
		ID:       4,
		Name:     "D",
		Address:  "XXX",
		Type:     "XXX",
		District: "TW",
		Geohash:  "wecpk8tne453",
	})
	idx.Index("5", Shop{
		ID:       5,
		Name:     "E",
		Address:  "XXX",
		Type:     "XXX",
		District: "QB",
		Geohash:  "wecnycjgz1u3",
	})
[...]

PR #1303, opened 2019-10-28
[GEO] Search all numeric type terms to match geo points

abhinavdangeti:
If a field (for example _all) were to index multiple numeric terms
(of type numeric and geopoint morton hash), verify all numeric
matches rather than just the last encountered entry to determine
if the entry morton unhashes into a geo point match.

Fixes: https://github.com/blevesearch/bleve/issues/1301

### DIFF
--- a/search/searcher/search_geoboundingbox.go
+++ b/search/searcher/search_geoboundingbox.go
@@ -224,7 +224,8 @@ func ComputeGeoRange(term uint64, shift uint,
 func buildRectFilter(dvReader index.DocValueReader, field string,
 	minLon, minLat, maxLon, maxLat float64) FilterFunc {
 	return func(d *search.DocumentMatch) bool {
-		var lon, lat float64
+		// check geo matches against all numeric type terms indexed
+		var lons, lats []float64
 		var found bool
 		err := dvReader.VisitDocValues(d.IndexInternalID, func(field string, term []byte) {
 			// only consider the values which are shifted 0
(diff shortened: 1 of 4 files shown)


mschoch:
So, I'm not sure that ensuring they all match is right either.

It seems to me that the issue is that a date/numeric/geo query on _all, is almost never what you want.

So, the question is what does this query mean?  We can probably all agree only checking the last value we saw doesn't seem correct.  But, it doesn't seem that checking *all* of them is right either?  No other query has those semantics (a term search succeeds if any of the terms indexed are the one you're looking for), so I would propose that *any* match is sufficient, and not *all*.

Thoughts?

abhinavdangeti:
Yes, fair. The PR does exactly that though, yes?
While the searcher iterates through all the potential hits, it returns on the first one that matches.

mschoch:
Oh let me review again, I thought I read the a comment or commit message indicating that this changed required all date/numeric/geo terms to satisfy the clause, not just one of them.

mschoch:
So, it sounds like we're all pretty close to agreement, let me see if I can summarize and possibly generalize...

1.  This change is better than the current behavior.  Just looking at the last geo location when there are multiple values makes no sense at all.  Each of them should be examined, as this is consistent with how all other fields work.

2.  There is a larger question of, is this really ever what the user wants?  I would argue that the answer is still "yes, sometimes".  If we forget specifics of numeric/date/geo for a moment, the basic operation of the bleve index is that a document has one or more "indexed values" for a document.  When you search for a term, you want to find documents that have this term, you don't care what the other terms are (again in this simple example).  So, if we extend that philosophy to composite fields, like _all, the same behavior works.  We assume the user has arranged that composite field such that it makes sense, the semantics are the same.  As an example, you could imagine one document model where they store all locations of a chain in a single document.  Including all those locations in the _all field now has some meaning.  I could search geo distance of say 10 miles, and answer the question, do any of the locations satisfy it?  Now, @sreekanth-cb also gave an example where it doesn't make sense, so to me the key is still that it is up to the user to arrange the composite fields intelligently.

3.  In the future, if we allow custom composite fields other than _all, the user would gain even more power to make these make sense.

4.  Once we have the versioning in place, I would propose we change the default behavior of numeric, date, and geo fields, so that `includeInAll` defaults to false.  The reason is that normally this doesn't make sense, and does waste space.  Users could still switch it on if they want.  But, changing this default will require a major version change so as to not impact users.

Issue #1305, opened 2019-11-06
Scorch persister stuck after segment creation fails?


tmm1:
On low memory devices, we see frequent memory allocation errors when submitting indexing batches to scorch:

<pre>
2019/11/03 08:18:50 [DVR] Fetching guide data for 262 stations in USA-WI48449-X @ 2019-11-18 2:00PM
2019/11/03 08:19:02 [DVR]   indexed 1761 airings (247 channels) [1s fetch, 10s index]
2019/11/03 08:19:03 [DVR]   indexed 112 movies (35 channels) [0s fetch, 0s index]
2019/11/03 08:19:03 [DVR] Fetching guide data for 262 stations in USA-WI48449-X @ 2019-11-18 8:00PM
2019/11/03 08:19:07 [DVR] Error indexing airings: error opening new segment at USA-WI48449-X.airings/store/00000000668c.zap, cannot allocate memory
2019/11/03 08:19:08 [DVR] Fetching guide data for 262 stations in USA-WI48449-X @ 2019-11-19 2:00AM
2019/11/03 08:19:12 [DVR] Error indexing airings: cannot allocate memory
2019/11/03 08:19:26 [IDX] Pruned 3839 expired airings from USA-WI48449-X in 10.116005734s.
</pre>

Then, a few hours later we experience a full crash of the process. The stack shows that the persister is sitting inside pausePersisterForMergerCatchUp() and the Batch is still stuck in prepareSegment()

<pre>
runtime/cgo: pthread_create failed: Resource temporarily unavailable
SIGABRT: abort
PC=0xb68bd6 m=2 sigcode=4294967290

goroutine 0 [idle]:
runtime: unknown pc 0xb68bd6
stack: frame={sp:0x66c9a87c, fp:0x0} stack=[0x6649b034,0x66c9ac34)
66c9a7fc:  00b6bcf1  00000000  00000000  00000000 
66c9a80c:  00000000  00000000  00000000  00000000 
66c9a81c:  00000000  00000000  00000000  00000000 
66c9a82c:  00000000  00000000  00000000  00000000 
66c9a83c:  00000000  00000000  00000000  00000000 
66c9a84c:  00000000  00000000  00000000  00000000 
66c9a85c:  00000000  00000000  00000000  00000000 
66c9a86c:  01afd358  66c9b300  00000001  01afd358 
66c9a87c: <01afd358  00b6bc7b  01afd358  66c9b300 
[...]

PR #1307, opened 2019-11-09
avoid writing nil entry to newSegments on zap.Open failures

tmm1:
i'm not certain what golang's behavior is with the previous syntax when an error occurs, but it seems safer to do it this way

/cc #1305


index/scorch/persister.go
```diff
@@ -553,10 +553,11 @@ func (s *Scorch) persistSnapshotDirect(snapshot *IndexSnapshot) (err error) {
 			}
 		}()
 		for segmentID, path := range newSegmentPaths {
-			newSegments[segmentID], err = zap.Open(path)
```
tmm1: I guess the `defer` loop above already ignores `nil` so this probably is fine as-is.

PR #1309, opened 2019-11-14
scorch: add backoff after merging errors

tmm1:
this adds a naive exponential backoff to prevent the 100% cpu usage issue noticed in #1305 

cc #1308

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -37,6 +37,7 @@ func (s *Scorch) mergerLoop() {
 		return
 	}
 
+	var errBackoff time.Duration
 OUTER:
 	for {
 		atomic.AddUint64(&s.stats.TotFileMergeLoopBeg, 1)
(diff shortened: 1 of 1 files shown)


tmm1:
@mschoch Curious what you think of this approach?

If this seems reasonable, I can add something similar to the persister as well.

There are also libraries like https://github.com/cenkalti/backoff we can re-use instead.

tmm1:
before:
```
2019/11/14 16:29:09 merging err: cannot allocate memory
2019/11/14 16:29:32 merging err: cannot allocate memory
2019/11/14 16:29:56 merging err: cannot allocate memory
2019/11/14 16:30:19 merging err: cannot allocate memory
2019/11/14 16:30:42 merging err: cannot allocate memory
```

after:
```
2019/11/14 16:35:30 merging err: cannot allocate memory (waiting 500ms to try again)
2019/11/14 16:36:09 merging err: cannot allocate memory (waiting 1s to try again)
2019/11/14 16:36:48 merging err: cannot allocate memory (waiting 2s to try again)
2019/11/14 16:37:28 merging err: cannot allocate memory (waiting 4s to try again)
2019/11/14 16:38:10 merging err: cannot allocate memory (waiting 8s to try again)
2019/11/14 16:38:57 merging err: cannot allocate memory (waiting 16s to try again)
2019/11/14 16:39:52 merging err: cannot allocate memory (waiting 32s to try again)
2019/11/14 16:41:03 merging err: cannot allocate memory (waiting 1m4s to try again)
2019/11/14 16:42:45 merging err: cannot allocate memory (waiting 1m4s to try again)
2019/11/14 16:44:27 merging err: cannot allocate memory (waiting 1m4s to try again)
```

tmm1:
This is working, and improves behavior on failure in that it prevents constant CPU/disk trashing by inserting sleeps between attempts.

But the underlying behavior is the same: the merger and persister will attempt the same operation over and over even if it keeps failing and never give up. So in that sense I'm not sure if backoff is the correct solution. Instead it would be better for the failing operation to be aborted and rolled back after N failures.

PR #1311, opened 2019-11-19
scorch/zap: disable mmap on 32-bit systems

tmm1:
skips the mmap() call on 32-bit systems, and falls back to direct file reads instead.

cc https://github.com/blevesearch/bleve/issues/1305#issuecomment-555282867 @mschoch @steveyen @sreekanth-cb

### DIFF
--- a/index/scorch/segment/zap/mem.go
+++ b/index/scorch/segment/zap/mem.go
@@ -1,0 +1,36 @@
+//  Copyright (c) 2017 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package zap
+
+import (
+	"io"
+	"io/ioutil"
+)
+
+func (s *SegmentBase) readMem(x, y uint64) []byte {
+	if s.mem == nil {
+		data, _ := ioutil.ReadAll(io.NewSectionReader(s.f, int64(x), int64(y-x)))
+		return data
+	}
+	return s.mem[x:y]
+}
+
+func (s *Segment) readMM(x, y int) []byte {
+	if s.mm == nil {
+		data, _ := ioutil.ReadAll(io.NewSectionReader(s.f, int64(x), int64(y-x)))
+		return data
+	}
+	return s.mm[x:y]
+}
--- a/index/scorch/segment/zap/build.go
+++ b/index/scorch/segment/zap/build.go
@@ -128,6 +128,7 @@ func InitSegmentBase(mem []byte, memCRC uint32, chunkFactor uint32,
 	dictLocs []uint64) (*SegmentBase, error) {
 	sb := &SegmentBase{
 		mem:               mem,
+		memSize:           len(mem),
 		memCRC:            memCRC,
 		chunkFactor:       chunkFactor,
 		fieldsMap:         fieldsMap,
--- a/index/scorch/segment/zap/docvalues.go
+++ b/index/scorch/segment/zap/docvalues.go
@@ -96,9 +96,9 @@ func (s *SegmentBase) loadFieldDocValueReader(field string,
 	var numChunks, chunkOffsetsPosition uint64
 
 	if fieldDvLocEnd-fieldDvLocStart > 16 {
-		numChunks = binary.BigEndian.Uint64(s.mem[fieldDvLocEnd-8 : fieldDvLocEnd])
+		numChunks = binary.BigEndian.Uint64(s.readMem(fieldDvLocEnd-8, fieldDvLocEnd))
 		// read the length of chunk offsets
-		chunkOffsetsLen := binary.BigEndian.Uint64(s.mem[fieldDvLocEnd-16 : fieldDvLocEnd-8])
+		chunkOffsetsLen := binary.BigEndian.Uint64(s.readMem(fieldDvLocEnd-16, fieldDvLocEnd-8))
 		// acquire position of chunk offsets
 		chunkOffsetsPosition = (fieldDvLocEnd - 16) - chunkOffsetsLen
 	} else {
@@ -114,7 +114,7 @@ func (s *SegmentBase) loadFieldDocValueReader(field string,
 	// read the chunk offsets
 	var offset uint64
 	for i := 0; i < int(numChunks); i++ {
-		loc, read := binary.Uvarint(s.mem[chunkOffsetsPosition+offset : chunkOffsetsPosition+offset+binary.MaxVarintLen64])
+		loc, read := binary.Uvarint(s.readMem(chunkOffsetsPosition+offset, chunkOffsetsPosition+offset+binary.MaxVarintLen64))
 		if read <= 0 {
 			return nil, fmt.Errorf("corrupted chunk offset during segment load")
 		}
@@ -145,7 +145,7 @@ func (di *docValueReader) loadDvChunk(chunkNumber uint64, s *SegmentBase) error 
 	curChunkEnd += end
 
 	// read the number of docs reside in the chunk
-	numDocs, read := binary.Uvarint(s.mem[destChunkDataLoc : destChunkDataLoc+binary.MaxVarintLen64])
+	numDocs, read := binary.Uvarint(s.readMem(destChunkDataLoc, destChunkDataLoc+binary.MaxVarintLen64))
 	if read <= 0 {
 		return fmt.Errorf("failed to read the chunk")
 	}
@@ -158,15 +158,15 @@ func (di *docValueReader) loadDvChunk(chunkNumber uint64, s *SegmentBase) error 
 		di.curChunkHeader = di.curChunkHeader[:int(numDocs)]
 	}
 	for i := 0; i < int(numDocs); i++ {
-		di.curChunkHeader[i].DocNum, read = binary.Uvarint(s.mem[chunkMetaLoc+offset : chunkMetaLoc+offset+binary.MaxVarintLen64])
+		di.curChunkHeader[i].DocNum, read = binary.Uvarint(s.readMem(chunkMetaLoc+offset, chunkMetaLoc+offset+binary.MaxVarintLen64))
 		offset += uint64(read)
-		di.curChunkHeader[i].DocDvOffset, read = binary.Uvarint(s.mem[chunkMetaLoc+offset : chunkMetaLoc+offset+binary.MaxVarintLen64])
+		di.curChunkHeader[i].DocDvOffset, read = binary.Uvarint(s.readMem(chunkMetaLoc+offset, chunkMetaLoc+offset+binary.MaxVarintLen64))
 		offset += uint64(read)
 	}
 
 	compressedDataLoc := chunkMetaLoc + offset
 	dataLength := curChunkEnd - compressedDataLoc
-	di.curChunkData = s.mem[compressedDataLoc : compressedDataLoc+dataLength]
+	di.curChunkData = s.readMem(compressedDataLoc, compressedDataLoc+dataLength)
 	di.curChunkNum = chunkNumber
 	di.uncompressed = di.uncompressed[:0]
 	return nil
--- a/index/scorch/segment/zap/merge.go
+++ b/index/scorch/segment/zap/merge.go
@@ -797,7 +797,7 @@ func (s *SegmentBase) copyStoredDocs(newDocNum uint64, newDocNumOffsets []uint64
 
 	storedOffset0New := uint64(w.Count())
 
-	storedBytes := s.mem[storedOffset0 : storedOffsetN+readN+metaLenN+dataLenN]
+	storedBytes := s.readMem(storedOffset0, storedOffsetN+readN+metaLenN+dataLenN)
 	_, err := w.Write(storedBytes)
 	if err != nil {
 		return err
@@ -806,7 +806,7 @@ func (s *SegmentBase) copyStoredDocs(newDocNum uint64, newDocNumOffsets []uint64
 	// remap the storedOffset's for the docs into new offsets relative
 	// to storedOffset0New, filling the given docNumOffsetsOut array
 	for indexOffset := indexOffset0; indexOffset <= indexOffsetN; indexOffset += 8 {
-		storedOffset := binary.BigEndian.Uint64(s.mem[indexOffset : indexOffset+8])
+		storedOffset := binary.BigEndian.Uint64(s.readMem(indexOffset, indexOffset+8))
 		storedOffsetNew := storedOffset - storedOffset0 + storedOffset0New
 		newDocNumOffsets[newDocNum] = storedOffsetNew
 		newDocNum += 1
--- a/index/scorch/segment/zap/posting.go
+++ b/index/scorch/segment/zap/posting.go
@@ -216,7 +216,7 @@ func (p *PostingsList) iterator(includeFreq, includeNorm, includeLocs bool,
 	// prepare the freq chunk details
 	if rv.includeFreqNorm {
 		var numFreqChunks uint64
-		numFreqChunks, read = binary.Uvarint(p.sb.mem[p.freqOffset+n : p.freqOffset+n+binary.MaxVarintLen64])
+		numFreqChunks, read = binary.Uvarint(p.sb.readMem(p.freqOffset+n, p.freqOffset+n+binary.MaxVarintLen64))
 		n += uint64(read)
 		if cap(rv.freqChunkOffsets) >= int(numFreqChunks) {
 			rv.freqChunkOffsets = rv.freqChunkOffsets[:int(numFreqChunks)]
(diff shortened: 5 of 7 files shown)


tmm1:
All the zap tests pass, and this works as expected on an linux/arm device that was previously experiencing issues in #1305 

However, performance is significantly worse without mmap. Given that zap can now gracefully handle both mapped and unmapped segments, I plan to improve the implementation in this PR to allow mapping segments on 32-bit instances up to a user-specified limit. mmap failures beyond this point can be gracefully ignored, allowing scorch to keep functioning at a lower throughput instead of getting stuck in a loop.

Ideally I would like to add a few new scorch configuration options for zap, including an optional mmap limit and a setting to ignore mmap failures. I would also like to expose some new stats, including the total number of bytes mapped into memory.

I am mixed, though, on whether these should be scorch options/stats, or part of zap directly. Since the address space is global to the process, it applies across all scorch indices. I use multiple scorch indices myself, so specifying a limit per index doesn't really help me (especially since the sizes of the indexes varies greatly). Plus, it's not obvious how scorch's existing config/stats variables could be shared across `zap.Open` in the current design.

I think as a first pass I will instead add some global stats/limits into the zap package.. Something like `zap.MaxMmapBytes` and `zap.CurrentMmapBytes`.

If anyone has a preference for API design here, please let me know. I would like to maximize the chances this direction can be merged back into master, and would like to prevent any wasted time on my part redoing interfaces and APIs.

tmm1:
I am now running this in production with:

>MmapIgnoreErrors = true
>MmapMaxBytes = 1GB

Everything is working as expected, with no more ENOMEM crashes due to mmap failures. I can observe via pmap that my process never uses more than 1GB of mapped files, and that it's VmSize is well below the kernel limits.

Further performance improvements here could be implemented using some sort of LRU, to unmap and map segments dynamically based on usage. For my needs, however, the naive approach here is good enough.

tmm1:
> Further performance improvements here could be implemented using some sort of LRU, to unmap and map segments dynamically based on usage.

I cleaned up the code to extract `loadMmap` and `unloadMmap` methods on `Segment`. But the challenge now is that most of the usage occurs directly from methods on `SegmentBase`, which cannot reference the `Segment` to try loading the mmap.

tmm1:
> But the challenge now is that most of the usage occurs directly from methods on `SegmentBase`, which cannot reference the `Segment` to try loading the mmap.

I ended up adding an `SegmentBase.mmapOwner` field with a minimal interface that lets me call `(*Segment).readMM()`

> Further performance improvements here could be implemented using some sort of LRU

An LRU would introduce a lot of complexity and it's unclear if that is necessary, especially since this is just optimizing for the 32-bit case. The immediate issue I'm noticing is that after a merge, the merged segments aren't loaded in memory when space becomes available (after the old snapshot is unmapped).

As a simple fix for this specific case, I'm experimenting with a call to `loadMmap()` from `readMM()`: https://github.com/fancybits/bleve/compare/zap-32bit-mode...fancybits:zap-mmap-load

I tried to minimize the overhead of these calls as much as possible, but it might still be too much to add to the hot path. Another approach might be simply to call `loadMmap()` on the segments of the latest snapshot after the old snapshot is closed.

tmm1:
> Another approach might be simply to call `loadMmap()` on the segments of the latest snapshot after the old snapshot is closed.

I gave this a shot in https://github.com/fancybits/bleve/compare/zap-32bit-mode...fancybits:zap-mmap-hint. I put it in each introducer, after `prevRoot.DecRef()` closes and unmaps the previous segments.

In order to bridge the scorch/zap boundary I had to add a public `(*zap.Segment).MmapHint()` for scorch to use. This is kinda ugly, but I think I still prefer this to the previous approach of trying to load during reads.

mschoch:
Thanks for all this.  I have some concerns, but its great to see you working on this.  I won't have time to do proper review until December, but hopefully @sreekanth-cb and @abhinavdangeti can take a look sooner.  Just wanted to let you know I'm not ignoring this.

abhinavdangeti:
Thanks for making this change @tmm1 . While this looks generally good to me, I'd wait to get one other opinion/review here as well.

index/scorch/segment/zap/mem.go
```diff
@@ -0,0 +1,99 @@
...
+func (s *SegmentBase) readMem(x, y uint64) []byte {
+	if s.mem == nil {
+		return s.mmapOwner.readMM(int64(x), int64(y))
+	}
+	return s.mem[x:y]
+}
+
+func (s *Segment) readMM(x, y int64) []byte {
+	if s.mm == nil {
+		data, _ := ioutil.ReadAll(io.NewSectionReader(s.f, x, y-x))
```
sreekanth-cb: ignoring the err here is preferred?

I may not have understood the thing here- hence a doubt,

Isn't this ioutil.ReadAll()  conveniently overrides the MmapMaxBytes or any such memory limits set? Or you were worried only about the mmap memory usage spikes?

My understanding is that - s.mm will be nil if your file size is bigger than MmapMaxBytes, and then
what if the size of the file is big, and then ioutil.ReadAll reads & copies the whole content to memory? And lets assume there are concurrent queries and mutations are happening in back ground?

Won't this approach will also have high memory requirements due to these "data" slices holding the whole file contents unlike that in mmap?

index/scorch/segment/zap/mem.go
```diff
@@ -0,0 +1,99 @@
...
+	MmapMaxBytes int64
+)
+
+type zapStats struct {
+	// Total number of bytes mapped into memory with mmap.
+	MmapCurrentBytes int64
+}
+
+type mmapOwner interface {
+	readMM(int64, int64) []byte
```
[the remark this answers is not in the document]
tmm1: The main reason was that all the callers in zap are passing in int64 and I didn't want to change all the callsites to add casts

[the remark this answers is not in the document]
tmm1: Alright I changed the signature to `start, end uint64`. Thanks for the feedback!

index/scorch/segment/zap/mem.go
tmm1: > Or you were worried only about the mmap memory usage spikes?

Correct.

So the underlying issue is not so much memory usage, but address space limitations. If a 32-bit process can only access 2GB of memory addresses, then mmap will start failing after 2GB of zap segments are mapped. And other methods like pthread_create also start failing because they need to reserve address space for stack, etc.

By limiting the address space used by mmap (via `MmapMaxBytes`), we ensure that part of the address space is available to the golang runtime. The runtime itself will reserve space for its own heap, and those addresses will be managed by the golang GC.

> s.mm will be nil if your file size is bigger than MmapMaxBytes

Yep, so if mmap fails or is over the limit, then `s.mm` is nil and that signals to the code to fallback to regular file access.

> what if the size of the file is big, and then ioutil.ReadAll reads & copies the whole content to memory

From what I can tell, `readMem` will never try to read the entire file. It only tries to read certain slices which are currently required. These will be copied onto the golang heap, but then as soon as they're not being used anymore the GC will reap them and re-use that memory for reads and segments.

> And lets assume there are concurrent queries and mutations are happening in back ground?

AFAICT, zap segments are immutable and zap only opens them in read-only mode.

>Won't this approach will also have high memory requirements due to these "data" slices holding the whole file contents unlike that in mmap?

Yes, basically we're trading off address space reservations in exchange for golang heap usage. By leaving part of the address space free, we allow golang heap to expand and the GC to manage which parts of the file are held in memory (vs assuming the entire thing can be mapped in like it does on 64-bit OS).

mschoch: Great questions @sreekanth-cb  and answers @tmm1 this helps my understanding already.

tmm1: > ignoring the err here is preferred?

Not sure about this one. It seems like if the file was able to be opened, and we checked its stat/size already, then reads should all succeed. I'm open to logging errors in this path, but I'm not sure where to send them.

tmm1:
> 1\. How important is the `MmapMaxBytes` and related functionality?  It seems likely to add significant complexity when debugging things as something things will be mmap'd and sometimes they might not.  My preference would be to have explicit modes, chosen by the user at the time they open the index, in BadgerDB they have options like: https://github.com/dgraph-io/badger/blob/407e5bde06c44e5503dc1eaf5a11cc9f2f348ac4/options/options.go#L23-L30

Very important IMHO. You need mmap for good performance, but you can't let mmap fill up the address space or else other parts of the golang runtime will fail. A mode that disables mmap altogether is not really tenable, as it will destroy performance even for small indices that can easily fit into RAM.

I'm not convinced there that much added complexity in terms of debugging. Everything goes through the `readMem` function, which either uses the mmap area directly or does a file read.

Another thing to keep in mind is that the mmap limits are global to the process. That's why it was easier for me to add the configuration option as a global into the zap package. Specifying limits per scorch index doesn't work well if you have multiple scorch indices open at the same time and they are all competing for the same limited mmap address space.



> 2\. How feasible would it be to have `mem` encapsulated only exposing readMem semantics?  I'm concerned that were just relying on convention for developers to not access it directly.

Not sure what you mean here.. everything already goes through `readMem`, and there's no way to prevent direct access to `mem` afaik (which is already a private variable).



> 3\. I share @sreekanth-cb's concerns about ignoring the error.  I don't know if it would actually work, but something like root force unmounting the filesystem containing the segment, causing errors on read.  I'm guessing the mmap would probably panic in that case anyway.

AFAIK the failure mode is the same. If you force unmount the filesystem, the mmap region becomes invalid and reading from it would raise a SEGV signal and crash the process.



> 4\. I guess I still didn't follow the `mmapOwner` part.  Like I understand that SegmentBase couldn't access Segment, but that's actually by design.  How does it work for the case where there only is a SegmentBase (segments not yet persisted to disk)?

A `SegmentBase` created directly will have `mmapOwner == nil` and `mem != nil`, so readMem will use `s.mem[x:y]` directly. Conversely when the `SegmentBase` was created by a `Segment`, `mmapOwner` is set and allows readMem to be propagated up to readMM.

mschoch:
> Very important IMHO....

I guess we just disagree over the performance an complexity trade-offs here.  I can't imagine this mode performing acceptably at all anyway, but I can imagine debugging it to not be something I'd want to do.

> Not sure what you mean here.. everything already goes through readMem, and there's no way to prevent direct access to mem afaik (which is already a private variable).

Everything goes through readMem right now, then tomorrow another pull request comes in that accesses mem directly and we have a problem.  To me having mem be a private member of something with much smaller surface area (a new struct with one method named readMem) would protect us from that scenario.

> AFAIK the failure mode is the same. If you force unmount the filesystem, the mmap region becomes invalid and reading from it would raise a SEGV signal and crash the process.

I don't think the failure mode would be the same at all.  The code as it exists in this PR would simply return nil, likely causing zap to crash upstream with a nil ptr, and the developer would have to spend time finding the root cause, instead of just getting an error returned.

To be more clear as to what I propose, readMem should also return an error, and we'd have to check/propogate that all the way back.

> A SegmentBase created directly will have mmapOwner == nil and mem != nil, so readMem will use s.mem[x:y] directly. Conversely when the SegmentBase was created by a Segment, mmapOwner is set and allows readMem to be propagated up to readMM.

I'll have to think about this more, I find it completely unreadable.

tmm1:
> I guess we just disagree over the performance an complexity trade-offs here. I can't imagine this mode performing acceptably at all anyway, but I can imagine debugging it to not be something I'd want to do.

FWIW its working well for me. My index only bloats to be bigger than RAM when adding data once a day, at which point performance degrades until all the merges are complete and everything can be mmap'd into RAM again.

If I had to disable mmap on 32-bit altogether instead, performance would be terrible all the time instead of only occasionally.



> To me having mem be a private member of something with much smaller surface area (a new struct with one method named readMem) would protect us from that scenario.

Got it. That's a good idea.



> I don't think the failure mode would be the same at all. The code as it exists in this PR would simply return nil, likely causing zap to crash upstream with a nil ptr, and the developer would have to spend time finding the root cause, instead of just getting an error returned.

That's fair. If you want the failure mode to be the same, then maybe we can catch the error and panic to mimic the panic that would happen if you read from the mmap.

> To be more clear as to what I propose, readMem should also return an error, and we'd have to check/propogate that all the way back.

Returning an error is possible too, but would require rewriting every single call site to deal with the extra return value. I'm not opposed to that.

> I'll have to think about this more, I find it completely unreadable.

I agree it's confusing. `SegmentBase` is used both with and without mmap, and some reads go through `Segment.readMm` whereas others go through `SegmentBase.readMem`. I'm open to suggestions to make this simpler and more straightforward.

mschoch:
@tmm1 sorry to seem like I'm taking a step back, but I think it would help us evaluate if this is a good solution to better understand more about which 32-bit platform(s) you're targeting, how much physical memory they have, the data sizes of the bleve index, and anything else you think we should know.

tmm1:
Our software runs on consumer NAS devices, from vendors like Synology, QNAP, Netgear, etc. We also support devices like the Raspberry Pi, as well as various Android TV STBs like the NVIDIA SHIELD.

All of these run various kinds of 32-bit ARM chips. Physical RAM is anywhere from 512MB to 4GB depending on the device. Most are compiled with stock kernels (VMSPLIT_3G) which provides user-space with 3GB of address space, but some vendors use VMSPLIT_2G.

Our index size depends per user, but is generally on the order of 500MB-1GB. We index new data every morning, and the segments on disk usually double during that time during merging. (Our product is a TV DVR, and our indices contain TV guide data which is refreshed every morning). This software is for personal in-home use, so apart from the reindexing of new data, there is very minimal read traffic during the day on each index.

The problem we've noticed so far occurs primarily on Synology ARM NAS which use VMSPLIT_2G. We have also noticed some issues on Android systems where the OS seems to restrict address space as well.

tmm1:
Hope all of you enjoyed some R&R over the holidays!

Let me know what I can do to move this forward.

tmm1:
Hi folks, looks like a bunch of merge conflicts are starting to pile up here. Is there any chance we can get this merged so I don't have to maintain my own fork and play catch-up constantly? We've been using this branch on hundreds of 32-bit ARM and Intel devices in production.

mschoch:
I wanted to provide an update on the status of this PR.  First, we have not forgotten about it, and we have a general consensus on wanting to support this feature.  I think the remaining concerns come back to two things:

1.  Are there performance implications even when not using File I/O due to extra work going through the Read method?
2.  Is this implementation the right way forward, or is there some cleaner way to abstract where the data is and how to access it so that going forward, the segment implementations aren't concerned with this detail (ideal).

With that in mind, I did some benchmarking of 3 implementations.

1.  Native slice access
2.  Concrete implementation, with simple conditional (if mmap = nil, do file I/O)
3.  Interface, with 2 different implementations

My testing showed that while 3 was ideal for coding purposes, there about an extra 4ns hit for every call, presumably the overhead of the dynamic dispatch.  Option 2 seemed to incur less of a penalty, but it's important to measure with the actual final implementation, as the function size affects in-lining, which directly affects these numbers.  These numbers may sound small, but the way the code is organized today, we assumed the data was mmapped, and so we have many individual calls during all operations, and each of these calls incurs any overhead.

Performance is however just one consideration, I did some looking around at other libraries, and really liked how BadgerDB allowed you to configure a similar choice:

https://github.com/dgraph-io/badger/blob/master/options/options.go#L23-L30

And their implementation when reading table data looks like a straightforward conditional:

https://github.com/dgraph-io/badger/blob/6001230ce640979998e40978f16fb6265a7475bb/table/table.go#L329-L342

Based on that I have a local prototype of new data-structure, basically implementing option 2 above, simple conditional choosing between slice access or File I/O.  I have it wired up and working locally, but it's on a highly experimental branch, so there are many other changes needed to wire it up to bleve now.

There is one other aspect to this PR, and that is that as I understand the use-case, configuring a static "segment loading mode" isn't what we want anyway.  We also need some way that an application can dynamically specify the behavior, mmap vs File I/O at segment load time.  In this way an application can use mmap the the system resources allow for it, and fall back to File I/O otherwise.  In this PR, it is accomplished via a package-level variable, and logic that works with this setting to toggle the behavior.  
I'd like to see about a way to do this without a package-level variable, and without putting that logic into bleve/zap, ideally leaving it as a function that the application can provide.

Anyway, I'm sure you'd prefer we just merge this as is, but I wanted to share with you our thinking, and that we have been doing some work on this in the background.  What I'd like to propose is that I'll see how much work there is to adapt the approach I have working locally to the current bleve.  And then if that's reasonable, produce a build and have someone at Couchbase see what perf impact there is for the cases not using File I/O (they have an application with repeatable performance benchmarks we can rely on)

tmm1:
> Based on that I have a local prototype of new data-structure, basically implementing option 2 above, simple conditional choosing between slice access or File I/O. I have it wired up and working locally, but it's on a highly experimental branch, so there are many other changes needed to wire it up to bleve now.

Can you share a link to this branch/approach?

I'd like to get sync'd up with bleve master which will require either rebasing my PR here or trying to integrate your work on a new approach.

mschoch:
The approach I mentioned is used in [Bluge](https://github.com/blugelabs/bluge)

The [bluge_segment_api](https://github.com/blugelabs/bluge_segment_api) defines a structure called [Data](https://github.com/blugelabs/bluge_segment_api/blob/master/data.go#L31).  The idea is that this one structure has two modes of operation, depending on how it was created (avoiding interface dynamic dispatch cost).  The two modes of operation are mmap, or regular file I/O, and all the methods on the struct attempt to provide consistent behavior for both modes.

Then [ice](https://github.com/blugelabs/ice) which is the bluge version of zap, has been refactored to work only with `*Data`.  Zap no longer knows (or cares) if the data has been memory-mapped or is using regular file I/O.

NOTE: that it is still up to the higher level code to choose whether or not to open a particular segment using mmap or file I/O.  In Bluge today, we always use mmap, but the thinking was that a custom [Directory](https://github.com/blugelabs/bluge/blob/master/index/directory.go#L37) implementation could track the stats you're interested in and make the decision.

mschoch:
I probably should have mentioned Bleve cannot just switch to using ice, because it has other changes to the file format (removed support for array positions).

Issue #1147, opened 2019-02-28
Support go.mod


fproulx-dfuse:
Hello,
We use go 1.11 (go.mod) and since some of your transitive dependencies have started to use it too.
When we `go mod tidy` your transitive dependencies follow and we can never be sure that updating them will not have an impact on bleve.

It would be great that you lock the stable dependencies for a tagged version of bleve....

mschoch:
Yes, it would be great.  Unfortunately, I don't know anything about Go modules.  The versions we recommend (and test against) are in the vendor/manifest file.

We will eventually start to support modules.  If the community help us get there faster that would be great.

prochac:
close this
https://github.com/blevesearch/bleve/issues/1350

PR #1312, opened 2019-11-19
Add go.mod

ntrrg:
Hello, this fixes errors for Go modules users


prochac:
https://github.com/blevesearch/bleve/issues/1147

bikbah:
@mschoch Hello! It would be great to review and merge this PR, in order to start using go modules. Any reasons for not doing this?

mschoch:
Hello @bikbah,  

We have an independent effort to adopt Go modules in-progress.  The reason it's somewhat more complicated is that Bleve has both a long history, and new improvements we can't ship without proper versioning.  This plan included a v2, and Go modules with v2 has some additional challenges.

The crucial problem we are solving is the ability for a single build of bleve to read/write more than one file format (necessary before we can introduce a breaking change).  The good news is that we are in the late stages of testing this now, and hope to communicate details to the community very soon.

mschoch:
Closing this as bleve now has a go.mod.

PR #1313, opened 2019-11-22
Api to trigger a force merge operation on an index

sreekanth-cb:
Introducing a new merge request api to enable customers
conveniently invoke a custom compaction/merge operation
on an existing scorch index. This api would let advanced
bleve users to compact their index all the way down to a
single file segment based index.
The api takes an optional mergePlanOptions argument which
would be applied for the forceful merge operations, while
the original mergePlanOptions property of the index remains
intact after the forceful merge cycle exits.
In the absense of a mergePlanOptions argument, default policy
would be an aggressive one to target a single file segment index
with a maximum segment size of 1B.
In the event of incoming mutations, the forceful merge
operations bail out early than strictly applying the
mergePlanOptions to the index.
Its the caller's responsibility to ensure that there is
enough disk space before invoking this compaction/merge
operation as it needs at least double the disk space
since its a full compaction.

During/right after a force merge, this also cleans up older snapshots(for faster rollbacks) kept to reduce the disk usage.

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -323,11 +323,14 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 
 	defer func() { _ = root.DecRef() }()
 
+	if nextMerge.creator == "" {
+		nextMerge.creator = "introduceMerge"
+	}
 	newSnapshot := &IndexSnapshot{
 		parent:   s,
 		internal: root.internal,
 		refs:     1,
-		creator:  "introduceMerge",
+		creator:  nextMerge.creator,
 	}
 
 	// iterate through current segments
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -28,16 +28,18 @@ import (
 	"github.com/blevesearch/bleve/index/scorch/segment/zap"
 )
 
+var forceMergeCreator = "introduceForceMerge"
+
 func (s *Scorch) mergerLoop() {
 	var lastEpochMergePlanned uint64
-	var activeMergeOptions *mergeplan.MergePlanOptions
-	var kickMsg *mergerKick
+	var ctrlMsg *mergerCtrl
 	mergePlannerOptions, err := s.parseMergePlannerOptions()
 	if err != nil {
 		s.fireAsyncError(fmt.Errorf("mergePlannerOption json parsing err: %v", err))
 		s.asyncTasks.Done()
 		return
 	}
+	ctrlMsgDflt := &mergerCtrl{options: mergePlannerOptions, doneCh: nil}
 
 OUTER:
 	for {
@@ -56,17 +58,15 @@ OUTER:
 			atomic.StoreUint64(&s.iStats.mergeEpoch, ourSnapshot.epoch)
 			s.rootLock.Unlock()
 
-			if ourSnapshot.epoch != lastEpochMergePlanned || kickMsg != nil {
-				if kickMsg == nil {
-					activeMergeOptions = mergePlannerOptions
-				} else {
-					activeMergeOptions = kickMsg.moptions
-				}
-
+			if ctrlMsg == nil && ourSnapshot.epoch != lastEpochMergePlanned {
+				ctrlMsg = ctrlMsgDflt
+			}
+			if ctrlMsg != nil {
 				startTime := time.Now()
 
 				// lets get started
-				err := s.planMergeAtSnapshot(ourSnapshot, activeMergeOptions, (kickMsg != nil))
+				err := s.planMergeAtSnapshot(ourSnapshot, ctrlMsg.options,
+					ctrlMsg.creator)
 				if err != nil {
 					atomic.StoreUint64(&s.iStats.mergeEpoch, 0)
 					if err == segment.ErrClosed {
@@ -80,10 +80,11 @@ OUTER:
 					continue OUTER
 				}
 
-				if kickMsg != nil {
-					close(kickMsg.doneCh)
-					kickMsg = nil
+				if ctrlMsg.doneCh != nil {
+					close(ctrlMsg.doneCh)
 				}
+
+				ctrlMsg = nil
 
 				lastEpochMergePlanned = ourSnapshot.epoch
 
@@ -105,7 +106,7 @@ OUTER:
 			case <-s.closeCh:
 				break OUTER
 			case s.persisterNotifier <- ew:
-			case kickMsg = <-s.mergerKickCh:
+			case ctrlMsg = <-s.mergerKickCh:
 				continue OUTER
 			}
 
@@ -114,7 +115,7 @@ OUTER:
 			case <-s.closeCh:
 				break OUTER
 			case <-ew.notifyCh:
-			case kickMsg = <-s.mergerKickCh:
+			case ctrlMsg = <-s.mergerKickCh:
 			}
 		}
 
@@ -146,25 +147,39 @@ func (s *Scorch) parseMergePlannerOptions() (*mergeplan.MergePlanOptions,
 	return &mergePlannerOptions, nil
 }
 
-type mergerKick struct {
-	moptions *mergeplan.MergePlanOptions
-	doneCh   chan struct{}
+type mergerCtrl struct {
+	creator string
+	options *mergeplan.MergePlanOptions
+	doneCh  chan struct{}
 }
 
-func (s *Scorch) RequestMerge(mo *mergeplan.MergePlanOptions) error {
+func (s *Scorch) RequestMerge(mo *mergeplan.MergePlanOptions,
+	doneCh chan struct{}) error {
 	// check whether force merge is already under processing
-	s.rootLock.Lock()
 	if atomic.LoadUint64(&s.stats.TotFileMergeForceOpsStarted) >
 		atomic.LoadUint64(&s.stats.TotFileMergeForceOpsCompleted) {
-		s.rootLock.Unlock()
+		if doneCh != nil {
+			close(doneCh)
+		}
 		return fmt.Errorf("compaction already in progress")
 	}
-	s.rootLock.Unlock()
 
 	go func() {
+		defer func() {
+			// doneCh for the caller to track the async merge completion
+			if doneCh != nil {
+				close(doneCh)
+			}
+		}()
+
+		if atomic.LoadUint64(&s.stats.TotFileMergeForceOpsStarted) >
+			atomic.LoadUint64(&s.stats.TotFileMergeForceOpsCompleted) {
+			return
+		}
 		atomic.AddUint64(&s.stats.TotFileMergeForceOpsStarted, 1)
 		for {
-			msg := &mergerKick{moptions: mo, doneCh: make(chan struct{})}
+			msg := &mergerCtrl{creator: forceMergeCreator,
+				options: mo, doneCh: make(chan struct{})}
 			// kick the merger
 			select {
 			case s.mergerKickCh <- msg:
@@ -204,7 +219,7 @@ func (s *Scorch) RequestMerge(mo *mergeplan.MergePlanOptions) error {
 }
 
 func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
-	options *mergeplan.MergePlanOptions, forceCompact bool) error {
+	options *mergeplan.MergePlanOptions, creator string) error {
 	// build list of zap segments in this snapshot
 	var onlyZapSnapshots []mergeplan.Segment
 	for _, segmentSnapshot := range ourSnapshot.segment {
@@ -321,11 +336,8 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 			oldNewDocNums: oldNewDocNums,
 			new:           seg,
 			notify:        make(chan *IndexSnapshot, 1),
+			creator:       creator,
 		}
-		if forceCompact {
-			sm.creator = "introduceForceMerge"
-		}
-
 		notifications = append(notifications, sm.notify)
 
 		// give it to the introducer
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -30,6 +30,8 @@ import (
 
 func (s *Scorch) mergerLoop() {
 	var lastEpochMergePlanned uint64
+	var activeMergeOptions *mergeplan.MergePlanOptions
+	var kickMsg *mergerKick
 	mergePlannerOptions, err := s.parseMergePlannerOptions()
 	if err != nil {
 		s.fireAsyncError(fmt.Errorf("mergePlannerOption json parsing err: %v", err))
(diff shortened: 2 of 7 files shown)


index/scorch/introducer.go
```diff
@@ -323,11 +323,15 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 
 	defer func() { _ = root.DecRef() }()
 
+	creator := "introduceMerge"
+	if nextMerge.forceCompact {
+		creator = "introduceForceMerge"
+	}
```
steveyen: looking this more, perhaps it feels better to add creator as param to introduceMerge() rather than forceCompact flag?

index/scorch/merge.go
```diff
@@ -129,8 +146,65 @@ func (s *Scorch) parseMergePlannerOptions() (*mergeplan.MergePlanOptions,
 	return &mergePlannerOptions, nil
 }
 
+type mergerKick struct {
+	moptions *mergeplan.MergePlanOptions
+	done     chan struct{}
```
steveyen: not sure if this is common bleve internals naming convention, but perhaps the ch suffix... like "doneCh"?

sreekanth-cb: oops..I missed this. thanks @steveyen !

index/scorch/introducer.go
sreekanth-cb: indeed... thats much cleaner :)

sreekanth-cb:
hey @mschoch, had you got a chance to look at this change set?

index_alias_impl.go
```diff
@@ -44,6 +44,13 @@ func NewIndexAlias(indexes ...Index) *indexAliasImpl {
 	}
 }
 
+func (i *indexAliasImpl) Indexes() []Index {
+	i.mutex.RLock()
+	rv := i.indexes
+	i.mutex.RUnlock()
+	return rv
+}
+
```
steveyen: btw, this change seems unrelated to theme of this force-merge change?

Also, it needs more documentation?  e.g., the locking here isn't enough, as the Add/Remove() methods can mutated the underlying i.indexes?  Need to warn the user to make all their add/remove() calls upfront before calling this Indexes() implementation as isn't safe in the face of concurrent add/remove() calls?

index/scorch/persister.go
```diff
@@ -789,20 +800,24 @@ var NumSnapshotsToKeep = 1
...
+func (s *Scorch) removeOldBoltSnapshots(forceCompact bool) (numRemoved int, err error) {
 	persistedEpochs, err := s.RootBoltSnapshotEpochs()
 	if err != nil {
 		return 0, err
 	}
 
-	if len(persistedEpochs) <= s.numSnapshotsToKeep {
+	numSnapshotsToKeep := s.numSnapshotsToKeep
+	if forceCompact {
+		numSnapshotsToKeep = 1
```
steveyen: i think i'd prefer instead more "powerful" API's rather than passing in forceCompact which feels like a special case?  like the caller might just pass in numSnapshotsToKeep as the parameter rather than forceCompact?

e.g., more like... if the incoming numSnapshotsToKeep param is 0, then perhaps default to s.numSnapshotsToKeep?

index/scorch/persister.go
```diff
@@ -212,7 +212,8 @@ OUTER:
 		case s.introducerNotifier <- w:
 		}
 
-		s.removeOldData() // might as well cleanup while waiting
+		forceCompact := s.isForceCompacted(ourSnapshot)
```
steveyen: somehow... this feels racey?  not sure...

...but thinking as I'm reading this codepath... that the force-compact happened long ago...

and it's somehow affecting how you want to cleanup old files right now?

wondering if we didn't change the removeOldData() behavior, would the other parts of the force-compact proposal still just work?

index/scorch/merge.go
```diff
@@ -129,8 +146,65 @@ func (s *Scorch) parseMergePlannerOptions() (*mergeplan.MergePlanOptions,
...
+			}
+		}
+	}()
+
+	return nil
+}
+
 func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
-	options *mergeplan.MergePlanOptions) error {
+	options *mergeplan.MergePlanOptions, forceCompact bool) error {
```
steveyen: one thought... thinking why not pass in the creator string as param?  The design "theory" (that's too fancy a word?) or the API "taste" there is that it'd be a more generalized API than a narrower "forceCompact" flag?

index/scorch/merge.go
```diff
@@ -129,8 +146,65 @@ func (s *Scorch) parseMergePlannerOptions() (*mergeplan.MergePlanOptions,
...
+
+func (s *Scorch) RequestMerge(mo *mergeplan.MergePlanOptions) error {
+	// check whether force merge is already under processing
+	s.rootLock.Lock()
+	if atomic.LoadUint64(&s.stats.TotFileMergeForceOpsStarted) >
+		atomic.LoadUint64(&s.stats.TotFileMergeForceOpsCompleted) {
+		s.rootLock.Unlock()
+		return fmt.Errorf("compaction already in progress")
+	}
+	s.rootLock.Unlock()
```
steveyen: now that time has passed since looking at this PR... i this is going to need a little bit more thought...

when i see lots of atomics being used, along with while they're embedded inside of locking... my gut feel worry instincts start tingling :-)

aside: it's slightly amazing that you all the way in Bangalore can cause this time-and-space delayed effect to happen halfway around the world!  we live in amazing times! :-)

quick thought is that reading these atomics while under a lock might not be helpful... especially if the mutations of these atomics are not under a lock.

And, then, if the next proposed move might be perhaps to refactor these atomics into the protection of a lock, then the thought would be... why would we need atomics?

anyways, stuff like these appear in my mind

...as

index/scorch/merge.go
```diff
@@ -54,11 +56,17 @@ OUTER:
 			atomic.StoreUint64(&s.iStats.mergeEpoch, ourSnapshot.epoch)
 			s.rootLock.Unlock()
 
-			if ourSnapshot.epoch != lastEpochMergePlanned {
+			if ourSnapshot.epoch != lastEpochMergePlanned || kickMsg != nil {
+				if kickMsg == nil {
```
steveyen: one refactoring thought...

...is that now this codeline has some "if kickMsg" somewhat woven throughout this codepath...

instead, perhaps always create a kickMsg instance by default, such as in the default case when kickMsg comes in as nil?  That way, you might be able to avoid these if-kickMsg checks which lessen some mental burden?

and, if so, perhaps rename the kickMsg?

you can see, then, we're suddenly in the difficult realm of naming things :-)

index/scorch/merge.go
sreekanth-cb: yes..there is a raciness issue there.. thanks @steveyen .
I need to add the exit on whether "force merge is already under processing" check inside the go-routine at line no:164.5 as well as there would be a random delay in the actual scheduling of a go-routine at runtime. 
So with the current code  - there could be N - concurrent go-routines all passing the original checks there at line:157 and triggering a forced merge routine..
Now it looks like, I don't really need a s.rootLock at all, all atomic checks will be enough...is what comes to mind.. With the additional check inside the routine at line:164.5, only one routine will win the force merge..

index_alias_impl.go
sreekanth-cb: this change is related, this help at a cbft level for triggering force merges to remote partitions, we need to know which all partitions are hidden behind an alias. Something similar to mocking the Multisearch sort mult index functionality at the application level..

yes, let me add more commentary there..

index/scorch/merge.go
sreekanth-cb: let me give a try though I may not be able to achieve what you suggest entirely... 

On other hand - what I feel is these changes are shallow , or mental burden is lesser as you already see where this kickMsg comes from or a != nil check happening right there .. So without even loading the whole code flow in mind you can follow whats happening...

But in the proposed elegant model of a generic flow - where values different across invocations from forceMerge/normal work cycle would impose more mental flow as you need to understand the whole code or load more items in mind to track whats happening there... just a thought :)

index/scorch/persister.go
```diff
@@ -245,6 +245,14 @@ func notifyMergeWatchers(lastPersistedEpoch uint64,
 	return watchersNext
 }
 
+func (s *Scorch) getNumSnapshotsToKeep(ourSnapshot *IndexSnapshot) int {
+	if ourSnapshot != nil &&
+		ourSnapshot.creator != forceMergeCreator {
+		return s.numSnapshotsToKeep
+	}
+	return 1
```
abhinavdangeti: Wondering if violating the numSnapshotsToKeep contract here in case of forced merge is OK?
I may be missing something obvious here, but this is something you're not doing below within pausePersisterForMergerCatchUp ?

sreekanth-cb: Indeed, the reason we are doing it here is that - Usually a full compaction will need 2X++ size requirement. In other words, we don't want to keep multiple copies of the same data.. So this is deliberate decision.

index/scorch/merge.go
```diff
@@ -129,8 +147,79 @@ func (s *Scorch) parseMergePlannerOptions() (*mergeplan.MergePlanOptions,
...
+
+func (s *Scorch) RequestMerge(mo *mergeplan.MergePlanOptions,
+	doneCh chan struct{}) error {
+	// check whether force merge is already under processing
+	if atomic.LoadUint64(&s.stats.TotFileMergeForceOpsStarted) >
+		atomic.LoadUint64(&s.stats.TotFileMergeForceOpsCompleted) {
+		if doneCh != nil {
+			close(doneCh)
+		}
+		return fmt.Errorf("compaction already in progress")
```
mschoch: maybe just minor naming thing here, the API operation is called merge, but the error message talks about compaction

index/scorch/merge.go
```diff
@@ -129,8 +147,79 @@ func (s *Scorch) parseMergePlannerOptions() (*mergeplan.MergePlanOptions,
...
+		defer func() {
+			// doneCh for the caller to track the async merge completion
+			if doneCh != nil {
+				close(doneCh)
+			}
+		}()
+
+		if atomic.LoadUint64(&s.stats.TotFileMergeForceOpsStarted) >
+			atomic.LoadUint64(&s.stats.TotFileMergeForceOpsCompleted) {
+			return
```
mschoch: this is the second time we've checked the stats to see if it's already running, however in the first case you get a helpful error message, and in this case you don't (yes I see the goroutine has no return, but i'm wondering what the thinking is for this difference in behavior)

index/scorch/merge.go
```diff
@@ -129,8 +147,79 @@ func (s *Scorch) parseMergePlannerOptions() (*mergeplan.MergePlanOptions,
...
+		}
+		atomic.AddUint64(&s.stats.TotFileMergeForceOpsStarted, 1)
```
mschoch: it seems to me that this use of atomics to avoid running multiple force merges isn't safe.  why couldn't 2 competing callers have both run through all the checks, and both gotten to this point, where 2 merges now start?

index/scorch/persister.go
mschoch: I share the concern, and don't understand the rationale.  What about a force merge implies that I want to violate the regular settings?

sreekanth-cb: Let me try to explain this with a merge policy for a single index file,
In this case, the entire data is getting rewritten to a single new file. So if we are keeping the older N snapshots, then this would result in a lot of space consumption due to data duplication. Also during merge phase - the disk consumption was going very high.

This extra space cost is  hard to justify when a customer try force merge for some potential benefits out it.

mschoch: But you're using phrase like "a lot of space", but what does that mean?  Maybe the user understands their space requirements and planned accordingly, but you've already decided for them to ignore the configuration that they specified.

It seems like your real complaint is that N snapshots != N useful snapshots.  Forced merge/compaction often results in useless snapshots.  I would prefer we try to address that issue separately, or at least make this change in behavior optional.

sreekanth-cb: I used the term a "lot of" because during my experiments, I saw the disk space spiking upto 3X or even more during the merge process. But eventually it dips to 2X or some bigger value. The eventual file size partly could be attributed to larger segment bloating issue. And as of today, a single segment merge was spiking the space usage more than 2X. 

The wrinkle I see(non convincing part) here is that - people often associate compaction/forceMerge with an index size reduction/space savings and such bigger requirements of space is contradicting that.

Secondly, another thought popping to mind while writing this is that (may be I have lost the rollback fundamentals from my mind)  keeping 3/N number of snapshots isn't a a big win after all. For eg: in cbft at **max** each batch/snapshot can contain only 200 mutations. 

So with the default snapshot keep count of 3 will save 600(3X200) mutations.

Faster rollback ability to 600 mutation isn't a huge win considering the time to re-index it?
So don't think its a significant trade off to drop older snapshots here..

Making this configurable is certainly an option too..

mschoch: So, in normal usage, keeping 3 snapshots should not mean anything close to 3x the size of the index.  Snapshots share older segments, and so in many typical cases, I would expect much of the data of the index to be shared in older segments that haven't changed enough to be rewritten.  Only a handful of segments would differ, and thus be kept around.  These would tend to be newer/smaller segments anyway.  There are of course other use cases where you're constantly modifying the same documents over and over, and yes in that case 3 old snapshots might mean having 3 complete sets of segments representing the data, and thus 3x the data.  However, I would not expect that to be common.

It's also possible that our current known issue of older/larger segments being less space efficient actually undermines this math, and is contributing to confusion.

Next, I would agree with you that generally people would expect *the result* of a merge/compaction to possibly save space.  I also think they would expect that *during* the operation it could in fact consume 2x more.  I think with appropriate documentation and explanation about the way merging works would convince most people.

I'm still holding on to my main objection.  The user requested N old snapshots in their configuration.  And this code intentionally ignores that configuration, for the stated reason of not taking too much space (even though the code has no idea how much space that is, or if the user has planned appropriately).  As a user, when code silently does something different than I asked, I find that to be a problem.

sreekanth-cb: I agree that overriding the `num snapshots to keep` is a side effect of this API, but my plan was to document that in the API specs. As the rewards there(by keeping 3/N older snapshots) anyway is of lesser significance to the user.

But any extra space mandates from the API call has a higher cost factor to take care of which might needs cluster size planning Or given the un bounded nature of size requirements as of today- we can't rule out the possibility of pushing them to buy more hardware, perhaps?. This concern outweighs the former for me.

I also agree that the effectiveness(return on investment) of existing `num snapshots to keep` feature is completely separate topic. 
But since we already touched that here-  I wasn't saying the upper bound was going to be strictly 3X or I didn't try to quantify that earlier during my trials with this API.

Even in a normal scenario,
If a single mutation gets indexed,  it could create a newer disk segment file, which could potentially trigger a merge operation. Merge can invalidate any random N number of disk files from different size tiers to result in a huge size duplication too.. (or there is no assurance that merge will happen only at the lower/bottom tier of smaller segments?)  
And let's say - if a single merge operation involved 1 older segment of gigantic size, then we will end up creating lot of space waste out of a single mutation ingestion. 

So on an average, the extra space occupied by these reasonably sized shared disk segments aren’t justifying the planned meagre performance boost for the seldom(debatable) occurring rollback that to upto a maximum point in history of a few mutations.?

index/scorch/merge.go
sreekanth-cb: true, you are right!

index/scorch/merge.go
sreekanth-cb: good catch! i missed it..

index/scorch/merge.go
```diff
@@ -129,8 +147,116 @@ func (s *Scorch) parseMergePlannerOptions() (*mergeplan.MergePlanOptions,
...
+	OverrideNumSnapshotsToKeep bool
+
+	// DoneCh indicates the completion of the asynchronous
+	// forced merge operation.
+	DoneCh chan struct{}
+}
+
+func (s *Scorch) RequestMerge(mr *MergeRequest) error {
+	// check whether force merge is already under processing
+	if atomic.LoadUint64(&s.stats.TotFileMergeForceOpsStarted) >
```
mschoch: I'm wondering why you still do this (ineffective in my opinion) check at the top, and then do the safe check (using lock) inside the goroutine you've spawned.  By doing it this way, you have to introduce the error channel and more complex error return handling code.  Couldn't you have placed the safe check (using lock) at the top of this function, and then you no longer need to check again inside the goroutine, as you've already ensured that any concurrent invocations have been stopped early.  Then the need for the error channel goes away too.

sreekanth-cb: hm, my worry there was - since the go routine spin ups and scheduling is an async process there could be many threads passing the first check and may end up seeing green signals for spinning up go routines until the first routine really goes gets to the execution to bump those stats..?

PR #1314, opened 2019-11-26
Remove dictionary rows when no document is referring to it

### LINKED ISSUE #374: Data leak with goleveldb backend?

tanzhang0504:
Hi - we have a project that uses bleve in-memory index based on gtreap implementation. We noticed a similar behavior that dictionary rows keep grow even after all the documents referring to them are deleted. I found https://github.com/blevesearch/bleve/issues/374, and seems that someone made a temporary only to leveldb, which does not apply to bleve in-memory index use-case.

So I've made this PR to systematically fix #374 for all different index store implementations. Looking forward to your feedback!

### DIFF
--- a/index/upsidedown/upsidedown_test.go
+++ b/index/upsidedown/upsidedown_test.go
@@ -16,6 +16,7 @@ package upsidedown
 
 import (
 	"log"
+	"os"
 	"reflect"
 	"regexp"
 	"strconv"
@@ -30,12 +31,23 @@ import (
 	"github.com/blevesearch/bleve/document"
 	"github.com/blevesearch/bleve/index"
 	"github.com/blevesearch/bleve/index/store/boltdb"
+	"github.com/blevesearch/bleve/index/store/goleveldb"
+	"github.com/blevesearch/bleve/index/store/gtreap"
 	"github.com/blevesearch/bleve/index/store/null"
 	"github.com/blevesearch/bleve/registry"
 )
 
 var testAnalyzer = &analysis.Analyzer{
 	Tokenizer: regexpTokenizer.NewRegexpTokenizer(regexp.MustCompile(`\w+`)),
+}
+
+var gtreapTestConfig = map[string]interface{}{
+	"path": "",
+}
+
+var levelTestConfig = map[string]interface{}{
+	"path":              "test_dir",
+	"create_if_missing": true,
 }
 
 func TestIndexOpenReopen(t *testing.T) {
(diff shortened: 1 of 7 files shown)


index/upsidedown/upsidedown_test.go
```diff
@@ -187,126 +199,119 @@ func TestIndexInsertThenDelete(t *testing.T) {
...
+	expectedCount := uint64(0)
+	validateDocCount(t, idx, expectedCount)
+	// 1 for version.
+	expectedRows := uint64(1)
+	validateRowCount(t, idx, expectedRows)
 
 	doc := document.NewDocument("1")
 	doc.AddField(document.NewTextField("name", []uint64{}, []byte("test")))
-	err = idx.Update(doc)
+	doc.AddField(document.NewTextField("name1", []uint64{}, []byte("test1")))
```
tanzhang0504: I enhanced the test to include one index that's shared by two documents, and the other is not. I'm also happy to revert to the original test setup if you think this hurts readability.

tmm1:
Similar issue was discussed in https://github.com/blevesearch/blevex/pull/30 and https://github.com/blevesearch/blevex/pull/44

tanzhang0504:
> Similar issue was discussed in [blevesearch/blevex#30](https://github.com/blevesearch/blevex/pull/30) and [blevesearch/blevex#44](https://github.com/blevesearch/blevex/pull/44)

Hi Aman, thanks for the pointers. I'm also wondering is it possible to not generate dictionary rows altogether? Seems like it's needed by bleve to look up the terms. I did an experiment to not insert dictionary rows (by removing this line - https://github.com/blevesearch/bleve/blob/5b9d7756e18c7c18b507d148ed4936896bd50370/index/upsidedown/upsidedown.go#L287). Then my inserted index can't be found when looking up with "+field:val" query.

tanzhang0504:
@mschoch Just wondering any comment on my PR? I saw the past discussions in https://github.com/blevesearch/blevex/pull/30 as pointed out by Aman, and seems like the suggested solution is to add a Compact() function for each DB store and rely on callers to invoke that function. Just wondering will that add too much burden on the caller instead of fixing the leak at the store level? Any feedback is greatly appreciated!

PR #1317, opened 2019-12-10
Add a Compact() function to gtreap store.

tanzhang0504:
This is a fix for memory leak from in-mem index (blevesearch/bleve#374). 

I made a fix in the underlying store - https://github.com/blevesearch/bleve/pull/1314. but seems like Compact() function is recommended by the owner - https://github.com/blevesearch/blevex/pull/30.


tanzhang0504:
I did not add batch_size as done by compact() in other two stores, because the in-memory operation is usually much faster, and the caller can adjust the frequency of calling compact() to limit the latency per compact(). PLMK your thoughts.

PR #1327, opened 2020-01-14
MB-37500 - Polygon query with invalid geohash gives results

sreekanth-cb:
Adding geohash length validation checks

### DIFF
--- a/geo/parse.go
+++ b/geo/parse.go
@@ -85,9 +85,11 @@ func ExtractGeoPoint(thing interface{}) (lon, lat float64, success bool) {
 			}
 		} else {
 			// geohash
-			lat, lon = DecodeGeoHash(geoStr)
-			foundLat = true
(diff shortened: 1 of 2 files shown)


geo/parse.go
```diff
@@ -85,9 +85,11 @@ func ExtractGeoPoint(thing interface{}) (lon, lat float64, success bool) {
 			}
 		} else {
 			// geohash
-			lat, lon = DecodeGeoHash(geoStr)
-			foundLat = true
-			foundLon = true
+			if len(geoStr) <= 12 {
```
abhinavdangeti: Think it'd do good to declare a const GEO_HASH_MAX_LENGTH = 12 and use that here.

sreekanth-cb: yep, let me replace with geoHashMaxLength.

PR #1328, opened 2020-01-16
Fix concurrent read write on analysis

aviadl:
This is the stack trace we got in our tests
```
fatal error: concurrent map read and map write

goroutine 7 [running]:
runtime.throw(0x17b1b17, 0x21)
	/usr/local/go/src/runtime/panic.go:774 +0x72 fp=0xc00c97f8b8 sp=0xc00c97f888 pc=0x431422
runtime.mapaccess2_faststr(0x15f4040, 0xc011202f90, 0x257c531, 0x1, 0xc00c97fa70, 0xcb5919)
	/usr/local/go/src/runtime/map_faststr.go:116 +0x48f fp=0xc00c97f928 sp=0xc00c97f8b8 pc=0x4150ff
github.com/blevesearch/bleve/analysis.TokenFrequencies.MergeAll(0xc011202f90, 0x14488dd, 0x2, 0xc01121cc00)
	/home/circleci/.go_workspace/pkg/mod/github.com/blevesearch/bleve@v0.0.0-20190812174355-c997533a776f/    +0x100 fp=0xc00c97fa38 sp=0xc00c97f928 pc=0x9d6770
github.com/blevesearch/bleve/document.(*CompositeField).Compose(0xc00e499d80, 0x14488dd, 0x2, 0x1, 0xc01121cc00)
	/home/circleci/.go_workspace/pkg/mod/github.com/blevesearch/bleve@v0.0.0-20190812174355-c997533a776f/document/field_composite.go:122 +0xf2 fp=0xc00c97fa80 sp=0xc00c97fa38 pc=0x9dd902
github.com/blevesearch/bleve/index/upsidedown.(*UpsideDownCouch).Analyze(0xc001dbda80, 0xc00e499d00, 0x2)
	/home/circleci/.go_workspace/pkg/mod/github.com/blevesearch/bleve@v0.0.0-20190812174355-c997533a776f/index/upsidedown/analysis.go:77 +0xc10 fp=0xc00c97ff28 sp=0xc00c97fa80 pc=0xc9b630
github.com/blevesearch/bleve/index.AnalysisWorker(0xc0004f1e00, 0xc0004f1e60)
	/home/circleci/.go_workspace/pkg/mod/github.com/blevesearch/bleve@v0.0.0-20190812174355-c997533a776f/index/analysis.go:106 +0x55 fp=0xc00c97ffd0 sp=0xc00c97ff28 pc=0x9e12b5
runtime.goexit()
	/usr/local/go/src/runtime/asm_amd64.s:1357 +0x1 fp=0xc00c97ffd8 sp=0xc00c97ffd0 pc=0x460741
created by github.com/blevesearch/bleve/index.NewAnalysisQueue
	/home/circleci/.go_workspace/pkg/mod/github.com/blevesearch/bleve@v0.0.0-20190812174355-c997533a776f/index/analysis.go:94 +0xc8
```

I think that the root cause is that iterating on the map, pointers change under the hood, which will lead to analysing the same document twice

### DIFF
--- a/index/upsidedown/upsidedown.go
+++ b/index/upsidedown/upsidedown.go
@@ -820,7 +820,8 @@ func (udc *UpsideDownCouch) Batch(batch *index.Batch) (err error) {
 
 	if numUpdates > 0 {
 		go func() {
-			for _, doc := range batch.IndexOps {
+			for k := range batch.IndexOps {
+				doc := batch.IndexOps[k]
 				if doc != nil {
 					aw := index.NewAnalysisWork(udc, doc, resultChan)
 					// put the work on the queue


aviadl:
@mschoch  @steveyen @abhinavdangeti  Can you please review this ?

abhinavdangeti:
Thanks for fix @aviadl. Merging this now.

sreekanth-cb:
@aviadl , @abhinavdangeti , Is this the right fix for the stack trace shown there?
Isn't the read/write happens here -
https://github.com/blevesearch/bleve/blob/master/analysis/freq.go#L93
https://github.com/blevesearch/bleve/blob/master/analysis/freq.go#L103

aviadl:
@sreekanth-cb from the way I understand the code
Each such map that we got the concurrent exception on should be used by one go routine
The reason for the panic is that iterating on the fields the way it was initially done, the same field was passed to different go routines (due to go nature of iterating maps)
My fix prevents that, so the code in freq.go will not fail now

mschoch:
Somehow I just saw this for the first time (bunch of github stuff was going to spam folder).  I also disagree this fix works (at least as explained here).

aviadl:
@mschoch you can look at the following unit test that illustrates the problem

```
func TestMapConncurrent(t *testing.T) {
	type g struct {
		n int
	}
	m := map[int]*g{}
	for i := 0; i < 10000; i++ {
		m[i] = &g{i}
	}
	for _, v := range m {
		go func() {
			fmt.Println(v.n)
		}()
	}
}
```
When looking at the prints, you will that the numbers are not unique as they should be
And BTW, since this change, we didnt encounter those failures

mschoch:
@aviadl that case I understand, however it does not obviously match the code in question (perhaps it does in some less obvious way)

aviadl:
The code in question is doing the exact same thing
It iterates on a map, and takes the value from the map and places it as a closure in a go routine
This pointer is over written, and since the inner object in Bleve case contains a map
we will have 2 go routines that will touch this map at the same time

mschoch:
The example shown here spawns a goroutine inside the for loop and incorrectly binds to a variable.  The code in question: https://github.com/blevesearch/bleve/blob/master/index/upsidedown/upsidedown.go#L821-L832 does not spawn a goroutine that I can see.

aviadl:
The object is placed in a queue 
Which in turn is run in a go routine

mschoch:
I supposed I just don't understand Go.  My understanding of the original is that the binding is incorrect because the actual spawning of the goroutine may happen at some later time after the value in the variable has already changed.  That clearly can't be what is going on in this case, because the goroutines in question are pooled, already spawn long ago, and that the line which creates the NewAnalysisWork(...) should (in my undertanding) correctly bind the variable in a deterministic way prior to the loop continuing.

PR #1332, opened 2020-02-03
Fix for numSnapshotsToKeep dilution with merge ops

sreekanth-cb:
The numSnapshotsToKeep logic is perceived to have two
shortcomings wrt the disk based segment merging.
1- It can result in index size amplification due to the
   duplicated segment data across index snapshots
   (kept for respecting numSnapshotsToKeep ) before and after a 
   disk based merge operation.

2- While retaining the mandated number of older snapshots, scorch isn’t
   considering the above mentioned data duplication across index
   snapshots and thus compromises the effectiveness of numSnapshotsToKeep.
   Essentially scorch isn’t retaining enough snapshots with non-duplicated data.

This fix aims to address that by accounting the recent/ongoing merge
operations to prevent the redundant snapshots from getting sneaked
into the numSnapshotsToKeep reserve.

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -159,7 +159,7 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 	// process tasks in serial for now
 	var notifications []chan *IndexSnapshot
 	var filenames []string
-	for _, task := range resultMergePlan.Tasks {
+	for tnum, task := range resultMergePlan.Tasks {
 		if len(task.Segments) == 0 {
 			atomic.AddUint64(&s.stats.TotFileMergePlanTasksSegmentsEmpty, 1)
 			continue
@@ -239,6 +239,11 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 			}
 
 			atomic.AddUint64(&s.stats.TotFileMergeSegments, uint64(len(segmentsToMerge)))
+		}
+
+		// update stats before the first introduction from the merge
+		if tnum == 0 {
+			atomic.StoreUint64(&s.stats.RecentMergeInProgEpoch, ourSnapshot.epoch)
 		}
 
 		sm := &segmentMerge{
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -801,9 +801,21 @@ func (s *Scorch) removeOldBoltSnapshots() (numRemoved int, err error) {
 	}
 
 	// make a map of epochs to protect from deletion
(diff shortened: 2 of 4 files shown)


mschoch:
Would benefit from having a test.

mschoch:
First thought is wondering if we're mixing up things that are useful stats and things which are actually just state for the scorch index.  In this case it seems to me that `RecentMergeInProgEpoch` isn't a particularly useful stat, it's just a useful value to keep and operate on atomically.  Is there a benefit to treating this as a stat that I'm not seeing?

sreekanth-cb:
By keeping this as a stat, cbft is easily getting access to it over the stats endpoint and thought this might help us to gain some insights in case of a forced merge operations in future.
When we fire the force merge(for a single file index) against a partition with lets say 19 files,  
there could be a merge task of 10 files and 9 files in the first merge cycle and a 2 files merge task in the second cycle by the merger.
By checking the stats before and the force merge api trigger and after, we could potentially debug something was the thinking behind this..

steveyen:
hmmm... here's how i'm starting to think about it...

the most recent segments that came from recent incoming mutations certainly have some interesting rollback information (that is, JSON blob info recorded via those SetInternal() calls).

And, in contrast... let's consider the segment that's the result or the output of the merger...  For example, say we have segments A, B, C which are input to the merger (which produces output segment "A+B+C")... then input segment C might not be worth keeping around, since segment C has the same rollback information (SetInternal() blob) as the merged segment "A+B+C"?

Not sure if segment A and if segment B are also worth keeping around, too, so that needs more thought?

So, while cleaning up old segments, if the system could somehow realize segment C (and others like it) is more worthy of deletion sooner than others... that might be the angle to think through?  

Not sure if that ought to be tracked as a stat, too?  Perhaps you'd just love to grab a blue crayon and mark segment C with blue grafitti so you can tell later that it's blue?  :-)

And, if that makes sense, how to express that big trail of thoughts concisely and how to name it... fun times :-)

sreekanth-cb:
> Not sure if segment A and if segment B are also worth keeping around, too, so that needs more thought?

Very good point! @steveyen
Indeed the proposed approach fails in such cases where the mutations in A and B are of updates nature on the same documents/VBs. We fail to retain it. The index snapshots which refers to A and B needs to be retained.

Hence closing this PR. Let me create a ticket to track this.

> So, while cleaning up old segments, if the system could somehow realize segment C (and others like it) is more worthy of deletion sooner than others... that might be the angle to think through?

Needs more thought on this as the segment clean ups goes hand in hand with that of index snapshots. As long as long as the index snapshots which contains A & B are retained, those index snapshots also has the reference to the segment C. So not sure how plausible it is to free the segment C alone in a sooner manner!

PR #1334, opened 2020-02-05
Enabling score/id based search_after/before paginations

sreekanth-cb:
This change attempts to enable users provide the last
fetched document's score or ID values as keys in the
search after/before fields to make the deep pagination
works for the default _score or _id based sort orders.

### DIFF
--- a/search/collector/topn.go
+++ b/search/collector/topn.go
@@ -17,6 +17,7 @@ package collector
 import (
 	"context"
 	"reflect"
+	"strconv"
 	"time"
 
 	"github.com/blevesearch/bleve/index"
@@ -90,6 +91,18 @@ func NewTopNCollectorAfter(size int, sort search.SortOrder, after []string) *Top
 	rv.searchAfter = &search.DocumentMatch{
 		Sort: after,
 	}
+
+	for pos, ss := range sort {
+		if ss.RequiresDocID() {
+			rv.searchAfter.ID = after[pos]
+		}
+		if ss.RequiresScoring() {
+			if score, err := strconv.ParseFloat(after[pos], 64); err == nil {
+				rv.searchAfter.Score = score
+			}
+		}
+	}
+
 	return rv
 }
 
--- a/test/tests/sort/data/d.json
+++ b/test/tests/sort/data/d.json
@@ -2,6 +2,6 @@
 	"id": "d",
 	"age": 65,
   "born": "1978-12-02",
-	"title": "agent",
+	"title": "agent d is desperately trying out to be successful rapster!",
 	"tags": ["cats"]
 }
--- a/test/tests/sort/data/e.json
+++ b/test/tests/sort/data/e.json
@@ -2,6 +2,6 @@
 	"id": "e",
 	"name": "nancy",
   "born": "1954-10-22",
-	"title": "rapstar",
+	"title": "rapstar nancy rapster",
 	"tags": ["pain"]
 }
--- a/test/tests/sort/data/f.json
+++ b/test/tests/sort/data/f.json
@@ -2,6 +2,6 @@
 	"id": "f",
 	"name": "frank",
 	"age": 1,
-	"title": "taxman",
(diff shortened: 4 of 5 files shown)


sreekanth-cb:
hey @mschoch , I am not sure whether I am overlooking something here Or any apparent flaws in this approach so that we didn't do this in the first iteration.

sreekanth-cb:
removed the poorly designed perf check `if len(sort.RequiredFields()) < len(after) {` as the RequiredFields() collective sortOrder checks ought to be costlier than just directly doing the boolean based RequiresDocID/RequiresScoring on every sort order item.

search/collector/topn.go
```diff
@@ -90,6 +91,19 @@ func NewTopNCollectorAfter(size int, sort search.SortOrder, after []string) *Top
 	rv.searchAfter = &search.DocumentMatch{
 		Sort: after,
 	}
+
+	for pos, ss := range sort {
+		if ss.RequiresDocID() {
+			rv.searchAfter.ID = after[pos]
+			continue
```
steveyen: oh, one thing -- perhaps get rid of this 'continue', just in case the ss object has both RequiresDocID() of true and RequiresScoring() of true?

sreekanth-cb: As of today, SortScore and SortDocID have these properties orthogonal or mutually exclusive. But perhaps in future if we have a new sort type, this could be problem then.

tylerkovacs:
Hi @sreekanth-cb, I think there's still one case where `search_after` is not working with scoring: when there are multiple documents with identical scores for a search.

Since this is not uncommon in my dataset, I always sort results using `["-_score", "_id"]` for consistent ordering. I'm trying to use `search_after` to paginate through the results, but it doesn't work in this case.

Here's a [branch on my fork that demonstrates the issue](https://github.com/blevesearch/bleve/compare/master...tylerkovacs:tyler/scoring-search-after?expand=1).

And the test result:
```
 $ go test ./test/... -v -run TestIntegration -testnum=18
=== RUN   TestIntegration
    TestIntegration: integration_test.go:49: using index type upside_down and kv type boltdb
    TestIntegration: integration_test.go:71: Running test: alias
    TestIntegration: integration_test.go:71: Running test: basic
    TestIntegration: integration_test.go:71: Running test: employee
    TestIntegration: integration_test.go:71: Running test: facet
    TestIntegration: integration_test.go:71: Running test: fosdem
    TestIntegration: integration_test.go:71: Running test: geo
    TestIntegration: integration_test.go:71: Running test: phrase
    TestIntegration: integration_test.go:71: Running test: sort
    TestIntegration: integration_test.go:150: test error - sort by score, with identical scores
    TestIntegration: integration_test.go:151: test 18 - expected hit 0 to have ID g got d
--- FAIL: TestIntegration (0.47s)
FAIL
FAIL	github.com/blevesearch/bleve/test	0.481s
FAIL
```

Since documents `f` and `g` are identical, the `search_after` should return the `g` document. But it skips over it an returns the next document (`d`) instead. This leads me to believe that for sorts that include score, the `search_after` implementation is only looking at the `score` and ignoring other fields. Is that possible?

As an aside, I don't have identical documents in my index but I do have documents that return identical scores for certain queries.

Would you prefer that I open an issue for this rather than commenting here?

sreekanth-cb:
@tylerkovacs, You are right.. I think this is pretty much possible as long as the field values we use for search_after or before won't have unique or duplicated values. It would result in many documents having the same score/rank in the result.

One work around popping to mind would be, along with score, use other meaningful ordering fields for sorting or ranking the result. And if you pass those field values in that order to the search_after/before [] parameter, it would help to create a unique score and order?

mschoch:
@sreekanth-cb my reading of what @tylerkovacs wrote is that they are already using `_id` field to force a unique sort order, but they are still getting incorrect results.  That is what I see on their branch, although I did not try running it myself.

I agree this should be a new issue.

PR #1335, opened 2020-02-13
Adding a getter method for TypeMappings

sreekanth-cb:
in the IndexMappingImpl


sreekanth-cb:
my bad! its not needed...

PR #1337, opened 2020-02-15
new segments now support 1-hit encoding

mschoch:
This commit alters the path for new segments, such that
they too should now consider, and when appropriate, use
the 1-hit encoding optimization.

### DIFF
--- a/index/scorch/segment/zap/new.go
+++ b/index/scorch/segment/zap/new.go
@@ -646,11 +646,15 @@ func (s *interim) writeDicts() (fdvIndexOffset uint64, dictOffsets []uint64, err
 			locs := s.Locs[pid]
 			locOffset := 0
 
+			var lastDocNum, lastFreq, lastNorm uint64
 			postingsItr := postingsBS.Iterator()
 			for postingsItr.HasNext() {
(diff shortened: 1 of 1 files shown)


mschoch:
First, this was the quick and dirty, minimal plausible change to get it working.  It's quite possible this is wrong with a fuller understanding of what is going on.  The question still remains why didn't @steveyen wire this up in the first place?

Also, I put the DONOTMERGE label on this, as we may want to wait and merge this into a v2 or v3 of scorch (I don't see any backwards compat issues, but since we have new versions coming anyway, maybe ship this with that?)

index/scorch/segment/zap/new.go
```diff
@@ -714,6 +731,9 @@ func (s *interim) writeDicts() (fdvIndexOffset uint64, dictOffsets []uint64, err
 
 			tfEncoder.Reset()
 			locEncoder.Reset()
+			lastDocNum = 0
+			lastFreq = 0
+			lastNorm = 0
```
mschoch: actually these are either in the wrong place, or not necessary in this verison, as these variables go out of scope here...

PR #1348, opened 2020-03-16
MB-38303 - merger leaks index snapshot refCounts on index closure

sreekanth-cb:
Upon an index closure event, the file merger skips the
reference count decrement operations at line no: https://github.com/blevesearch/
bleve/blob/05d86ea8f6e30456949f612cf68cf4a27ce8c9c5/index/scorch/merge.go#L261
leading to the leaking of index snapshots. This will further
lead to the segment file leaking as those segment files will
never get ceremoniously un-mapped or handle closed in the event
of an index closure.

details ref - https://issues.couchbase.com/browse/MB-38303

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -159,6 +159,21 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 	// process tasks in serial for now
 	var notifications []chan *IndexSnapshot
 	var filenames []string
+	// clean up any pending notifications from introducer on exit
+	// from an index closure.
+	defer func() {
+		for _, notification := range notifications {
+			select {
+			case newSnapshot := <-notification:
+				atomic.AddUint64(&s.stats.TotFileMergeIntroductionsDone, 1)
+				if newSnapshot != nil {
+					_ = newSnapshot.DecRef()
+				}
+			default:
+			}
+		}
+	}()
+
 	for _, task := range resultMergePlan.Tasks {
 		if len(task.Segments) == 0 {
 			atomic.AddUint64(&s.stats.TotFileMergePlanTasksSegmentsEmpty, 1)
@@ -217,6 +232,7 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 				s.unmarkIneligibleForRemoval(filename)
 				atomic.AddUint64(&s.stats.TotFileMergePlanTasksErr, 1)
 				if err == segment.ErrClosed {
+					// handle any pending index snapshot introduction notifications on exit
 					return err
 				}
 				return fmt.Errorf("merging failed: %v", err)
@@ -253,6 +269,7 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 		// give it to the introducer
 		select {
 		case <-s.closeCh:
+			// handle any pending index snapshot introduction notifications on exit
 			_ = seg.Close()
 			return segment.ErrClosed
 		case s.merges <- sm:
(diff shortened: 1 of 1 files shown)


index/scorch/merge.go
```diff
@@ -263,6 +263,17 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 	}
 
 	for _, notification := range notifications {
+		// update the idex snapshot reference counts for all
+		// the merge introductions which are already done.
+		if len(notification) == cap(notification) {
```
steveyen: Hi -- not sure if I understand the big picture here on this change yet, but looking at this from a very narrow-minded way (e.g., looking at the code through a mechanistic "paper towel roll" point of view)...

This kind of len(ch) == cap(ch) feels like an uncommon and perhaps even a frowned-upon pattern in golang?  Wondering, instead, if the notification channel is full then the reading from this channel might be still the right thing to do?

index/scorch/merge.go
```diff
@@ -263,6 +263,17 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
...
+			atomic.AddUint64(&s.stats.TotFileMergeIntroductionsDone, 1)
+			if newSnapshot != nil {
+				_ = newSnapshot.DecRef()
+			}
+			continue
+		}
+
 		select {
 		case <-s.closeCh:
 			atomic.AddUint64(&s.stats.TotFileMergeIntroductionsSkipped, 1)
```
steveyen: One thought... instead, if we end up here in this closeCh case... perhaps this would be an additional place to invoke the newSnapshot.DecRef()?

sreekanth-cb: indeed, I had that thought.. But couldn't visualise it fully - as we may again meet the scenario where we wait for reading from the notification channel?  newSnapshot := <-notification and a bail out needed there too..

index/scorch/merge.go
sreekanth-cb: hm..i don't think I am able to follow you..

> Wondering, instead, if the notification channel is full then the reading from this channel might be still the right thing to do?

"len(notification) == cap(notification)" is my attempt at checking whether notification channel is full or not.

index/scorch/merge.go
```diff
@@ -159,6 +159,20 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 	// process tasks in serial for now
 	var notifications []chan *IndexSnapshot
 	var filenames []string
+	// clean up any pending notifications from introducer on exit
+	// from an index closure.
+	defer func() {
+		for _, notification := range notifications {
+			if len(notification) > 0 {
```
steveyen: imho, using len(ch) ... (or, as an aside, the related cap(ch))... is a bit of a worrisome no-no in golang, usually indicating a troublesome area...

...because in-between checking the len(ch) and the reading of that ch, theoretically, some other goroutine might sneak in and change the channel's state.  Now, that may not be the case here...

but instead, maybe use the select { case <-ch ...; default } idiom instead?

index/scorch/merge.go
```diff
@@ -159,6 +159,20 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
...
+				newSnapshot := <-notification
+				atomic.AddUint64(&s.stats.TotFileMergeIntroductionsDone, 1)
+				if newSnapshot != nil {
+					_ = newSnapshot.DecRef()
+				}
+			}
+		}
+	}()
+
```
steveyen: also, another thought... now that you have a defer here that consumes the notifications... wondering if the other loop in the main codepath that consumes the notifications might be removable or consolidate'able?

sreekanth-cb: Not sure of removing it as we need the blocking behaviour there for the merge introductions, I admit that I am very poor in visualising any potential consolidations here.. :)

PR #1349, opened 2020-03-16
Initialise the root snapshot's DecRef errs with Close()

sreekanth-cb:
During the scorch's Close call, the root snapshot's
DecRef return errs are ignored. Fixing that to bubble up
the errors so that the higher levels may listen and
act on it.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -263,7 +263,7 @@ func (s *Scorch) Close() (err error) {
 		err = s.rootBolt.Close()
 		s.rootLock.Lock()
 		if s.root != nil {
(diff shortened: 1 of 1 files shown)


index/scorch/scorch.go
```diff
@@ -263,7 +263,7 @@ func (s *Scorch) Close() (err error) {
 		err = s.rootBolt.Close()
```
abhinavdangeti: This err gets overwritten. Is that ok? Should we just return here if this err isn't nil perhaps?

sreekanth-cb: i didn't see the above initialisation.. oops..

Issue #570, opened 2017-04-04
[]string in, string out


porjo:
I have a situation where if I store a []string with a single element I get back a value of type string, whereas if I store a []string with 2 or more elements, I get back a value of type []interface{} 

Is this a property of Go language, or something about Bleve implementation?

This example code below  demonstrates the issue. It should produce output:
```
abc result type string
xyz result type []interface {}
```

```Go
package main

import (
	"fmt"
	"log"
	"reflect"

	"github.com/blevesearch/bleve"
)

type doc struct {
	Name       string
	Recipients []string
}

func main() {
	mapping := bleve.NewIndexMapping()
	index, err := bleve.NewMemOnly(mapping)

	// single string
	id := "1234"
	d1 := doc{Name: "abc", Recipients: []string{"a@example.com"}}
	err = index.Index(id, d1)
	if err != nil {
		log.Fatal(err)
	}

	query := bleve.NewMatchQuery("abc")
	search := bleve.NewSearchRequest(query)
	search.Fields = []string{"Name", "Recipients"}
	searchResults, err := index.Search(search)
	if err != nil {
		log.Fatal(err)
	}

	hit := searchResults.Hits[0]
	fmt.Printf("abc result type %s\n", reflect.TypeOf(hit.Fields["Recipients"]))

	// multi strings
	id = "5678"
	d2 := doc{Name: "xyz", Recipients: []string{"a@example.com", "b@example.com"}}
	err = index.Index(id, d2)
	if err != nil {
		log.Fatal(err)
	}

	query = bleve.NewMatchQuery("xyz")
	search = bleve.NewSearchRequest(query)
	search.Fields = []string{"Name", "Recipients"}
	searchResults, err = index.Search(search)
	if err != nil {
		log.Fatal(err)
	}

	hit = searchResults.Hits[0]
	fmt.Printf("xyz result type %s\n", reflect.TypeOf(hit.Fields["Recipients"]))
}
```

mschoch:
The problem is that stored fields in bleve are kind of broken.  The implementation cannot always return to you the exact original value (which is what users want).  On top of that, support for arrays was added later, and as you found, an array with one element is confused with there being no array at all.

I'm fine with leaving this issue open to track it, but it's going to take a bleve2 type overhaul of stored fields to address this properly.

pbnsilva:
What is the best way to get an array back in the current version?

mschoch:
[...]

PR #1355, opened 2020-03-18
Offline RollbackPoints/Rollback apis for scorch index

sreekanth-cb:
This change comprises of enabling the rollback related scorch APIs
to work only with closed/offline indexes.
Both the API signatures have modified to accept an index path
pointing to the scorch index location to accommodate this change.

The two API approach(RollbackPoints api for fetching the rollback points
and the Rollback api for performing the actual rollback) is retained in
favour of modularity and simplicity of usage.

The Rollback api would make the meta/bolt store with the given
rollback point as the most recent epoch to be loaded.

During the process of restoring the index state to the relevant
rollback point, the new Rollback implementation cleans up the
irrelevant epoch data from the bolt store. This is done to favour
a lean version of persisted epochs in root bolt anytime to further
lead a more efficient disk clean up/utilisation.
Also, this helps to better safe guard against any potential
recursive rollback request as we aren't introducing any newer
data into the system.

This Rollback implementation overrides the numSnapShotsToKeep
config in certain special cases of a rollback point.

### DIFF
--- a/index/scorch/snapshot_rollback.go
+++ b/index/scorch/snapshot_rollback.go
@@ -17,6 +17,7 @@ package scorch
 import (
 	"fmt"
 	"log"
+	"os"
 
 	"github.com/blevesearch/bleve/index/scorch/segment"
 	bolt "github.com/etcd-io/bbolt"
@@ -34,13 +35,22 @@ func (r *RollbackPoint) GetInternal(key []byte) []byte {
 // RollbackPoints returns an array of rollback points available for
 // the application to rollback to, with more recent rollback points
 // (higher epochs) coming first.
-func (s *Scorch) RollbackPoints() ([]*RollbackPoint, error) {
-	if s.rootBolt == nil {
-		return nil, fmt.Errorf("RollbackPoints: root is nil")
+func RollbackPoints(path string) ([]*RollbackPoint, error) {
+	if len(path) == 0 {
+		return nil, fmt.Errorf("RollbackPoints: invalid path")
+	}
+
+	rootBoltPath := path + string(os.PathSeparator) + "root.bolt"
+	rootBoltOpt := &bolt.Options{
+		ReadOnly: true,
+	}
+	rootBolt, err := bolt.Open(rootBoltPath, 0600, rootBoltOpt)
+	if err != nil || rootBolt == nil {
+		return nil, err
 	}
 
 	// start a read-only bolt transaction
-	tx, err := s.rootBolt.Begin(false)
+	tx, err := rootBolt.Begin(false)
 	if err != nil {
 		return nil, fmt.Errorf("RollbackPoints: failed to start" +
 			" read-only transaction")
@@ -49,6 +59,7 @@ func (s *Scorch) RollbackPoints() ([]*RollbackPoint, error) {
 	// read-only bolt transactions to be rolled back
 	defer func() {
 		_ = tx.Rollback()
+		_ = rootBolt.Close()
 	}()
 
 	snapshots := tx.Bucket(boltSnapshotsBucket)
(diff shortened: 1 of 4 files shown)


index/scorch/snapshot_rollback.go
```diff
@@ -105,69 +116,85 @@ func (s *Scorch) RollbackPoints() ([]*RollbackPoint, error) {
...
-// Rollback atomically and durably (if unsafeBatch is unset) brings
-// the store back to the point in time as represented by the
-// RollbackPoint. Rollback() should only be passed a RollbackPoint
-// that came from the same store using the RollbackPoints() API.
-func (s *Scorch) Rollback(to *RollbackPoint) error {
-	if to == nil {
-		return fmt.Errorf("Rollback: RollbackPoint is nil")
+// Rollback atomically and durably brings the store back to the point
+// in time as represented by the RollbackPoint.
+// Rollback() should only be passed a RollbackPoint that came from the
```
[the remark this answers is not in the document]
sreekanth-cb: indeed, thats a wrinkle there... lemme add a validation there..

index/scorch/snapshot_rollback.go
```diff
@@ -105,69 +116,85 @@ func (s *Scorch) RollbackPoints() ([]*RollbackPoint, error) {
...
+// same store using the RollbackPoints() API along with the index path.
+func Rollback(path string, to *RollbackPoint) error {
+	if to == nil || len(path) == 0 {
```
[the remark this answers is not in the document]
sreekanth-cb: done!

index/scorch/snapshot_rollback.go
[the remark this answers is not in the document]
sreekanth-cb: Given the validations in place to guard against a possible index damage from an invalid rollback point, I don't see/foresee any problems with this destructive approach to bolt store.. But the pros are like 
-lean boltdb/store contents to manage
-faster disk clean ups on next index open as there is lesser epoch on bolt

Same time, I am unable to understand the benefits out of an extra epoch getting stacked to front approach too..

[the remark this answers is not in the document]
mschoch: 1.  Is a lean boltdb really a concern? Once you open it and things start running, its going to return the NumSnapshotsToKeep.  And I've never seen a root.bolt ever get very large.  Does it in the use cases you support?

2. To me the only reason there is faster disk cleanup is that you reduce the root.bolt to a single snapshot (going below the users configuration).

3.  You had also mentioned some recursive rollback case, but it seemed to me you always process the RollbackPoints linearly looking for what you want, so I didn't see possibility of loop or recursion.

4.  To me, the best argument for your approach is that it is simple, and we got this wrong the first time.  I still like the idea of introducing a new snapshot that matches the rollback point, because then we go over the desired NumSnapshotsToKeep, instead of below, and once you open it, it returns to normal.  But I won't hold this up, and will let others give their input.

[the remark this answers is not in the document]
sreekanth-cb: > Is a lean boltdb really a concern? Once you open it and things start running, its going to return the NumSnapshotsToKeep. And I've never seen a root.bolt ever get very large. Does it in the use cases you support?

Less is better is the idea there.
We have seen boltdb containing thousands of items.. Its just about storing only the bare needed things in bolt at least whenever we get a clean up opportunity in form of rollback. 


> To me the only reason there is faster disk cleanup is that you reduce the root.bolt to a single snapshot (going below the users configuration).

We don't always go below the user configured numSnapshotsToKeep or to a single snapshot. We keep deleting epochs until we find the asked epoch. Going below user configuration happens if the asked epoch was the last remaining one in the bolt or the asked epoch was one among the retained numSnapshotsToKeep epochs.


> You had also mentioned some recursive rollback case, but it seemed to me you always process the RollbackPoints linearly looking for what you want, so I didn't see possibility of loop or recursion.

Repetitive rollback requests from KV was referred as the recursive possibility there. We should refrain from adding data into the system in such weird system behaviour.

Destructive approach fits the easy mental model of each epoch in bolt stores a linear progressive state of the index and the insert approach break it.   ;)

If the main concern is only about NumSnapshotsToKeep, 
then it isn't a significant worry to me as long as there is no math/logical relation exposed on the precise rollback-ability gained out of a chosen number of NumSnapshotsToKeep. 
Now there is an argument of what if once we fix it?   

But then we go below the NumSnapshotsToKeep only in two corner cases and that too under a successful rollback. So the rollback purpose is served. And then index open would start of with a clean disk of necessary files.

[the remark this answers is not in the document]
sreekanth-cb: One other thought popping to mind is - in the insert epoch approach ,
aren't we spoiling the quality/effectiveness of the retained  epochs for NumSnapshotsToKeep ?

As after a successful rollback, index starts from the rollbacked point and the epochs behind the newly introduced one aren't really helping/contributing to any of the next imminent rollback request to a point since the last successful rollback point?  

Epochs behind the recent rollback introduced one are going to be useful only for those cases - when the next upcoming rollback request is asking to rollback to even an older rollback point than the last rollback point, which is less often than rollback to a recent point since the last successful rollback point.. ? (debatable)

index/scorch/snapshot_rollback.go
```diff
@@ -105,69 +116,97 @@ func (s *Scorch) RollbackPoints() ([]*RollbackPoint, error) {
...
-			filename := zapFileName(segSnap.id)
-			s.ineligibleForRemoval[filename] = true
+		sc := snapshots.Cursor()
+		for sk, _ := sc.Last(); sk != nil; sk, _ = sc.Prev() {
+			_, snapshotEpoch, err := segment.DecodeUvarintAscending(sk)
+			if err != nil {
+				continue
+			}
+			if snapshotEpoch == to.epoch {
+				found = true
```
steveyen: two thoughts here...
- if it's found, wonder if you can break out of this loop early?
- that is, instead of grabbing all the persistedEpochs
  wondering if you instead need the "epochs we're thinking that we're gonna delete"? 
- wondering if the == epoch check needs to be a >= or > kind of comparison instead?

steveyen: also, not sure if if bolt snapshot cursor needs to be closed?

sreekanth-cb: cursors doesn't seem to have a close method.

sreekanth-cb: > wondering if the == epoch check needs to be a >= or > kind of comparison instead?

I think, once we have the early exit logic that should be taken care.

steveyen: one thing about numbers is I tend to favor comparisons rather than equality...

like with equality, it ought to be ok when the epochs #'s are cleanly increasing by 1... like, 2,3,4,5,6,7, etc

but, if they're not only just +1 increasing (not sure why that might be the case) but if for whatever reason the bolt stuff has some missing #'s... and the epoch we're trying to get to is that #570 then the range comparison feels better to me?

mschoch: @steveyen on this one I disagree.  If you use a comparison operator here you are presuming something which may not be the case.  To me, the "fuzziness" part of, I want something in this range, is done externally when the caller evaluates which rollback points are available, and what metadata is associated with each point.  In this method call, they are choosing a specific rollback point of the known rollback points.  I think it should either rollback to exactly that point, or return an error if it cannot.

steveyen: oh, right!  headslap!  yah -- the evaluation is done by the caller of RollbackPoints.

PR #1357, opened 2020-03-20
chore: fixes bbolt url

mohsensamiei:


mschoch:
Was there a reason you closed this?  It seems like we've been importing the wrong URL all along, and the tests were passing.

mohsensamiei:
Yes, I did not have enough confidence and I felt maybe I was wrong
Thanks for your feedback 😊

mschoch:
Let's leave it open, I'll take a closer look on Monday.  Thanks.

PR #1359, opened 2020-03-23
Fix for snapshot leak by merger routine on scorch index Close

sreekanth-cb:
Improving ownership handling during merge introductions
    
    As a part of notifying the merger during the merge
    introductions, the introducer bumps the newer
    snapshot's reference count. But in situations
    where the index is getting closed, there is a
    chance that the merge requested merger(file/in memory)
    routines would have already exited.
    This would cause the newly introduced, ref count
    bumped snapshot to leak as there is none to decrement
    the bumped up reference count.
    
    The fix is about tighening the merge introduction
    notifications handling at the merger side.
    As the introducer non-preemptively handles the
    intro notify channel, its safe for the merger
    to blockingly awaits for each of those merge introduction
    notifications and decrements the reference counts
    where applicable.
    
    Unlike earlier, merger is processing each of the merge
    tasks and awaits its introductions to complete before
    proceeding to the next merge task. We don't expect any serious
    performance wrinkles here as the introducer ought to be
    faster with the merge introductions compared to the merge
    tasks.
    
    Added a few introduction related stats to track any
    performance considerations raising out of this.

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -312,6 +312,8 @@ func (s *Scorch) introducePersist(persist *persistIntroduction) {
 	close(persist.applied)
 }
 
+// The introducer should definitely handle the segmentMerge.notify
+// channel before exiting the introduceMerge.
 func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 	atomic.AddUint64(&s.stats.TotIntroduceMergeBeg, 1)
 	defer atomic.AddUint64(&s.stats.TotIntroduceMergeEnd, 1)
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -157,22 +157,7 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 	atomic.AddUint64(&s.stats.TotFileMergePlanTasks, uint64(len(resultMergePlan.Tasks)))
 
 	// process tasks in serial for now
-	var notifications []chan *IndexSnapshot
 	var filenames []string
-	// clean up any pending notifications from introducer on exit
-	// from an index closure.
-	defer func() {
-		for _, notification := range notifications {
-			select {
-			case newSnapshot := <-notification:
-				atomic.AddUint64(&s.stats.TotFileMergeIntroductionsDone, 1)
-				if newSnapshot != nil {
-					_ = newSnapshot.DecRef()
-				}
-			default:
-			}
-		}
-	}()
 
 	for _, task := range resultMergePlan.Tasks {
 		if len(task.Segments) == 0 {
@@ -232,7 +217,6 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
 				s.unmarkIneligibleForRemoval(filename)
 				atomic.AddUint64(&s.stats.TotFileMergePlanTasksErr, 1)
 				if err == segment.ErrClosed {
-					// handle any pending index snapshot introduction notifications on exit
 					return err
 				}
 				return fmt.Errorf("merging failed: %v", err)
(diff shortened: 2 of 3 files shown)


index/scorch/introducer.go
```diff
@@ -438,8 +436,17 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 		_ = rootPrev.DecRef()
 	}
 
-	// notify requester that we incorporated this
-	nextMerge.notify <- newSnapshot
+	// check for index closure before bumping the reference count,
+	// else this snapshot may leak if the requester has already
+	// exited the work loop.
+	select {
+	case <-s.closeCh:
```
steveyen: hmmm -- a perhaps more robust way might be instead like...
```
newSnapshot.AdRef()
... do stuff
select {
  case <-s.closeCh:
     newSnapshot.DecRef()
  case nextMerge.notify <- newSnapshot:
    // no-op
}
close(nextMerge.notify)
```

And, anytime we respond to the closing of the closeCh, look for these patterns where we still own the ref-count?

steveyen: "more robust" might be ill-worded... perhaps "a little more idiomatic-feeling to me" on how to do this might be the better description

sreekanth-cb: yea..I totally agree that this looks more idiomatic. 
Have one doubt here,
iirc => select randomly picks one out of a 'n' valid/successful cases. 
Here : 
"case nextMerge.notify <- newSnapshot: " will always be true as notify is a buffered channel.
So we still have a chance of leaking the bumped index snapshot if the select ends up randomly picking up the second notify ch push case?

mschoch:
Am I wrong that the race is a deeper issue, and this doesn't really fix it, just might hide the most egregious cases?

index/scorch/introducer.go
sreekanth-cb: done!

sreekanth-cb: done!

sreekanth-cb:
> latest version looks good to me -- especially, if you've double-checked on the ch receiver side that the ch is either non-buffered or gets drained by the receiver (even in the midst of or after a close)

The summers in kerala woke me up in the early early morning, I just feel its not quite right yet.
Before going back to bed, wanted to comment that here and then saw that steve already was asking about it.

I prefer this than the current version as the notify channel is buffered one.
```
select {
	case <-s.closeCh:
		// handle the ref count
		_ = newSnapshot.DecRef()
	default:
                 nextMerge.notify <- newSnapshot:
	}
```
@steveyen , Isn't this better as it totally avoids the select randomly picking non Close case when 
case <-s.closeCh: is already true?

abhinavdangeti:
@sreekanth-cb the length of nextMerge.notify ch is 1.

Meaning .. if you move `nextMerge.notify <- newSnapshot` to the default case, that race window would come back again where the select enters the default case and nextMerge.notify isn't listening anymore due to closeCh closing.

sreekanth-cb:
Thanks @abhinavdangeti for the quick response!

default case would enter only if all the other case statements are non true - is my understanding. So the moment we land there - index is not closed yet.. So the requester ought to be live/listening there as Channel write is an atomic operation.
Once we write its to channel, the requester will block until it reads it even if an index close happens on the very next moment...

mschoch:
I was about to try and debate the specifics of your logic, which I still think are subtltley flawed.

But, you keep preferring this solution and your reasoning is, "as it totally avoids the select randomly picking non Close case when case <-s.closeCh: is already true".

So, I'd like to first understand why that is so bad in this case?  All it changes is who does the decref right?

steveyen:
hi @sreekanth-cb -- yah, i'm with @abhinavdangeti on this one...
```
10 select {
20	case <-s.closeCh:
30		// handle the ref count
40		_ = newSnapshot.DecRef()
50	default:
60                 nextMerge.notify <- newSnapshot
70	}
```
I added line numbers to clarify...

Right at line 55, I can imagine the CPU being taken away and reassigned to another goroutine?

Perhaps that other goroutine decides to shut everything down and starts going through the close steps.

Then 5 minutes later, the slightly evil OS scheduler finally reassigns the CPU back to our goroutine at line 60.

So, I guess the mental trick that I think about... is in the whitespace in-between the lines-of-code is where the CPU potentially gets ripped away and evil things start happening elsewhere?

The more idiomatic approach, in contrast, doesn't have a so-called "line 55" problem?

sreekanth-cb:
@mschoch , 

Receiver will block if there was a successful notify channel write before the index closure.

The channel is buffered and hence the channel write will happen if it gets picked randomly.

But if the index close happens before that - receiver would exit and no receiver side handling will happen for the snapshot passed to the receiver over the randomly picked channel case.

ref - https://github.com/blevesearch/bleve/blob/master/index/scorch/merge.go#L172

mschoch:
I'm with @steveyen and @abhinavdangeti too, this issue is that while select is atomic and the channel write is atomic, they are two separate operations.  The "line 55 problem" as steve put it.

I still think the fix is that the on the receiver side, if it gets a close, while waiting to be notified, it simply cannot exit right away, it has to wait for the other process to either send the message, or close the channel, depending on which path it closed.  If it got the message, then the decref is then it's repsonsibility.

steveyen:
Just thought of one other possibility -- is to perhaps keep this file as-is, so that it just does a naked send as it did before...

    nextMerge.notify <- newSnapshot

...and perhaps instead the fix only needs to be on the side of whoever (let's call it goroutine X) is reading or receiving from the nextMerge.notify channel?

e.g., that is, perhaps the fix is that goroutine X must always consume the nextMerge.notify channel first.  Then after it's drained the notify channel, then X can look at the closeCh, if that's still needed?

If that's too confusing (it might be to me!), then please place this idea far far away down into the circular file.

steveyen:
Also, while I'm on it... wondering if that nextMerge.notify channel actually needs to be buffered?

If not, perhaps it becomes easier to reason about.

sreekanth-cb:
> Also, while I'm on it... wondering if that nextMerge.notify channel actually needs to be buffered?
> 
> If not, perhaps it becomes easier to reason about.

I think we did a buffered channel to have a faster/unblocked merge iteration over the merge tasks..
Else before processing the next merge task, we have to ACK the notification from introducer..

sreekanth-cb:
done @steveyen , added comments!

index/scorch/merge.go
```diff
@@ -262,35 +246,33 @@ func (s *Scorch) planMergeAtSnapshot(ourSnapshot *IndexSnapshot,
...
-			if newSnapshot != nil {
-				_ = newSnapshot.DecRef()
-			}
+		introStartTime := time.Now()
+		// it is safe to blockingly wait for the merge introduction
+		// here as the introducer is bound to handle the notify channel.
+		newSnapshot := <-sm.notify
+		introTime := uint64(time.Since(introStartTime))
+		atomic.AddUint64(&s.stats.TotFileMergeZapIntroductionTime, introTime)
+		if atomic.LoadUint64(&s.stats.MaxFileMergeZapIntroductionTime) < introTime {
```
abhinavdangeti: We should perhaps consider adding histograms like we did with moss into scorch.
This'd be a good candidate - wouldn't need stats for TotTime and a MaxTime then.

sreekanth-cb: I am not familiar with the histogram of moss, the stats idea is to inspect a potential future worry on any perf impacts due to this tight stepping together between merger-introducer duo here..

sreekanth-cb:
> I'm a little concerned if blocking merger tasks on the notification from the introducer could possible cause some other issue(s), but otherwise this looks OK to me to get some testing done.

My only concern here is on the perf impacts out of this as these intro wait times weren't getting interleaved between merge tasks earlier.

The indexing could potentially get slower by ~TotFileMergeZapIntroductionTime roughly or more precisely by TotFileMergeZapIntroductionTime- Cumulative Sum(last task's intro time).

But thats the price we have to pay for getting this simplified is my thinking here.

PR #1377, opened 2020-04-20
add CI workflow for code coverage

mschoch:


mschoch:
So, this seems to work, and reports the same 66% that was getting reported before (using the more complicated set of scripts).  So, I suggest we adopt this version.

PR #1383, opened 2020-04-30
remove old geo functions

mschoch:
the older slower implementation was kept around in the test
functions to compare with the newer version.  this makes
sense as an initial validation, but presumably once satisfied
this new version is correct, we should discard the old.

### DIFF
--- a/search/searcher/search_geoboundingbox_test.go
+++ b/search/searcher/search_geoboundingbox_test.go
@@ -19,11 +19,9 @@ import (
 	"testing"
 
 	"github.com/blevesearch/bleve/document"
-	"github.com/blevesearch/bleve/geo"
 	"github.com/blevesearch/bleve/index"
 	"github.com/blevesearch/bleve/index/store/gtreap"
 	"github.com/blevesearch/bleve/index/upsidedown"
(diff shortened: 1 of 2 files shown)


mschoch:
Hey @sreekanth-cb I've been going through all our tests looking for major slowdowns and this one showed up.  It seems like we have some tests which kept the old slow implementation for comparison purposes.  To me that makes sense for an initial validation before making the switch, but something we should get rid of once we're satisfied the new version is correct.  Can you review to make sure we're not losing anything by getting rid of this old code?

mschoch:
Another idea, if you think the comparisons are important, because we think they are correct.  Perhaps we can use the old code one more time, to generate a table, and validate against that table going forward.  As opposed to relying on the old slow code every time we run the tests.

sreekanth-cb:
> Another idea, if you think the comparisons are important, because we think they are correct. Perhaps we can use the old code one more time, to generate a table, and validate against that table going forward. As opposed to relying on the old slow code every time we run the tests.

I think, I am inclined to this option.

mschoch:
Added no not merge label, as Sreekanth has requested we keep the test in table form (instead of code).  So there is more work to be done.

PR #1386, opened 2020-04-30
replace testidx with proper ioutil.TempDir

mschoch:


mschoch:
I have some further cleanup planned for some follow-on PRs.

test/versus_score_test.go
```diff
@@ -48,9 +49,19 @@ func TestDisjunctionSearchScoreIndexWithCompositeFields(t *testing.T) {
 
 func disjunctionQueryiOnIndexWithCompositeFields(indexName string,
 	t *testing.T) []*search.DocumentMatch {
+	tmpIndexPath, err := ioutil.TempDir("", "bleve-testidx")
```
abhinavdangeti, line 52: Why not use the same createTmpIndexPath(..)-defer cleanupTmpIndexPath(..) code path here?

index_test.go
```diff
@@ -42,15 +43,30 @@ import (
 	"github.com/blevesearch/bleve/index/upsidedown"
 )
 
+type Fatalfable interface {
```
abhinavdangeti, line 46: I read Fatalf-able as Fatal-Fable :)

test/versus_score_test.go
mschoch: It's in a different package, and this is the only usage in this package, so I figured more clear to do it directly.  If you think copy/pasting those funcs is better, I don't feel too strongly about it.

index_test.go
mschoch: Yeah at first I didn't like it, but it grew on me.

test/versus_score_test.go
abhinavdangeti: Fair enough :)

abhinavdangeti:
I'm good with this PR!

Issue #1390, opened 2020-05-04
Document mapping's analyzer isn't inherited if field analyzer isn't specified


abhinavdangeti:
Take a look at the following example, where I have 

- IndexMapping
  - DefaultAnalyzer: standard
  - DefaultMapping
    - DefaultAnalyzer: keyword
    - Field: city, Analyzer: ""

I expect the city's analyzer to be "keyword" and not "standard".

The following test case breaks ..
```
func TestAnalyzerInheritance(t *testing.T) {
	dMapping := mapping.NewDocumentStaticMapping()
	dMapping.DefaultAnalyzer = keyword.Name

	fMapping := mapping.NewTextFieldMapping()
	dMapping.AddFieldMappingsAt("city", fMapping)

	idxMapping := NewIndexMapping()
	idxMapping.DefaultMapping = dMapping

	tmpIndexPath := createTmpIndexPath(t)
	idx, err := New(tmpIndexPath, idxMapping)
	if err != nil {
		t.Fatal(err)
	}

	defer func() {
		err := idx.Close()
		if err != nil {
			t.Fatal(err)
		}
	}()

	doc := map[string]interface{}{
		"city": "San Francisco",
	}

	if err = idx.Index("doc", doc); err != nil {
		t.Fatal(err)
	}

	q := NewTermQuery("San Francisco")
	q.SetField("city")

	res, err := idx.Search(NewSearchRequest(q))
	if err != nil {
		t.Fatal(err)
	}

	if len(res.Hits) != 1 {
		t.Fatalf("unexpected number of hits: %v", len(res.Hits))
	}
}
```

PR #1391, opened 2020-05-04
MB-38957: Document mapping's analyzer to be inherited correctly

abhinavdangeti:
Check if the top level document mapping's (default mapping or a
type mapping) default analyzer is available while determining the
analyzer to use for a child field in the event that neither the child
field has an analyzer set nor any of the intermediate child mappings.

Fixes: https://github.com/blevesearch/bleve/issues/1390

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/blevesearch/snowballstem v0.9.0
 	github.com/blevesearch/zap/v11 v11.0.7
 	github.com/blevesearch/zap/v12 v12.0.7
+	github.com/couchbase/ghistogram v0.1.0 // indirect
 	github.com/couchbase/moss v0.1.0
 	github.com/couchbase/vellum v1.0.1
 	github.com/golang/protobuf v1.3.2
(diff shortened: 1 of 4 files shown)


abhinavdangeti:
> One thought is to use the sub-test functionality for table-driven tests. You may have seen some commits from me recently converting to that style. It's not a huge deal, but makes it easier to run individual tests from the table, and makes the output easier to read.

Not sure I follow what you mean by sub-test functionality for table-driven tests.

>Another thing, perhaps include a comment somewhere explaining the assertion. If I understand it correctly, we can't easily just check which analyzer was used. Instead, our check is based on keyword analyzer working differently enough that it affects the search results. This may not be obvious to someone else reading the code later.
Oh, another idea, since the assertion is kind of non-obvoius. You could have an actual "unit test" in the mapping package that just tests the defaultAnalyzerName function. That might be cleaner?

Yea this makes sense, let me add tests just testing the affected API within mapping/..

> Finally, I see the PR also alters the go.mod in an unexpected way, can you explain that?

The entry in go.mod is unnecessary and also unrelated to this change. It snuck in when I ran go test ./... I suppose.

abhinavdangeti:
Thanks @mschoch for pointing out the sub-test functionality way. This should now be ready for your review.

PR #1392, opened 2020-05-06
MB-37962 - graceful handling of the json encode errors

sreekanth-cb:
Handle the json encode errors than panicing
in the response encoder.

### DIFF
--- a/http/util.go
+++ b/http/util.go
@@ -16,6 +16,7 @@ package http
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
@@ -29,14 +30,17 @@ func showError(w http.ResponseWriter, r *http.Request,
 }
 
 func mustEncode(w io.Writer, i interface{}) {
-	if headered, ok := w.(http.ResponseWriter); ok {
+	var headered http.ResponseWriter
+	var ok bool
+	if headered, ok = w.(http.ResponseWriter); ok {
 		headered.Header().Set("Cache-Control", "no-cache")
 		headered.Header().Set("Content-type", "application/json")
 	}
 
 	e := json.NewEncoder(w)
 	if err := e.Encode(i); err != nil {
-		panic(err)
+		showError(headered, nil, fmt.Sprintf("json encode, err: %v",
+			err), http.StatusInternalServerError)
 	}
 }
 


mschoch:
At a minimum we need to change the method name as well.  In Go, the prefix "must" is a convention used for methods that either succeed or panic.

This code is so old, I don't remember why we ever did it this way instead of returning an error and letting the caller decide what to do.

sreekanth-cb:
Is there a name you prefer @mschoch , just encode or encodeResponse ?

mschoch:
Looking closer, I'm now -1 on this change.  By the time we've encountered the error, we may have already started writing the response.  If we then call `showError` we may attempt to set a header again which is invalid.  Exactly how it fails at that point is unclear, since nothing on this path returns error.

Do you have a bug where this actually happens?

sreekanth-cb:
Yes, the MB is put in the PR title - https://issues.couchbase.com/browse/MB-37962

thats right, then one option would be to switch to json.Marshal beforehand for detecting any marshal errors and set the headers accordingly () ?  

Not sure whether the lack of the streamed write capabilities is such a big concern though here?

mschoch:
The error is coming from the I/O (timeout), so I don't think marshaling first would have helped.

sreekanth-cb:
yikes... thats true..

mschoch:
This is actually a good example of why the `bleve/http` package should not exist.

The HTTP handler API is such that we cannot return the error for the caller to decide what to do.  Our choices are to panic (undesirable because the entire process exits), log it (something bleve generally doesn't do), or do nothing.  Writing the error the response in not possible for cases like this, where the error is our inability to write the response.

sreekanth-cb:
Shall we use HTTP Trailer headers here? but that will be unconventional too..

I just wanted a better handling of errors here than panicking.. 
The issue in CB is a bigger one and that mostly has to do with the way the system/node is overloaded I guess.. The fix for it would be a broader one with more resource pruning/judicious usage across the stack.

mschoch:
The issue is that we cannot write successfully to the socket, so it doesn't matter whether we call it header/body/trailer-header.  As I said before, the choices I see are:
- panic (what we do today)
- log (not possible today without further changes, only top-level bleve pkg has access to a configured logger)
- do nothing
- add yet another callback to let the user choose between the first 3 options (dislike this because I would prefer we remove this package instead)

I don't see any other choices, do you?

sreekanth-cb:
Strangely, this package has it's own logger. 
https://github.com/blevesearch/bleve/blob/9b6e89951d4e9207356a833a7d99577cfe249208/http/util.go#L45

Now, I am thinking between "do nothing Vs Panic", and finding it difficult to decide.. 
Perhaps eventually has to choose the "do nothing" option..

mschoch:
@sreekanth-cb my honest opinion is that these handlers should move into FTS, then you can do proper logging, and not panic.  Do you see any problem with that?

mschoch:
Oh, another idea, since the assertion is kind of weak.  You could have an actual "unit test" in the mapping package that just tests the `defaultAnalyzerName` function.  That might be cleaner?

mschoch:
Sorry @sreekanth-cb I posted some comments I intended for a PR by @abhinavdangeti

sreekanth-cb:
Closing this PR as this needs to be handled at the app level.

PR #1398, opened 2020-05-20
use inline comparison operators instead of strings.Compare

tylerkovacs:
In the comments from strings.Compare:

```
// Compare is included only for symmetry with package bytes.
// It is usually clearer and always faster to use the built-in
// string comparison operators ==, <, >, and so on.
```

And another comment within the implementation:
```
// NOTE(rsc): This function does NOT call the runtime cmpstring function,
// because we do not want to provide any performance justification for
// using strings.Compare. Basically no one should use strings.Compare.
```

### DIFF
--- a/search/sort.go
+++ b/search/sort.go
@@ -233,7 +233,11 @@ func (so SortOrder) Compare(cachedScoring, cachedDesc []bool, i, j *DocumentMatc
 		} else {
 			iVal := i.Sort[x]
 			jVal := j.Sort[x]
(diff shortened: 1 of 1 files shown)


tylerkovacs:
I ran a few tests and average sort time was about 8% faster for sorting 1,000,000 `DocumentMatch`es. Results were quite variable though so I'm not sure the results are predictable. Stil, the comments from the source were enough to convince me that this change seemed worthwhile.

PR #1399, opened 2020-05-20
use parallel Quicksort in MultiSearch

tylerkovacs:
I noticed that MultiSearch fully utilizes all cores on
my machine for only the first 40% of total search time.
Then only a single core is fully utilized for the remainder.

I tracked it down to the call to sort.Sort. I frequently run
searches that return over 1 million hits. Replacing the call
to sort.Sort with a parallel Quicksort dropped sort
latency on my queries from 18 seconds to 5 seconds.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,20 +5,32 @@ go 1.13
 require (
 	github.com/RoaringBitmap/roaring v0.4.21
 	github.com/blevesearch/blevex v0.0.0-20190916190636-152f0fe5c040
+	github.com/blevesearch/cld2 v0.0.0-20200327141045-8b5f551d37f5 // indirect
 	github.com/blevesearch/go-porterstemmer v1.0.3
 	github.com/blevesearch/segment v0.9.0
 	github.com/blevesearch/snowballstem v0.9.0
 	github.com/blevesearch/zap/v11 v11.0.7
 	github.com/blevesearch/zap/v12 v12.0.7
-	github.com/couchbase/ghistogram v0.1.0 // indirect
 	github.com/couchbase/moss v0.1.0
 	github.com/couchbase/vellum v1.0.1
+	github.com/cznic/b v0.0.0-20181122101859-a26611c4d92d // indirect
+	github.com/cznic/mathutil v0.0.0-20181122101859-297441e03548 // indirect
+	github.com/cznic/strutil v0.0.0-20181122101858-275e90344537 // indirect
+	github.com/facebookgo/ensure v0.0.0-20200202191622-63f1cf65ac4c // indirect
+	github.com/facebookgo/stack v0.0.0-20160209184415-751773369052 // indirect
+	github.com/facebookgo/subset v0.0.0-20200203212716-c811ad88dec4 // indirect
 	github.com/golang/protobuf v1.3.2
+	github.com/ikawaha/kagome.ipadic v1.1.2 // indirect
+	github.com/jmhodges/levigo v1.0.0 // indirect
 	github.com/kljensen/snowball v0.6.0
 	github.com/rcrowley/go-metrics v0.0.0-20190826022208-cac0b30c2563
+	github.com/remyoudompheng/bigfft v0.0.0-20200410134404-eec4a21b6bb0 // indirect
 	github.com/spf13/cobra v0.0.5
 	github.com/steveyen/gtreap v0.1.0
 	github.com/syndtr/goleveldb v1.0.0
+	github.com/tebeka/snowball v0.4.2 // indirect
+	github.com/tecbot/gorocksdb v0.0.0-20191217155057-f0fad39f321c // indirect
+	github.com/twotwotwo/sorts v0.0.0-20160814051341-bf5c1f2b8553
 	github.com/willf/bitset v1.1.10
 	go.etcd.io/bbolt v1.3.4
 	golang.org/x/text v0.3.0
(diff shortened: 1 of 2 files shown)


tylerkovacs:
This is a game-changer for me but I'm not sure about your policy for introducing new dependencies.

The changes to `go.sum` seem odd to me since I only added a single dependency. I'm quite new to Go modules so some guidance would be appreciated!

abhinavdangeti:
For the new library dependency, it would need more looking into.

We didn't previously commit go.sum into the project.
@mschoch if we plan to keep it that way - I suppose it's worth adding to .gitignore.

tylerkovacs:
Thanks @abhinavdangeti. I've just updated the commit to remove the change to `go.sum`.

tylerkovacs:
As an aside, there are many parallel sort implementationst to choose from. I tried the one in this branch based on the benchmarks for sorting a `[]string` here:
https://github.com/jfcg/sorty

tylerkovacs:
> A new entry will need to go into go.mod however.

Is that typically done by running `go mod tidy`? I ran that and it added a line for the new dependency but also updated a number of other lines. Here's the diff:

```
diff --git a/go.mod b/go.mod
index d38cf8f9..3eb9a8f6 100644
--- a/go.mod
+++ b/go.mod
@@ -5,20 +5,32 @@ go 1.13
 require (
        github.com/RoaringBitmap/roaring v0.4.21
        github.com/blevesearch/blevex v0.0.0-20190916190636-152f0fe5c040
+       github.com/blevesearch/cld2 v0.0.0-20200327141045-8b5f551d37f5 // indirect
        github.com/blevesearch/go-porterstemmer v1.0.3
        github.com/blevesearch/segment v0.9.0
        github.com/blevesearch/snowballstem v0.9.0
        github.com/blevesearch/zap/v11 v11.0.7
        github.com/blevesearch/zap/v12 v12.0.7
-       github.com/couchbase/ghistogram v0.1.0 // indirect
        github.com/couchbase/moss v0.1.0
        github.com/couchbase/vellum v1.0.1
+       github.com/cznic/b v0.0.0-20181122101859-a26611c4d92d // indirect
+       github.com/cznic/mathutil v0.0.0-20181122101859-297441e03548 // indirect
+       github.com/cznic/strutil v0.0.0-20181122101858-275e90344537 // indirect
+       github.com/facebookgo/ensure v0.0.0-20200202191622-63f1cf65ac4c // indirect
+       github.com/facebookgo/stack v0.0.0-20160209184415-751773369052 // indirect
+       github.com/facebookgo/subset v0.0.0-20200203212716-c811ad88dec4 // indirect
        github.com/golang/protobuf v1.3.2
+       github.com/ikawaha/kagome.ipadic v1.1.2 // indirect
+       github.com/jmhodges/levigo v1.0.0 // indirect
        github.com/kljensen/snowball v0.6.0
        github.com/rcrowley/go-metrics v0.0.0-20190826022208-cac0b30c2563
+       github.com/remyoudompheng/bigfft v0.0.0-20200410134404-eec4a21b6bb0 // indirect
        github.com/spf13/cobra v0.0.5
        github.com/steveyen/gtreap v0.1.0
        github.com/syndtr/goleveldb v1.0.0
+       github.com/tebeka/snowball v0.4.2 // indirect
+       github.com/tecbot/gorocksdb v0.0.0-20191217155057-f0fad39f321c // indirect
+       github.com/twotwotwo/sorts v0.0.0-20160814051341-bf5c1f2b8553
        github.com/willf/bitset v1.1.10
        go.etcd.io/bbolt v1.3.4
        golang.org/x/text v0.3.0
```

abhinavdangeti:
Yea, the rest (which are marked as indirect) are indirect dependencies, looks like they're projects twotwotwo/sorts imports within it.

mschoch:
Let's continue to leave go.sum out.  It is not permanent, it is just due to the fact that I have to hand-edit go.mod when we do releases (because we have a circular dependecy between zap and bleve).  Fixing that is also in the works for 2.x.

Regarding the other entries added by `go mod tidy` some of them are indirect from the new sort lib, but some are because of the config/config_XYZ.go dependencies.  They designed `go mod tidy` to add everything you need, without regard for build tags.  But I've been leaving those out because the whole reason we put them behind build tags in the first place was we didn't want most users to depend on them.  Again, there is a longer-term fix planned (deprecate the config folder).  For now I guess we just have to live with it...

mschoch:
As for the dependency, the team will discuss it, but I see another alternative to consider.  Both stdlib sort.Sort and twotwotwo/sorts.QuickSort conform to the same interface.  So it would be straightforward to add a `SortFunc func(data sort.Interface)` field to bleve.SearchRequest.  It can just default to `sort.Sort`, but that would let you change it for your app.  Thoughts?

tylerkovacs:
> So it would be straightforward to add a `SortFunc func(data sort.Interface)` field to bleve.SearchRequest. It can just default to `sort.Sort`, but that would let you change it for your app. Thoughts?

On one hand, it seems likely that people using `MultiSearch` are doing so because they want to use all available cores to minimize search latency. In that case, defaulting to a parallel sort implementation seems less surprising, and it's one less thing for people to configure. That's just my assumption and you'd know better than me.

On the other hand, I like that this approach does not introduce a new dependency. And I see no reason why it would not work for my use case. If you go that route, the only thing that I would suggest is to document it somewhere. The main `README.md` seems like a reasonable place to start collecting these thoughts.

mschoch:
I think it's a bit less black and white than that.  At least one other application I know of using aliases for search (and thus MultiSearch) have the opposite problem.  The earlier phase results in too many concurrent searches, which end up  fighting over the buffer cache, causing all the searches to run slower.

I guess I'm increasingly thinking it should be configurable.  But that doesn't mean we shouldn't have good (or better) defaults too.

tylerkovacs:
Sounds good to me. I'll close this PR and open another for the interface.

PR #1400, opened 2020-05-20
Add SortImpl to SearchRequest

tylerkovacs:
This allows SearchRequests to specify the sort implementation (defaults to sort.Sort) to use
when sorting hits.

### DIFF
--- a/search.go
+++ b/search.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"sort"
 	"time"
 
 	"github.com/blevesearch/bleve/analysis"
@@ -264,6 +265,7 @@ func (h *HighlightRequest) AddField(field string) {
 // Score controls the kind of scoring performed
 // SearchAfter supports deep paging by providing a minimum sort key
 // SearchBefore supports deep paging by providing a maximum sort key
+// SortImpl specifies the sort implementation to use for sorting results.
 //
 // A special field named "*" can be used to return all fields.
 type SearchRequest struct {
@@ -279,6 +281,8 @@ type SearchRequest struct {
 	Score            string            `json:"score,omitempty"`
 	SearchAfter      []string          `json:"search_after"`
 	SearchBefore     []string          `json:"search_before"`
+
+	SortImpl func(sort.Interface) `json:"-"`
 }
 
 func (r *SearchRequest) Validate() error {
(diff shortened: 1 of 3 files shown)


tylerkovacs:
There's one thing that's puzzling me. In `index_impl.go`, I can only figure out where to use the new option in the case that the search is reversed.

If the search is not reversed, I think sorting happens in `NewTopNCollectorAfter` or `NewTopNCollector` but it's not clear to me where the sort is actually happening. I'll keep looking.

mschoch:
@tylerkovacs that's correct.  For regular search of a single index, we don't explicitly sort the results.  Instead, as hit's come in to the collector they're put into a heap (usually, though the impl can vary) and the heap uses the sort value of the incoming item to place it in the right spot.  That way, if it's not in the top N at that time it can be discarded right away.  So the results are kept in order as they're being built.

The reversing is a special case when performing a searchBefore type search, here basically just intentionally perform the search backwards, and reverse it at the end.  I don't actually recall if we really need to resort, or if they could just be reversed here.  Seems like it could be worst case for some sort if it's exactly reversed...

As for MultiSearch, because we got the top N from each index, we have to merge these results and sort again to get the final top N.

So I think this is all the appropriate places/

search.go
```diff
@@ -606,3 +610,22 @@ func MemoryNeededForSearchResult(req *SearchRequest) uint64 {
 
 	return uint64(estimate)
 }
+
+// SetSortFunc sets the sort implementation to use when sorting hits.
```
mschoch, line 614: Comment function name doesn't match actual function name.  Though, personally I do like SortFunc better than SortImpl.  What do others think?

search.go
```diff
@@ -606,3 +610,22 @@ func MemoryNeededForSearchResult(req *SearchRequest) uint64 {
...
+// SearchRequests can specify a custom sort implementation to meet
+// their needs. For instance, by specifying a parallel sort
+// that uses all available cores.
+func (r *SearchRequest) SetSortImpl(s func(sort.Interface)) {
+	r.SortImpl = s
+}
+
+// GetSortFunc returns the sort implementation to use when sorting hits.
+// Defaults to sort.Sort.
+func (r *SearchRequest) GetSortImpl() func(data sort.Interface) {
```
mschoch, line 625: Typically in Go, the convention is not to have methods for the form GetField(), instead they use just Field() and SetField().

tylerkovacs:
@mschoch Any more thoughts on this PR? Your comments have been addressed.

mschoch:
@sreekanth-cb @abhinavdangeti I have approved this, in order to keep things moving, please either review, ask for more time, or ignore.  If I don't hear anything we'll merge in 24 hours.

PR #1401, opened 2020-05-22
improve scorch config parsing of forced zap version

mschoch:
previous version did not correctly handle integer
values stored in a float64 due to JSON parsing

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -144,8 +144,8 @@ func configForceSegmentTypeVersion(config map[string]interface{}) (string, uint3
 		return "", 0, nil
 	}
 
-	forcedSegmentType, ok2 := config["forceSegmentType"].(string)
-	if !ok2 {
+	forcedSegmentType, ok := config["forceSegmentType"].(string)
+	if !ok {
 		return "", 0, fmt.Errorf(
 			"forceSegmentVersion set to %d, must also specify forceSegmentType", forcedSegmentVersion)
 	}
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -139,17 +139,18 @@ func NewScorch(storeName string,
 // configForceSegmentTypeVersion checks if the caller has requested a
 // specific segment type/version
 func configForceSegmentTypeVersion(config map[string]interface{}) (string, uint32, error) {
-	forcedSegmentVersion, ok := config["forceSegmentVersion"].(int)
-	if ok {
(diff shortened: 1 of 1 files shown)


index/scorch/scorch.go
```diff
@@ -139,17 +139,18 @@ func NewScorch(storeName string,
...
-			return "", 0, fmt.Errorf(
-				"forceSegmentVersion set to %d, must also specify forceSegmentType", forcedSegmentVersion)
-		}
+	forcedSegmentVersion, err := parseToInteger(config["forceSegmentVersion"])
+	if err != nil {
+		return "", 0, nil
+	}
 
-		return forcedSegmentType, uint32(forcedSegmentVersion), nil
+	forcedSegmentType, ok2 := config["forceSegmentType"].(string)
```
abhinavdangeti, line 147: Now that there's no "ok" anymore, "ok2" should be the new "ok" :-D

index/scorch/scorch.go
```diff
@@ -139,17 +139,18 @@ func NewScorch(storeName string,
...
-	forcedSegmentVersion, ok := config["forceSegmentVersion"].(int)
-	if ok {
-		forcedSegmentType, ok2 := config["forceSegmentType"].(string)
-		if !ok2 {
-			return "", 0, fmt.Errorf(
-				"forceSegmentVersion set to %d, must also specify forceSegmentType", forcedSegmentVersion)
-		}
+	forcedSegmentVersion, err := parseToInteger(config["forceSegmentVersion"])
+	if err != nil {
+		return "", 0, nil
```
sreekanth-cb, line 144: in case of err, its returning nil?

mschoch: Yes, because error in this case just means that it was not int or float64, in which case we do not proceed to process the settings (which I believe is consistent with what we did before).

sreekanth-cb: I am not very sure of this.  Giving some hints about the incorrect format looks cleaner to me than not giving any clues to the user of what is going wrong.
Also, unable to comprehend the importance of maintaining consistency with what we did before for a small period of time between v11 - v12/13..

mschoch: I'm not arguing for consistency, I'm simply saying I'm not changing the behavior (other than adding support for float64).  If you'd like to improve the error handling, maybe you should create your own PR?

mschoch:
Just wasted half a day yesterday testing, because this PR hasn't been merged, bleve-blast wasn't using my forceSegmentVersion configuration.

PR #1404, opened 2020-05-25
optimize TermFacets

tylerkovacs:
The existing `TermFacets` implementation was O(n) when adding new TermFacets. All TermFacets were stored in a slice and the code iterated over the entire slice to determine if a term already existed.

This was further aggravated in `MultiSearch` where TermFacets for searches across each Index
were merged together, resulting in several iterations over the same slice.

This change adds a map to `TermFacets` so that existence checks are done in O(1).


tylerkovacs:
I ran some benchmarks on my index using the following code:

```
t1 := time.Now()
searchRequest := bleve.NewSearchRequest(bleve.NewMatchAllQuery())
f := bleve.NewFacetRequest(fieldName, limit)
searchRequest.AddFacet(fieldName, f)
fmt.Printf("latency for %s limit %d is %s\n", fieldName, limit, time.Since(t1).String())
```

I have one field in my index with cardinality around 120k. Here are the before/after results for that field:
```
Before: 29.674101683s
After:  384.708349ms
```

The change also benefits fields with much lower cardinatlity with the only downside being (what should be very modest) increased memory usage (to store the lookup map) while generating serach result.

mschoch:
Yeah, the original implementation was never intended to be used for a field with more than a handful of values.

Unfortunately as I understand it, though you made an attempt to parse the JSON in a backward compatible way, it's still a breaking change from the Go perspective (top-level FacetResult has struct definition change).  So, this would have to wait for a bleve 2.x, if agree to proceed.

tylerkovacs:
Yes, I'd figured this would have to wait since it is a breaking change. I'll maintain the code on my fork since I can't wait for 2.x to get these changes into production.

abhinavdangeti (approved):
Thanks for this contribution @tylerkovacs .
Looks good to me.

As this is a public API altering change, targeting this for the next major release v2.3.

sreekanth-cb (approved):

PR #1405, opened 2020-05-25
optimize FacetsBuilder

tylerkovacs:
Replace O(n) implementation in FacetsBuilder.UpdateVisitor

The existing implementation iterated over every facet in the search request and called `UpdateVisitor` along with the field name.

The new implementation maintains a map of FacetBuilders referencing each field and only calls `UpdateVisitor` on the relevant FacetBuilders.

### DIFF
--- a/search/facet/facet_builder_datetime.go
+++ b/search/facet/facet_builder_datetime.go
@@ -87,23 +87,21 @@ func (fb *DateTimeFacetBuilder) Field() string {
 	return fb.field
 }
 
-func (fb *DateTimeFacetBuilder) UpdateVisitor(field string, term []byte) {
-	if field == fb.field {
-		fb.sawValue = true
-		// only consider the values which are shifted 0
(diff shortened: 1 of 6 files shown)


tylerkovacs:
For me, a typical search request generates 30 facets on an index of size 500,000. With the existing implementation, total seach time for a simple `MatchAll` query that generated all facets took 9 seconds. With this change, the same query takes 6 seconds.

tylerkovacs:
Any thoughs on this PR? I don't mind maintaining it on my own branch but it results in such a big improvement that it would be nice to get it on master so that everyone can benefit.

abhinavdangeti (approved):
Thanks for this contribution @tylerkovacs .
Looks good to me. 

As this is a public API altering change, targeting this for the next major release v2.3.

sreekanth-cb (approved):

PR #1407, opened 2020-05-26
allow segment type/version override to work

### LINKED ISSUE #1406: Builder always uses zap v11
No matter what configuration you pass in, zap v11 is used due to a bug:

https://github.com/blevesearch/bleve/blob/master/index/scorch/builder.go#L109-L111

mschoch:
logic was reversed, preventing correct configuration
from actually switching the segment plugin used

fixes #1406

### DIFF
--- a/index/scorch/builder.go
+++ b/index/scorch/builder.go
@@ -106,7 +106,7 @@ func (o *Builder) parseConfig(config map[string]interface{}) (err error) {
 	if forcedSegmentType != "" && forcedSegmentVersion != 0 {
 		segPlugin, err := chooseSegmentPlugin(forcedSegmentType,
 			uint32(forcedSegmentVersion))
-		if err != nil {
+		if err == nil {
 			o.segPlugin = segPlugin
 		}
 	}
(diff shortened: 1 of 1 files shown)


index/scorch/builder.go
```diff
@@ -106,7 +106,7 @@ func (o *Builder) parseConfig(config map[string]interface{}) (err error) {
 	if forcedSegmentType != "" && forcedSegmentVersion != 0 {
 		segPlugin, err := chooseSegmentPlugin(forcedSegmentType,
 			uint32(forcedSegmentVersion))
-		if err != nil {
+		if err == nil {
```
sreekanth-cb, line 109: good catch!
Errs is returned on line no:104 and above, but here at line no:109 error is ignored?

mschoch:
@sreekanth-cb untested, but I changed it around again, do you prefer this?  It's the same challenge as that other PR, the config is optional, and involves multiple settings, so error reporting is complicated by differentiating invalid configs from omitted ones.

PR #1410, opened 2020-06-01
MatchOperator should be that type instead of int

ethervoid:
Right now `MatchOperator` constants are ints and when you're trying to
assign them to a variable you get an int. If you try to use those
variables as an operator you get the following error:

```
cannot use operator (type int) as type MatchQueryOperator in argument to q.SetOperator
```

So, to fix this problem we just need to create `MatchOperator` constants

### DIFF
--- a/search/query/match.go
+++ b/search/query/match.go
@@ -37,9 +37,9 @@ type MatchQueryOperator int
 
 const (
 	// Document must satisfy AT LEAST ONE of term searches.
-	MatchQueryOperatorOr = 0
+	MatchQueryOperatorOr = MatchQueryOperator(0)
 	// Document must satisfy ALL of term searches.
-	MatchQueryOperatorAnd = 1
+	MatchQueryOperatorAnd = MatchQueryOperator(1)
 )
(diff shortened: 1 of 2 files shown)


mschoch:
I'm fine with the change, but technically it does break the API since it changes the type of an exported identifier.  It is inside a sub-package, so probably not as big a deal, but still seems like something that should be a 2.0 change, since it can be worked around anyway.

ethervoid:
Read about release of version 2, would be these changes included?

mschoch:
I would be OK adding this to v2.  It's relatively small, but since its a breaking change, this is a good opportunity.  Thoughts @abhinavdangeti @sreekanth-cb?

abhinavdangeti:
Ya good time to merge I think.

mschoch:
This has been merged to the 2.x branch, and I've updated the documentation to mention it.  Thanks for the contribution and the reminder ping @ethervoid

ethervoid:
@mschoch  thank you for taking care :)

PR #1414, opened 2020-06-14
Shutdown analysis queue

vdimir:
Hi!

In package `bleve` function `init` runs several goroutines for procession analysis queue.
In some cases we need to shutdown this goroutines. For example, it triggers goleak tool.

```
package main

import (
	"testing"
	"go.uber.org/goleak"
)

func main() {
	var idx bleve.Index
	_ = idx
}

func TestMain(m *testing.M) {
	goleak.VerifyTestMain(m)
}
```

Causes:
```
goleak: Errors on successful test run: found unexpected goroutines:
...
```

Also, when `SetAnalysisQueueSize` is called analysis queue size not set to specified size, but increased on this number because exited workers isn't closed.

This PR adds `Shutdown` function that closes analysis queue from user's code and closes queue when new size is set.

### DIFF
--- a/config.go
+++ b/config.go
@@ -43,7 +43,14 @@ type configuration struct {
 }
 
 func (c *configuration) SetAnalysisQueueSize(n int) {
+	if c.analysisQueue != nil {
+		c.analysisQueue.Close()
+	}
 	c.analysisQueue = index.NewAnalysisQueue(n)
+}
+
+func (c *configuration) Shutdown() {
+	c.SetAnalysisQueueSize(0)
 }
 
 func newConfiguration() *configuration {


vdimir:
Also there are other question. Should we always run analysis queue by default? Maybe it is better not run in it `init`, but allow user to start process? But this PR provide conservative solution that leaves all functionality as is.

paskal (approved):

mschoch:
I think there is already consensus that the design is not a good one.  But fixing it properly will be a breaking change, so that will have to wait for 2.0.

I would be OK with this as an interim solution, thoughts @abhinavdangeti @sreekanth-cb ?

vdimir:
Any updates on this?

abhinavdangeti (approved):
Ok with this. This can be merged after one other approval from the team.

mschoch (approved):

PR #1416, opened 2020-06-15
fix stats handling of term field reader advancing backwards

### LINKED ISSUE #1415: term field reader stats out of sync when advancing backwards
The scorch implementation of `TermFieldReader` `Advance()` has special handling for the case when it is advancing backwards.  Essentially it cheats, and just creates a new `TermFieldReader`, and advances that forward.  This works, but the stats tracking is not handled correctly, and this leads to the `TotTermSearchersStarted` and `TotTermSearchersFinished` stats to get out of sync.

mschoch:
The current implementation creates a new term field reader,
which replaces the original one, this causes the stats
tracking start/finished to get out of sync.

This fixes the issue by recording the original term field
reader as finished when replacing with the new one.

fixes #1415


index/scorch/snapshot_index_tfr.go
```diff
@@ -133,6 +133,9 @@ func (i *IndexSnapshotTermFieldReader) Advance(ID index.IndexInternalID, preAllo
 		if err != nil {
 			return nil, err
 		}
+		// a new term searcher is started as a result of this, so consider the first closed
```
sreekanth-cb, line 136: one doubt,  now it looks like we are re allocating another tfr at line no: 131. 
Is there a way we could reuse/reassign the existing tfr here? 

What will happen if we call Close before seeking backwards? if thats possible then the stats will correct itself?

mschoch: Of course there is no need to allocate a new tfr.  @steveyen's comment is admitting he took a shortcut.  Properly resetting all the state so that it would work correctly was going to be hard, and Advance going backwards was a corner case (required for correctness, but not deemed that important) he chose to take this shortcut instead.

We could call close at 130.5, but then if any problem happens and we return an error, we already closed the TFR, which seems weird.  The way I coded this, I only adjust the stat after we're sure the new TFR was created and we're guaranteed to do the replacement.  It's all a bit hacky because of the shortcut, but I think this way is slightly better.

sreekanth-cb: Couldn't we call i.Close after a successful creation of i2?

mschoch: Yeah, I think that makes sense.  It's kind of late for me.  Let me look at it again in the morning, but I think so.

mschoch: I've updated it to call Close() but have to swallow the returned error.  I like calling close because it feels a bit cleaner and future-proof.  But, only reasonable thing to do is to swallow the error, and that makes it a lot less future-proof.  Anyway, this seems better than before?

PR #1417, opened 2020-06-18
MB-40007 -Obsolete segment file leaks during merge introductions

sreekanth-cb:
During merge segment introduction, if the segment contents
becomes totally obsolete, then the introducer skips
its introduction to the root. But it wasn't handling the clean
up ceremonies for the segment like Closing and cleaning
the entries from the ineligibleToRemove file list.

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -357,8 +357,10 @@ func (s *Scorch) planMergeAtSnapshot(ctx context.Context,
 			old:           oldMap,
 			oldNewDocNums: oldNewDocNums,
 			new:           seg,
-			notifyCh:      make(chan *notify),
+			notifyCh:      make(chan *mergeTaskIntroStatus),
 		}
+
+		s.fireEvent(EventKindMergeTaskIntroductionStart, 0)
 
 		// give it to the introducer
 		select {
@@ -372,23 +374,25 @@ func (s *Scorch) planMergeAtSnapshot(ctx context.Context,
 		introStartTime := time.Now()
 		// it is safe to blockingly wait for the merge introduction
 		// here as the introducer is bound to handle the notify channel.
-		notify := <-sm.notifyCh
+		introStatus := <-sm.notifyCh
 		introTime := uint64(time.Since(introStartTime))
 		atomic.AddUint64(&s.stats.TotFileMergeZapIntroductionTime, introTime)
 		if atomic.LoadUint64(&s.stats.MaxFileMergeZapIntroductionTime) < introTime {
 			atomic.StoreUint64(&s.stats.MaxFileMergeZapIntroductionTime, introTime)
 		}
 		atomic.AddUint64(&s.stats.TotFileMergeIntroductionsDone, 1)
-		if notify != nil && notify.iss != nil {
-			_ = notify.iss.DecRef()
-			if notify.skipped {
-				// decrement the ref counts on skipping introduction.
+		if introStatus != nil && introStatus.indexSnapshot != nil {
+			_ = introStatus.indexSnapshot.DecRef()
+			if introStatus.skipped {
+				// close the segment on skipping introduction.
 				s.unmarkIneligibleForRemoval(filename)
-				_ = seg.DecRef()
+				_ = seg.Close()
 			}
 		}
 
 		atomic.AddUint64(&s.stats.TotFileMergePlanTasksDone, 1)
+
+		s.fireEvent(EventKindMergeTaskIntroduction, 0)
 	}
 
 	// once all the newly merged segment introductions are done,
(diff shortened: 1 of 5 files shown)


index/scorch/merge.go
```diff
@@ -371,18 +374,25 @@ func (s *Scorch) planMergeAtSnapshot(ctx context.Context,
...
 		}
 		atomic.AddUint64(&s.stats.TotFileMergeIntroductionsDone, 1)
-		if newSnapshot != nil {
-			_ = newSnapshot.DecRef()
+		if introStatus != nil && introStatus.indexSnapshot != nil {
+			_ = introStatus.indexSnapshot.DecRef()
+			if introStatus.skipped {
+				// decrement the ref counts on skipping introduction.
+				s.unmarkIneligibleForRemoval(filename)
+				_ = seg.DecRef()
```
abhinavdangeti, line 389: I found this like a bit misleading - so and had to look around the code a bit to understand why this is needed.
It makes sense - I feel seg.Close() would look better (on line 368 above too) although it's the same thing. I'll leave that up to you.

sreekanth-cb: done!

mschoch: Funny you should mention that.  

In the bluge codebase yesterday I just made a similar change.  We (IMHO) make the same mistake with *IndexSnapshot's all over the place as well.  We directly call DecRef(), when we should call Close() and let decRef() be an implementation detail of Close()

In the case of *IndexSnapshot's, I think the correct pattern is to have a single `getCurrent()` which uses RLock() to copy the root pointer, adds a ref, and returns it to you.  From that point, the caller simply uses and closes when done.  That way the ref counting is hidden from the user.

I'm not exactly sure that same pattern works for Segments, but I think the principals are the same.

sreekanth-cb:
> The changes look good, and the test makes sense.
> 
> I have one question, is it possible to run a variation on this test, with the events in place, but the rest of the fix not in place. So that we could see if we get a segment file in the DEL state?

@mschoch , I missed this.. though on friday i started doing something locally but completely slipped out of mind..

```root@donut:~# lsof  | grep DEL
packageki  4949                   root  DEL       REG                8,2             2101088 /usr/lib/x86_64-linux-gnu/libapt-inst.so.2.0.0
packageki  4949                   root  DEL       REG                8,2             2100848 /usr/lib/x86_64-linux-gnu/libapt-pkg.so.5.0.1
gmain      4949  4950             root  DEL       REG                8,2             2101088 /usr/lib/x86_64-linux-gnu/libapt-inst.so.2.0.0
gmain      4949  4950             root  DEL       REG                8,2             2100848 /usr/lib/x86_64-linux-gnu/libapt-pkg.so.5.0.1
gdbus      4949  4951             root  DEL       REG                8,2             2101088 /usr/lib/x86_64-linux-gnu/libapt-inst.so.2.0.0
gdbus      4949  4951             root  DEL       REG                8,2             2100848 /usr/lib/x86_64-linux-gnu/libapt-pkg.so.5.0.1
sample    23495                   root  DEL       REG                8,2             7209069 /tmp/bleve-scorch-test-TestObsoleteSegmentMergeIntroduction/000000000003.zap
sample    23495 23496             root  DEL       REG                8,2             7209069 /tmp/bleve-scorch-test-TestObsoleteSegmentMergeIntroduction/000000000003.zap
sample    23495 23497             root  DEL       REG                8,2             7209069 /tmp/bleve-scorch-test-TestObsoleteSegmentMergeIntroduction/000000000003.zap
sample    23495 23498             root  DEL       REG                8,2             7209069 /tmp/bleve-scorch-test-TestObsoleteSegmentMergeIntroduction/000000000003.zap
sample    23495 23499             root  DEL       REG                8,2             7209069 /tmp/bleve-scorch-test-TestObsoleteSegmentMergeIntroduction/000000000003.zap
sample    23495 23500             root  DEL       REG                8,2             7209069 /tmp/bleve-scorch-test-TestObsoleteSegmentMergeIntroduction/000000000003.zap
sample    23495 23501             root  DEL       REG                8,2             7209069 /tmp/bleve-scorch-test-TestObsoleteSegmentMergeIntroduction/000000000003.zap
sample    23495 23502             root  DEL       REG                8,2             7209069 /tmp/bleve-scorch-test-TestObsoleteSegmentMergeIntroduction/000000000003.zap
sample    23495 23503             root  DEL       REG                8,2             7209069 /tmp/bleve-scorch-test-TestObsoleteSegmentMergeIntroduction/000000000003.zap


I did get this with the sample program on debian. Not sure why there are many entries for the same file. Its the 3rd zap file as we imagined!

ref - https://play.golang.org/p/G2Q5unb_Cmj

PR #1419, opened 2020-06-19
Switching to atomic ops instead of locks for events

sreekanth-cb:


index/scorch/scorch.go
```diff
@@ -157,29 +153,15 @@ func configForceSegmentTypeVersion(config map[string]interface{}) (string, uint3
...
-	s.pauseCount++
-	s.pauseLock.Unlock()
-}
-
-func (s *Scorch) decrPause() {
-	s.pauseLock.Lock()
-	s.pauseCount--
-	s.pauseLock.Unlock()
+	return atomic.LoadUint64(&s.stats.TotEventTriggerStarted) -
+		atomic.LoadUint64(&s.stats.TotEventTriggerCompleted)
```
mschoch, line 157: So, there is a trick to doing this right.  Remember that the two operations are independently atomic, but any amount of time could pass between them.  Consider the case where we first get the events started, and it is 27, then 5 minutes passes by, and we get the number of events completed.  But in those 5 minutes many more events have started and completed, so we get 35.  Now when you do the subtraction, you overflow the uint64 and return an unexpected value.

The trick is to always load the second operand of your subtraction first.  That way, even if some arbitrary amount of time passes between these two atomic ops, you don't overflow.

```
eventsCompleted := atomic.LoadUint64(&s.stats.TotEventTriggerCompleted)
eventsStarted := atomic.LoadUint64(&s.stats.TotEventTriggerStarted)
return eventsStarted - eventsCompleted
```

index/scorch/scorch.go
```diff
@@ -157,29 +153,15 @@ func configForceSegmentTypeVersion(config map[string]interface{}) (string, uint3
 }
 
 func (s *Scorch) paused() uint64 {
```
mschoch, line 155: I prefer a name like `NumEventsBlocking()` as `paused` implies a particular use case that only really applies to one kind of event.

index/scorch/scorch.go
sreekanth-cb: i had that sense of wrong doing here.. But this soln is pretty cool.. 👍

index/scorch/scorch.go
sreekanth-cb: done!

PR #1421, opened 2020-06-22
[BP] MB-40007 -Obsolete segment file leaks during merge introductions

sreekanth-cb:
During merge segment introduction, if the segment contents
becomes totally obsolete, then the introducer skips
its introduction to the root. But it wasn't handling the clean
up ceremonies for the segment like Closing and cleaning
the entries from the ineligibleToRemove file list.

### DIFF
--- a/index/scorch/merge_test.go
+++ b/index/scorch/merge_test.go
@@ -1,0 +1,154 @@
+//  Copyright (c) 2020 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


mschoch:
Do you also plan to backport the event lock->atomic change as well?  Wouldn't want to get a perf surprise in the backport.

mschoch:
On it's own probably doesn't make a difference no.  I was just thinking that the more events we add over time, the more it will add up.  I don't know how long this release will be maintained though.

index/scorch/merge_test.go
```diff
@@ -0,0 +1,292 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = DestroyTest(testConfig)
+	if err != nil {
+		t.Fatal(err)
+	}
+	log.Printf("path cleaned")
+	var in chan struct{}
+	_ = <-in
```
abhinavdangeti, line 291: What's this channel for? And the log statement above?

sreekanth-cb: oops.. something local changes snuck in..cleaned it.

PR #1428, opened 2020-07-16
Improve Readme example

refs:
The current example can only be run once, running again attempts to create an index in the same location, ending in an error:

```
panic: cannot create new index, path already exists

goroutine 1 [running]:
main.main()
        /Users/code/refs/bleve/main.go:21 +0x1bf
exit status 2
```

The proposed fix attempts to `Open` if an error occurred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Discuss usage and development of bleve in the [google group](https://groups.goog
 ## Indexing
 
 ```go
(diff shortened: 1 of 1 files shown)


iredmail:
Dear @abhinavdangeti 

Could you help review this PR? should be either merged or closed.

PR #1430, opened 2020-07-20
new FormatSpecifier option for FieldMapping

mschoch:
This new FormatSpecifier option allows us to convert numeric
values into strings suitable for indexing in a text field.

Format specifiers are relatively limited, but do allow for basic
conversion, as well as fixed width padding.

### DIFF
--- a/mapping/field.go
+++ b/mapping/field.go
@@ -59,6 +59,8 @@ type FieldMapping struct {
 	// DocValues, if true makes the index uninverting possible for this field
 	// It is useful for faceting and sorting queries.
 	DocValues bool `json:"docvalues,omitempty"`
+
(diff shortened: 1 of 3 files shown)


mapping/field.go
```diff
@@ -59,6 +59,8 @@ type FieldMapping struct {
 	// DocValues, if true makes the index uninverting possible for this field
 	// It is useful for faceting and sorting queries.
 	DocValues bool `json:"docvalues,omitempty"`
+
+	FormatSpecifier string `json:"format_specifier,omitempty"`
```
abhinavdangeti, line 63: Some commentary here that "%f" is the thing to use for correct translation of numeric (float) values?

PR #1435, opened 2020-08-04
MB-40730 - Sort by field returns incorrect results

sreekanth-cb:
Fixing the filterTermsByType method by resetting
the terms only when valid prefix coded, zero shifted
values are found to guard against any inadvertent overrides
of the original field values.

### DIFF
--- a/search/sort.go
+++ b/search/sort.go
@@ -427,7 +427,8 @@ func (s *SortField) filterTermsByType(terms [][]byte) [][]byte {
 				allTermsPrefixCoded = false
 			}
 		}
-		if allTermsPrefixCoded {
+		// reset the terms only when valid zero shift terms are found.
+		if allTermsPrefixCoded && len(termsWithShiftZero) > 0 {
 			terms = termsWithShiftZero
 			s.tmp = termsWithShiftZero[:0]
 		}


mschoch:
I don't think this is a good assumption, it is violated already here:
https://github.com/blevesearch/bleve/blob/master/document/field_geopoint.go#L34

sreekanth-cb:
yea...right  :) 
@mschoch , as the precision/shift is a variable here, looks like it doesn't make sense to add a check for precision on code, right? (like divisibility checks for both 4 & 9 on the shift parameter here - https://github.com/blevesearch/bleve/blob/master/numeric/prefix_coded.go#L103)

sreekanth-cb:
@mschoch , @abhinavdangeti , can I have your comments here please. 
Perhaps remove the modular division checks as the `conditional reset of terms` would fix the immediate issue we have at hand?

sreekanth-cb:
The modulo operations `shift&3 != 0 or shift&9 != 0` was intent to check whether the first byte (shift) is a multiple of the two precision steps we use today.  4 for numeric and 9 for geo.

But this isn't good as we could change the precision of encodings anytime, So let us remove that?

I tried to keep the search level change minimum here, hence I am inclined to keep the current fix of not resetting terms without finding a valid zero shift numeric term.

Are you particular about moving to this newer bluge implementation in bleve too?

mschoch:
Personally, I don't like restricting the shift values, because the original design was that those could be changed.  Although the likelihood of changing it now seems low (even though it could be a good short term fix to date/number indexing).

I only mentioned the bluge change because I think it also solves the problem for most cases.  Tokens like 'RUB' and 'TUB' would only get thrown away if there was also a token starting with <space>.  As I understand it, tokens like 'RUB' and 'TUB' are most likely to occur when the the keyword analyzer is in use (preserving capital letters).  But, the keyword analyzer also only produces a single token.  Thus you wouldn't have 'RUB' and '<space><any other N bytes where N is the expected number>'.

Do you not think this approach helps? Or do you see other drawbacks?

sreekanth-cb:
Not very sure of the keyword analyser part.. meaning whether it covers all the potential cases..

We can see tokens like Rot, Rut, Rop etc all fails the validity checks..

So if the sourceValues contains different texts out of which one happen to be any of these false positives with 
<space><any other N bytes where N is the expected number> is still a possibility?

mschoch:
I'm not sure I follow what you're saying.  The logic I proposed is to only remove terms when they're ALL numeric AND there is at least one with zero-padding.  If a field contains Rot, Rut, or Rop they would not get removed with the logic I proposed, because while these terms are valid numeric encodings, they are the ONLY term, and therefore the requirement that one of the terms be zero-padded is not satisfied.

As I said before, this whole thing is a heuristic, there are always going to be ways to defeat it.  If you want to remove all possibility you should remove support for the AUTO mode from Couchbase FTS.

A further reason I think the logic I proposed makes sense is that since it only removes non-zero padded terms when there is at least one zero-padded terms, it means that we only remove terms when there is at least more than one term.  This helps in practice, because when sorting on a text term, it almost never makes sense to sort on an analyzed field (or any field with more than one term).  Why?  Because ultimately we have choose one of those terms anyway to sort on, and we give the user very little control over that (default is first term, and I think we sort min/max as well).  The point is sorting on a field with multiple values is almost always a mistake anyway, and that is the only way the new logic I propose could fail...

sreekanth-cb:
oops..my copy paste example didn't work last time.. I was referring to a case with multiple tokens where the one of them is special false positive like above and others are a valid term.. 

As you said, there will always be exceptions.

The current change set here tried to incorporate both the fixes (improve the validations as well as the reset of terms), 
I am not clear on your review suggestions here.

mschoch:
Sorry I don't understand.

mschoch:
My belief is that the `shift&3 != 0 AND shift&9 != 0` operations should be removed for 2 reasons:

1.  They limit valid shifts going forward.  So, if we do that we should remove all places where the shift is configurable (even though it isn't used today).

2.  These operations are not required to give correct behavior for any known/reasonable inputs.

PR #1438, opened 2020-08-13
MB-40916: Do not recycle unadorned term field readers for optimized conjunctions/disjunctions

abhinavdangeti:
When a conjunction or a disjunction query is eligible to be optimized
to conjunctionUnadorned or disjunctionUnadorned, roaring.Bitmaps are
created by AND-ing or OR-ing the individual bitmaps. These'll need
to be released - for GC.

### DIFF
--- a/index/scorch/optimize.go
+++ b/index/scorch/optimize.go
@@ -266,10 +266,6 @@ OUTER:
 			bm.And(actualBM)
 		}
 
-		// On closure, do not recycle this optimized TermFieldReader -
-		// as the roaring.Bitmap created above needs to be released.
-		oTFR.doNotRecycle = true
-
 		oTFR.iterators[i] = segment.NewUnadornedPostingsIteratorFromBitmap(bm)
 	}
 
@@ -402,10 +398,6 @@ func (o *OptimizeTFRDisjunctionUnadorned) Finish() (rv index.Optimized, err erro
 		}
 
 		bm.AddMany(docNums)
-
-		// On closure, do not recycle this optimized TermFieldReader -
-		// as the roaring.Bitmap created above needs to be released.
-		oTFR.doNotRecycle = true
 
 		oTFR.iterators[i] = segment.NewUnadornedPostingsIteratorFromBitmap(bm)
 	}
--- a/index/scorch/optimize.go
+++ b/index/scorch/optimize.go
@@ -283,7 +283,7 @@ func (s *IndexSnapshotTermFieldReader) optimizeDisjunctionUnadorned(
 	octx index.OptimizableContext, minChildCardinality uint64) (index.OptimizableContext, error) {
 	if octx == nil {
 		octx = &OptimizeTFRDisjunctionUnadorned{
-			snapshot: s.snapshot,
+			snapshot:            s.snapshot,
 			minChildCardinality: minChildCardinality,
 		}
 	}
--- a/index/scorch/optimize.go
+++ b/index/scorch/optimize.go
@@ -266,6 +266,10 @@ OUTER:
 			bm.And(actualBM)
 		}
 
+		// On closure, do not recycle this optimized TermFieldReader -
+		// as the roaring.Bitmap created above needs to be released.
+		oTFR.doNotRecycle = true
+
 		oTFR.iterators[i] = segment.NewUnadornedPostingsIteratorFromBitmap(bm)
 	}
 
@@ -398,6 +402,10 @@ func (o *OptimizeTFRDisjunctionUnadorned) Finish() (rv index.Optimized, err erro
 		}
 
 		bm.AddMany(docNums)
+
+		// On closure, do not recycle this optimized TermFieldReader -
+		// as the roaring.Bitmap created above needs to be released.
+		oTFR.doNotRecycle = true
 
 		oTFR.iterators[i] = segment.NewUnadornedPostingsIteratorFromBitmap(bm)
 	}
--- a/index/scorch/optimize.go
+++ b/index/scorch/optimize.go
@@ -165,16 +165,8 @@ func (o *OptimizeTFRConjunctionUnadorned) Finish() (rv index.Optimized, err erro
 
 	// We use an artificial term and field because the optimized
 	// termFieldReader can represent multiple terms and fields.
-	oTFR := &IndexSnapshotTermFieldReader{
-		term:               OptimizeTFRConjunctionUnadornedTerm,
-		field:              OptimizeTFRConjunctionUnadornedField,
-		snapshot:           o.snapshot,
-		iterators:          make([]segment.PostingsIterator, len(o.snapshot.segment)),
-		segmentOffset:      0,
-		includeFreq:        false,
-		includeNorm:        false,
-		includeTermVectors: false,
-	}
+	oTFR := o.snapshot.unadornedTermFieldReader(
+		OptimizeTFRConjunctionUnadornedTerm, OptimizeTFRConjunctionUnadornedField)
 
 	var actualBMs []*roaring.Bitmap // Collected from regular posting lists.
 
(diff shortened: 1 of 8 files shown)


index/scorch/optimize.go
```diff
@@ -262,6 +262,10 @@ OUTER:
 			bm.And(actualBM)
 		}
 
+		// On closure, do not recycle this optimized TermFieldReader -
+		// as the roaring.Bitmap created above needs to be released.
+		oTFR.doNotRecycle = true
```
sreekanth-cb, line 267: not sure whether we need this extra field as it is already using a specific fieldName/term for all the optimisations.. So actual recycle logic could be somewhere inside the recycleTermFieldReader.

Lesser surface area for changes and no need to set this extra flag..

abhinavdangeti: Hmm, we can avoid recycling only when a new instance of roaring.Bitmap is created. There's several cases above where we just end up reusing the bitmaps and not setting up new instances.

sreekanth-cb: In all such applicable cases, we would have created a new IndexSnapshotTermFieldReaders with that customised field/terms? So it can still handled in a single place.

Another thought is -  Do we really need to throw away the newly created IndexSnapshotTermFieldReaders ?
Can't we enable the reuse for them too? ie can't we alloc them or ask from the index snapshot the same way we do for other TFRs?

sreekanth-cb: I am taking back my thought on re usability/recycling here. Mostly this will only add more IS TFRs pointers for GC to track and hence pain rather a win.

abhinavdangeti:
Hey @mschoch , let me know your thoughts on the latest patch set. Cheers.

index/scorch/snapshot_index_tfr.go
```diff
@@ -181,6 +181,13 @@ func (i *IndexSnapshotTermFieldReader) Count() uint64 {
 func (i *IndexSnapshotTermFieldReader) Close() error {
 	if i.snapshot != nil {
 		atomic.AddUint64(&i.snapshot.parent.stats.TotTermSearchersFinished, uint64(1))
+		if bytes.Equal(i.term, OptimizeTFRConjunctionUnadornedTerm) ||
+			bytes.Equal(i.term, OptimizeTFRDisjunctionUnadornedTerm) {
+			// Reset iterators before recycling unadorned term field readers,
+			// so as to release the roaring.Bitmaps set up by AND-ing or OR-ing
+			// individual bitmaps for GC.
+			i.iterators = i.iterators[:0]
```
mschoch, line 189: I'm not sure I follow this part, is it because these bitmaps are on the heap, while in the other case the bitmaps are backed by bytes mmap'd from disk?

I really liked the way you organized the rest of the code into the unadornedTermFieldReader method, as it reduced duplication and the possibility that some new field doesn't get initialized correctly down the road.  But here I don't like that IndexSnapshotTermFieldReader has to know anything about optimizations...

abhinavdangeti: I was thinking that if we don't let go of these iterators here, the bitmaps would never be released - but I was wrong. This bit here is unnecessary. Let me update the last patchset.

mschoch: So, hate to go back and forth on this, but now it seems like there is something to this.

It seems that when we recycle the TFR, we don't reset any of the slices inside (dicts, postings, iterators).  Now, at first I thought, maybe that was on purpose, but the TFRs are shared across the entire snapshot, so you can't safely reuse any of those fields, you have to reinitialize them anyway (different field, term or both).  To be clear, the only thing we're reusing is the struct itself, not the contents.

So, that means a no-longer-used struct sits in slice, possibly holding on to references to other things.  Is that a problem?  Let's look closer, the slice of dictionaries does use some memory, but complex to build FST is already being cached inside of zap, and bytes back it are mmap'd, so it's just the size of the Dictionary structs that really counts.  Next, the PostingsLists, again the structs themselves, and the roaring.Bitmap structs would take some memory, but, in the case of normal TFRs the bytes backing the roaring are mmap'd.  But, in the optimization case, these are on the heap.  Finally, the PostingsIterators, another struct, possibly another bitmap, and several more things that are generally mmap'd.

In conclusion, it seems like there is sufficient reason to reset all of these slices before recycling, and in all cases, not just the optimization case.  Does this make sense?

abhinavdangeti: In today's code we recycle the TFRs without actually reseting any of the dicts, postings and iterators. When the TFRs are picked up for re-use, we only re-use dicts, but overwrite postings and iterators and at this point we release the references from the previous use.

So reseting the postings and iterators slices shouldn't change anything - and is probably the better thing to do.
As for the dicts slice, if we were to reset them - we'd incur additional overhead of having to set them as well during reuse.

mschoch: Ah you're correct, the dicts slice is reused, so we should not reset that one.

Do you agree that resetting the other slices potentially frees up memory sooner?  It eventually gets freed, but not until someone else reuses that instance and overwrites the pointer...

abhinavdangeti: Yes, agreed. Let me push up another patch-set here.

abhinavdangeti: One more thing, looks like we use the previous postings entry to decide on the next one on re-use. So it's actually just the iterators we can reset safely.

abhinavdangeti: So unfortunately, we're back to the original comment here - it makes sense to clear iterators only for unadorned term field readers :/

mschoch: Well I've misread the code 3 times now, so my opinion isn't very helpful.

sreekanth-cb:
For the optimised TFR, not resetting the iterator's slice of before adding them back for recycling is sub optimal and then it also brings us back to the original point of effective re usability here. 

Looks like we aren't re cycling much stuffs here with optimised TFRs and making it to follow the current design of reusability is throwing an equal amount of design/layering concerns..

mschoch:
@sreekanth-cb agreed.  I now think we just shouldn't recycle these.  So, what is the cleanest way to accomplish that without it's own set of design/layering concerns.

sreekanth-cb:
@mschoch , IndexSnapshotTermFieldReader's Close() implementation deciding it's own re-usability isn't looking that bad to me.  We aren't touching the older style of TFR reusability, rather TFRs are smarter enough to just don't stack these special cases for reuse.

For example, having an extra heuristic (as we just thrown away one :) ) decider at line: 183.5 at snapshot_index_tfr.go 
https://github.com/blevesearch/bleve/blob/master/index/scorch/snapshot_index_tfr.go#L183

i.isReusable() / anyBetterNamedFunc() { 
   i.snapshot.recycleTermFieldReader(i)
}

But I know, you already disliked this idea earlier. So essentially I don't have any other cleaner idea..

mschoch:
So, to me, adding another boolean to the struct reinforces the reality that you have 2 different types hiding inside one Go type.  Additionally, the optimized case has uninitialized fields (dicts, postings).  These are intermediate structures used in the creation of a normal TFR, that aid in faster reuse, but do not exist in the optimized case.

A cleaner solution would be if a TFR was recyclable or not based on some internal property.  For example, a clean solution (that probably won't work for other reasons) would be `if i.snapshot != nil { /* then recycle */ }`  That doesn't work in practice because the snapshot is checked for other purposes during optimization itself, but it would be cleaner because the snapshot is actually required to perform the recycling, so the check makes sense even without knowing anything about optimized TFRs.

The recycling could key on the nil dicts and postings, but I don't like that either as that seems a consequence of the way the optimized TFR is built, not some essential property of recycling.

I don't have a suggestion yet, just sharing my thinking.

mschoch:
OK, how about this:

We add a `recycle bool` to the TFR struct, and the only place you're allowed to set it to true is inside: `func (i *IndexSnapshot) allocTermFieldReaderDicts(field string)`  We can't enforce that, but the idea being that recycle is always false, unless you got it from the code managing the pool.

abhinavdangeti:
Ya that sounds ok to me. So this will take us back to adding a conditional to Close() before recycling, but I agree the - recycle boolean flag looks a little more neater with it being set only within allocTermFieldReaderDicts.

sreekanth-cb:
If we look into the commit history here,
there is one each with titles 
Release.... , 
Re use..., 
Do not Recycle..., 
but our original change isn't that confusing/complex. 
So would be great if you squash these commits with a better commit msg. 
I prefer a commit msg which focuses on not recycling TFR part than first the commit message here at the top which overly focus on the AND/OR optimisations and roaring Bitmaps.. I like your last commit msg.

abhinavdangeti:
Yea will squash them all before merging, will set the commit message to reflect the latest.

PR #1439, opened 2020-08-13
Skip the score:none optimisations with composite queries

sreekanth-cb:
when any of the sub queries are comprised of composites
like disjuncts queries. (eg:fuzzy clauses).

In the current implementation, the Optimize(kind string,
octx OptimizableContext) (OptimizableContext, error) method
is not returning an empty context when the it fails to optimise.

And that causes the higher composite searcher level nil checks against
the octx futile and it incorrectly proceeds with the incomplete optimizable
context to result in wrong results.

For example,

Consider the query with the option score:"none":

"query": {
   "conjuncts": [
         {
           "field": "name",
           "match": "hotel"
         },
         {
          "field": "directions",
          "match": "56",
          "fuzziness": 1
         }
   ]
}

One of the conjunct searcher would be a disjunct heap/slice searcher
wrapping all the candidate term searcherss of the fuzzy clause.
And within the Optimize() implementations they only handle an edge case
of a single child searcher and silently return back the incoming octx as it is
even when it is not performing the optimisations too.

This causes higher layers to use the incomplete octx inadvertently
to result in incorrect hits eventually.


mschoch:
@sreekanth-cb one more thought on this.  You had that example in Couchbase illustrating incorrect results with 4 docs.  Could we convert that to a Bleve unit test and add it here?

sreekanth-cb:
@mschoch , done!

PR #1446, opened 2020-08-18
MB-40962: Dropping the heuristic to skip disjunction optimizations

abhinavdangeti:
+ Getting rid of the minChildCardinality heuristic for optimized
  unadorned disjunctions where we previously used to skip the
  optimization in case of low-cardinality bitmaps.
+ Removing this heuristic showed tremendous improvement for certain
  types of queries which involved a conjunction of disjunctions of
  fuzzy queries.
+ However, here's something worth noting from the commit that introduced
  this heuristic .. https://github.com/blevesearch/bleve/commit/58e6641d13844e212687bd35823914c5e26b04dd
  ```
   Regarding perf microbenchmarks (bleve-query) on a 200K en-wiki docs scorch index...

    for a high number of high-frequency terms...

    - wildcard search on "th*" (~31K hits)...
        before the change, with normal scoring         -  ~4.7 q/sec
        w/ NoScore:true                                -  ~5.1 q/sec
        w/ NoScore:true & unadorned disj. optimization - ~11.6 q/sec

    for a low number of high-frequency terms...

    - query-string search on "http www com" (~25K hits)...
        before the change, with normal scoring         - ~190 q/sec
        w/ NoScore:true                                - ~260 q/sec
        w/ NoScore:true & unadorned disj. optimization - ~415 q/sec

    for a low number of low-frequency terms...

    - query-string search on "marty shoch" (207 hits)...
        before the change, with normal scoring         - ~15.0K q/sec
        w/ NoScore:true                                - ~21.3K q/sec
        w/ NoScore:true & unadorned disj. optimization - ~16.1K q/sec
  ```

### DIFF
--- a/index/scorch/optimize.go
+++ b/index/scorch/optimize.go
@@ -37,11 +37,7 @@ func (s *IndexSnapshotTermFieldReader) Optimize(kind string,
 	}
 
 	if OptimizeDisjunctionUnadorned && kind == "disjunction:unadorned" {
-		return s.optimizeDisjunctionUnadorned(octx, OptimizeDisjunctionUnadornedMinChildCardinality)
-	}
-
-	if OptimizeDisjunctionUnadorned && kind == "disjunction:unadorned-force" {
-		return s.optimizeDisjunctionUnadorned(octx, 0)
+		return s.optimizeDisjunctionUnadorned(octx)
 	}
 
 	return nil, nil
@@ -272,11 +268,10 @@ OUTER:
 // term-vectors are not required, and instead only the internal-id's
 // are needed.
 func (s *IndexSnapshotTermFieldReader) optimizeDisjunctionUnadorned(
-	octx index.OptimizableContext, minChildCardinality uint64) (index.OptimizableContext, error) {
+	octx index.OptimizableContext) (index.OptimizableContext, error) {
 	if octx == nil {
 		octx = &OptimizeTFRDisjunctionUnadorned{
-			snapshot:            s.snapshot,
-			minChildCardinality: minChildCardinality,
+			snapshot: s.snapshot,
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


abhinavdangeti:
Hey @mschoch , @sreekanth-cb , I've pulled in details and added them to the commit message (above) from @steveyen 's commit that shows the reason why the heuristic was introduced for unadorned disjunctions. Perhaps this changes things or perhaps it doesn't, but definitely warrants another conversation around this.

mschoch:
@abhinavdangeti thanks for digging that up, it certainly helps quantify things.

To me, it's not that the heuristic is wrong, it's that it's implemented at such a low-level it lacks sufficient context to make the decision.

Specifically, now that we have optimizations that can stack together up the query tree, making a decision based on the cardinalities involved in one clause, ends up affecting the performance of the entire query.  A higher-level conjunction/disjunction may fail to be optimized because this one clause skipped the optimization.

Unfortunately, I don't see a simply solution at the moment.  These optimizations happen from the bottom up.  We're forced to decide, optimize this clause or not, without any idea if a higher-level optimization is in play.

A cleaner design would be for us to make better use of the Query/Searcher split.  The searchers should be focused on execution, and queries should be focused on representing the logical structure of the query.  Some sort of optimization phase could make transformations to the query hierarchy, perhaps aided by statistics from the term dictionary.  The benefit is that it has that global view of the whole query, and isn't limited by only seeing a small piece.

I think it would also offer an opportunity to simplify the optimization code paths.  We don't have to build the unoptimized searchers first, and then optimized them.  We can just directly expose methods to build unadorned conjunction/disjunction bitmaps for multiple field/terms at a time.

For now, it seems like we can disable (but not remove) the heuristic by changing the 256 value to 0.  Couchbase perf tests won't show regression, because we don't have any historical runs with score=none.

To me it seems like the benefits exceed the downsides for real-world queries.

What is everyone else thinking?

mschoch:
Also, I'm fine with just removing it too, I see that's how it's already coded.

abhinavdangeti:
Ya think I agree with you.
Since it was coded in such a way to be easily removed, i just removed it for now.

sreekanth-cb (approved):

PR #1447, opened 2020-08-19
simplify compute geo range

mschoch:
attempt to go back to a simpler compute geo range
but not give up the perf improvements

### DIFF
--- a/search/searcher/search_geoboundingbox.go
+++ b/search/searcher/search_geoboundingbox.go
@@ -100,6 +100,7 @@ func NewGeoBoundingBoxSearcher(indexReader index.IndexReader, minLon, minLat,
 
 var geoMaxShift = document.GeoPrecisionStep * 4
 var geoDetailLevel = ((geo.GeoBits << 1) - geoMaxShift) / 2
+type closeFunc func() error
 
 func ComputeGeoRange(term uint64, shift uint,
 	sminLon, sminLat, smaxLon, smaxLat float64, checkBoundaries bool,
@@ -120,10 +121,24 @@ func ComputeGeoRange(term uint64, shift uint,
 		return rv
 	}
 
-	var fieldDict index.FieldDictContains
-	var isIndexed filterFunc
+	isIndexed, closeF, err := buildIsIndexedFunc(indexReader, field)
+	if closeF != nil {
+		defer func() {
+			cerr := closeF()
+			if cerr != nil {
+				err = cerr
+			}
+		}()
+	}
+
+	onBoundary, notOnBoundary = computeGeoRange(term, shift, sminLon, sminLat, smaxLon, smaxLat, checkBoundaries, onBoundary, notOnBoundary, makePrefixCoded, isIndexed)
+
+	return onBoundary, notOnBoundary, nil
+}
+
+func buildIsIndexedFunc(indexReader index.IndexReader, field string) (isIndexed filterFunc, closeF closeFunc, err error) {
 	if irr, ok := indexReader.(index.IndexReaderContains); ok {
-		fieldDict, err = irr.FieldDictContains(field)
+		fieldDict, err := irr.FieldDictContains(field)
 		if err != nil {
 			return nil, nil, err
 		}
@@ -132,22 +147,18 @@ func ComputeGeoRange(term uint64, shift uint,
 			found, err := fieldDict.Contains(term)
 			return err == nil && found
 		}
-	}
 
-	defer func() {
-		if fieldDict != nil {
+		closeF = func() error {
 			if fd, ok := fieldDict.(index.FieldDict); ok {
-				cerr := fd.Close()
-				if cerr != nil {
-					err = cerr
+				err := fd.Close()
+				if err != nil {
+					return err
 				}
 			}
+			return nil
 		}
-	}()
-
-	if isIndexed == nil {
+	} else if indexReader != nil {
 		isIndexed = func(term []byte) bool {
-			if indexReader != nil {
 				reader, err := indexReader.TermFieldReader(term, field, false, false, false)
 				if err != nil || reader == nil {
 					return false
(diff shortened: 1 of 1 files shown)


mschoch:
OK, this requires a bit of background.

I've been running lots of static analysis tools against Bluge, and when I got to ComputeGeoRange I was really sad.  The function had something like 4 inline closures 2 of which recursively called each other.  Needless to say this function violated the length a cyclomatic complexity rules.  But, beyond that, the function was really a mess (sorry @steveyen) 

I would see things like this:

https://github.com/blevesearch/bleve/blob/master/search/searcher/search_geoboundingbox.go#L195

The function literally starts by check `if err != nil` before it's even done anything? If you look harder you'll see other similar oddities that can only be explained by all the shared state between these functions.

So, I went back to the drawing board.  How did the code end up like this?  Good question, see: https://github.com/blevesearch/bleve/pull/1172

In this @steveyen observes a few things:
- many parameters passed down recursively
- slices returned, and then appended to combine in to master slice
- many allocations from prefix coded numbers -> []byte

His changes in that PR address it, but in the process they made the code a lot less maintainable.  I complained about this in the PR, but ultimately approved the PR.

I've now attempted to right this wrong...

First, I decided to assume that having lots of parameters passed recursively wasn't really the problem.  And in fact, the other 2 issues were the real issue.  So, with that in mind, I could avoid the shared state of Steve's solution, by adding even more parameters.  In my solution, I have the two slices up front, that get appended to as we go (no more intermediate slices which have to be copied).  Second, I simply borrowed the prefix coded reuse solution , as was able to pass that function down as yet another argument.

Perf tests at this point seemed promising, so I went ahead and tried to finish the PR.

One more thing remained, after Steve's PR, @sreekanth-cb added the ability to filter terms which are "actually used in the index".  This too was able to be passed in as a function.  And I further took the opportunity to extract the building of this function into a separate method.  Note: this has 3 modes (direction dictionary lookup, manual search for term, or always assume it's in the index)

Finally, what does the perf look like.  Here is my before (on my laptop):

```
$ go test -run=xxx -bench=. -benchmem
goos: linux
goarch: amd64
pkg: github.com/blevesearch/bleve/search/searcher
BenchmarkComputeGeoRangePt01-8   	  120304	      9908 ns/op	     208 B/op	       4 allocs/op
BenchmarkComputeGeoRangePt1-8    	   28938	     40643 ns/op	   16341 B/op	      25 allocs/op
BenchmarkComputeGeoRange10-8     	      99	  10150456 ns/op	 8503322 B/op	      73 allocs/op
BenchmarkComputeGeoRange100-8    	      19	  54421748 ns/op	42778612 B/op	      86 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/searcher	6.428s
```

And here is my after:

```
go test -run=xxx -bench=. -benchmem       
goos: linux
goarch: amd64
pkg: github.com/blevesearch/bleve/search/searcher
BenchmarkComputeGeoRangePt01-8   	   99495	     10380 ns/op	     208 B/op	       4 allocs/op
BenchmarkComputeGeoRangePt1-8    	   28658	     41054 ns/op	   16341 B/op	      25 allocs/op
BenchmarkComputeGeoRange10-8     	     100	  10081498 ns/op	 8503335 B/op	      73 allocs/op
BenchmarkComputeGeoRange100-8    	      24	  50566678 ns/op	42778634 B/op	      86 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/searcher	6.051s
```

Good news, the allocations are the same, and the perf is almost as good, but ever so slightly slower (in some cases, slightly faster others).

But, the code is much easier to follow: `relateAndRecurse`, `computeGeoRange`, and `buildIsIndexedFunc` are separate functions, with clear parameters/returns connecting them.

mschoch:
Updated perf from latest commit:

```
go test -run=xxx -bench=. -benchmem
goos: linux
goarch: amd64
pkg: github.com/blevesearch/bleve/search/searcher
BenchmarkComputeGeoRangePt01-8   	  109303	      9409 ns/op	     208 B/op	       4 allocs/op
BenchmarkComputeGeoRangePt1-8    	   31486	     39468 ns/op	   16341 B/op	      25 allocs/op
BenchmarkComputeGeoRange10-8     	     116	   9712282 ns/op	 8503324 B/op	      73 allocs/op
BenchmarkComputeGeoRange100-8    	      24	  48345987 ns/op	42778614 B/op	      86 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/searcher	8.058s
```

Now on my laptop it's the same or faster than before, same number of allocs, but much better organized code (IMHO).

mschoch:
Probably the geoRangeCompute should encapsulate the isIndexed functionality as well, and just have it's own Close() method hiding those details as well, but it's already so much better than before...

mschoch:
@sreekanth-cb @abhinavdangeti if it helps I can do a code walkthough of this change.  Looking at the diff it can be overwhelming, but like Steve's original change, it's a very mechanical set of changes, not touching the logic.  So it might be easier to review with that part explained.

search/searcher/search_geoboundingbox.go
```diff
@@ -157,68 +165,15 @@ func ComputeGeoRange(term uint64, shift uint,
...
-		var upperMax uint64
-		if shift < 63 {
-			upperMax = term | ((uint64(1) << (shift + 1)) - 1)
-		} else {
-			upperMax = 0xffffffffffffffff
-		}
-
-		lowerMax := split - 1
-
-		level := (GeoBitsShift1 - shift) >> 1
```
abhinavdangeti, line 209: Now that you aren't using this variable `GeoBitsShift1`, you could remove its declaration at line 27 in this file.

mschoch: That is still used elsewhere in the file (line 44)

abhinavdangeti: Ah ok - in that case an op can be saved below on line 251 within geoRangeCompute's relateAndRecurse(..) while initializing level ..
```
level := (GeoBitsShift1 - res) >> 1
```

mschoch:
Perf with latest commit:

```
go test -run=xxx -bench=. -benchmem
goos: linux
goarch: amd64
pkg: github.com/blevesearch/bleve/search/searcher
BenchmarkComputeGeoRangePt01-8   	  114841	      9301 ns/op	     208 B/op	       4 allocs/op
BenchmarkComputeGeoRangePt1-8    	   31824	     37815 ns/op	   16341 B/op	      25 allocs/op
BenchmarkComputeGeoRange10-8     	     124	   9430192 ns/op	 8503338 B/op	      73 allocs/op
BenchmarkComputeGeoRange100-8    	      24	  47343916 ns/op	42778611 B/op	      86 allocs/op
PASS
ok  	github.com/blevesearch/bleve/search/searcher	6.999s
```

search/searcher/search_geoboundingbox.go
mschoch: See if the most recent commit addresses what you meant.

mschoch:
So, I also tried Sreekanth's observation, that since level is compute by both calls to relateAndRecurse, we can compute it once and pass it in.  But, it seems to be in the noise vs signal.  Sometimes faster, sometimes slower...  I wonder if adding that additional argument spills out of registers or something.

Issue #1455, opened 2020-08-26
Embedded anonymous struct pointers don't elide field name


bmhatfield:
Looking through the code, and at this commit in particular, https://github.com/blevesearch/bleve/commit/9f31ea6805b221152f05a342c725ba531db2eaec - it appears to me that there's an intent to elide the field name of an embedded type.

This works fine in practice if it's something like:

```go
type Inner struct {
  Field string
}

type Outer struct {
  Inner
}
```

for which the discovered property name will be `Field`. However, if the embed is changed to a pointer:

```go
type Outer struct {
  *Inner
}
```

then the property name is interpreted as `Inner.Field`.

Some of our structs are not embeddable as a non-pointer (because they are proto messages, which trips up `go vet`'s `copylocks` checker).

It would be very helpful if we could embed pointers anonymously as well. It's fairly challenging to have the document mappings change depending on whether or not it's a pointer.

 I have been poking at this trying to see if I can send a PR, but I can't quite see where the property name interpretation is happening just yet.

mschoch:
Based on the commit message in the commit you linked to it sounds like this was on purpose, as it specifically describes non-struct behavior as different, even when it's a pointer to the struct.

Further, it's highly unlikely we'll make a big change here, as many people now rely on the existing behavior of the mapping (regardless of whether they like it or not).

My best suggestion is for you to manually map your documents, and index them using the `IndexAdvanced` method.  It's slightly more work, but it's a one time thing, and it gives you full control over how every aspect of your documents are indexed.  As a bonus, your code won't have to use reflection, as you can work directly with your types.

bmhatfield:
Understood. Thanks for the note. I'll look into manual mapping and IndexAdvanced. Thank you!

PR #1456, opened 2020-08-26
walkDocument: elide anonymous field name for struct pointers

bmhatfield:
This fixes the issue discussed in https://github.com/blevesearch/bleve/issues/1455


bmhatfield:
Based upon the comment in https://github.com/blevesearch/bleve/issues/1455#issuecomment-681178413, this appears to be a load-bearing issue and this PR would break some folks. Closing.

PR #1465, opened 2020-09-17
escape the input text in HTML highlighter when adding <mark> tags, fixes #1464

### LINKED ISSUE #1464: HTML highlighter should escape its input
Hi!

I'm trying out HTML highlighter. I believe the highlighter takes plain text as input and outputs HTML text. Thus, it should treat its input as plain text and properly escape it when adding `<mark>` tags.

Let's say I have three documents:

    id: "doc1", text: "Lorem ipsum dolor sit amet"
    id: "doc2", text: "Thelma & Louise"
    id: "doc3", text: "I <3 Blancmange"
    id: "doc4", text: "The <script> tag is potentially dangerous"

If I search for `Lorem`, I get a nice HTML fragment `<mark>Lorem</mark> ipsum dolor sit amet`.

However, if I search for `Thelma`, I get `<mark>Thelma & Louise</mark>`. I would expect to get `<mark>Thelma &amp; Louise</mark>`.

Searching for `tag` would result in e.g. `The <script> <mark>tag</mark> is potentially dangerous`, which wound hide the rest of the HTML page after "The".

Am I possibly missing something, or should the highlighter do the escaping?

tuomassalo:


ingenthr:
@mschoch and @sreekanth-cb, see note on CLA submission on @tuomassalo's other PR.

mschoch (commented):
Thanks, this is certainly an improvement, however I'm wondering if we also need to escape the matched search term itself.  While the returned output might end up less desirable in some cases, it would seem that without that, there is still the possibility of some security issue.  Thoughts?

mschoch:
Thanks again @ingenthr we hadn't gotten to that yet, but a ping was helpful, as we've fallen behind on community PRs again.

tuomassalo:
@mschoch Certainly the search term should also be escaped, if it can contain something like `<>&`. I apologize for missing that!

This seems trivial to fix (add `html.EscapeString` to [this line](https://github.com/blevesearch/bleve/pull/1465/files#diff-80e0440c09a298ab0e1fab58106eae31R63)). However, I'm not familiar enough with the codebase or its test suite that I could write a test against the change.

mschoch:
@tuomassalo I've added another commit.  It adds HTML escaping to the highlighted term itself.  I also reorganized the unit tests and added 2 new ones which exercise the HTML escaping.  The first has characters that need escaping around the term, and the second one also has a character needing escaping inside the term.

Let me know if this makes sense to you.

sreekanth-cb (approved):

abhinavdangeti (approved):

PR #1466, opened 2020-09-17
fix tests on go 1.15

tuomassalo:
`go test` fails on my macos (go 1.15.2), complaining:

    mapping/mapping_test.go:962:31: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)

This PR fixes the issue by just doing what the error suggests. But I didn't dig into whether this is the thing that should be done and what the code actually did before this change.

### DIFF
--- a/mapping/mapping_test.go
+++ b/mapping/mapping_test.go
@@ -959,7 +959,7 @@ func TestMappingForGeo(t *testing.T) {
 	expect = append(expect, []float64{-71.34, 41.12})
 
 	for i, geopoint := range geopoints {
-		doc := document.NewDocument(string(i))
+		doc := document.NewDocument(fmt.Sprint(i))
 		err := mapping.MapDocument(doc, geopoint)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


ingenthr:
@mschoch and @sreekanth-cb, note that @tuomassalo has submitted a CLA via email.  Amy Tu at Couchbase has it if you'd like to check details.

mschoch:
Thanks @ingenthr I don't think anyone had checked yet.

The issue found here was a change introduced by the Go 1.15 release.  It is mentioned in the release notes.  Basically it is a language construct that is frequently misused, so they have made it a warning in `go vet`.  Further, this vet test is included with `go test`, leading to the issue reported here:  https://golang.org/doc/go1.15#vet

I've also created a PR to add go 1.15.x to our CI testing matrix here:
https://github.com/blevesearch/bleve/pull/1468

mschoch (approved):

PR #1470, opened 2020-09-25
New stats for reclaimable disk space from scorch index

sreekanth-cb:
Adding a new stat `num_bytes_used_disk_by_root_reclaimable`
which hints about the approximate amount of reclaimable disk space
from a scorch index. This could turn out to be a useful insight
into knowing the amount of disk space wasted due to tombstoned/obsoleted
contents across the segments and could potentially act as a trigger input
for the forceMerge api.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -515,6 +515,26 @@ func (s *Scorch) diskFileStats(rootSegmentPaths map[string]struct{}) (uint64,
 	return numFilesOnDisk, numBytesUsedDisk, numBytesOnDiskByRoot
 }
 
+// reClaimableDocsRatio gives a ratio about the obsoleted or
+// reclaimable documents present in a scorch index.
+func (s *Scorch) reClaimableDocsRatio() float64 {
+	var totalCount, liveCount uint64
+	s.rootLock.RLock()
+	indexSnapshot := s.root
+	s.rootLock.RUnlock()
+	for _, segmentSnapshot := range indexSnapshot.segment {
+		if _, ok := segmentSnapshot.segment.(segment.PersistedSegment); ok {
+			totalCount += uint64(segmentSnapshot.FullSize())
+			liveCount += uint64(segmentSnapshot.Count())
+		}
+	}
+
+	if totalCount > 0 {
+		return float64(totalCount-liveCount) / float64(totalCount)
+	}
+	return 0
+}
+
 func (s *Scorch) rootDiskSegmentsPaths() map[string]struct{} {
 	rv := make(map[string]struct{}, len(s.root.segment))
 	for _, segmentSnapshot := range s.root.segment {
@@ -556,6 +576,9 @@ func (s *Scorch) StatsMap() map[string]interface{} {
 	m["num_bytes_used_disk"] = numBytesUsedDisk
 	// total disk bytes by the latest root index, exclusive of older snapshots
 	m["num_bytes_used_disk_by_root"] = numBytesOnDiskByRoot
+	// num_bytes_used_disk_by_root_reclaimable is an approximation about the
+	// reclaimable disk space in an index. (eg: from a full compaction)
+	m["num_bytes_used_disk_by_root_reclaimable"] = uint64(float64(numBytesOnDiskByRoot) * s.reClaimableDocsRatio())
 	m["num_files_on_disk"] = numFilesOnDisk
 	m["num_root_memorysegments"] = m["TotMemorySegmentsAtRoot"]
 	m["num_root_filesegments"] = m["TotFileSegmentsAtRoot"]
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -515,41 +515,17 @@ func (s *Scorch) diskFileStats(rootSegmentPaths map[string]struct{}) (uint64,
 	return numFilesOnDisk, numBytesUsedDisk, numBytesOnDiskByRoot
 }
 
-// reClaimableDocsRatio gives a ratio about the obsoleted or
-// reclaimable documents present in a scorch index.
-func (s *Scorch) reClaimableDocsRatio() float64 {
-	var totalCount, liveCount uint64
-	s.rootLock.RLock()
-	indexSnapshot := s.root
-	s.rootLock.RUnlock()
-	for _, segmentSnapshot := range indexSnapshot.segment {
-		if _, ok := segmentSnapshot.segment.(segment.PersistedSegment); ok {
-			totalCount += uint64(segmentSnapshot.FullSize())
-			liveCount += uint64(segmentSnapshot.Count())
-		}
-	}
-
-	if totalCount > 0 {
-		return float64(totalCount-liveCount) / float64(totalCount)
-	}
-	return 0
-}
-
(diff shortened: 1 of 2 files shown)


abhinavdangeti (approved):
@sreekanth-cb Assuming you've made sure that you wouldn't need to acquire a lock for s.root.segment access..

mschoch:
> @sreekanth-cb Assuming you've made sure that you wouldn't need to acquire a lock for s.root.segment access..

I think we need to acquire read lock to copy `s.root` (pointer to current snapshot) into our own variable, then we can work with that snapshot without locks after that (contents are immutable).  As another goroutine (introducer) acquires a lock to update `s.root`

mschoch (commented):
Still have my doubts about how we have separate stats for the "root", given that the segments in the root may not even all be on disk.

Also, wondering if Steve would suggest we expose the raw values directly, and let the computation be made on the outside, as opposed to computing the estimate internally.

abhinavdangeti (commented):
Yea, just checked the rest of the code myself, you do need to acquire the s.rootLock.RLock for s.root's access.

sreekanth-cb:
thats right! need locks.

sreekanth-cb:
> Still have my doubts about how we have separate stats for the "root", given that the segments in the root may not even all be on disk.

We already have a stat to track the disk size of root, so thinking is that this would be a natural extension to it.

index/scorch/scorch.go
```diff
@@ -515,6 +515,26 @@ func (s *Scorch) diskFileStats(rootSegmentPaths map[string]struct{}) (uint64,
...
+		}
+	}
+
+	if totalCount > 0 {
+		return float64(totalCount-liveCount) / float64(totalCount)
+	}
+	return 0
+}
+
 func (s *Scorch) rootDiskSegmentsPaths() map[string]struct{} {
```
abhinavdangeti, line 538: As part of this change, would you also move the acquiring of RLock into this function?
I feel it's better readable here than in the calling function like we do in StatsMap() API below - avoids any future mistakes.

sreekanth-cb: hm.. actually I feel the otherway.. the lock should be at the StatsMap() as that assures that the whole op is atomic..
Else it is not guaranteed from a single index snapshot.

So, let me take out the lock and call from parent like this rootDiskSegmentsPaths.
It has its own ugliness... :)

mschoch:
> > Still have my doubts about how we have separate stats for the "root", given that the segments in the root may not even all be on disk.
> 
> We already have a stat to track the disk size of root, so thinking is that this would be a natural extension to it. Or I think that doesn't matter right, as these disk related stats of root captures that aspect of it.

My point is that I think the "root" disk stats are already misleading.  The intent of them, as I understood it, was that it would let a customer know how much space is used just by the current snapshot.  But I feel this stat is actually misleading, as it may under-count the space, because if some of the segments are not yet persisted, that space isn't accounted for.

Then this stat takes that value (possibly under-counted IMHO) and multiplies it to something else.

index/scorch/scorch.go
mschoch: Seems like if we have multiple stats using the current snapshot, we should get the lock once, grab a copy of the pointer to that snapshot, and pass it around, so we minimize lock contention.

sreekanth-cb:
hm... my perspective is that it reflects the current value/present state. 

And if we apply your logic to other stats then those looks incorrect too.

total_compaction_bytes_written (merges/disk writes may be in flights)
number of introductions or every other total_* stats looks like falls in that category to me... :(

index/scorch/scorch.go
sreekanth-cb: hm.. not sure whether I am overlooking something here.. But will a copy of index snapshot works? 
Since if the root changes, then there is a chance that disk files/stats becomes obsolete due to clean ups.. 
So looks like we need to readLock until the disk stats are taken?
(i am on a friday beer)

mschoch:
> hm... my perspective is that it reflects the current value/present state.
> 
> And if we apply your logic to other stats then those looks incorrect too.
> 
> total_compaction_bytes_written (merges/disk writes may be in flights)
> number of introductions or every other total_* stats looks like falls in that category to me... :(

The values are not wrong, they are easy to misinterpret.  A user looking at the "root" stats while indexing is in progress might think wow, that space value is low, stop indexing.  And then every time they stop indexing, the number jumps up.  The reason being that under load, some segments are always in memory and "root" seems smaller.  But as soon as you stop indexing, all the segments make it to disk, and the number jumps.

Anyway, I have nothing better to propose, I'm just sharing the concern for misinterpretation.

index/scorch/scorch.go
abhinavdangeti: > hm.. actually I feel the otherway.. the lock should be at the StatsMap() as that assures that the whole op is atomic..
> Else it is not guaranteed from a single index snapshot.
> 
> So, let me take out the lock and call from parent like this rootDiskSegmentsPaths.
> It has its own ugliness... :)

Ok in that case, per our naming - suffix LOCKED to these function names then?

mschoch: I don't think we should hold the lock, I think we just have to also increment the refcount.  In Bluge, I cleaned up scorch to always go through this helper, because I didn't want anyone touching `s.root` directly:

```
func (s *Writer) currentSnapshot() *Snapshot {
	s.rootLock.RLock()
	var rv *Snapshot
	if s.root != nil {
		rv = s.root
		if rv != nil {
			rv.addRef()
		}
	}
	s.rootLock.RUnlock()
	return rv
}
```

Obviously some of the type names have changed, but you can see the pattern.  Anyone that wants to access the current snapshot, goes through this method, and simply calls Close() when done.

sreekanth-cb: good idea!, utility is already there. Just need to use it!

index/scorch/scorch.go
```diff
@@ -515,9 +515,26 @@ func (s *Scorch) diskFileStats(rootSegmentPaths map[string]struct{}) (uint64,
...
+		}
+	}
+
+	if totalCount > 0 {
+		return float64(totalCount-liveCount) / float64(totalCount)
+	}
+	return 0
+}
+
+func rootDiskSegmentsPaths(indexSnapshot *IndexSnapshot) map[string]struct{} {
```
mschoch, line 535: So, I think this refactoring is good, but not quite complete.

Now that this function works on `*IndexSnapshot`, I think it should be method on that type.  Second, the function now misnamed, as it returns the disk segment paths for whichever snapshot we pass in.  We will just use it to find the disk segment paths of the root, but it works for any snapshot.

So, together this would give us:

```func (i *IndexSnapshot) diskSegmentPaths() map[string]struct{}`

mschoch (changes requested):

index/scorch/scorch.go
sreekanth-cb: absolutely make sense.
 Shall we make reClaimableDocsRatio also of IndexSnapshot  type?

mschoch: Yeah I think that makes sense too.

mschoch (approved):

abhinavdangeti (approved):

PR #1476, opened 2020-10-01
Fix Dropped searcher Test Errors

alrs:
This fixes three dropped `err` variables in `search/searcher`.

### DIFF
--- a/search/searcher/search_conjunction_test.go
+++ b/search/searcher/search_conjunction_test.go
@@ -394,6 +394,9 @@ func testScorchCompositeSearchOptimizationsHelper(
 			}
 
 			next, err = cs.Next(ctx)
+			if err != nil {
+				t.Fatal(err)
+			}
 			i++
 		}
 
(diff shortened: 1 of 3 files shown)


search/searcher/search_conjunction_test.go
```diff
@@ -394,6 +394,9 @@ func testScorchCompositeSearchOptimizationsHelper(
 			}
 
 			next, err = cs.Next(ctx)
+			if err != nil {
+				t.Fatal(err)
+			}
```
mschoch, line 399: This one is incorrect (or at least not following the intended pattern).  The err != nil case will exit the for loop.  However, the `err != nil` case should be immediately after the for loop.

search/searcher/search_term_range_test.go
```diff
@@ -253,6 +253,9 @@ func TestTermRangeSearchTooManyTerms(t *testing.T) {
 		got = append(got, extId)
 		ctx.DocumentMatchPool.Put(next)
 		next, err = searcher.Next(ctx)
+		if err != nil {
+			t.Fatal(err)
+		}
```
mschoch, line 258: This one is not needed, as in the previous case, `err != nil` will immediately exit the loop, and in this case we already have the `err != nil` case afterwards.

mschoch (changes requested):

PR #1478, opened 2020-10-01
Index is always left closed when opening process fails

pavelbazika:
When index opening fails in some later stage, after storage is opened, `nil` and `error` are returned, but the index remains open. There is no way for the user to close the index, because of `nil`.

The fix closes the index when `nil` is going to be returned.


pavelbazika:
Bug found, will open a new one with correct fix.

Issue #1350, opened 2020-03-16
RFC - Bleve v1.0.0 and Go Modules


mschoch:
The Bleve project achieved wide adoption prior to the introduction of Go modules.  As a result, the Bleve team has been careful to not introduce breaking changes.  However, now that Go modules have matured, the community has made it clear that the Bleve project should also adopt them.

While evaluating this change, we determined that this was also the right time to use Go modules to address an important shortcoming in Bleve.  Specifically, Bleve has been limited to a single file format version within a given build.  This is a critical limitation for applications which need to offer backwards compatibility to their end users.  This set of changes leverages Go modules to address this issue.

**NOTE: This proposal will end support for GOPATH builds of bleve and related projects inside the blevesearch organization on github.**

### Support Go Modules

On the surface, this can appear as simple as adding a `go.mod` file to the project.  However, when combined with the other changes below, adopting Go modules really requires you to reevaluate all of your dependencies.

Review the proposed changes to [blevesearch/bleve](https://github.com/blevesearch/bleve/pull/1342)

### Breakout Zap into separate Module

Bleve has an indexing scheme named `scorch`, and scorch was designed to work with different segment implementations.  As of today there is just a single implementation named `zap`.  Zap was also designed to have a version number, so that the file format could evolve over time.  Unfortunately, although the segment abstraction was put in place, the code was physically nested inside the same repository.  This meant that supporting more than one version at a time would either require if/else logic or copy/paste large sections of code, neither of which would be easy to maintain over time.

The much cleaner alternative was to promote `zap` to become its own module.  This allows us to use major versions to delineate incompatible changes with the file format, and use standard git branching and tagging to develop new versions.

The version of `zap` used in Bleve today is `11` and has been moved to it's own repo here: 

[zap v11.0.1](https://github.com/blevesearch/zap/tree/v11.0.1)

### Multiple Versions of Zap supported

A newer version of `zap` which offers significant space savings is introduced as `12` and this code now exists here:

[zap v12.0.1](https://github.com/blevesearch/zap/tree/v12.0.1)

As mentioned in the previous section, by having `zap` as its own module, and by releasing major versions when the file format changes, it becomes straightforward for Bleve's `scorch` index scheme to import multiple `zap` versions into a single build.  This allows Bleve to introduce a new file format `zap v12` which provides significant space savings.  Crucially, an application can start leveraging this new format, while not giving up the ability to read the previous format.

[...]

PR #1482, opened 2020-10-06
add support for zap v15 and update zap version

mschoch:


mschoch:
Failures expected because of circular deps

go.mod
```diff
@@ -8,10 +8,11 @@ require (
...
 	github.com/blevesearch/snowballstem v0.9.0
-	github.com/blevesearch/zap/v11 v11.0.10
-	github.com/blevesearch/zap/v12 v12.0.10
-	github.com/blevesearch/zap/v13 v13.0.2
-	github.com/blevesearch/zap/v14 v14.0.1
+	github.com/blevesearch/zap/v11 v11.0.11
+	github.com/blevesearch/zap/v12 v12.0.11
+	github.com/blevesearch/zap/v13 v13.0.3
+	github.com/blevesearch/zap/v14 v14.0.2
+        github.com/blevesearch/zap/v15 v15.0.0
```
abhinavdangeti, line 15: Indentation here and below looks off.

abhinavdangeti (approved):

go.mod
mschoch: Yeah I copy/pasted and it converted tabs to spaces.  Pushed a commit using tab.

index/scorch/segment_plugin.go
```diff
@@ -23,13 +23,15 @@ import (
...
 	zapv14 "github.com/blevesearch/zap/v14"
+	zapv15 "github.com/blevesearch/zap/v15"
 )
 
 var supportedSegmentPlugins map[string]map[uint32]segment.Plugin
 var defaultSegmentPlugin segment.Plugin
 
 func init() {
 	ResetPlugins()
+	RegisterPlugin(zapv15.Plugin(), false)
```
tmm1, line 34: Am I reading this correctly that zap v11 is still used by default? How does one tell bleve to use v15?

mschoch: Yes, we made a decision earlier that the defaults are essentially part of the API, so changing the file format would be equivalent to a breaking change.

We are [planning bleve 2.x now](https://github.com/blevesearch/bleve/issues/1495), and scorch will become the default index type, and the latest zap will also become the default at that time.

One can force the segment type and version used with the `forceSegmentType` and `forceSegmentVersion` config settings.  They are described here: https://github.com/blevesearch/bleve/issues/1350

Issue #1491, opened 2020-11-05
Configurable bbolt flock timeouts?


tmm1:
Sometimes randomly, for reasons I am still not quite sure of, I will see a deadlock trying to open an index:

```
goroutine 1 [sleep]:
time.Sleep(0x2faf080)
	runtime/time.go:188 +0xc0
github.com/etcd-io/bbolt.flock(0x4000b12800, 0x1, 0x0, 0x180, 0x40000aa210)
	github.com/etcd-io/bbolt@v1.3.3/bolt_unix.go:40 +0x8c
github.com/etcd-io/bbolt.Open(0x4000b719e0, 0x1c, 0x180, 0x22951e0, 0xa, 0x4000b719e0, 0x1c)
	github.com/etcd-io/bbolt@v1.3.3/db.go:223 +0x170
github.com/blevesearch/bleve/index/scorch.(*Scorch).openBolt(0x4000171400, 0x400071b6b0, 0x4000171400)
	github.com/blevesearch/bleve@v0.8.2-0.20191030071327-189ee421f71e/index/scorch/scorch.go:205 +0x424
github.com/blevesearch/bleve/index/scorch.(*Scorch).Open(0x4000171400, 0x6, 0x4000630ae0)
	github.com/blevesearch/bleve@v0.8.2-0.20191030071327-189ee421f71e/index/scorch/scorch.go:160 +0x28
github.com/blevesearch/bleve.openIndexUsing(0x4000b717c0, 0x14, 0x400071b6b0, 0x40006f60e0, 0x0, 0x0)
	github.com/blevesearch/bleve@v0.8.2-0.20191030071327-189ee421f71e/index_impl.go:174 +0x468
github.com/blevesearch/bleve.OpenUsing(...)
	github.com/blevesearch/bleve@v0.8.2-0.20191030071327-189ee421f71e/index.go:294
```

bbolt provides a timeout argument to its `flock` method, but it is unused. Can we add a scorch option to be able to pass this through?

meirwah:
I have seen it too:
```
#	0x44ed5c	time.Sleep+0x12c											/usr/local/go/src/runtime/time.go:105
#	0xef7689	github.com/etcd-io/bbolt.flock+0x99									/home/circleci/.go_workspace/pkg/mod/github.com/etcd-io/bbolt@v0.0.0-20190809044356-35b666109344/bolt_unix.go:40
[...]

PR #1492, opened 2020-11-05
add optional bolt_timeout config option for scorch

tmm1:
cc #1491


mschoch:
We merged #1553 instead which is similar but uses a string config parameter.

PR #1494, opened 2020-11-10
bleve v2.0.0 proposal PR

mschoch:
This PR currently includes the bare minimum set of changes to avoid circular dependencies with zap.

This is accomplished by refactoring the `blevesearch/bleve/index` and `blevesearch/index/scorch/segment` packages out into a separate modules `blevesearch/bleve_index_api` and `blevesearch/scorch_segment_api` respectively.

Further, a new module `blevesearch/zapx` is introduced.  This repository is a duplicate of `blevesearch/zap` up to today, with all major version branches (v11.x, v12.x, v13.x, v14.x and master) updated to no longer depend on bleve.  This solution was chose we intend for the major version number to mostly strongly indicate file format compatibility.  Compatibility with major version of bleve (or lack of such a dependency) is therefore another dimension, and we felt most clearly represented with a different package name. 

Along the way a few other minor changes were required, in all such cases the focus was to keep these as small as possible.


mschoch:
Keep in mind this is a large change, so please review carefully.  I attempted to clean it up as best I could, but in a change this large sometimes things are missed.

mschoch:
Links to new repos:

- https://github.com/blevesearch/bleve_index_api
- https://github.com/blevesearch/scorch_segment_api
- https://github.com/blevesearch/zapx

mschoch:
So, one of the wrinkles you'll see upon closer review is that we have a few places where we type assert that a `index.Document` is really type `*document.Document`.  As implemented in this PR, this is safe for all known index implementations, because both `upsidedown` and `scorch` are still in the bleve repo, and use `document.Document` directly.

However, one of my additional refactors I'd like to consider is moving `upsidedown` out, and this means it can no longer use `document.Document` as that brings circular dependency back.  I'll continue the experiment, but it most likely will mean new methods on the `index.Document` type, and changing those type assertions to go through the interface...

mschoch:
OK, now looks like moving `upsidedown` out is not possible without substantially more refactoring, as it currently uses the registry package to instantiate the key/value store implementations.  That seems like too big a change for this round.

Instead I'll focus on trying to remove any `upsidedown` specific things from the `bleve_index_api` interfaces.

analysis/freq.go
```diff
@@ -123,13 +36,13 @@ func TokenFrequency(tokens TokenStream, arrayPositions []uint64, includeTermVect
...
+				curr.SetFrequency(curr.Frequency() + 1)
 			} else {
-				rv[string(token.Term)] = &TokenFreq{
+				rv[string(token.Term)] = &index.TokenFreq{
 					Term:      token.Term,
-					Locations: []*TokenLocation{&tls[tlNext]},
-					frequency: 1,
+					Locations: []*index.TokenLocation{&tls[tlNext]},
 				}
+				rv[string(token.Term)].SetFrequency(1)
```
sreekanth-cb, line 45: As rv looks like map[string]* and  there is string(byte) conversion during look ups,
can we just access it only once for setting it like on the original code with a fully initialized &index.TokenFreq{}. (saves an extra map look up and the string-[]byte conversion).

analysis/freq.go
```diff
@@ -138,12 +51,12 @@ func TokenFrequency(tokens TokenStream, arrayPositions []uint64, includeTermVect
...
-				curr.frequency++
+				curr.SetFrequency(curr.Frequency() + 1)
 			} else {
-				rv[string(token.Term)] = &TokenFreq{
-					Term:      token.Term,
-					frequency: 1,
+				rv[string(token.Term)] = &index.TokenFreq{
+					Term: token.Term,
 				}
+				rv[string(token.Term)].SetFrequency(1)
```
sreekanth-cb, line 59: same comment as above.

analysis/freq.go
mschoch: I don't think the string is actually converted, because of this optimization in the runtime: https://github.com/golang/go/commit/f5f5a8b6209f84961687d993b93ea0d397f5d5bf

But I'll rewrite and see if you like that better.

index/upsidedown/upsidedown.go
```diff
@@ -616,7 +615,7 @@ func encodeFieldType(f document.Field) byte {
 	return fieldType
```
abhinavdangeti, line 615: This function isn't needed anymore yes?

abhinavdangeti (commented):
Looks mostly OK to me. Couple things though -
- I'm guessing you'll set up release tag(s) for bleve_index_api and scorch_segment_api - think we should keep their go versions at 1.13 as well (rather than 1.15).
- Is there a reason why you included go.sum into this change, I don't remember your reason for not committing it in earlier?

index/upsidedown/upsidedown.go
mschoch: pushed another commit removing it

mschoch:
> I'm guessing you'll set up release tag(s) for bleve_index_api and scorch_segment_api - think we should keep their go versions at 1.13 as well (rather than 1.15).

I expect once these interfaces packages are finalized, they will be tagged v1.0.0. What reason is there for them to be something else?

> Is there a reason why you included go.sum into this change, I don't remember your reason for not committing it in earlier?

Yes this is intentional.  Best practices is to include go.sum.  In the past we did not because our go.mod was committed after manual edits (forward declarations of module versions that did not yet exist).  Therefore everyone's go.sum would change after doing a basic build locally, which meant every PR was going to propose a change to go.sum.  I wished to avoid that.  However, now that that problem has gone away, we should adopt the best practices.

abhinavdangeti (approved):
Fine with me.

sreekanth-cb (approved):

mschoch:
@abhinavdangeti @sreekanth-cb last call before we merge to master, requesting final re-review.

abhinavdangeti (approved):
Looks good to me, rebase.squash.merge away.

PR #1496, opened 2020-11-11
fix analyzer lookup when field name contains dot

mschoch:
At query time, we sometimes attempt to lookup the correct
analyzer for a field.  When doing so, the "." character can
be interpretted as a path separater.  Doing this can result
in lookup failure in cases where the user specified a field
name which also contains a dot.

This fix addresses the case where the path contains a dot,
but the next element does not match a mapping at this level.
Previously this ended all looukp in this mapping, now we
attempt to match the remaining path as a whole, without
splitting on the dot.

This fix is intended to fix most common cases where a user has
given a field a name with a dot.  However, ambigutities
between custom field names containing dots, and actual mapping
paths can still happen, and must be manually avoided.

### DIFF
--- a/mapping/mapping_test.go
+++ b/mapping/mapping_test.go
@@ -1151,3 +1151,23 @@ func TestDefaultAnalyzerInheritance(t *testing.T) {
 		t.Fatalf("Expected analyzer: xyz to be inherited by field, but got: '%v'", analyzer)
 	}
 }
+
+func TestWrongAnalyzerSearchableAs(t *testing.T) {
+	fieldMapping := NewTextFieldMapping()
+	fieldMapping.Name = "geo.accuracy"
+	fieldMapping.Analyzer = "xyz"
+
+	nestedMapping := NewDocumentMapping()
+	nestedMapping.AddFieldMappingsAt("accuracy", fieldMapping)
+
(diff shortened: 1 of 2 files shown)


mschoch:
I did a force push, which I usually avoid, to correct a typo in the commit message.

mapping/mapping_test.go
```diff
@@ -1151,3 +1151,23 @@ func TestDefaultAnalyzerInheritance(t *testing.T) {
...
+
+	docMapping := NewDocumentMapping()
+	docMapping.AddSubDocumentMapping("geo", nestedMapping)
+
+	indexMapping := NewIndexMapping()
+	indexMapping.AddDocumentMapping("brewery", docMapping)
+
+	analyzerName := indexMapping.AnalyzerNameForPath("geo.geo.accuracy")
+	if analyzerName != "xyz" {
+		t.Errorf("expected analyzer name `xyz`, got `%s", analyzerName)
```
abhinavdangeti, line 1171: Missed a trailing back-quote after %s

abhinavdangeti (approved):
Looks good to me!

abhinavdangeti (approved):

PR #1498, opened 2020-11-19
reduce index and segment API surface area

mschoch:
Logic inside segment implementations has been
moved up the stack into scorch, so that fewer
methods are required in the API.

- Iterator() becomes AutomatonIterator(nil, nil, nil)
- RangeIterator(start, end) now computes an end key
  compatible with traditional inclusive start, exclusive
  end keys.  Then becomes:
  AutomatonIterator(nil, start, endExclusive)
- PrefixIterator(prefix) now computes an end key
  and becomes:
  AutomatonIterator(nil, prefix, prefixEnd)

NumericRangeSearcher updated to no longer require
the IndexReaderOnly interface.  This interface
was only used for non-scorch indexes, but only
scorch actually implemented it.  Removing this
also meant we can remove the OnlyIterator method
from the segment dictionary API.

### DIFF
--- a/index/scorch/snapshot_index.go
+++ b/index/scorch/snapshot_index.go
@@ -194,7 +194,11 @@ func (i *IndexSnapshot) FieldDict(field string) (index.FieldDict, error) {
 	}, false)
 }
 
-func exclusiveEndFromInclusiveEnd(inclusiveEnd []byte) []byte {
+// calculateExclusiveEndFromInclusiveEnd produces the next key
+// when sorting using memcmp style comparisons, suitable to
+// use as the end key in a traditional (inclusive, exclusive]
+// start/end range
+func calculateExclusiveEndFromInclusiveEnd(inclusiveEnd []byte) []byte {
 	rv := inclusiveEnd
 	if len(inclusiveEnd) > 0 {
 		rv = make([]byte, len(inclusiveEnd))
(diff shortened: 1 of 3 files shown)


mschoch:
The code computing the range iterator end key originated here: https://github.com/blevesearch/zap/blob/master/dict.go#L149-L158

However, I believe that implementation was incorrect.  In the case where we cannot increment the last byte, we should append 0, not 0xff, the next key is the first longer key, which would be the same key, but with an extra 0 byte.

The code computing the prefix iterator end key originate here: https://github.com/blevesearch/scorch_segment_api/blob/85deb08a9c2d7b648ac5ddef52ed18e96a842ec8/regexp.go#L65-L75 and should be unmodified.

mschoch:
NOTE, if this gets approved, we can remove all these variations of the methods from the API.  At that time I would propose that we rename `AutomatonIterator(automaton, start, end)` to simply `Iterator(automaton, start, end)`

index/scorch/snapshot_index.go
```diff
@@ -190,21 +190,58 @@ func (i *IndexSnapshot) newIndexSnapshotFieldDict(field string,
...
 func (i *IndexSnapshot) FieldDictRange(field string, startTerm []byte,
 	endTerm []byte) (index.FieldDict, error) {
 	return i.newIndexSnapshotFieldDict(field, func(i segment.TermDictionary) segment.DictionaryIterator {
-		return i.RangeIterator(string(startTerm), string(endTerm))
+		endTermExclusive := exclusiveEndFromInclusiveEnd(endTerm)
+		return i.AutomatonIterator(nil, startTerm, endTermExclusive)
 	}, false)
 }
 
+// incrementBytesPrefix produces the first key that does not
```
abhinavdangeti, line 222: incrementBytesPrefix should be exclusiveEndFromPrefix here, although you're incrementing bytes on the prefix :)

Possibly a name change - maybe obtainExclusiveEndKeyForPrefix ?

mschoch: i went with calculate, not obtain, as that felt more correct to me

sreekanth-cb (approved):

abhinavdangeti (approved):
Looks good.

PR #1504, opened 2020-11-26
Improve the effectiveness of rollback points in scorch

sreekanth-cb:
Scorch originally tries to save the most recent `numSnapshotsToKeep`
snapshots for better rollback-ability. But this is turning out to
be ineffective in capturing any reasonable history of the index.

Proposed approach tries to,
-Save the most recent snapshot/epoch.
-Evenly distribute the remaining `numSnapshotsToKeep-1` points at
 regular intervals in the whole set of persisted list of epochs.

Another potential benefit here is that,

In a heavy traffic system with continuous mutations, the list
of persisted epochs tend to be a lengthy one. This happens as
the removeOldData() clean up happens only when a persister
caught up with the current root.
Evenly distributing rollback points within a lengthy
persisted epoch list is a better bet for rollbackability.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -807,6 +807,33 @@ func (s *Scorch) removeOldData() {
 // rollback'ability.
 var NumSnapshotsToKeep = 1
 
+// getProtectedSnapshots always protects the most recent epoch
+// from the persisted list of epochs. After that, it tries to
+// pick up epochs which are located at even intervals from the list.
+// It helps to evenly distribute the rollback restore points in a given
+// range of persisted epochs and thereby increasing the effectiveness
+// rollback points.
+func getProtectedSnapshots(persistedEpochs []uint64,
+	numSnapshotsToKeep int) map[uint64]struct{} {
+	// always save the most recent epoch.
+	protectedEpochs := make(map[uint64]struct{}, numSnapshotsToKeep)
+	protectedEpochs[persistedEpochs[0]] = struct{}{}
+	if numSnapshotsToKeep == 1 {
+		return protectedEpochs
+	}
+	// save the remaining epochs which are at even intervals
+	// in the whole list of persisted epochs.
+	interval := len(persistedEpochs) / numSnapshotsToKeep
+	pos := interval
+
+	for pos < len(persistedEpochs) && len(protectedEpochs) < numSnapshotsToKeep {
+		protectedEpochs[persistedEpochs[pos]] = struct{}{}
+		pos += interval
+	}
+
+	return protectedEpochs
+}
+
 // Removes enough snapshots from the rootBolt so that the
 // s.eligibleForRemoval stays under the NumSnapshotsToKeep policy.
 func (s *Scorch) removeOldBoltSnapshots() (numRemoved int, err error) {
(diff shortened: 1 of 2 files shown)


mschoch:
So, I think I understand the logic here, but I'm not as clear how to reason about it in practice.

First, my understanding is that we always protect the most recent persisted snapshot (no issue here).  Then, we look at the list of persisted snapshots, and try to distributed the `numSnapshotsToKeep` evenly spaced across this list.  And, the reasoning for this makes some sense for me.

But, my concern is this.  What determines how long that list is in the first place?  Will there be 3 items there?  Or 10, or 100?  It seems it would be partly be a function of `numSnapshotsToKeep`, but what else?  Does rate of ingest, size of batches, persister pause settings, or other things affect the length of this list?  My worry is that "how far back the rollback goes" now depends on the length of this list.  And yet, I don't know how to think about that list...

sreekanth-cb:
One consistent behavior observed is that - the list is going to be even in the order of 100s when there are steady mutations keep coming in.  

Lengthy persisted epoch list (deprived old data cleanups) mainly happens due to the fact that the persister is busy doing in-memory merges, persists ops, and the latest root catch-ups. So doesn't get time to clean up the older persisted epochs.

There could many factors contributing to the length of the non-eligibleRemoval list of persisted epochs like the slow queries, trailing merger etc.

`numSnapshotsToKeep` primarily dictates the list length in a steady-state system (in terms of low/no mutations and other background operations like compactions/persist cycles)

sreekanth-cb:
It is a minimum surface area attempt to improve things than the current situation. One thing I noted is that

sreekanth-cb:
> My worry is that "how far back the rollback goes" now depends on the length of this list. And yet, I don't know how to think about that list...

It was always dependent on the length of the list, meaning if the length is shorter than the number of items asked to save, then it is just the shorter list. In other cases, it was just the first `N` items in the list. 
Now rather than picking the most recent consecutive items, we are picking items from even intervals.

mschoch:
> It was always dependent on the length of the list

I disagree.  If `numSnapshotsToKeep` was 3, then "how far back the rollback goes" is never further back than `curr-3`, no matter how long the list was.  Now, if the list is 100, the last snapshot might be something like `curr-66` (0, -33, and -66 were kept approximately).  However, if the list was 1000, the last snapshot might be something like `curr-666` (0, -333 and -666 were kept).  So, "how far back the rollback goes" is now different, depending on the length of the list.

sreekanth-cb:
The list of persisted epochs after update operations on 300K items in travel-sample bucket in local/mac environment is as below. (~30 length)
```
2020/11/27 09:18:52 persistedEpochs epochs2: [6329 6328 6213 5732 5169 4572 4134 3837 3634 3474 3297 3120 2955 2804 2779 2731 2676 2607 2525 2448 2370 2300 2232 2159 2081 1977 1855 1715 1112 443] 
```
And the new logic picked the epochs,
```
stored epochs2: map[2370:{} 3297:{} 6329:{}] 
```
which ought to be more efficient than `latest-3` for rollback ability.

In a highly mutated environment (lengthy list) we leverage the number of snapshots we have for better rollback ability and in a stable environment (persisted list if having shorter lengths like 3-5) it almost turns out to be the same as that of the former approach.

abhinavdangeti (approved):

abhinavdangeti:
Abandoning in favor of https://github.com/blevesearch/bleve/pull/1776

PR #1507, opened 2020-11-30
move the plugin interface into scorch

mschoch:


mschoch:
NOTE: this will fail until we point to versions with all the other changes merged/tagged, but let's you see the complete picture

mschoch:
> NOTE: this will fail until we point to versions with all the other changes merged/tagged, but let's you see the complete picture

Actually I see now that it won't fail, but will in some intermediate states once the segment.Plugin is removed.  Anyway, hopefully you see where this is going.

index/scorch/segment_plugin.go
```diff
@@ -26,25 +28,63 @@ import (
 	zapv15 "github.com/blevesearch/zapx/v15"
 )
 
-var supportedSegmentPlugins map[string]map[uint32]segment.Plugin
-var defaultSegmentPlugin segment.Plugin
+// Plugin represents the essential functions required by a package to plug in
+// it's segment implementation
+type Plugin interface {
```
sreekanth-cb, line 33: Feels like this name `Plugin` becomes a very generic name now, without leaving any clues?
Any thoughts?

sreekanth-cb (commented):
.

index/scorch/segment_plugin.go
mschoch: Originally I was going make it not exported, but we use it as an argument to the `RegisterPlugin` method.  We could change it to `SegmentPlugin`.  But, then should also change the method names to match?

mschoch: Alternatively, if you don't like `Plugin` at all, maybe `SegmentImplementation`?  That ends up kind of long, and `SegmentImpl` feels very much like a Java name...

sreekanth-cb: I liked `SegmentPlugin` though. .

mschoch: OK, I will go with `SegmentPlugin`, do you think the method names should change as well?  

RegisterPlugin -> RegisterSegmentPlugin
ResetPlugins -> ResetSegmentPlugins

sreekanth-cb (approved):

PR #1511, opened 2020-12-02
remove Advanced() method from scorch

mschoch:
after the API is updated, scorch no longer requires
this useless method to satisfy the interface


mschoch:
NOTE: this will fail until the API changes are merged and adopted

sreekanth-cb (approved):

PR #1518, opened 2020-12-08
Introducing indexingOptions for configurable frequency/norm.

sreekanth-cb:
Adding an indexing option to skip freq/norm processing with
indexes for use cases where score relevancy isn't needed
for the search requests.

A special frequency value of zero/0 is used for skipping
the freq/norm.


abhinavdangeti (approved):
Looks OK to me.

Issue #1529, opened 2020-12-29
html escaping (caught mistake)


mschoch:
I was in the process of porting this fix to bluge:

https://github.com/blevesearch/bleve/pull/1465

But, while the conversation indicated I had pushed another commit making an important change, the history shows I seem to have pushed some possibly unrelated commit to this branch.

Reviewing my machine locally, it seems that the fact that I never clean up my local branches shows that I do still have the missing commit here:

```
diff --git a/search/highlight/format/html/html.go b/search/highlight/format/html/html.go
index 259a0379..5bf16ca9 100644
--- a/search/highlight/format/html/html.go
+++ b/search/highlight/format/html/html.go
@@ -60,7 +60,7 @@ func (a *FragmentFormatter) Format(f *highlight.Fragment, orderedTermLocations h
                // start the <mark> tag
                rv += a.before
                // add the term itself
-               rv += string(f.Orig[termLocation.Start:termLocation.End])
+               rv += html.EscapeString(string(f.Orig[termLocation.Start:termLocation.End]))
                // end the <mark> tag
                rv += a.after
                // update current
diff --git a/search/highlight/format/html/html_test.go b/search/highlight/format/html/html_test.go
index 6bb49dfe..0652d477 100644
--- a/search/highlight/format/html/html_test.go
+++ b/search/highlight/format/html/html_test.go
@@ -21,11 +21,13 @@ import (
        "github.com/blevesearch/bleve/search/highlight"
 )
 
-func TestHTMLFragmentFormatter1(t *testing.T) {
+func TestHTMLFragmentFormatter(t *testing.T) {
        tests := []struct {
                fragment *highlight.Fragment
                tlm      search.TermLocationMap
                output   string
+               start string
+               end string
        }{
                {
                        fragment: &highlight.Fragment{
@@ -43,25 +45,9 @@ func TestHTMLFragmentFormatter1(t *testing.T) {
                                },
                        },
                        output: "the <b>quick</b> brown fox",
+                       start: "<b>",
+                       end: "</b>",
                }
-       }
-
-       emHTMLFormatter := NewFragmentFormatter("<b>", "</b>")
-       for _, test := range tests {
-               otl := highlight.OrderTermLocations(test.tlm)
-               result := emHTMLFormatter.Format(test.fragment, otl)
-               if result != test.output {
-                       t.Errorf("expected `%s`, got `%s`", test.output, result)
-               }
-       }
-}
-
[...]

PR #1530, opened 2020-12-29
also escape the matched term

mschoch:
depending on the analyzer used, it is possible for the search
term itself to contain characters which should be escaped
before generating html output. this commit adds escaping
on the term itself, and adds some unit tests as well.


mschoch:
As reported in #1529 this commit was accidentally left out of a previous PR.

sreekanth-cb (approved):

abhinavdangeti (approved):

PR #1533, opened 2021-01-05
Accommodate field names with whitespaces in query string syntax

abhinavdangeti:
+ Update the the yacc rules while parsing query string syntax to
  accommodate a "phrase" as a field name. This will allow for the
  user to use field names that have white spaces in them.

  For example:
    - Index definition has a field "with JSON" searchable as "with JSON"
    - With a match query - {"field": "with JSON", "match": "reserved"}
    - This is previously not supported within the query string syntax,
      but is now possible ..
            +------------------------+
            | "with JSON":"reserved" |
            +------------------------+

+ query_string.y.go is generated using goyacc based on new rules added
  to query_string.y.
+ This addresses https://issues.couchbase.com/browse/MB-41264.


abhinavdangeti:
Ya fair, let me give that a try tomorrow.

mschoch (approved):

PR #1540, opened 2021-01-13
Adding back an exported Analyze(..) API for scorch & upsidedown

abhinavdangeti:
Note that these APIs have different signatures for the two.

### DIFF
--- a/test/tests/sort/searches.json
+++ b/test/tests/sort/searches.json
@@ -170,14 +170,14 @@
 		}
 	},
   {
-    "comment": "sort by age, descending, missing first, _id, ascending",
+    "comment": "sort by age, descending, missing first, id, ascending",
 		"search": {
 			"from": 0,
 			"size": 10,
 			"query": {
 				"match_all":{}
 			},
-      "sort": [{"by":"field","field":"age","missing":"first","desc":true},{"by":"field","field":"_id"}]
+      "sort": [{"by":"field","field":"age","missing":"first","desc":true},{"by":"id","desc":false}]
 		},
 		"result": {
 			"total_hits": 6,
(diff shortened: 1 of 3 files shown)


mschoch (changes requested):
Woah this is wrong.  At best, we add Analyze() back, **just** as a wrapper around the analyze() method.  We should be reintroducing any of these types.  There should be no index/analysis package?  what is going on here?

abhinavdangeti:
Ah shoot, you're right. There's an AnalyzedTokenFrequencies() API in bleve_index_api that I can use. I'm going to overwrite this patch, one sec.

mschoch (commented):
better, but I'd still like a clear explanation of how this is used before we add it back

mschoch:
Also, that test is failing and the name of test doesn't have the "_id" change we put in.  Is this change not based on the 2.x branch?

abhinavdangeti:
> Also, that test is failing and the name of test doesn't have the "_id" change we put in. Is this change not based on the 2.x branch?

It is. Seems like this is breaking on some builds only.

mschoch:
But, I changed the name of the test to include "_id" and the name of the test printed doesn't have it.

mschoch:
Oh I see, it's actually a different test, with a similar name.  We basically have a bunch of these and will just find them as they fail...

test/tests/sort/searches.json
```diff
@@ -170,14 +170,14 @@
 		}
 	},
   {
-    "comment": "sort by age, descending, missing first",
+    "comment": "sort by age, descending, missing first, _id, ascending",
```
sreekanth-cb, line 173: I guess this is for making the test work for both scorch and upsidedown?

mschoch (approved):

PR #1543, opened 2021-01-16
Simplify defer statements

amnonbc:
A small PR, which replaces 
```
	defer func() {
		if cerr := fieldDict.Close(); cerr != nil && err == nil {
			err = cerr
		}
	}()
```

with the obvious `defer fieldDict.Close()`

I am not too keen on the  former version as it is hard to read.
A reader would assume that the if `fieldDict.Close()` returns an error
then this error would be returned by the enclosing function.
But closer examination and experimentation shows that the this has no effect
on the value returned by the closing function (unless `err` is declared as a named result parameter).

See https://play.golang.org/p/NEpYGcw_bko

In scorch, the `Close()` method is a noop which always returns nil, which 
makes the logic here even harder to justify.


mschoch:
The more complicated defer which returns an error does actually.  Here is an updated example, note that `err` being a named return value makes a difference.

https://play.golang.org/p/xzavWdRDHI3

amnonbc:
> The more complicated defer which returns an error does actually work. Here is an updated example, note that `err` being a named return value makes a difference.
> 
> https://play.golang.org/p/xzavWdRDHI3

True. There is even an example to illustrate this in the [Go Language Spec](https://golang.org/ref/spec#Defer_statements)

But `findFuzzyCandidateTerms()` `NewRegexpStringSearcher()`, `NewTermPrefixSearcher()` and `NewTermRangeSearcher()` do not have named return values.

And then Scorch's dictionary Close always returns nil.

Anyway, this was just a "drive-by" commit. 
I'll close it, as I don't want to waste your time.

mschoch:
findFuzzyCandidateTerms does have a named return value:

```
func findFuzzyCandidateTerms(indexReader index.IndexReader, term string,
	fuzziness int, field, prefixTerm string) (rv []string, err error)
```

However, you are right that others do not, and should be updated.  However, I think they should be updated to use a name return value and properly catch and return the error condition.

amnonbc:
I'm closing this, as it is not a massively important change,
it no longer fulfils the title of simplifying defer statements in the code.

PR #1549, opened 2021-01-22
introduce online backup for scorch indexes

mschoch:
approach used is to acquire an index snapshot
and reuse existing code for persisting
snapshot segments and recording snapshot
information into the root bolt, with minimal changes
to record items into a new backup directory

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -429,7 +429,7 @@ func (s *Scorch) persistSnapshotMaybeMerge(snapshot *IndexSnapshot) (
 }
 
 func prepareBoltSnapshot(snapshot *IndexSnapshot, tx *bolt.Tx, path string,
-	segPlugin SegmentPlugin) ([]string, map[uint64]string, error) {
+	segPlugin SegmentPlugin, backupHelper func(string, uint64) (string, error)) ([]string, map[uint64]string, error) {
 	snapshotsBucket, err := tx.CreateBucketIfNotExists(boltSnapshotsBucket)
 	if err != nil {
 		return nil, nil, err
@@ -482,6 +482,12 @@ func prepareBoltSnapshot(snapshot *IndexSnapshot, tx *bolt.Tx, path string,
 		switch seg := segmentSnapshot.segment.(type) {
 		case segment.PersistedSegment:
 			segPath := seg.Path()
+			if backupHelper != nil {
+				segPath, err = backupHelper(segPath, segmentSnapshot.id)
+				if err != nil {
+					return nil, nil, fmt.Errorf("error backing up segment: %v", err)
+				}
+			}
 			filename := strings.TrimPrefix(segPath, path+string(os.PathSeparator))
 			err = snapshotSegmentBucket.Put(boltPathKey, []byte(filename))
 			if err != nil {
@@ -535,7 +541,7 @@ func (s *Scorch) persistSnapshotDirect(snapshot *IndexSnapshot) (err error) {
 		}
 	}()
 
-	filenames, newSegmentPaths, err := prepareBoltSnapshot(snapshot, tx, s.path, s.segPlugin)
+	filenames, newSegmentPaths, err := prepareBoltSnapshot(snapshot, tx, s.path, s.segPlugin, nil)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


mschoch:
CI will fail until we merge the new optional API, and adopt it here.

mschoch:
The test is admittedly weak, as it's so low volume that nothing is in flight, we're just copying the file.  But it is a start.  Something with higher volume trying to get it so that at least one segment might be in-memory still would be nice.  And also it would be nice to assert that the original index isn't blocked by the operation, but these are harder to test.

index/scorch/snapshot_index.go
```diff
@@ -744,6 +748,89 @@ func (i *IndexSnapshot) reClaimableDocsRatio() float64 {
...
+
+	_, _, err = prepareBoltSnapshot(i, tx, path, i.parent.segPlugin, backupHelper)
+	if err != nil {
+		_ = tx.Rollback()
+		return fmt.Errorf("error backing up index snapshot: %v", err)
+	}
+
+	// commit bolt data
+	err = tx.Commit()
+	if err != nil {
```
sreekanth-cb, line 822: one doubt, for such commit errors - what is the best action for the already copied data? leave there as it is? or cleanup to avoid disk usage..

index/scorch/persister.go
```diff
@@ -482,6 +482,12 @@ func prepareBoltSnapshot(snapshot *IndexSnapshot, tx *bolt.Tx, path string,
 		switch seg := segmentSnapshot.segment.(type) {
 		case segment.PersistedSegment:
 			segPath := seg.Path()
+			if backupHelper != nil {
+				segPath, err = backupHelper(segPath, segmentSnapshot.id)
```
sreekanth-cb, line 486: I may not be on the same page here, don't we need to handle the backup for the memory segments too? as those are anyways getting persisted as a part of the prepareBoltSnapshot call.

index/scorch/snapshot_index.go
```diff
@@ -744,6 +748,89 @@ func (i *IndexSnapshot) reClaimableDocsRatio() float64 {
...
+		return 0, err
+	}
+	defer source.Close()
+
+	destination, err := os.Create(dst)
+	if err != nil {
+		return 0, err
+	}
+	defer destination.Close()
+	nBytes, err := io.Copy(destination, source)
```
sreekanth-cb, line 772: nBytes, err := io.Copy(destination, source) Vs return io.Copy(destination, source) ...perhaps..

index/scorch/snapshot_index.go
mschoch: Yeah that is a good question.  I think it might be cleaner for the higher level function to remove the whole dir on error (as opposed to tracking individual file names again for cleanup purposes.  Do you have any opinion on that?

index/scorch/persister.go
mschoch: We do, if you expand the section below, you'll see the type assertion already handles segment.UnpersistedSegment.  And on that path, backup and regular persist are identical (just write segment to the incoming path) so no changes were necessary.

mschoch: We do, if you expand the section below, you'll see the type assertion already handles segment.UnpersistedSegment. And on that path, backup and regular persist are identical (just write segment to the incoming path) so no changes were necessary.

index/scorch/snapshot_index.go
mschoch: ok sure, i honestly copy/pasted this from some Go 101 copy file example.

index/scorch/snapshot_index.go
sreekanth-cb: Hm.. I thought since the Backup API is making the os.MkdirAll(path..) call, it made sense to clean it up when it fails to backup.

mschoch: Right, so when I said "higher level function" I meant the Backup() method at the bleve level (this comment is inside the scorch level).  I agree it should be at the same level as the MkdirAll call.

iredmail:
Dear @mschoch @abhinavdangeti 

Do you have any further update of this online backup feature? :)

abhinavdangeti:
We've already supported online copy-ing of scorch indexes as of v2.1.0.
Ref - https://github.com/blevesearch/bleve/pull/1605

PR #1556, opened 2021-02-12
switch to blevesearch vellum

mschoch:


mschoch:
CI will not work until we release new versions of scorch_segment_api and zapx and adopt them here.

mschoch:
Ah, the CI still works because the argument passed is received as an interface, and both the old and new package satisfy that interface.

abhinavdangeti (approved):

PR #1557, opened 2021-02-17
switch to segment.Bitmap interface

mschoch:


mschoch:
For the full context around this change see: https://github.com/blevesearch/scorch_segment_api/pull/11

PR #1562, opened 2021-02-24
add Go 1.16 to the test matrix

amnonbc:
(We should probably remove 1.13 too).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.13.x, 1.14.x, 1.15.x]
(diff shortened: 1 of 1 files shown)


mschoch (approved):
@abhinavdangeti @sreekanth-cb can you comment on the need to keep testing 1.13 from Couchbase perspective?  I am also in favor of dropping it.

mschoch:
@abhinavdangeti @sreekanth-cb I see that the next release of Couchbase will still use Go 1.13, so I would propose that we leave 1.13 in the testing matrix, until we have the final release of Bleve that it will ship with.  After that time we will drop 1.13 from Bleve master testing matrix, does that work for you?

abhinavdangeti:
@mschoch sounds good to me.

abhinavdangeti (approved):

Issue #1495, opened 2020-11-10
v2.0.0 Proposal (DRAFT)


mschoch:
This is the initial draft proposal for a Bleve v2.0.0 release.

- [x]  Fix circular dependency issue between zap and bleve (https://github.com/blevesearch/bleve/pull/1494)
- [ ]  Review new API modules (these should change slowly going forward, but allow bleve and zap to advance easily)
- [ ]  Officially deprecate the `upsidedown` index (still supported, but time limited)
- [ ]  Move `upsdiedown` index into separate module
- [ ]  Make default Bleve index created with `New()` use Scorch and zap v15

jkassis:
I oppose deprecation of upsidedown. While scorch might win on speed, size, etc., it has no mechanism to force non-blocking (or blocking) commits to persistence *on disk*.

So while it is a win for dedicated/standalone search applications without requirement for strong consisteency and where data lives in scorch and scorch alone... it is a fail for applications that use bleve as a full-text index of data that resides in other data stores.

For example i might need a badger db to retain my data and serve pk queries and a bleve index to serve queries that rely on complex criteria with strong consistency requirements.

I was deep into this in August (https://github.com/blevesearch/bleve/issues/1396) and was determined to fix Scorch. The Scorch code was quite complex to my eye and it wasn't clear to me how to make it work. I soon ran out of time / juice.

Today I'm revisiting this to *replace* scorch with upsidedown backed by badger so that my scorch data and badger data gets committed in the same badger commit.

Leaving this use case behind seems like a big loss for bleve.

jkassis:
If your comment here still applies there is no way to block on persistence to disk... https://github.com/blevesearch/bleve/issues/1396#issuecomment-676578352

jkassis:
Sorry Marty. I just updated my comment after reviewing that thread more thoroughly. The issue was about backups. The `unsafe_batch` option gives me the ability to block until data is on disk. That *is* good, though I have yet to test the impact on performance of this. In any case, i still have an unmet online backup / RAFT compaction requirement.

I have a RAFT the guarantees that calls to my service that write data happen in series. Each call to a write operation modifies badger and bleve.

My scorch option (as i understand it) is call Batch with unsafe_batch == false to get the Bleve data to disk and then call Badger to commit the txn. Here i have to decide how to manage scorch rollback in case the Badger txn fails.

If I'm using upsidedown to commit to badger in the same txn i have Badger to guarantee the mutual consistency of the index and data.

[...]

PR #1568, opened 2021-03-04
Allow bulk command to specify document id instead of using a random id

blamoo:
This change allows me to specify the document id as a string before each line, like this:

Bulk file:
```json
"doc1"
{"name": "Chico Anysio"}
"doc2"
{"name": "Jô Soares"}
"doc3"
{"name": "Paulo Silvino"}
"doc4"
{"name": "Agildo Ribeiro"}
"doc5"
{"name": "José Vasconcellos"}
```

Output:
```
PS D:\Users\Blamoo\Projects\bleve> bleve bulk ../brazilian-comedians.index ../brazilian-comedians.ndjson
Indexing: ../brazilian-comedians.ndjson
Indexing batch (0 docs)...

PS D:\Users\Blamoo\Projects\bleve> bleve query ../brazilian-comedians.index name:*
5 matches, showing 1 through 5, took 998.4µs
    1. doc1 (0.171398)
        name
                Chico Anysio
    2. doc4 (0.171398)
        name
                Agildo Ribeiro
    3. doc3 (0.171398)
        name
                Paulo Silvino
    4. doc5 (0.171398)
        name
                José Vasconcellos
    5. doc2 (0.171398)
        name
                Jô Soares
```

### DIFF
--- a/cmd/bleve/cmd/bulk.go
+++ b/cmd/bleve/cmd/bulk.go
@@ -20,11 +20,13 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 )
 
 var batchSize int
+var named bool
 
 // bulkCmd represents the bulk command
 var bulkCmd = &cobra.Command{
@@ -62,6 +64,25 @@ var bulkCmd = &cobra.Command{
 					batch = idx.NewBatch()
 				}
 
+				docID := ""
+
+				if named {
+					n, _ := r.ReadBytes('\n')
+					if len(n) == 0 {
+						break
+					}
+
+					var err error
+					if parseJSON {
+						err = json.Unmarshal(n, &docID)
+						if err != nil {
+							return fmt.Errorf("error parsing JSON: %v", err)
+						}
+					} else {
+						docID = strings.TrimRight(string(n), "\r\n")
+					}
+				}
+
 				b, _ := r.ReadBytes('\n')
 				if len(b) == 0 {
 					break
@@ -77,7 +98,14 @@ var bulkCmd = &cobra.Command{
 					}
 				}
 
-				docID := randomString(5)
+				if docID == "" {
+					if named {
+						return fmt.Errorf("document name is empty")
+					}
+
+					docID = randomString(5)
+				}
+
 				err = batch.Index(docID, doc)
 				if err != nil {
 					return err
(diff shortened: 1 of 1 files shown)


mschoch:
As of v2.0.0 the  bulk command is deprecated: https://github.com/blevesearch/bleve/issues/1495

As we plan to remove this in a future release, I do think we have any plans to update the functionality.  We recommend users implement this type of logic in their own application.

blamoo:
I see this command more like a prototyping/learning tool than a production-ready tool... But I don't get what you mean by saying it's "extremely limited because they assume and only work with the default index mapping" in the v2 proposal.

These sub-commands allow me to both create complex indexes by passing a complex JSON mapping and index complex JSON documents... Isn't that enough to access most of Bleve's complex features? I mean, it's exactly the same structure used in a lot of tests.

mschoch:
> But I don't get what you mean by saying it's "extremely limited because they assume and only work with the default index mapping" in the v2 proposal.

I think I misspoke here, as you are right, that is not correct.  The `bleve create` command allows you to provide a custom mapping JSON file.  Instead, what I would say is that the limitation is that the mappings you can use are limited to what was compiled in at the time you build the bleve command.  Now, one might say, that's ok, that's a reasonable limiation for prototyping/learning.  But the problem is that we've been working to remove unnecessary dependencies in the project, and the bleve tool is problematic in this area.  Today it uses a long list of forced imports to include a variety of functionality out of the box.

> These sub-commands allow me to both create complex indexes by passing a complex JSON mapping and index complex JSON documents... Isn't that enough to access most of Bleve's complex features?

The problem is that this is highly subjective.  It gives preference to languages that we can easily support without external dependencies, and excludes languages which need them.

To me the decision to remove these commands is clear because in practice I can never actually use these commands when I wan to.  For example, when a user encounters some problem, and they share their index with me, can i run `bleve search` to test it out? No, instead I get an error message that the analyzer named "XYZ" cannot be found.  Because in practice, everyone uses custom analyzers.

I also think that this PR shows why keeping these tools around just for prototyping/learning is problematic.  We have received many PRs like this one that seek to support some new file format, or some new representation of the data.  In the end, there is no standard format that will satisfy everyone, and we have no interest in building out this tool to support all such use cases.  Our belief is that it's much easier for users to build their own 100-line program to do exactly what they want.

> I mean, it's exactly the same structure used in a lot of tests.

I don't follow what this has to do with keep/removing/enhancing the bleve command-line tooling.

blamoo:
> But the problem is that we've been working to remove unnecessary dependencies in the project, and the bleve tool is problematic in this area. Today it uses a long list of forced imports to include a variety of functionality out of the box.

Thinking that way, in my opinion it might even make more sense to remove the CLI tool completely from the main code and move it to a separate project.

Well, in such scenario, this PR does not make sense at all. Thank you anyways.

PR #1570, opened 2021-03-17
introduce new hierarchy token filter

mschoch:
This token filter takes a value representing hierarchical
informtion separated by a delimiter and produces tokens
for each level of the hierarchy.

For example, consider a field with the value "a/b/c".
This token filter is useful when the interpretation of
this is "a" is the most general categorization, "a/b" is more
specific, and finally "a/b/c" is the most specific.  Often
times one may wish to query for documents satisfying "a" or
"a/b".  This token filter generates tokens for these values
so that an exact term query can be used at search time.

Required configuration:

delimter - used for splitting (optional) and joining the parts

Option configuration:

max - the maximum number of parts to join

splitInput - if each input term should be split by the delimiter,
this defaults to true, allowing simpler usage by pairing it with
the single tokenizer

### DIFF
--- a/analysis/token/hierarchy/hierarchy.go
+++ b/analysis/token/hierarchy/hierarchy.go
@@ -1,0 +1,92 @@
+package hierarchy
+
+import (
+	"bytes"
+	"fmt"
+	"math"
+
+	"github.com/blevesearch/bleve/v2/analysis"
+	"github.com/blevesearch/bleve/v2/registry"
+)
+
+const Name = "hierarchy"
+
+type HierarchyFilter struct {
+	maxLevels  int
+	delimiter  []byte
+	splitInput bool
+}
+
+func NewHierarchyFilter(delimiter []byte, maxLevels int, splitInput bool) *HierarchyFilter {
+	return &HierarchyFilter{
+		maxLevels:  maxLevels,
+		delimiter:  delimiter,
+		splitInput: splitInput,
+	}
+}
+
+func (s *HierarchyFilter) Filter(input analysis.TokenStream) analysis.TokenStream {
+	rv := make(analysis.TokenStream, 0, s.maxLevels)
+
+	var soFar [][]byte
+	for _, token := range input {
+		if s.splitInput {
+			parts := bytes.Split(token.Term, s.delimiter)
+			for _, part := range parts {
+				soFar, rv = s.buildToken(rv, soFar, part)
+				if len(soFar) >= s.maxLevels {
+					return rv
+				}
+			}
+		} else {
+			soFar, rv = s.buildToken(rv, soFar, token.Term)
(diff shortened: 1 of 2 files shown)


mschoch:
Ooops, I have some copy/paste errors in the config.

abhinavdangeti (commented):
I think I like most of it.

Only one cause for concern though -> the definition of the "delimiter". It appears to have 2 -

1. The delimiting character to look for within a token.
2. And also, the "glue" character to concatenate tokens.

So, in this situation ..
```
{
    input: analysis.TokenStream{
            &analysis.Token{
                    Term: []byte("a/b"),
            },
            &analysis.Token{
                    Term: []byte("c/d"),
            },
    },
    delimiter: "/",
}
```

In this situation should the user expect -

- a, a/b, a/b/c, a/b/c/d
- a, a/b, c, c/d

I suppose it's the former that's the output with the current algorithm.

Also, not sure I understand the purpose of "splitInput".

mschoch:
I see, so you've made different assumptions than me with respect to the purpose of the component.

I was trying to leave it open ended as to whether or not this filter did any splitting or not (it defaults to on, but can be turned off).  Because of that decision (supporting already tokenized data), I build tokens across the token boundary (a/b/c, and a/b/c/d in your examples).

You seem to have an example in mind where the user did some tokenization, but then didn't want this filter to build tokens across that boundary (c is start of new beginning).  Are you thinking that someone would have multiple hierarchies represented in a single field?  And they might first tokenize (say on space), and then have this filter also split on another character (like /)  So the original field value is something like "Linux/Ubuntu x86/x86_64"  And you would expect to get the following tokens:

Linux
Linux/Ubuntu
x86
x86/x86_64

abhinavdangeti:
> Are you thinking that someone would have multiple hierarchies represented in a single field?

I was yes. If it's this output that the user expects ..
```
Linux
Linux/Ubuntu
x86
x86/x86_64
```
A field modeled this way would work fine with the current code -
```
"field": ["Linux/Ubuntu", "x86/x86_64"]
```
But not this ..
```
"field": "Linux/Ubuntu x86/x86_64"
```

My thought was about - if this is something we should even support?

mschoch:
So, an array of values, and a single value that is tokenized will behave differently.

If you have an array of values, I think my version still works correctly (they are not consecutive tokens in the token stream).

So if you want to support an array, I think this already works correctly.  If you want to support some other token separated single value, then this version does not work.

sreekanth-cb (approved):
I am fine with the current implementation though Abhi's point is valid but may not be the most generic use case.

Not sure hierarchy is the best name for this, too generic? as this is a very specific or lean version(depth only) of hierarchy. But I dont have a better suggestion too.

abhinavdangeti (approved):
I'm OK with this proposal @mschoch .

Issue #1593, opened 2021-05-26
Can I use HTML highlighting with the default (empty string) field?


danielhjacobs:
I'm sure I'm missing something obvious, but I'm trying to highlight my search results and am running into a problem. I can't seem to find a way to show a highlighted result.

My code:

    package main

    import (
        "fmt"`
        "github.com/blevesearch/bleve/v2"
    )

    func main() {
        mapping := bleve.NewIndexMapping()
        index, err = bleve.New("example.bleve", mapping)
        if err != nil {
            fmt.Println(err)
            return
        }
        data := "DNA does stuff..."

        // index some data
        index.Index("id", data)

        // search for some text
        query := bleve.NewMatchQuery("DNA")
        search := bleve.NewSearchRequest(query)
        search.Fields = []string{"*"}
        search.Highlight = bleve.NewHighlightWithStyle("html")
        search.Highlight.AddField("")
        searchResults, err := index.Search(search)
        if err != nil {
            fmt.Println(err)
            return
        }
        if len(searchResults.Hits) > 0 {
            fmt.Println(searchResults.Hits[0].Fragments[""][0])
            fmt.Println(searchResults.Hits[0].Fields[""])
        }
        fmt.Println(searchResults)
    }

Based on the [documentation](http://blevesearch.com/docs/Highlight%20Matches%20in%20Results/) the output I expected was:

    <mark>DNA</mark> does stuff...
    DNA does stuff...
    1 matches, showing 1 through 1, took 136.685µs
        1. id (0.216978)
	
		    <mark>DNA</mark> does stuff...
		    
But there were no marks around any results, as if I hadn't even included the highlight. My result was:

    DNA does stuff...
    DNA does stuff...
    1 matches, showing 1 through 1, took 136.685µs
        1. id (0.216978)
	
		    DNA does stuff...
    
I see the documentation says the field needs to be Stored and IncludeTermVectors, but since [those default to true](https://blevesearch.com/docs/Index-Mapping/#fieldmappings:fac957943db74e6a57605072a04b8308), I didn't think I had to be explicit about that.

mschoch:
So, in general we discourage people from indexing flat data without field names.  In theory an empty field name is valid, and shouldn't cause problems, but it looks like for some reason it does here.

So, if you'd like a quick fix, I'd suggest changing the data you pass to the Index method to look something like:

```
    data := map[string]string {
        "data": "DNA does stuff...",
    }
```

[...]

PR #1594, opened 2021-05-26
fix Complete method to hand empty field name

mschoch:
Previously the implementation of Complete used the zero
value of a string to detect and handle the first time
though a loop iteration.  If however, we decide to support
the empty field name, this change is required to correctly
differentiate between first iteration of the loop and a
field with empty name.

relates to #1593

### DIFF
--- a/search/search.go
+++ b/search/search.go
@@ -265,14 +265,12 @@ func (dm *DocumentMatch) Complete(prealloc []Location) []Location {
 		}
 		prealloc = prealloc[:nlocs]
 
-		var firstField = true
 		var lastField string
 		var tlm TermLocationMap
 		var needsDedupe bool
 
 		for i, ftl := range dm.FieldTermLocations {
-			if firstField || lastField != ftl.Field {
-				firstField = false
+			if i == 0 || lastField != ftl.Field {
 				lastField = ftl.Field
 
 				if dm.Locations == nil {
--- a/search/search.go
+++ b/search/search.go
@@ -265,12 +265,14 @@ func (dm *DocumentMatch) Complete(prealloc []Location) []Location {
 		}
 		prealloc = prealloc[:nlocs]
 
+		var firstField = true
 		var lastField string
 		var tlm TermLocationMap
 		var needsDedupe bool
 
 		for i, ftl := range dm.FieldTermLocations {
-			if lastField != ftl.Field {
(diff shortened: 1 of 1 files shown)


mschoch:
Essentially, the big decision we should make is if we should allow empty string as a valid field name in Bleve.  If so, we need this or something like it (current code assumed field name cannot be empty, and it is used to represent the state prior to first loop iteration).

However, there are other places in the code where we assume field name cannot be empty.  Two known today are:
- Queries have field, for scoping to a single field, empty string is interpreted to mean user doesn't care and we substitute the index mappings default field (usually _all).
- Term vectors optionally store a field (because composite fields need to know which real field it came from), but non-composite fields omit this because the field is implied by the context, this makes it impossible to search in the _all field, find a location and have it refer to data in the empty field name.

There may be others...

search/search.go
```diff
@@ -265,12 +265,14 @@ func (dm *DocumentMatch) Complete(prealloc []Location) []Location {
 		}
 		prealloc = prealloc[:nlocs]
 
+		var firstField = true
 		var lastField string
 		var tlm TermLocationMap
 		var needsDedupe bool
 
 		for i, ftl := range dm.FieldTermLocations {
-			if lastField != ftl.Field {
```
abhinavdangeti, line 273: Hmm, wouldn't have guessed that `lastField != ftl.Field` was simply only a zero value check. Would've guessed that it could do with a redundant field check as well (if array was sorted).

How about just checking for `i == 0` instead of introducing the new variable `firstField` for this?

mschoch: Oh, sorry if I gave that impression.  It is **NOT** simply/only a zero value check.  It is intended to (and functions correctly) process a list sorted by field.  Thus every time a new field is encountered, some additional work must be done.  The check correctly detects that.

However, the author of this code believed that the empty field name was not valid, therefore we will never encounter an empty field name, except on that first run.  The code actually relies on this for correctness, because the initializaiton of `tlm` only happens when cross this "i see a new field" boundary.  So now, if we think the authors assumption was not valid, and we want to allow for empty string, we cannot rely on the zero value to track the state for us.

mschoch: And yes, we can use `i`, good idea.

abhinavdangeti (approved):

PR #1605, opened 2021-06-28
Introduce online Copy for scorch indexes.

sreekanth-cb:
Approach used is to acquire an index snapshot
and reuse existing code for persisting
snapshot segments and recording snapshot
information into the root bolt.

### DIFF
--- a/index_test.go
+++ b/index_test.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"log"
 	"math"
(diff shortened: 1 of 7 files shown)


index_test.go
```diff
@@ -2261,3 +2262,246 @@ func TestIndexMappingDocValuesDynamic(t *testing.T) {
...
+
+	// now create a copy of the index, and repeat assertions on it
+	copyableIndex, ok := idx.(IndexCopyable)
+	if !ok {
+		t.Fatal("index doesn't support copy")
+	}
+	backupIndexPath := createTmpIndexPath(t)
+	defer cleanupTmpIndexPath(t, backupIndexPath)
+
+	getWriter := func(filePath string) io.WriteCloser {
```
mschoch, line 2428: So, I still strongly dislike the design that requires the caller to pass this function to CopyTo().

Let me start by asking, what would a user expect the API to look like?  I feel pretty confident it would be something like:

```
func (b *bleveIndexImplStruct) CopyTo(destinationPath string) error
```

The end-user doesn't expect to pass a function that creates writers inside that directory, we should do that part on their behalf.

What if all of `getWriter()` here were internal to bleve.  And then when the user calls CopyTo(path), we internally use this getWriter implementation?  It seems like the only logic inside getWriter has to do with internal implemenetation choices made by bleve about file/dir naming.

index_test.go
```diff
@@ -2261,3 +2262,246 @@ func TestIndexMappingDocValuesDynamic(t *testing.T) {
...
+		var path string
+		if strings.Contains(filePath, "store") ||
```
mschoch, line 2430: This seems a little bit brittle, for example, what if the destination of the copy contains the word "store", then all attempts to getWriter will mistakenly use this logic.  It would seem more correct to check for suffice "/store".

mschoch (changes requested):
So, I'm fine with the core logic, I just think the end-user API should be more user-friendly, and it seems like that shouldn't be hard to change by rearranging things.

index_test.go
sreekanth-cb: @mschoch, I completely agree that the current API is CB-specific and can't be used by other users due to its ugliness.

The reason I ended up here is that - 

If it accepts a path, then cbft has to copy/duplicate all the disk segment files into a temp location. And then perform a stream write.  Wanted to save the additional disk requirements as well as the extra stream write/copy. 
With the current api, we can directly write to stream without demanding additional disk resources. 

Not sure whether we can achieve that if we accept a CopyTo(destinationPath string)?

mschoch: OK, so I'm going to try to make it work, because I think that getting the API right is the most important thing.

mschoch: @sreekanth-cb why can't this work?  https://play.golang.org/p/vqwsUB7h487

sreekanth-cb: I see what you meant now. 
CBFT can directly call the scorch level implementation which takes a getWriter and bleve users can invoke the Index level API which takes a path.
Sounds good! 👍

mschoch: @sreekanth-cb I'm still not clear what cbft does differently, such that it needs to do that?  Can you link me to the right part of the cbft review?  I did try to find, but did not...

sreekanth-cb: hm,  Looks like I am circling around the same problems here. 

The above approach won't help as the index impl's meta data is associated with the bleve.Index implementation and my use case can't get that copied by directly calling any APIs from the index implementations like scorch which takes a getWriter() callback.

Then the only way would be to directly call index level CopyTo(path) which won't solve the additional disk space requirements. Let's sync on this in a call.

sreekanth-cb:
@mschoch , changed the signature to accept a path here.

mschoch (approved):
Looks ok, but my review of bleve_index_api includes comments to move one thing into this PR...

sreekanth-cb:
@mschoch, moved the FileSystemDirectory here.

abhinavdangeti:
Once the bleve_index_api and zapx changes are merged - we'll need to set up new versions and update the go.mod here absorbing the right version of the two before committing this.

abhinavdangeti:
@sreekanth-cb Rebasing your change on master branch (which now includes https://github.com/blevesearch/bleve/pull/1612) should allow for the checks to pass.

PR #1608, opened 2021-07-21
MB-47476: JSON marshalers cannot work on `0xff` (HighTerm)

abhinavdangeti:
Using unicode/utf8's MaxRune as the `HighTerm` for sort while
missing values.

### DIFF
--- a/search/sort.go
+++ b/search/sort.go
@@ -27,7 +27,7 @@ import (
 	"github.com/blevesearch/bleve/v2/numeric"
 )
 
-var HighTerm = string(utf8.MaxRune)
+var HighTerm = strings.Repeat(string(utf8.MaxRune), 3)
 var LowTerm = string([]byte{0x00})
 
 type SearchSort interface {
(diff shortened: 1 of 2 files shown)


search/sort.go
```diff
@@ -21,12 +21,13 @@ import (
...
 	"sort"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/blevesearch/bleve/v2/geo"
 	"github.com/blevesearch/bleve/v2/numeric"
 )
 
-var HighTerm = strings.Repeat(string([]byte{0xff}), 10)
+var HighTerm = string(utf8.MaxRune)
```
mschoch, line 30: one thought is that we repeat the max rune 3x, that way get 12 bytes (approximately as long as the byte sequence we had before)

abhinavdangeti: Ok, sounds good.

sreekanth-cb (approved):

Issue #1609, opened 2021-07-22
Close of closed channel in mergerLoop


remLse:
Hello there.
The tests for my code sometimes fail due to something I believe to be a bug in bleve. It's non-deterministic, I get the bug about 10% of the times I run my tests. I was using v2.0.1 and upgraded to v2.0.6, which did not help. Here is the error message:
```
panic: close of closed channel

goroutine 453216 [running]:
github.com/blevesearch/bleve/v2/index/scorch.(*closeChWrapper).close(0xc000320c00)
	/home/cbc/go/pkg/mod/github.com/blevesearch/bleve/v2@v2.0.6/index/scorch/merge.go:228 +0x9b
github.com/blevesearch/bleve/v2/index/scorch.(*Scorch).planMergeAtSnapshot(0xc00005b400, 0xde4e10, 0xc0001a6010, 0xc0000a0000, 0xc00027a500, 0x0, 0x0)
	/home/cbc/go/pkg/mod/github.com/blevesearch/bleve/v2@v2.0.6/index/scorch/merge.go:405 +0x208a
github.com/blevesearch/bleve/v2/index/scorch.(*Scorch).mergerLoop(0xc00005b400)
	/home/cbc/go/pkg/mod/github.com/blevesearch/bleve/v2@v2.0.6/index/scorch/merge.go:68 +0x8b7
created by github.com/blevesearch/bleve/v2/index/scorch.(*Scorch).Open
	/home/cbc/go/pkg/mod/github.com/blevesearch/bleve/v2@v2.0.6/index/scorch/scorch.go:186 +0x168
```
Here are a few leads I gathered:
- The relevant channel is created at bleve/v2/index/scorch.go:272.
- The channel is passed to `s.segPlugin.Merge()` in the same file on line 322. If the defaultSegmentPlugin is used (which is the case for me), this close channel is passed to zapx/v15/merge.go:39 and passed along a few functions from there, but zapx does not seem to close the channel itself. So I don't think the bug is there.
- `closeChWrapper.close()` and `closeChWrapper.listen()` are each called exactly once, namely on lines 273 and 275.

[...]

PR #1613, opened 2021-08-05
Address panic on closing a previously closed channel

abhinavdangeti:
+ Fixes: https://github.com/blevesearch/bleve/issues/1609

  panic: close of closed channel

    goroutine 68 [running]:
    github.com/blevesearch/bleve/v2/index/scorch.(*closeChWrapper).close(0xc000346b80)
            /Users/runner/work/bleve/bleve/index/scorch/merge.go:228 +0x99
    github.com/blevesearch/bleve/v2/index/scorch.(*closeChWrapper).listen(0xc000346b80)
            /Users/runner/work/bleve/bleve/index/scorch/merge.go:235 +0x1cd
    created by github.com/blevesearch/bleve/v2/index/scorch.(*Scorch).planMergeAtSnapshot
            /Users/runner/work/bleve/bleve/index/scorch/merge.go:275 +0x4ba

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -209,32 +209,32 @@ func (s *Scorch) parseMergePlannerOptions() (*mergeplan.MergePlanOptions,
 }
 
 type closeChWrapper struct {
-	ch1     chan struct{}
-	ctx     context.Context
-	closeCh chan struct{}
+	ch1      chan struct{}
+	ctx      context.Context
+	closeCh  chan struct{}
+	cancelCh chan struct{}
 }
 
 func newCloseChWrapper(ch1 chan struct{},
 	ctx context.Context) *closeChWrapper {
-	return &closeChWrapper{ch1: ch1,
-		ctx:     ctx,
-		closeCh: make(chan struct{})}
+	return &closeChWrapper{
+		ch1:      ch1,
+		ctx:      ctx,
+		closeCh:  make(chan struct{}),
+		cancelCh: make(chan struct{}),
+	}
 }
 
 func (w *closeChWrapper) close() {
-	select {
-	case <-w.closeCh:
-	default:
-		close(w.closeCh)
-	}
+	close(w.closeCh)
 }
 
 func (w *closeChWrapper) listen() {
 	select {
 	case <-w.ch1:
-		w.close()
+		close(w.cancelCh)
 	case <-w.ctx.Done():
-		w.close()
+		close(w.cancelCh)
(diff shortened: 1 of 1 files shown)


sreekanth-cb (changes requested):
@abhinavdangeti , I don't understand the fix here as I see certain functionality getting removed.

For eg: now there is no way to cancel a force merge operation as the listen() function does nothing other than just waiting on certain events?

Also, note that the issue wasn't happening with a concurrent force merge and scorch close operation. 

Let's sync on later in a call. (maybe there is something in your fix which I am overlooking)

abhinavdangeti:
Ah I see. Ok so, will need to edit this fix. The panic occurs whenever the index is closed when force merge is in flight or if the operation is canceled - and then the defer cw.close() runs.

mschoch (changes requested):
I recommend you fix the design instead.

abhinavdangeti:
> I recommend you fix the design instead.

Ya but sreekanth pointed out that I didn't take into account that there needs to be a way to cancel a running operation. So we'll need to support closing of this channel from a couple of places.

mschoch:
I need to review how it's used further, but this is more along the lines of what I was thinking...

mschoch (commented):
OK, I thought adding another channel was going to be the solution, but having read through things more, I'm not not sure this is right either.  Maybe we'll have to meet on Monday...

mschoch:
Latest thought, why is this needed?

```
defer cw.close()
```

abhinavdangeti:
```
defer cw.close()
```

The defer is needed to close the listen() go routine.

mschoch:
Got it, so now I think I understand your solution.  Allow me to paraphrase and see if you agree:

The problem was that we shared the closeCh with 2 purposes, one to serve as indicator downstream to stop early (triggered by either ch1 or the context doneness) and also with ending the listen goroutine.

This current PR at the time I'm writing this, splits these into two separate channels.  This allows the outer defer-close to cleanup the listen goroutine, independently of channel we passed downstream as a signal.

@abhinavdangeti  Does the align with your thinking?

abhinavdangeti:
@mschoch That's accurate, yes.

mschoch (approved):

mschoch:
Approved, but would like Sreekanth's thoughts as well.  To me this is a better solution than refcouting our close calls, but I suppose it could be debatable...

sreekanth-cb (approved):
Looks good.

Issue #1623, opened 2021-09-14
Cannot search against indexes in folders without write access


ckollett:
If a Bleve index is in a folder that the current user does not have write access to, searching fails when you attempt to open the index, even with the read only option selected as follows:
`index, err := bleve.OpenUsing(pathToIndex, map[string]interface{}{"read_only": true})`

The underlying issue seems to be that, by default, BoltDB will always attempt to open files using CREATE permissions. Here is the code from BoltDB that does this:
`db.openFile(path, flag|os.O_CREATE, mode)`

Given that BoltDB allows calling code to override the default function used to open the bolt file, it is possible to work around this issue in Bleve by creating a function that will open the bolt file without requiring CREATE permission.

For more information, please see [my post to the Bleve Google Group](https://groups.google.com/g/bleve/c/b89GwDqlL4I).

abhinavdangeti:
So `openFile(path, os.O_RDONLY|os.O_CREATE, mode)` should fail with "_permission denied_" only if the file at the path you are attempting to access does not exist.

So the bleve.Index you're attempting to access hasn't been created yet?
 
Would you share the following ..

- the path you're passing in for the index
- the permissions of the directories in your path (note that they should at least be 0500) 
- exact error message you see

ckollett:
Thanks for your response.

I am working on Windows, and can reproduce this behavior using either a UNC path or a mapped drive.

The index that I am attempting to access exists in a folder with permissions rwxr-xr-x (755). The root.bolt file itself has permissions rw-r--r-- (644). I am not the owner of these files so I cannot write to the folder or modify the file.

The exact error message is:
```
error(*io/fs.PathError) *{Op: "open", Path: "\\\\<path to index>\\bleve_index\\store\\root.bolt", Err: error(syscall.Errno) ERROR_ACCESS_DENIED (5)}
```

I can easily reproduce this behavior with the following code. When I run this the first code block successfully opens my root.bolt file, whereas the second block will throw an error.
```
// First try opening the file with read-only permissions
_, err1 := os.OpenFile(filePath, os.O_RDONLY, 0600)
if err1 == nil {
    fmt.Println("Successfully opened file " + filePath)
} else {
    fmt.Println("Error opening file " + filePath)
}
```
```
// Now try opening with read only and create permissions
_, err2 := os.OpenFile(filePath, os.O_RDONLY|os.O_CREATE, 0600)
if err2 == nil {
    fmt.Println("Successfully opened file " + filePath)
[...]

PR #1624, opened 2021-09-14
In RDONLY mode, use O_CREATE for bolt.Open(..) only when file doesn't exist

abhinavdangeti:
+ By default, etcd-io/bbolt appends an O_CREATE flag to the flags passed in
  no matter which mode the user attempts to open the db ..
    https://github.com/etcd-io/bbolt/blob/v1.3.5/db.go#L197...L213
+ This appears to be bolt's design decision where they let you create a
  file that is only writeable by the process if it doesn't exist - even
  when it is OPEN-ed in READ-ONLY mode.
+ However, using an O_CREATE on a file that's owned by a different user
  but READ-able by all users, seems to be causing a permissions error.
    - While I cannot seem to reproduce this issue on osX, this seems to
      manifest on windows which complains ERROR_ACCESS_DENIED.

+ The proposal here is to overwrite Bolt.Options' OpenFile with a custom
  implementation that appends the O_CREATE file only if the file does
  not exist - sticking to bolt's design decision.
+ Should fix https://github.com/blevesearch/bleve/issues/1623

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -212,6 +212,15 @@ func (s *Scorch) openBolt() error {
 	var rootBoltOpt = *bolt.DefaultOptions
 	if s.readOnly {
 		rootBoltOpt.ReadOnly = true
+		rootBoltOpt.OpenFile = func(path string, flag int, mode os.FileMode) (*os.File, error) {
+			// Bolt appends an O_CREATE flag regardless.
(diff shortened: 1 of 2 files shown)


ckollett:
Hi, thanks again for taking this on. Is there anything I can do to help nudge this pull request along? I can contribute a test, for example.

sreekanth-cb (approved):

PR #1652, opened 2022-02-22
Fix for potential file handle leaks in merger.

sreekanth-cb:
There is a potential file handles leak in an update heavy use case like below,

The new segment files that are created during a merge cycle
are protected from accidental clean up by the concurrent persister
routine over a removal eligibility status set over here -
https://github.com/blevesearch/bleve/blob/master/index/scorch/merge.go#L324

And these merger introduced files are getting removal eligibility
status reset by the persister here -
https://github.com/blevesearch/bleve/blob/master/index/scorch/persister.go#L665

Note that not every epoch/root instance undergoes the processing of the persister
and hence such snapshots that escaped the persister reset process (that
would have flipped the segment files to eligible for removal) would contain segment
files ineligible for removal.

Now, if a new segment introduction comes that would completely invalidate
the current ineligible for removal segment contents, then the new index snapshot
won't be having the current segment without any live data.
ref - https://github.com/blevesearch/bleve/blob/master/index/scorch/introducer.go#L162

But such segments will eternally linger around in the `ineligibleForRemoval` slice of
scorch resulting in segment file handle leaks like in this case.

This change attempts to address this by flipping the removal ineligibility
status of the newly created merged file right after the segment introduction.

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -16,6 +16,7 @@ package scorch
 
 import (
 	"fmt"
+	"path/filepath"
 	"sync/atomic"
 
 	"github.com/RoaringBitmap/roaring"
@@ -164,6 +165,11 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 			root.segment[i].segment.AddRef()
 			newSnapshot.offsets = append(newSnapshot.offsets, running)
 			running += newss.segment.Count()
+		} else {
+			if seg, ok := newss.segment.(segment.PersistedSegment); ok {
+				filename := filepath.Base(seg.Path())
+				s.unmarkIneligibleForRemoval(filename)
+			}
 		}
 
 		if isMemorySegment(root.segment[i]) {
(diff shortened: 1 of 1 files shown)


abhinavdangeti:
> Note that not every epoch/root instance undergoes the processing of the persister
and hence such snapshots that escaped the persister reset process that
would have flipped the segment files to eligible for removal.

@sreekanth-cb by this do you mean the root snapshot that is updated by the introducer?

sreekanth-cb:
> > Note that not every epoch/root instance undergoes the processing of the persister
> > and hence such snapshots that escaped the persister reset process that
> > would have flipped the segment files to eligible for removal.
> 
> @sreekanth-cb by this do you mean the root snapshot that is updated by the introducer?

All root snapshots are updated by the introducer, but only the introducer sees every root.
Persister don't necessarily operate on every root. It just picks up the current root for its persist cycle. 
So, once the merger introduced files gets thrown out of the new root by the introducer due to outdated content, then the persister won't get a chance to unflip it ever.

abhinavdangeti (approved):

Issue #1653, opened 2022-02-22
`TermFacets.Terms()` can now panic


bmhatfield:
Prior to this PR: https://github.com/blevesearch/bleve/pull/1404, a user interacting with `search.FacetResult` was safe to range over `Terms`, because Go will do the right thing when ranging over a nil slice. As of 1404, the interface changed to require a call into `.Terms()`, which does not have a nil guard. If `FacetResult.Terms` is nil, this will result in a panic. It's possible to guard this from calling code, but should not be necessary.

Note, also: 1404 was a backwards incompatible change which required code updates to address in calling code.

abhinavdangeti:
@bmhatfield yes we understand it's a breaking change - reason why the last release was a major version change. I've advertised it here - https://github.com/blevesearch/bleve/releases/tag/v2.3.0

Placed a nil guard within the Terms() API with this - https://github.com/blevesearch/bleve/pull/1654

PR #1654, opened 2022-02-22
Place a nil guard within TermFacets' Terms() API

abhinavdangeti:
For https://github.com/blevesearch/bleve/issues/1653

### DIFF
--- a/search/facets_builder.go
+++ b/search/facets_builder.go
@@ -127,6 +127,9 @@ type TermFacets struct {
 }
 
 func (tf *TermFacets) Terms() []*TermFacet {
+	if tf == nil {
+		return []*TermFacet{}
+	}
 	return tf.termFacets
 }
 


bmhatfield (approved):
I'm not an official approver but this was the exact PR I was going to put up before I got distracted with some things on my side.

search/facets_builder.go
```diff
@@ -127,6 +127,9 @@ type TermFacets struct {
 }
 
 func (tf *TermFacets) Terms() []*TermFacet {
+	if tf == nil {
+		return []*TermFacet{}
```
sreekanth-cb, line 131: Shouldn't  the`tf` instance needs to be initialized correctly and then the default return would send back the right tf.termFacets slice?

sreekanth-cb (commented):
.

search/facets_builder.go
abhinavdangeti: Oh it's initialized alright - but only when available. This is to handle the case when there aren't any term facets available. As pointed out https://github.com/blevesearch/bleve/issues/1653, previously this didn't need a nil guard - because TermFacets within FacetResult wasn't a pointer but now it is.

sreekanth-cb (approved):

PR #1666, opened 2022-03-18
Adding configurable default threshold for field TFR cache

### LINKED ISSUE #1243: allocTermFieldReaderDicts leaks PostingsIterator
We've noticed a memory leak, but it never was much an issue as those machines are in a "chaos monkey" type cluster where they are restarted on period basis. The leak is getting worse as we have more usage.

We are doing parallel searches on a few thousands indexes (about 10,000 moderate sized). Those are scorch indexes. We have many small indexes, because we create them with realtime data being indexed and where the new index has to be shared in a cluster, so we cannot have one very large index.

We think we have root caused the leak to be when we have queries with AND / OR and each index keeps `PostingIterator` objects in a pre-allocation pool (I guess for performance optimization), but this pool grows, instead of coming to a steady state / size.

We have run `pprof` and the allocation occurs in the `zap/posting.go` at line 225

Is there a way to cap the size or the pool in `allocTermFieldReaderDicts` , if that's indeed the root cause?

-----

```
github.com/blevesearch/bleve/index/scorch/segment/zap.(*PostingsList).iterator+0x701    /bleve/index/scorch/segment/zap/posting.go:225
github.com/blevesearch/bleve/index/scorch/segment/zap.(*PostingsList).Iterator+0x7d    /bleve/index/scorch/segment/zap/posting.go:153
github.com/blevesearch/bleve/index/scorch.(*IndexSnapshot).TermFieldReader+0x2f1    /bleve/index/scorch/snapshot_index.go:474
github.com/blevesearch/bleve/search/searcher.NewTermSearcherBytes+0xb7            /bleve/search/searcher/search_term.go:46
github.com/blevesearch/bleve/search/searcher.NewTermSearcher+0x11c            /bleve/search/searcher/search_term.go:41
github.com/blevesearch/bleve/search/query.(*TermQuery).Searcher+0x6d            /bleve/search/query/term.go:60
github.com/blevesearch/bleve/search/query.(*ConjunctionQuery).Searcher+0x149        /bleve/search/query/conjunction.go:58
github.com/XYZ/search.RunSingleIndexQuery+0x143                    /search/streamquery.go:192
```

metonymic-smokey:
Signed-off-by: Aditi Ahuja <aditi.ahuja@couchbase.com>

Fixes #1243 
This PR adds a parameter for the maximum field TFR length of an index. This is used to check for the length of the field TFRs of the index when recycling.
This can be overwritten by the user at the library level by modifying the exported variable or the index level by modifying kvconfig.

### DIFF
--- a/index/scorch/snapshot_index.go
+++ b/index/scorch/snapshot_index.go
@@ -51,6 +51,15 @@ type asynchSegmentResult struct {
 
 var reflectStaticSizeIndexSnapshot int
 
+// DefaultFieldTFRCacheThreshold limits the number of TermFieldReaders(TFR) for
+// a field in an index snapshot. Without this limit, when recycling TFRs, it is
+// possible that a very large number of TFRs may be added to the recycle
+// cache, which could eventually lead to significant memory consumption.
+// This threshold can be overwritten by users at the library level by changing the
+// exported variable, or at the index level by setting the FieldTFRCacheThreshold
+// in the kvConfig.
+var DefaultFieldTFRCacheThreshold uint64 = 10
+
 func init() {
 	var is interface{} = IndexSnapshot{}
 	reflectStaticSizeIndexSnapshot = int(reflect.TypeOf(is).Size())
@@ -558,6 +567,15 @@ func (i *IndexSnapshot) allocTermFieldReaderDicts(field string) (tfr *IndexSnaps
 	}
 }
 
+func (i *IndexSnapshot) getFieldTFRCacheThreshold() uint64 {
+	if i.parent.config != nil {
+		if _, ok := i.parent.config["FieldTFRCacheThreshold"]; ok {
+			return i.parent.config["FieldTFRCacheThreshold"].(uint64)
+		}
+	}
+	return DefaultFieldTFRCacheThreshold
+}
+
 func (i *IndexSnapshot) recycleTermFieldReader(tfr *IndexSnapshotTermFieldReader) {
 	if !tfr.recycle {
 		// Do not recycle an optimized unadorned term field reader (used for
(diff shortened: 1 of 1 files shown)


sreekanth-cb (approved):
Look OK to me.

index/scorch/snapshot_index.go
```diff
@@ -51,6 +51,15 @@ type asynchSegmentResult struct {
...
 var reflectStaticSizeIndexSnapshot int
 
+// DefaultFieldTFRCacheThreshold limits the number of TermFieldReaders(TFR) for
+// a field in an index snapshot. Without this limit, when recycling TFRs, it is
+// possible that a very large number of TFRs may be added to the recycle
+// cache, which could eventually lead to significant memory consumption.
+// This threshold can be overwritten by users at the library level by changing the
+// exported variable, or at the index level by setting the FieldTFRCacheThreshold
+// in the kvConfig.
+var DefaultFieldTFRCacheThreshold uint64 = 10
```
abhinavdangeti, line 61: There a reason why you chose 10?

abhinavdangeti (commented):
Curious on why the default of 10, looks fine otherwise.

abhinavdangeti (approved):

index/scorch/snapshot_index.go
metonymic-smokey: Not really, I just picked it arbitrarily 😬 
I'll consider choosing another value with some deliberation, thanks!

sreekanth-cb: I don't think there is a universal perfect number here. And once the applications realize such a magic number that works for them through trials then they could configure it either index level or library level depending on their use case.
Have the same thoughts for cbft too.

Thejas-bhat (approved):

PR #1669, opened 2022-03-25
Adding GeoJSON spatial capability

sreekanth-cb:
This change introduces the capability to index and
query various spatial shapes specified by the
GeoJSON standard.
ref - https://datatracker.ietf.org/doc/html/rfc7946

For convenience, it also supports two additional shapes
like circle and envelope(bounded rectangle).

Changes included here are,

-A field type "geoshape" for specifying any of the
 geojson shapes/circle/envelope.
-A new geoshape query for performing geo queries
 across any of the geoJSON shapes.
-Three main geo query operations between shapes are,

"intersects" - finding all the documents that
 intersect with the query shape.

"contains" - finding all the documents that
 contain the query shape.

"within" - finding all the documents that reside
within the query shape.


geo/geojson_shapes_impl.go
```diff
@@ -0,0 +1,1817 @@
...
+
+		var typ string
+
+		if val, ok := t["type"]; ok {
+			typ = strings.ToLower(val.(string))
+		} else {
+			continue
+		}
+
+		if typ == PointType {
```
abhinavdangeti, line 433: A switch-case might look cleaner here.

abhinavdangeti (commented):
Looks generally OK to me - No red flags.

I do however want to take another pass at geo/geo_s2_util.go on all the contains/intersect logic. @sreekanth-cb if you think a walk-through of that part would help me understand faster, I'd appreciate it.

geo/geojson_shapes_impl.go
```diff
@@ -0,0 +1,1805 @@
...
+
+			for _, s2pgn := range s2pgns {
+				if s2pgn.Contains(s2p2) {
+					polygonsWithIn[pgnIndex] = struct{}{}
+					continue nextPolygon
+				}
+			}
+		}
+
+		rv := len(p2.Vertices) == len(polygonsWithIn)
```
sreekanth-cb, line 1210: cut rv ?

abhinavdangeti (approved):

abhinavdangeti:
I'll recommend a squash and merge to generate 1 commit for easy back porting :)

PR #1683, opened 2022-05-18
Add search highlight formatter: plain.

iredmail:
This new formatter `plain` behaves doesn't do any escape (like `html` formatter), it simply adds highlight tags around matched terms. It's very useful in web development when you do escape in HTML template engine.

### DIFF
--- a/search/highlight/format/plain/plain.go
+++ b/search/highlight/format/plain/plain.go
@@ -1,0 +1,89 @@
+//  Copyright (c) 2014 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package plain
+
+import (
+	"github.com/blevesearch/bleve/v2/registry"
+	"github.com/blevesearch/bleve/v2/search/highlight"
+)
+
+const Name = "plain"
+
+const defaultPlainHighlightBefore = "markbefore"
+const defaultPlainHighlightAfter = "markafter"
+
+type FragmentFormatter struct {
+	before string
+	after  string
+}
+
+func NewFragmentFormatter(before, after string) *FragmentFormatter {
+	return &FragmentFormatter{
+		before: before,
+		after:  after,
+	}
+}
+
+func (a *FragmentFormatter) Format(f *highlight.Fragment, orderedTermLocations highlight.TermLocations) string {
+	rv := ""
+	curr := f.Start
+	for _, termLocation := range orderedTermLocations {
+		if termLocation == nil {
+			continue
+		}
+		// make sure the array positions match
+		if !termLocation.ArrayPositions.Equals(f.ArrayPositions) {
+			continue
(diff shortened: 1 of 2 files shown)


iredmail:
Any interest? @sreekanth-cb @abhinavdangeti

iredmail:
Any interest? @sreekanth-cb @abhinavdangeti

abhinavdangeti:
@iredmail I don't quite see the need for this - have you seen the "simple" highlighter.
I'd like to hear why you'd prefer this over simple.

iredmail:
Do i understand correctly that `simple` is highlighter, not formatter, and it doesn't add tag/text like `defaultHTMLHighlightBefore` and `defaultHTMLHighlightAfter` in the `html` formatter?

iredmail:
Thanks for the reply.

We're working on a web application and display bleve search result on a web page with [pongo2 template engine](https://github.com/flosch/pongo2), and of course we do html escape in pongo2 template files. If we use the default `html` formatter, it will lead to chaos in our (pongo2) template files because we have to remember explicitly which ones are escaped by bleve `html` formatter, and not double escape in pongo2 template. This generates additional mental burden for a simple work. So we prefer bleve to return unescaped plain text and always escape them with pongo2.

abhinavdangeti:
Like I mentioned in my earlier comment - you will be able to overwrite the values used by the before/after tags by setting them in the config.

iredmail:
You didn't get me.
It's easy to set the before/after tag, but what i want to avoid is html escaping done by the `html` formatter.

search/highlight/format/plain/plain.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	// add any remaining text after the last token
+	rv += string(f.Orig[curr:f.End])
+
+	return rv
+}
+
+func Constructor(config map[string]interface{}, cache *registry.Cache) (highlight.FragmentFormatter, error) {
+	before := defaultPlainHighlightBefore
+	beforeVal, ok := config["before"].(string)
```
abhinavdangeti, line 75: You cannot use the same settings' keys as html here - "before" and "after" need to be replaced with some other keys.

search/highlight/format/plain/plain.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"github.com/blevesearch/bleve/v2/registry"
+	"github.com/blevesearch/bleve/v2/search/highlight"
+)
+
+const Name = "plain"
+
+const defaultPlainHighlightBefore = "markbefore"
+const defaultPlainHighlightAfter = "markafter"
```
abhinavdangeti, line 25: Perhaps different defaults here - like "<start>" and "<end>"

abhinavdangeti (changes requested):

search/highlight/format/plain/plain.go
```diff
@@ -0,0 +1,89 @@
+//  Copyright (c) 2014 Couchbase, Inc.
```
abhinavdangeti, line 1: 2022

abhinavdangeti (changes requested):

iredmail:
Fixed copyright year in all 3 formatters.

abhinavdangeti (changes requested):
Please don't change the copyright year on existing files.

iredmail:
> Please don't change the copyright year on existing files.

Reverted. Sorry about this.

abhinavdangeti (commented):
There's unresolved comments still.

search/highlight/format/plain/plain.go
abhinavdangeti: The config is something that is shared. After a second thought I suppose it's ok to share these settings across both the highlight formatters.

abhinavdangeti (approved):

PR #1687, opened 2022-05-30
Fixing the docValue callback handling for spatial filtering

sreekanth-cb:
-This change updates the geoshape value extraction across
 docValue callbacks.
-Updated the multipolygon tests


search/searcher/search_geoshape.go
```diff
@@ -86,16 +87,17 @@ func buildRelationFilterOnShapes(dvReader index.DocValueReader, field string,
...
 					} else if startReading && !finishReading {
-						if bytes.Equal(geo.GlueBytes, term[len(term)-geo.GlueBytesOffset:]) {
+						if len(term) > geo.GlueBytesOffset &&
+							bytes.Equal(geo.GlueBytes, term[len(term)-geo.GlueBytesOffset:]) {
 							term = term[:len(term)-geo.GlueBytesOffset]
+							term = append(termSeparatorSplitSlice, term...)
 							dvShapeValue = append(dvShapeValue, term...)
 							finishReading = true
 						} else {
+							term = append(termSeparatorSplitSlice, term...)
```
Thejas-bhat, line 99: A minor thing, this statement and the same statement at line 95 could be inside the else if block as a common thing (instead of repeated it in both the blocks)

Thejas-bhat: On second thoughts, maybe the else block is not needed here? The statements in it could be kept as-is after the if sub block since these append ops are anyways going to happen

Thejas-bhat (approved):

PR #1699, opened 2022-06-06
Added more tests for geo spatial queries

metonymic-smokey:
This PR adds more tests for the geo spatial querying feature, ported from CockroachDB.


search/searcher/search_contains_test.go
```diff
@@ -0,0 +1,1040 @@
...
+		DocShapeVertices [][][]float64
+		DocShapeName     string
+		Expected         []string
+		Desc             string
+		QueryType        string
+	}{
+		{
+			QueryShape:       rightRectPoint,
+			DocShapeVertices: rightRect,
+			DocShapeName:     "polygon1",
```
sreekanth-cb, line 44: DocShapeName looks un used and contains an invalid value?

search/searcher/search_contains_test.go
```diff
@@ -0,0 +1,1040 @@
...
+	for _, test := range tests {
+		doc := document.NewDocument(test.DocShapeName)
+		doc.AddField(document.NewGeoShapeFieldWithIndexingOptions("geometry", []uint64{},
+			[][][][]float64{{{test.DocShapeVertices}}}, "point", document.DefaultGeoShapeIndexingOptions))
+		err := i.Update(doc)
+		if err != nil {
+			t.Errorf(err.Error())
+		}
+
+		indexReader, err := i.Reader()
```
sreekanth-cb, line 199: closing the index reader is till needed? and the same applies to many other test cases?

search/searcher/search_contains_test.go
```diff
@@ -0,0 +1,1040 @@
...
+			QueryType:        "contains",
+		},
+	}
+	i := setupIndex(t)
+
+	for _, test := range tests {
+		doc := document.NewDocument(test.DocShapeName)
+		doc.AddField(document.NewGeoShapeFieldWithIndexingOptions("geometry", []uint64{},
+			[][][][]float64{test.DocShapeVertices}, "polygon", document.DefaultGeoShapeIndexingOptions))
+		err := i.Update(doc)
```
sreekanth-cb, line 64: line 64-92 looks repeated across many tests and similar patterns are visible across tests.
Can this be another function with the right arguments and reduce the boilerplate code?

search/searcher/search_contains_test.go
```diff
@@ -0,0 +1,1040 @@
...
+func TestPolygonEnvelopeContains(t *testing.T) {
+	tests := []struct {
+		QueryShape       [][][]float64
+		DocShapeVertices [][]float64
+		DocShapeName     string
+		Expected         []string
+		Desc             string
+		QueryType        string
+	}{
+		{
```
sreekanth-cb, line 471: you could have working cases/positive cases as well here?

search/searcher/search_contains_test.go
```diff
@@ -0,0 +1,1040 @@
...
+			t.Errorf(err.Error())
+		}
+		err = indexReader.Close()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+func TestPolygonMultiPointContains(t *testing.T) {
```
sreekanth-cb, line 780: include some hits/positive cases as well?

search/searcher/search_intersects_test.go
```diff
@@ -0,0 +1,1784 @@
...
+	index "github.com/blevesearch/bleve_index_api"
+)
+
+func setupIndex(t *testing.T) index.Index {
+	analysisQueue := index.NewAnalysisQueue(1)
+	i, err := scorch.NewScorch(
+		gtreap.Name,
+		map[string]interface{}{
+			"path":          "",
+			"spatialPlugin": "s2",
```
sreekanth-cb, line 33: you may not need this flag for geojson?

search/searcher/search_intersects_test.go
```diff
@@ -0,0 +1,1784 @@
...
+			t.Errorf(err.Error())
+		}
+		err = indexReader.Close()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+func TestPolygonIntersects(t *testing.T) {
```
sreekanth-cb, line 1118: perhaps add a test for the second polygon inside the hole of the first polygon?

search/searcher/search_contains_test.go
metonymic-smokey: I use it here - https://github.com/metonymic-smokey/bleve/blob/dd5928c00253f2a0c9fd50e1c88cbe2e87b974c1/search/searcher/search_contains_test.go#L61 
Why is the value invalid?

search/searcher/search_contains_test.go
metonymic-smokey: WRT Polygon 'contains' queries, when multipoint is the doc shape, this is what the spec mentions:
"NA.  MultiPoint is a non-closed shape."
So I think with such a combination, the result will always be nil.

sreekanth-cb: So, the question would be why do we even have a test case if there are no results expected. I mean the whole test case itself is a non-existing case given that points are non-closed shapes.

metonymic-smokey: I added it so that in case of any changes in the future which might change this rule/break this functionality, such a UT would come in handy to spot that.

There are other UTs for combinations which will always be nil according to the current spec. If required, I'll remove them too.

search/searcher/search_intersects_test.go
metonymic-smokey: Added in a recent commit. PTAL and let me know. Thanks!

metonymic-smokey:
> @metonymic-smokey I'd rename the new files you're adding to better explain what they're meant for - so perhaps geoshape_contains_test.go, geoshape_intersects_test.go etc.

Thanks @abhinavdangeti! Done it in a recent commit.

search/searcher/search_contains_test.go
metonymic-smokey: In a recent WIP commit, I have attempted to do something like this for 1 func. I'm not _entirely_ sure about it. Could you PTAL and let me know?
thanks!

search/searcher/geoshape_contains_test.go
```diff
@@ -0,0 +1,1091 @@
...
+
+var (
+	leftRectEdgeMultiPoint [][]float64   = [][]float64{{-1, 0.2}, {-0.9, 0.1}}
+	leftRectWithHole       [][][]float64 = [][][]float64{{{-1, 0}, {0, 0}, {0, 1}, {-1, 1}, {-1, 0}},
+		{{-0.75, 0.25}, {-0.75, -0.75}, {-0.25, 0.75}, {-0.25, 0.25}, {-0.74, 0.25}}}
+	leftRectEdgePoint  []float64   = []float64{-1, 0.2}
+	leftRectMultiPoint [][]float64 = [][]float64{{0.5, 0.5}, {-0.9, 0.1}}
+)
+
+func testCaseSetup(t *testing.T, docShapeName, docShapeType string, docShapeVertices [][][][]float64,
```
sreekanth-cb, line 33: Can this function be generic enough to be used across all test cases?

search/searcher/search_contains_test.go
sreekanth-cb: my bad.. I mistaken that for the shape type that was getting parameterised later to the runXXX functions.. My thinking was that all test specific values would be driven by the table and the rest would be a few utility functions that are invoked with those test entries from the tables.

search/searcher/geoshape_contains_test.go
metonymic-smokey: yes, I think so.
Now that you've approved it, I'll add it to other functions as well.
This was a WIP commit.
thanks!

metonymic-smokey: Added it for the others.
Thanks!

search/searcher/geoshape_contains_test.go
```diff
@@ -0,0 +1,938 @@
...
+		Expected         []string
+		Desc             string
+		QueryType        string
+	}{
+		{
+			QueryShape:       leftRectEdgeMultiPoint,
+			DocShapeVertices: leftRectWithHole,
+			DocShapeName:     "polygon1",
+			Expected:         []string{"polygon1"},
+			Desc:             "multi point inside polygon with hole",
```
sreekanth-cb, line 515: can we have a test where at least one point is inside the hole and others are inside the polygon?

search/searcher/geoshape_contains_test.go
```diff
@@ -0,0 +1,938 @@
...
+		})
+
+		err = closeFn()
+		if err != nil {
+			t.Errorf(err.Error())
+		}
+	}
+}
+
+func TestMultiPointLinestringContains(t *testing.T) {
```
sreekanth-cb, line 556: Since the XXXX_LinestringContains mostly serves as a reference returning nothing, can all these variations could be made to a single test TestLinestringContains() with all entries in a single table?

search/searcher/geoshape_contains_test.go
```diff
@@ -0,0 +1,938 @@
...
+			}
+		})
+		err = closeFn()
+		if err != nil {
+			t.Errorf(err.Error())
+		}
+	}
+}
+
+func TestGeometryCollectionPolygonContains(t *testing.T) {
```
sreekanth-cb, line 892: There could be a TestGeometryCollectionMultiPolygonContains test where as the various shapes inside the GC would be covered by multiple polygons in the multipolygon?

search/searcher/geoshape_within_test.go
```diff
@@ -0,0 +1,1318 @@
...
+		})
+
+		err = closeFn()
+		if err != nil {
+			t.Errorf(err.Error())
+		}
+	}
+}
+
+func TestMultiPointMultiLinestringWithin(t *testing.T) {
```
sreekanth-cb, line 551: Same here for LinestringWithin tests as well?

search/searcher/geoshape_contains_test.go
metonymic-smokey: I'm not sure it's possible to combine those functions since the querying functions and fields change based on query shape.

search/searcher/geoshape_contains_test.go
metonymic-smokey: Added in a recent commit. PTAL and let me know. Thanks!

search/searcher/geoshape_contains_test.go
metonymic-smokey: I'm not sure these can be combined since the query shapes and query functions are different for each.

sreekanth-cb (approved):

PR #1702, opened 2022-06-09
A new stat for bytes read off the disk

Thejas-bhat:


index/scorch/scorch.go
```diff
@@ -525,6 +535,14 @@ func (s *Scorch) Stats() json.Marshaler {
 	return &s.stats
 }
 
+func (s *Scorch) BytesRead() uint64 {
```
sreekanth-cb, line 538: BytesRead is a generic term, could be more specific about the query time stats.

index/scorch/scorch.go
```diff
@@ -582,7 +600,9 @@ func (s *Scorch) StatsMap() map[string]interface{} {
 	m["index_time"] = m["TotIndexTime"]
 	m["term_searchers_started"] = m["TotTermSearchersStarted"]
 	m["term_searchers_finished"] = m["TotTermSearchersFinished"]
+	m["num_bytes_read_query_time"] = m["TotBytesReadQueryTime"]
 	m["num_plain_text_bytes_indexed"] = m["TotIndexedPlainTextBytes"]
+	m["num_analysis_bytes_indexed"] = m["TotIndexedAnalysisBytes"]
```
sreekanth-cb, line 605: It's good to have both the stats share the same prefix like "num_bytes_".
So, please reword the stats..

index/scorch/snapshot_index.go
```diff
@@ -505,46 +522,78 @@ func (i *IndexSnapshot) InternalID(id string) (rv index.IndexInternalID, err err
...
 	rv.currPosting = nil
 	rv.currID = rv.currID[:0]
-
 	if rv.dicts == nil {
-		rv.dicts = make([]segment.TermDictionary, len(i.segment))
-		for i, segment := range i.segment {
+		rv.dicts = make([]segment.TermDictionary, len(is.segment))
+		for i, segment := range is.segment {
+			var prevBytesRead uint64
+			if segP, ok := segment.segment.(bytesOffDiskStats); ok {
```
sreekanth-cb, line 548: bytesOffDiskStats needs a better name.

index/scorch/snapshot_index.go
```diff
@@ -505,46 +522,78 @@ func (i *IndexSnapshot) InternalID(id string) (rv index.IndexInternalID, err err
...
+				prevBytesRead = segP.BytesRead()
+			}
 			dict, err := segment.segment.Dictionary(field)
 			if err != nil {
 				return nil, err
 			}
+			if segP, ok := segment.segment.(bytesOffDiskStats); ok {
```
sreekanth-cb, line 555: Had we remember the segP last time, could have avoided the recheck of the interface here.

index/scorch/snapshot_index.go
```diff
@@ -505,46 +522,78 @@ func (i *IndexSnapshot) InternalID(id string) (rv index.IndexInternalID, err err
...
+				atomic.AddUint64(&is.parent.stats.TotBytesReadQueryTime, segP.BytesRead()-prevBytesRead)
+			}
 			rv.dicts[i] = dict
 		}
 	}
 
-	for i, segment := range i.segment {
+	for i, segment := range is.segment {
+		var prevBytesReadPL uint64
+		if _, ok := segment.segment.(bytesOffDiskStats); ok {
```
sreekanth-cb, line 564: do we need this check here?

index/scorch/snapshot_index.go
```diff
@@ -505,46 +522,78 @@ func (i *IndexSnapshot) InternalID(id string) (rv index.IndexInternalID, err err
...
+				prevBytesReadPL = postings.BytesRead()
+			}
+		}
 		pl, err := rv.dicts[i].PostingsList(term, segment.deleted, rv.postings[i])
 		if err != nil {
 			return nil, err
 		}
 		rv.postings[i] = pl
+		var prevBytesReadItr uint64
+		if _, ok := segment.segment.(bytesOffDiskStats); ok {
```
sreekanth-cb, line 575: segment.segment.(bytesOffDiskStats) do we need it?

sreekanth-cb (commented):
few comments..

Thejas-bhat (commented):
Thanks sreekanth, I've included your suggestions

index/scorch/merge.go
```diff
@@ -426,6 +435,16 @@ type segmentMerge struct {
 	notifyCh      chan *mergeTaskIntroStatus
 }
 
+func cumulateBytesRead(sbs []segment.Segment) uint64 {
+	rv := uint64(0)
```
sreekanth-cb, line 439: var rv uint64?

index/scorch/scorch.go
```diff
@@ -387,13 +387,23 @@ func (s *Scorch) Batch(batch *index.Batch) (err error) {
...
+	}
 	for itemsDeQueued < numUpdates {
 		result := <-resultChan
 		resultSize := result.Size()
 		atomic.AddUint64(&s.iStats.analysisBytesAdded, uint64(resultSize))
 		totalAnalysisSize += resultSize
 		analysisResults[itemsDeQueued] = result
 		itemsDeQueued++
+		result.VisitFields(func(f index.Field) {
+			atomic.AddUint64(&s.stats.TotIndexedAnalysisBytes,
```
sreekanth-cb, line 404: better stats wordings

TotBytesIndexedAfterAnalysis?
TotBytesReadDuringQueryTime Or TotBytesReadAtQueryTime 

so that both stats share the same prefix.

index/scorch/snapshot_index.go
```diff
@@ -146,10 +148,18 @@ func (i *IndexSnapshot) newIndexSnapshotFieldDict(field string,
 	results := make(chan *asynchSegmentResult)
 	for index, segment := range i.segment {
 		go func(index int, segment *SegmentSnapshot) {
+			var prevBytesRead uint64
+			if seg, ok := segment.segment.(diskStatsReporter); ok {
```
sreekanth-cb, line 152: same interface checks were repeated at 152 and 159.
Can't we make that to 1?

index/scorch/snapshot_index.go
```diff
@@ -526,25 +544,54 @@ func (i *IndexSnapshot) TermFieldReader(term []byte, field string, includeFreq,
...
+		}
 		rv.iterators[i] = pl.Iterator(includeFreq, includeNorm, includeTermVectors, rv.iterators[i])
+
+		if postings, ok := pl.(diskStatsReporter); ok &&
+			prevBytesReadPL < postings.BytesRead() {
+			atomic.AddUint64(&is.parent.stats.TotBytesReadQueryTime,
+				postings.BytesRead()-prevBytesReadPL)
+		}
+
+		if itr, ok := rv.iterators[i].(diskStatsReporter); ok &&
```
sreekanth-cb, line 588: avoid repeated interface checks for the same vars..

sreekanth-cb (commented):
minor...

index/scorch/merge.go
Thejas-bhat: done.

index/scorch/scorch.go
Thejas-bhat: done.

index/scorch/snapshot_index.go
Thejas-bhat: yeah done, my bad.

index/scorch/snapshot_index.go
Thejas-bhat: there is a re-initialisation of the var rv.iterators[i], so i can't reuse the same variable over here, since the bytes read as part of the iterator initialisation won't be part of the old variable.

index/scorch/scorch.go
```diff
@@ -387,13 +387,23 @@ func (s *Scorch) Batch(batch *index.Batch) (err error) {
 	analysisResults := make([]index.Document, int(numUpdates))
 	var itemsDeQueued uint64
 	var totalAnalysisSize int
+	analysisBytes := func(tokMap index.TokenFrequencies) (rv uint64) {
```
abhinavdangeti, line 390: Should we add a flag here to calculate this only when the flag is set? 

Feel like this is something we don't always necessarily need to do - hope none of this calculation shows up in CPU profiles?

abhinavdangeti: Not sure if this is enough, we've to record copies - stored values and doc values. Also term vectors. I don't see you doing this in your zap PR.

abhinavdangeti (changes requested):

index/scorch/scorch.go
Thejas-bhat: Just to close the thread, these two changes are going to be included in upcoming PRs.

index/scorch/merge.go
```diff
@@ -352,6 +353,14 @@ func (s *Scorch) planMergeAtSnapshot(ctx context.Context,
 				atomic.AddUint64(&s.stats.TotFileMergePlanTasksErr, 1)
 				return err
 			}
+
+			switch segI := seg.(type) {
+			case segment.DiskStatsReporter:
```
abhinavdangeti, line 358: Per my latest conversation with Jon, we shouldn't be accounting for content written to disk by the merger. More conversation needed here I suppose.

https://docs.google.com/document/d/1pmrasugnCIAXUNgR14h_tix_ZxU3V28unEQYhnz1u34/edit#heading=h.w7mwv55b14ou

sreekanth-cb: That is the case here as well, we are only accounting for the BytesRead from disk for the new merge created segment. And this would anyway be needed/accounted for the next query on this newly formed segment.

sreekanth-cb (commented):
.

index/scorch/merge.go
sreekanth-cb: We need to carry further the stats from the older segments that got merged into the newly formed segment so that the accounting of stats is cumulative. And hence the need for the SetBytesRead() api. 
If we fail to consider the stats from the pre-merge segments, then the bytes accounting go wrong.

There could be a better way to handle this?

index/scorch/snapshot_index.go
```diff
@@ -424,6 +435,11 @@ func (i *IndexSnapshot) Document(id string) (rv index.Document, err error) {
 	segmentIndex, localDocNum := i.segmentIndexAndLocalDocNumFromGlobal(docNum)
 
 	rvd := document.NewDocument(id)
+	var prevBytesRead uint64
+	seg, ok := i.segment[segmentIndex].segment.(segment.DiskStatsReporter)
```
abhinavdangeti, line 439: Rename ok to say diskStatsAvailable so in line 472 below we know why things are ok.

index/scorch/snapshot_index.go
```diff
@@ -526,25 +545,54 @@ func (i *IndexSnapshot) TermFieldReader(term []byte, field string, includeFreq,
 	rv.currID = rv.currID[:0]
 
 	if rv.dicts == nil {
-		rv.dicts = make([]segment.TermDictionary, len(i.segment))
-		for i, segment := range i.segment {
+		rv.dicts = make([]segment.TermDictionary, len(is.segment))
+		for i, segment := range is.segment {
+			var prevBytesRead uint64
+			segP, ok := segment.segment.(diskStatsReporter)
```
abhinavdangeti, line 551: same here, rename ok to diskStatsAvailable.

index/scorch/snapshot_index.go
```diff
@@ -661,10 +709,18 @@ func (i *IndexSnapshot) documentVisitFieldTermsOnSegment(
 	}
 
 	if ssvOk && ssv != nil && len(vFields) > 0 {
+		var prevBytesRead uint64
+		ssvp, ok := ssv.(segment.DiskStatsReporter)
+		if ok {
```
abhinavdangeti, line 714: ditto

index/scorch/snapshot_index_tfr.go
```diff
@@ -76,6 +76,11 @@ func (i *IndexSnapshotTermFieldReader) Next(preAlloced *index.TermFieldDoc) (*in
 	}
 	// find the next hit
 	for i.segmentOffset < len(i.iterators) {
+		prevBytesRead := uint64(0)
+		itr, ok := i.iterators[i.segmentOffset].(segment.DiskStatsReporter)
```
abhinavdangeti, line 80: ditto

abhinavdangeti (dismissed):

PR #1704, opened 2022-06-15
Adding input validations for spatial coordinates.

sreekanth-cb:


geo/parse.go
```diff
@@ -272,29 +275,38 @@ func extract4DCoordinates(thing interface{}) (rv [][][][]float64) {
...
 	var coordValue interface{}
-	var typ string
+
 	if thingVal.Kind() == reflect.Map {
 		iter := thingVal.MapRange()
 		for iter.Next() {
 			if iter.Key().String() == "type" {
-				typ = iter.Value().Interface().(string)
-				// make the shape type case insensitive.
-				typ = strings.ToLower(typ)
```
abhinavdangeti, line 289: Would we be giving up on the case insensitivity here ?

geo/parse.go
```diff
@@ -383,23 +395,68 @@ func ExtractGeoShapeCoordinates(coordValue interface{},
...
+		// ignore the contents with invalid entry.
+		if len(coords) == 0 {
+			return nil, typ, false
+		}
+
+		if len(coords) > 2 && typ == EnvelopeType {
+			return nil, typ, false
+		}
+
+		if len(coords) < 2 && typ != MultiPointType {
```
abhinavdangeti, line 422: `len(coords) < 2 && typ == MultiPointType` ?

geo/parse.go
sreekanth-cb: Nope, it is handled in the new function in parse.go line no:301.

abhinavdangeti: Ack.

sreekanth-cb (commented):
.

geo/parse.go
```diff
@@ -383,23 +395,68 @@ func ExtractGeoShapeCoordinates(coordValue interface{},
...
 		typ == EnvelopeType {
-		rv = [][][][]float64{{extract2DCoordinates(coordValue)}}
+		coords := extract2DCoordinates(coordValue)
+
+		// ignore the contents with invalid entry.
+		if len(coords) == 0 {
+			return nil, typ, false
+		}
+
+		if len(coords) > 2 && typ == EnvelopeType {
```
abhinavdangeti, line 418: len(coords) != 2 && typ == EnvelopeType

geo/parse.go
abhinavdangeti: Ah in that case - I recommend updating the earlier check too :)

abhinavdangeti: This is to avoid 2 unnecessary checks for EnvelopeType when len(coords) < 2

abhinavdangeti (changes requested):

geo/parse.go
abhinavdangeti: type == EnvelopeType && len(coords) != 2

geo/parse.go
sreekanth-cb: I realize the readability problem here. :)

abhinavdangeti (approved):
Cheers!

PR #1724, opened 2022-08-24
When using cjk analyzer with `output_unigram`, an unigram before Japanese punctuation is not indexed

kaakaa:
## Summary
When I do indexing and searching Japanese text with cjk analyzer with `output_unigram` config, the word before Japanese punctuation characters (e.g.: `、`,`。`) cannot be searched. For example, I got no results with the search query `ちは` from sentence of `こんにちは、世界`, but I got a result with the search query `ちは` from sentence of `こんにちは世界`.

I look into this, and I found that CJKBigramFilter doesn't flush unigram before punctuations correctly. This PR fixed it.

---

Sample code to reproduce this issue is here:
https://gist.github.com/kaakaa/ca8c20821ef610b098851f487eb61ea5


abhinavdangeti (approved):

abhinavdangeti:
Thanks for this contribution @kaakaa .
Let's wait for at least one more review from the team.

PR #1729, opened 2022-09-14
allow custom analyzers

antonpriyma:
In our project we need to implement analyzer with some custom logic. It's only possible with interface

### DIFF
--- a/analysis/type.go
+++ b/analysis/type.go
@@ -72,13 +72,17 @@ type TokenFilter interface {
 	Filter(TokenStream) TokenStream
 }
 
-type Analyzer struct {
+type Analyzer interface {
+	Analyzer([]byte) TokenStream
+}
+
+type DefaultAnalyzer struct {
 	CharFilters  []CharFilter
 	Tokenizer    Tokenizer
 	TokenFilters []TokenFilter
 }
 
-func (a *Analyzer) Analyze(input []byte) TokenStream {
+func (a *DefaultAnalyzer) Analyze(input []byte) TokenStream {
 	if a.CharFilters != nil {
 		for _, cf := range a.CharFilters {
 			input = cf.Filter(input)


abhinavdangeti:
@antonpriyma You shouldn't need what you're proposing here.

If you want to implement an analyzer with custom logic, you can freely do so by creating a constructor with your custom logic/components with this signature. You can even create your custom components - CharFilter, Tokenizer and TokenFilter are interfaces.

```
type AnalyzerConstructor func(config map[string]interface{}, cache *Cache) (*analysis.Analyzer, error)
```

And then add that to the registry during init -
```
registry.RegisterAnalyzer(AnalyzerName, AnalyzerConstructor)
```

This would make the analyzer available for use.

antonpriyma:
> o implement an analyzer with custom logic, you can freely do so by creating a constructor with your custom logic/components with t

Yes, it's possible, but i can't do something like this: 

`
func AnalyzerConstructor(config map[string]interface{}, cache *registry.Cache) (*analysis.Analyzer, error) {
 languages := []lingua.Language{
  lingua.Thai,
 }

 detector := lingua.NewLanguageDetectorBuilder().
  FromLanguages(languages...).
  Build()

 thAnalyzer, err := th.AnalyzerConstructor(config, cache)
 if err != nil {
  return nil, fmt.Errorf("failed to build th analyzer: %w", err)
 }

 defaultAnalyzer, err := standard.AnalyzerConstructor(config, cache)
 if err != nil {
  return nil, fmt.Errorf("failed to build default analyzer: %w", err)
 }

 return &DetectLangAnalyzer{
  detector:        detector,
  DefaultAnalyzer: defaultAnalyzer,
  ThAnalyzer:      thAnalyzer,
 }, nil
}

func init() {
 registry.RegisterAnalyzer(AnalyzerName, AnalyzerConstructor)
}

type DetectLangAnalyzer struct {
 detector        lingua.LanguageDetector
 DefaultAnalyzer analysis.Analyzer
 ThAnalyzer      analysis.Analyzer
}

func (a *DetectLangAnalyzer) Analyze(input []byte) analysis.TokenStream {
 if language, _ := a.detector.DetectLanguageOf(string(input)); language == lingua.Thai {
  return a.ThAnalyzer.Analyze(input)
 }

 return a.DefaultAnalyzer.Analyze(input)
}`

abhinavdangeti:
@antonpriyma Thanks for sharing the test sample. Seems a reasonable to me.
Would you consider adding a unit test demonstrating the need for this changes?

We'll consider this change for our next release.

abhinavdangeti (approved):

Thejas-bhat (approved):

sreekanth-cb:
@antonpriyma, can you please add the Thai language support as well to bleve so that it would be a good demonstration for the current proposed changes?

PR #1730, opened 2022-09-15
MB-53702: Unit test highlighting a geo issue

abhinavdangeti:


abhinavdangeti:
@sreekanth-cb am I missing something obvious here or could this be a legit issue?
Please feel free to add commits to this PR if it turns out a bug.

sreekanth-cb:
This polygon isn't following the right-hand rule or the CCW ordering for the outer loop.

A linear ring MUST follow the right-hand rule with respect to the  area it bounds, i.e., exterior rings should be counterclockwise, and the holes should be clockwise.'

abhinavdangeti:
Phew, thanks @sreekanth-cb!

PR #1735, opened 2022-10-05
MB-40604: Support IPRange query

abhinavdangeti:
Related: https://github.com/blevesearch/bleve/pull/1546

### DIFF
--- a/search_test.go
+++ b/search_test.go
@@ -1889,3 +1889,47 @@ func TestHightlightingWithHTMLCharacterFilter(t *testing.T) {
 			gotFragment, expectedFragment)
 	}
 }
+
+func TestIPRangeQuery(t *testing.T) {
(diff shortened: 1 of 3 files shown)


search_test.go
```diff
@@ -1889,3 +1889,47 @@ func TestHightlightingWithHTMLCharacterFilter(t *testing.T) {
...
+	}
+
+	defer func() {
+		err = idx.Close()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	ipContent := "1.2.3.4"
```
Thejas-bhat, line 1915: A small doubt, do you think we can use a dummy (but a more realistic) IP address over here as an example?

abhinavdangeti: Ack.

Issue #1731, opened 2022-09-22
Significant query throughput drop observed with v2.3.4


Thejas-bhat:
We introduced changes that intend to publish IO stats (read/write) for tracking system usage with [v2.3.4](https://github.com/blevesearch/bleve/releases/tag/v2.3.4).

- https://github.com/blevesearch/bleve/pull/1702
- https://github.com/blevesearch/bleve/pull/1710

For queries with high hit cardinality (order of 100K+ hits), we've observed a considerable query throughput drop in our in-house testing.

 ![image](https://user-images.githubusercontent.com/35959007/191663254-b3f96abd-bdff-4057-bf69-a7ff2906683d.png)

We are actively working on this issue to optimizing this stats calculation and perhaps even making this cost estimation on-demand to mitigate this regression.

The fix will be targeted for the v2.3.5 release.

abhinavdangeti:
https://github.com/blevesearch/bleve/pull/1738

PR #1738, opened 2022-10-12
Optimising IO Stats Computation

Thejas-bhat:
The current way of IO Stats computation is causing too much of overhead:
- In case of _very high_ hit cardinality, the current implementation results in too many atomic.AddUint64 operations on an index level stat and that too its potentially invoked by a large number of independent queries running in the system (which are running as parallel threads).
- This essentially causes the stats updation to move from a per hit to a per query level (at the end of collector phase i.e.). So, DocumentMatch is used to track the bytes read information which will be aggregated in the collector level and the updating the same in the stats map using a callback.
- Furthermore, adding an extra interface "DiskStatsReporter" and introducing type assertions regarding that for the implementing struct causes additional overhead as well. So, the existing PostingsIterator, PostingsList and Segment interfaces now have DiskStatsReporter as an additional requirement to supported thereby avoiding the interface assertions altogether.

The issue which mentions the regression:
https://github.com/blevesearch/bleve/issues/1731

### DIFF
--- a/index/scorch/reader_test.go
+++ b/index/scorch/reader_test.go
@@ -153,6 +153,7 @@ func TestIndexReader(t *testing.T) {
 	if err != nil {
 		t.Errorf("unexpected error: %v", err)
 	}
+	match.BytesRead = 0
 	if !reflect.DeepEqual(expectedMatch, match) {
 		t.Errorf("got %#v, expected %#v", match, expectedMatch)
 	}
--- a/index/scorch/reader_test.go
+++ b/index/scorch/reader_test.go
@@ -153,6 +153,8 @@ func TestIndexReader(t *testing.T) {
 	if err != nil {
 		t.Errorf("unexpected error: %v", err)
 	}
+	// Ignoring the BytesRead value, since it doesn't have
+	// relevance in this type of test
 	match.BytesRead = 0
 	if !reflect.DeepEqual(expectedMatch, match) {
 		t.Errorf("got %#v, expected %#v", match, expectedMatch)
--- a/index/scorch/segment_plugin.go
+++ b/index/scorch/segment_plugin.go
@@ -94,6 +94,20 @@ func RegisterSegmentPlugin(plugin SegmentPlugin, makeDefault bool) {
 	}
 }
 
+// This exported method allows the application
+// to set the state as to whether it needs
+// disk stats to be tracked while indexing and querying
+//
+// The stats said above refer to num_bytes_read_at_query_time
+// and num_bytes_written_at_index_time
+func (s *Scorch) enableIOStatsComputation(state bool) {
+	zapv15.AccountIOStats = state
+}
+
+func accountIOStats() bool {
+	return zapv15.AccountIOStats
+}
+
 func SupportedSegmentTypes() (rv []string) {
 	for k := range supportedSegmentPlugins {
 		rv = append(rv, k)
--- a/index/scorch/segment_plugin.go
+++ b/index/scorch/segment_plugin.go
@@ -94,20 +94,6 @@ func RegisterSegmentPlugin(plugin SegmentPlugin, makeDefault bool) {
 	}
 }
 
-// This exported method allows the application
-// to set the state as to whether it needs
-// disk stats to be tracked while indexing and querying
-//
-// The stats said above refer to num_bytes_read_at_query_time
-// and num_bytes_written_at_index_time
-func (s *Scorch) enableIOStatsComputation(state bool) {
-	zapv15.AccountIOStats = state
-}
-
-func accountIOStats() bool {
-	return zapv15.AccountIOStats
-}
-
 func SupportedSegmentTypes() (rv []string) {
 	for k := range supportedSegmentPlugins {
 		rv = append(rv, k)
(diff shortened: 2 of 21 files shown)


index/scorch/segment_plugin.go
```diff
@@ -94,6 +94,20 @@ func RegisterSegmentPlugin(plugin SegmentPlugin, makeDefault bool) {
 	}
 }
 
+// This exported method allows the application
+// to set the state as to whether it needs
+// disk stats to be tracked while indexing and querying
+//
+// The stats said above refer to num_bytes_read_at_query_time
+// and num_bytes_written_at_index_time
+func SetDiskStatsCollection(state bool) {
```
Thejas-bhat, line 103: Must be made made part of scorch options rather than an exported method.

index/scorch/segment_plugin.go
```diff
@@ -94,6 +94,20 @@ func RegisterSegmentPlugin(plugin SegmentPlugin, makeDefault bool) {
...
+	zapv15.AccountIOStats = state
+}
+
+func collectDiskStats() bool {
```
Thejas-bhat, line 107: collectDiskStats() -> accountIOStats()

Thejas-bhat (commented):
Suggestions

index/scorch/snapshot_index.go
```diff
@@ -149,7 +149,11 @@ func (i *IndexSnapshot) newIndexSnapshotFieldDict(field string,
 	for index, segment := range i.segment {
 		go func(index int, segment *SegmentSnapshot) {
 			var prevBytesRead uint64
-			seg, diskStatsAvailable := segment.segment.(diskStatsReporter)
+			var seg diskStatsReporter
+			var diskStatsAvailable bool
+			if accountIOStats() {
+				seg, diskStatsAvailable = segment.segment.(diskStatsReporter)
+			}
 			if diskStatsAvailable {
```
abhinavdangeti, line 157: Move this snippet into within accountIOStats() ?

index/scorch/snapshot_index.go
```diff
@@ -436,7 +440,11 @@ func (i *IndexSnapshot) Document(id string) (rv index.Document, err error) {
 
 	rvd := document.NewDocument(id)
 	var prevBytesRead uint64
-	seg, diskStatsAvailable := i.segment[segmentIndex].segment.(segment.DiskStatsReporter)
+	var seg segment.DiskStatsReporter
+	var diskStatsAvailable bool
+	if accountIOStats() {
+		seg, diskStatsAvailable = i.segment[segmentIndex].segment.(segment.DiskStatsReporter)
+	}
 	if diskStatsAvailable {
```
abhinavdangeti, line 448: ditto^ and also everywhere else.

abhinavdangeti (changes requested):
You can update the zap v15 version based on the SHA for now.
We can set up formal release tags for zap mid next week.

index/scorch/reader_test.go
```diff
@@ -153,6 +153,7 @@ func TestIndexReader(t *testing.T) {
 	if err != nil {
 		t.Errorf("unexpected error: %v", err)
 	}
+	match.BytesRead = 0
```
abhinavdangeti, line 156: Why is this step necessary?

index/scorch/snapshot_index.go
```diff
@@ -436,7 +431,9 @@ func (i *IndexSnapshot) Document(id string) (rv index.Document, err error) {
 
 	rvd := document.NewDocument(id)
 	var prevBytesRead uint64
-	seg, diskStatsAvailable := i.segment[segmentIndex].segment.(segment.DiskStatsReporter)
+	var seg segment.DiskStatsReporter
+	var diskStatsAvailable bool
```
abhinavdangeti, line 435: This following type assertion is now not needed yes?

index/scorch/snapshot_index.go
```diff
@@ -471,6 +468,7 @@ func (i *IndexSnapshot) Document(id string) (rv index.Document, err error) {
 	if err != nil {
 		return nil, err
 	}
+
 	if diskStatsAvailable {
```
abhinavdangeti, line 472: Flag not necessary?

test/versus_test.go
```diff
@@ -343,12 +343,15 @@ func testVersusSearches(vt *VersusTest, searchTemplates []string, idxA, idxB ble
 				if len(hitB.FieldTermLocations) == 0 {
 					hitB.FieldTermLocations = nil
 				}
+				// hitA.BytesRead = 0
+				// hitB.BytesRead = 0
```
abhinavdangeti, line 347: Drop these comments.

test/versus_test.go
```diff
@@ -386,7 +389,7 @@ func hitsById(res *bleve.SearchResult) map[string]*search.DocumentMatch {
 		hit.Score = math.Trunc(hit.Score*1000.0) / 1000.0
 		hit.IndexInternalID = nil
 		hit.HitNumber = 0
-
+		hit.BytesRead = 0
```
abhinavdangeti, line 392: Again, why is this necessary?

abhinavdangeti (changes requested):

test/versus_test.go
Thejas-bhat: BytesRead doesn't relevance to that particular test, so I've disabled comparison over that. I've included comments for such settings.

index/scorch/snapshot_index.go
Thejas-bhat: Yeah these were minor rebase discrepancies. I've fixed them

PR #1752, opened 2022-11-04
Returning bytesRead as part of searchResult

Thejas-bhat:
Refactoring the bytesRead accounting to now be accounted at a per query level. The searchResult returned by SearchInContext now has an extra field called bytesRead which helps to track the cost associated with the query. 

The implementation to achieve this basically involves passing a context.Context which has a callback that is used to track or update the bytes read associated with some operation. This change of adding an extra parameter context to the searcher is an invasive one, and hence involves updating a lot of existing files to accommodate the same. However, the good thing is that the adding of context to searcher is in a way a future proof change, which hopefully might be useful to do any more useful things down at the searcher level and relay them back to the indexReader level.


document/document.go
```diff
@@ -30,9 +30,18 @@ func init() {
...
-	id              string  `json:"id"`
-	Fields          []Field `json:"fields"`
-	CompositeFields []*CompositeField
+	id                string  `json:"id"`
+	Fields            []Field `json:"fields"`
+	CompositeFields   []*CompositeField
+	storedFieldsBytes uint64
+}
+
+func (d *Document) SetStoredFieldsBytes(val uint64) {
```
abhinavdangeti, line 39: Drop this method you can updated storedFieldsBytes by simply exporting it (like CompositeFields above).

go.mod
```diff
@@ -40,3 +40,17 @@ require (
 	github.com/spf13/pflag v1.0.5 // indirect
 	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a // indirect
 )
+
+replace github.com/blevesearch/zapx/v15 => ../zapx
```
abhinavdangeti, line 44: Remove all these replaces.

index/scorch/snapshot_index.go
```diff
@@ -430,13 +432,17 @@ func (i *IndexSnapshot) Document(id string) (rv index.Document, err error) {
 	segmentIndex, localDocNum := i.segmentIndexAndLocalDocNumFromGlobal(docNum)
 
 	rvd := document.NewDocument(id)
-	prevBytesRead := i.segment[segmentIndex].segment.BytesRead()
+	var totalStoredBytes int
```
abhinavdangeti, line 435: remove variable.

index/scorch/snapshot_index.go
```diff
@@ -430,13 +432,17 @@ func (i *IndexSnapshot) Document(id string) (rv index.Document, err error) {
...
 
 	err = i.segment[segmentIndex].VisitDocument(localDocNum, func(name string, typ byte, val []byte, pos []uint64) bool {
 		if name == "_id" {
 			return true
 		}
 
+		// track uncompressed stored fields bytes as part of IO stats.
+		// However, ideally we'd need to track the compressed on-disk value
+		// Keeping that TODO for now until we have a cleaner way.
+		totalStoredBytes += len(val)
```
abhinavdangeti, line 445: rvd.StoredFieldsBytes += len(val)

Separate question - shouldn't you be doing this only when the field is actually "stored"?

index/upsidedown/index_reader.go
```diff
@@ -15,11 +15,12 @@
 package upsidedown
 
 import (
+	"context"
 	"reflect"
 
 	"github.com/blevesearch/bleve/v2/document"
 	index "github.com/blevesearch/bleve_index_api"
-	"github.com/blevesearch/upsidedown_store_api"
+	store "github.com/blevesearch/upsidedown_store_api"
```
abhinavdangeti, line 23: This changes seems unrelated?

index/upsidedown/reader.go
```diff
@@ -22,7 +22,7 @@ import (
 
 	"github.com/blevesearch/bleve/v2/size"
 	index "github.com/blevesearch/bleve_index_api"
-	"github.com/blevesearch/upsidedown_store_api"
+	store "github.com/blevesearch/upsidedown_store_api"
```
abhinavdangeti, line 25: Unrelated change?

abhinavdangeti (changes requested):
Mostly nits, I'll make another pass over this soon.
Meanwhile you should get some perf results with these changes.

document/document.go
```diff
@@ -30,9 +30,18 @@ func init() {
...
+	d.storedFieldsBytes = val
+}
+
+func (d *Document) GetStoredFieldsBytes() uint64 {
```
abhinavdangeti, line 43: This method can go away too, with storedFieldsBytes exported.

index/upsidedown/index_reader.go
Thejas-bhat: This popped up because of the vscode extension I believe, removing it.

index/upsidedown/reader.go
Thejas-bhat: This popped up because of the vscode extension I believe, removing it.

document/document.go
Thejas-bhat: Done.

document/document.go
Thejas-bhat: This Get type of API might be necessary since the struct is hidden behind the Document interface.

go.mod
Thejas-bhat: Yeah just got pushed to the review as part of "git add .". Removing them.

index/scorch/snapshot_index.go
Thejas-bhat: This callback is called only for the stored fields. The VisitDocument API ends up calling the VisitStoredFields in the zap layer which loads the bytes pertaining to the stored fields, uncompresses them and iterates over each of these fields and applies this callback.

index/scorch/snapshot_index.go
Thejas-bhat: Done.

abhinavdangeti:
@Thejas-bhat 

- bleve_index_api@v1.0.5
- scorch_segment_api/v2@v2.1.4
- zapx/v11@11.3.7
- zapx/v12@12.3.7
- zapx/v13@13.3.7
- zapx/v14@14.3.7
- zapx/v15@15.3.7

index/scorch/snapshot_index.go
```diff
@@ -37,6 +38,7 @@ import (
 // re usable, threadsafe levenshtein builders
 var lb1, lb2 *lev.LevenshteinAutomatonBuilder
 
+type diskStatsReporter segment.DiskStatsReporter
```
abhinavdangeti, line 41: Is this step necessary, can't we just type assert to segment.DiskStatsReporter?

index_impl.go
```diff
@@ -469,7 +469,20 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
 	}()
 
-	searcher, err := req.Query.Searcher(indexReader, i.m, search.SearcherOptions{
+	// This callback and variable handles the tracking of bytes read
+	//  1. as part of creation of tfr and its Next() calls which is
+	//     accounted by invoking this callback when the TFR is closed.
+	//  2. the docvalues portion (accounted in collector) and the retrieval
+	//     of stored fields bytes (by LoadAndHighlightFields)
+	var totalBytesRead uint64
+	SendBytesRead := func(bytesRead uint64) {
```
abhinavdangeti, line 478: Does this closure need to be exported? Would the compiler complain on sendBytesRead?

abhinavdangeti (commented):
Looking good, couple minor comments.

index_impl.go
Thejas-bhat: Aah, my bad.

index/scorch/snapshot_index.go
Thejas-bhat: Yeah at a couple of places, there are library name and variable name conflicts, for eg. `indexSnapshot.segment.segment `.

abhinavdangeti: Ah variable and import conflict. Should update the variable name then (it looks quite confusing :)). I'll make that change separately.

abhinavdangeti (approved):

PR #1766, opened 2022-12-05
Account for bytesRead in the event of MultiSearch

abhinavdangeti:
MultiSearch runs when the index is partitioned, and the results of individual search requests need to
be merged before sending the response to the caller.


Thejas-bhat (approved):

abhinavdangeti:
Ah, this is already accounted for in the Merge(..). So this change is not needed.
cc @Thejas-bhat

PR #1769, opened 2022-12-06
Correcting bytes read value with respect to metadata loading

Thejas-bhat:


index/scorch/snapshot_segment.go
```diff
@@ -35,6 +35,12 @@ type SegmentSnapshot struct {
 	deleted *roaring.Bitmap
 	creator string
 
+	// this flag is needed to identify whether this
+	// segment was mmaped recently, in which case
+	// we consider the loading cost of the metadata
+	// as part of IO stats.
+	mmaped uint64
```
abhinavdangeti, line 42: Move this to be the first attribute of this struct. Also, a uint32 would suffice?

index/scorch/merge.go
```diff
@@ -429,6 +430,7 @@ type segmentMerge struct {
 	oldNewDocNums map[uint64][]uint64
 	new           segment.Segment
 	notifyCh      chan *mergeTaskIntroStatus
+	mmaped        bool
```
abhinavdangeti, line 433: I'd rather we keep the type of this variable as uint for consistency.

abhinavdangeti (commented):
Jon should be able to get numbers from your toy build by Thursday.

index/scorch/merge.go
Thejas-bhat: Done.

index/scorch/snapshot_segment.go
Thejas-bhat: yeah it's sufficient, will do.

abhinavdangeti (approved):
Tentatively approving this.
Let's wait on the results before merging.

PR #1776, opened 2022-12-28
Improving rollback effectiveness 

Thejas-bhat:
Essentially, this PR aims to improve the existing rollback effectiveness by collecting the possible rollback points which are spread across a configurable time interval.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -516,6 +516,20 @@ func prepareBoltSnapshot(snapshot *IndexSnapshot, tx *bolt.Tx, path string,
 		return nil, nil, err
 	}
 
+	// Storing the timestamp at which the current indexSnapshot
+	// was persisted, useful when you want to spread the
+	// numSnapshotsToKeep reasonably better than consecutive
+	// epochs.
+	currTimeStamp := time.Now()
+	timeStampBinary, err := currTimeStamp.MarshalText()
+	if err != nil {
+		return nil, nil, err
+	}
+	err = metaBucket.Put(boltMetaDataTimeStamp, timeStampBinary)
+	if err != nil {
+		return nil, nil, err
+	}
+
 	// persist internal values
 	internalBucket, err := snapshotBucket.CreateBucketIfNotExists(boltInternalKey)
 	if err != nil {
@@ -682,6 +696,7 @@ var boltInternalKey = []byte{'i'}
 var boltMetaDataKey = []byte{'m'}
 var boltMetaDataSegmentTypeKey = []byte("type")
 var boltMetaDataSegmentVersionKey = []byte("version")
+var boltMetaDataTimeStamp = []byte("timeStamp")
 
 func (s *Scorch) loadFromBolt() error {
 	return s.rootBolt.View(func(tx *bolt.Tx) error {
@@ -871,24 +886,58 @@ func (s *Scorch) removeOldData() {
 // rollback'ability.
 var NumSnapshotsToKeep = 1
 
+var RollbackPointsSamplingRate = 1 * time.Hour
+
+// getProtectedEpochs aims to fetch the epochs keep based on a timestamp basis.
+// It tries to get NumSnapshotsToKeep snapshots, each of which are separated
+// by a time duration of RollbackPointsSamplingRate
+func (s *Scorch) getProtectedEpochs(persistedSnapshots []*snapshotMetaData) map[uint64]struct{} {
+	// make a map of epochs to protect from deletion
+	protectedEpochs := make(map[uint64]struct{}, s.numSnapshotsToKeep)
+
+	protectedEpochs[persistedSnapshots[0].epoch] = struct{}{}
+	nextSnapshotToProtect := persistedSnapshots[0].timeStamp.Add(RollbackPointsSamplingRate)
+	protectedSnapshots := 1
+	lastProtectedSnapshot := 0
+
+	for i := 1; i < len(persistedSnapshots); i++ {
+		if persistedSnapshots[i].timeStamp.After(nextSnapshotToProtect) {
+			protectedEpochs[persistedSnapshots[i-1].epoch] = struct{}{}
+			nextSnapshotToProtect =
+				persistedSnapshots[i-1].timeStamp.Add(RollbackPointsSamplingRate)
+			protectedSnapshots++
+			lastProtectedSnapshot = i - 1
+			if protectedSnapshots >= s.numSnapshotsToKeep {
+				return protectedEpochs
+			}
+		}
+	}
+
+	// The worst case when there aren't enough snapshots which are
+	// RollbackPointsSamplingRate apart from each other, then just
+	// protect the next s.numSnapshotsToKeep - lastProtectedSnapshot
+	// of snapshots.
+	for _, snapshot := range persistedSnapshots[lastProtectedSnapshot+
+		1 : lastProtectedSnapshot+1+s.numSnapshotsToKeep] {
+		protectedEpochs[snapshot.epoch] = struct{}{}
+	}
+	return protectedEpochs
+}
+
 // Removes enough snapshots from the rootBolt so that the
 // s.eligibleForRemoval stays under the NumSnapshotsToKeep policy.
 func (s *Scorch) removeOldBoltSnapshots() (numRemoved int, err error) {
-	persistedEpochs, err := s.RootBoltSnapshotEpochs()
+	persistedSnapshots, err := s.rootBoltSnapshotEpochTimeStamps()
 	if err != nil {
 		return 0, err
 	}
 
-	if len(persistedEpochs) <= s.numSnapshotsToKeep {
+	if len(persistedSnapshots) <= s.numSnapshotsToKeep {
 		// we need to keep everything
 		return 0, nil
 	}
 
-	// make a map of epochs to protect from deletion
-	protectedEpochs := make(map[uint64]struct{}, s.numSnapshotsToKeep)
-	for _, epoch := range persistedEpochs[0:s.numSnapshotsToKeep] {
-		protectedEpochs[epoch] = struct{}{}
-	}
+	protectedEpochs := s.getProtectedEpochs(persistedSnapshots)
 
 	var epochsToRemove []uint64
 	var newEligible []uint64
@@ -994,6 +1043,45 @@ func (s *Scorch) removeOldZapFiles() error {
 	s.rootLock.RUnlock()
 
 	return nil
+}
+
+type snapshotMetaData struct {
+	epoch     uint64
+	timeStamp time.Time
+}
+
+func (s *Scorch) rootBoltSnapshotEpochTimeStamps() ([]*snapshotMetaData, error) {
+	var rv []*snapshotMetaData
+	err := s.rootBolt.View(func(tx *bolt.Tx) error {
+		snapshots := tx.Bucket(boltSnapshotsBucket)
+		if snapshots == nil {
+			return nil
+		}
+		sc := snapshots.Cursor()
+		for sk, _ := sc.Last(); sk != nil; sk, _ = sc.Prev() {
+			_, snapshotEpoch, err := decodeUvarintAscending(sk)
+			if err != nil {
+				continue
+			}
+
+			snapshot := snapshots.Bucket(sk)
+			metaBucket := snapshot.Bucket(boltMetaDataKey)
+			if metaBucket == nil {
+				continue
+			}
+			timeStampBytes := metaBucket.Get(boltMetaDataTimeStamp)
+			var timeStamp time.Time
+			err = timeStamp.UnmarshalText(timeStampBytes)
+			if err != nil {
+				continue
+			}
+			rv = append(rv, &snapshotMetaData{
+				epoch:     snapshotEpoch,
+				timeStamp: timeStamp})
+		}
+		return nil
+	})
+	return rv, err
 }
 
 func (s *Scorch) RootBoltSnapshotEpochs() ([]uint64, error) {


index/scorch/persister.go
```diff
@@ -871,24 +886,79 @@ func (s *Scorch) removeOldData() {
...
+var RollbackSamplingInterval = 1 * time.Hour
+
+// getProtectedEpochs aims to fetch the epochs keep based on a timestamp basis.
+// It tries to get NumSnapshotsToKeep snapshots, each of which are separated
+// by a time duration of RollbackSamplingInterval.
+func (s *Scorch) getProtectedEpochs(
+	persistedSnapshots []*snapshotMetaData) map[uint64]struct{} {
+
+	// make a map of epochs to protect from deletion
+	protectedEpochs := make(map[uint64]struct{}, s.numSnapshotsToKeep)
```
abhinavdangeti, line 905: I don't think providing a length parameter while "make"ing over a map does anything does it? Just this would suffice ..

    protectedEpochs := make(map[uint64]struct{})

index/scorch/persister.go
```diff
@@ -871,24 +886,79 @@ func (s *Scorch) removeOldData() {
...
+	protectedEpochs[persistedSnapshots[0].epoch] = struct{}{}
+	nextSnapshotToProtect :=
+		persistedSnapshots[0].timeStamp.Add(s.rollbackSamplingInterval)
+	protectedSnapshots := 1
```
abhinavdangeti, line 909: No need for this, you can do len(protectedEpochs) to determine the number of protectedSnapshots.

index/scorch/persister.go
```diff
@@ -996,6 +1066,45 @@ func (s *Scorch) removeOldZapFiles() error {
 	return nil
 }
 
+type snapshotMetaData struct {
+	epoch     uint64
+	timeStamp time.Time
+}
+
+func (s *Scorch) rootBoltSnapshotEpochTimeStamps() ([]*snapshotMetaData, error) {
```
abhinavdangeti, line 1074: I'd suggest a better name, perhaps `rootBoltSnapshotMetaData()` owing to it's return values.

index/scorch/persister.go
```diff
@@ -996,6 +1066,45 @@ func (s *Scorch) removeOldZapFiles() error {
...
+			}
+			timeStampBytes := metaBucket.Get(boltMetaDataTimeStamp)
+			var timeStamp time.Time
+			err = timeStamp.UnmarshalText(timeStampBytes)
+			if err != nil {
+				continue
+			}
+			rv = append(rv, &snapshotMetaData{
+				epoch:     snapshotEpoch,
+				timeStamp: timeStamp})
```
abhinavdangeti, line 1101: nit:
```
                    rv = append(rv, &snapshotMetaData{
				epoch:     snapshotEpoch,
				timeStamp: timeStamp,
                    })
```

index/scorch/persister.go
```diff
@@ -871,24 +886,79 @@ func (s *Scorch) removeOldData() {
...
+	protectedSnapshots := 1
+	lastProtectedSnapshot := 0
+
+	for i := 1; i < len(persistedSnapshots); i++ {
+		// Finding that snapshot whose timestamp is just above
+		// the required timestamp, because the snapshot at the previous
+		// index is bound to be the closest possible snapshot to the
+		// required timestamp of nextSnapshotToProtect
+		if persistedSnapshots[i].timeStamp.After(nextSnapshotToProtect) {
+			protectedEpochs[persistedSnapshots[i-1].epoch] = struct{}{}
```
abhinavdangeti, line 918: Hmm this doesn't feel right, shouldn't you be adding to the map ..
```
protectedEpochs[persistedSnapshots[i].epoch] = struct{}{}
```

index/scorch/persister.go
```diff
@@ -871,24 +886,79 @@ func (s *Scorch) removeOldData() {
...
+	}
+
+	// The worst case when there aren't enough snapshots which are
+	// RollbackPointsSamplingRate apart from each other, then just
+	// protect the next s.numSnapshotsToKeep - protectedSnapshots
+	// of snapshots.
+	start := lastProtectedSnapshot + 1
+	end := start + s.numSnapshotsToKeep - protectedSnapshots
+
+	// If we don't have enough snapshots, just take all of them.
```
abhinavdangeti, line 936: Not a good idea, this'd violate the numSnapshotsToKeep contract. Also I feel the math here has been over engineered, and can be simplified.

abhinavdangeti (changes requested):

index/scorch/persister.go
Thejas-bhat: Done.

index/scorch/persister.go
Thejas-bhat: protectedSnapshots over here answers as to how many snapshots have been protected until now.

index/scorch/persister.go
Thejas-bhat: nextSnapshotToProtect essentially indicates the timestamp of the next snapshot we *want* to protect. There can be a situation that we don't have a snapshot in our list of persisted snapshots, with the exact timestamp that we're trying to find, during which we just protect the snapshot closest to the required timestamp, but less than it. So, we are trying to find the first timestamp (at index `i`) that's greater than the nextSnapshotToProtect and we can say for sure that the snapshot at `i-1` has a timestamp lesser than nextSnapshotToProtect which is bound to be the closest one.

index/scorch/persister.go
Thejas-bhat: Over here, I've favoured having enough (`numSnapshotsToKeep`) number of snapshots over the concept of picking specific rollback points that are separated by `rollbackSamplingInterval` duration. So, if we don't have "enough" snapshots in the persisted snapshots list to guarantee that we can get `numSnapshotsToKeep` number of protected snapshots, then we just protect the next numSnapshotsToKeep - protectedSnapshots number of snapshots.

However I'm still looking out for more edge cases that need to handled, and I'll add comments for the same.

index/scorch/persister.go
Thejas-bhat: Done.

index/scorch/persister.go
Thejas-bhat: Done.

index/scorch/persister.go
```diff
@@ -871,30 +888,118 @@ func (s *Scorch) removeOldData() {
...
 var NumSnapshotsToKeep = 1
 
+// RollbackSamplingInterval controls how far back we are looking
+// in the history to get the rollback points.
+// For example, the default value of 10 minutes ensures that the
+// protected snapshots (NumSnapshotsToKeep = 3) are:
+//    the very latest snapshot(ie the current one),
+//    the snapshot that was persisted 10 minutes before the current one,
+//	  the snapshot that was persisted 20 minutes before the current one
+var RollbackSamplingInterval = 2 * time.Minute
```
abhinavdangeti, line 898: Thought we'd default it to 0? So today's behavior will remain unchanged with this change. The user will optionally be able to configure this. For couchbase we'll set the config within cbft.

index/scorch/persister.go
```diff
@@ -871,30 +888,118 @@ func (s *Scorch) removeOldData() {
...
+
+// Controls what portion of the earlier rollback points to retain during
+// a infrequent/sparse mutation scenario
+var RollbackRetentionFactor = float64(0.5)
```
abhinavdangeti, line 902: Not sure I understand the purpose of this.

index/scorch/scorch.go
```diff
@@ -57,15 +57,18 @@ type Scorch struct {
...
-	closeCh            chan struct{}
-	introductions      chan *segmentIntroduction
-	persists           chan *persistIntroduction
-	merges             chan *segmentMerge
-	introducerNotifier chan *epochWatcher
-	persisterNotifier  chan *epochWatcher
-	rootBolt           *bolt.DB
-	asyncTasks         sync.WaitGroup
+	numSnapshotsToKeep       int
+	rollbackRetentionFactor  float64
```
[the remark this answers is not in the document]
Thejas-bhat: I've updated the description with more explanation about the latest patch. 

rollbackRetentionFactor refers to how much of the rollback checkpoints to conserve when we run the rootBoltSnapshotsMetaData after a long period of no mutations. This is because after this long time period of no mutations, because we delete the snapshots that are older than the numSnapshotsToKeep * rollbackSamplingInterval from the boltDB, we could stand to lose a lot of the rollback checkpoints and hence its necessary to retain a certain portion of it so that we maintain the effectiveness of these checkpoints (atleast to a certain degree).

index/scorch/scorch.go
```diff
@@ -57,15 +57,18 @@ type Scorch struct {
...
+	checkPoints              []*snapshotMetaData
```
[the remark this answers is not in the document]
Thejas-bhat: Yeah I've updated the reasoning behind this in the description.

index/scorch/persister.go
Thejas-bhat: Yep, my bad, I've included this in the recent patch

index/scorch/persister.go
Thejas-bhat: My bad, I've updated the description which clarifies this part

iamrajiv:
@Thejas-bhat One way to handle sparse/infrequent mutations is to use a hybrid approach that combines snapshot-based and log-based techniques. Snapshots are taken at regular intervals, and incremental logs are stored between snapshots. During a rollback, the system uses the latest snapshot and replays the mutations from the logs to restore the system to the desired state. This approach reduces data storage and provides recovery from failures/rollbacks. The retention policy can be dynamically adjusted based on mutation frequency and traffic patterns, rather than using a fixed policy.

iamrajiv (commented):
@Thejas-bhat One way to handle sparse/infrequent mutations is to use a hybrid approach that combines snapshot-based and log-based techniques. Snapshots are taken at regular intervals, and incremental logs are stored between snapshots. During a rollback, the system uses the latest snapshot and replays the mutations from the logs to restore the system to the desired state. This approach reduces data storage and provides recovery from failures/rollbacks. The retention policy can be dynamically adjusted based on mutation frequency and traffic patterns, rather than using a fixed policy.

Thejas-bhat:
@iamrajiv, thanks for pointing it out. The thing is that what you described as "replaying the snapshots from the logs" is something that a datastore type of a component would possibly do. 

The replaying part from the logs, basically means indexing back the corresponding set of documents by calling the indexing APIs. However, the way in which the "replay" happens could vary and made tunable as per the application, such as the number of batches you're using in this replay, equal sized vs variable sized batches etc. before passing on to bleve to index and recover back to the desired state (which need not be the same as the saved snapshots).
This particular scenario is something that a datastore, that can stream documents to be indexed, would be apt for.

index/scorch/persister.go
```diff
@@ -113,6 +114,7 @@ OUTER:
 		select {
 		case <-s.closeCh:
 			break OUTER
+
```
abhinavdangeti, line 117: Drop this (unrelated) new line.

index/scorch/persister.go
```diff
@@ -871,30 +888,124 @@ func (s *Scorch) removeOldData() {
...
+	numSnapshotsProtected := 1
+
+	// traverse the list in reverse order, older timestamps to newer ones.
+	for i := ptr - 1; i >= 0; i-- {
+		// If we find a timeStamp which is the next datapoint in our
+		// timeseries of snapshots, and newer by RollbackSamplingInterval duration
+		// (comparison in terms of minutes), which is the interval of our time
+		// series. In this case, add the epoch rv
+		if int(snapshots[i].timeStamp.Sub(snapshots[ptr].timeStamp).Minutes()) >
+			int(interval.Minutes()) {
```
abhinavdangeti, line 928: Why cast both the Minutes() above into int - we can simply leave them in the float64 format as it is just a greater than comparison.

index/scorch/persister.go
```diff
@@ -871,30 +888,124 @@ func (s *Scorch) removeOldData() {
...
+			rv[snapshots[i+1].epoch] = snapshots[i+1].timeStamp
+			ptr = i + 1
+			numSnapshotsProtected++
+		} else if int(snapshots[i].timeStamp.Sub(snapshots[ptr].timeStamp).Minutes()) ==
+			int(interval.Minutes()) {
```
abhinavdangeti, line 933: Ditto.

index/scorch/scorch.go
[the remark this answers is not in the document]
abhinavdangeti: Hmm, so this is the situation when there hasn't been a lot of mutations for a long period - in this situation I feel we will simply not hit a rollback situation at all in realistic situations. So that considered do we really need to preserve older snapshots in the event of no recent snapshots - whose epochs fall within `numSnapshotsToKeep * rollbackSamplingInterval`.

abhinavdangeti (commented):
@Thejas-bhat a couple more questions around the retention scheme you've used. Perhaps we can chat on this tomorrow.

index/scorch/persister.go
```diff
@@ -996,6 +1108,97 @@ func (s *Scorch) removeOldZapFiles() error {
...
 }
 
+// In low/sparse mutation scenarios, because there can be
+// a scenario where all (or most/lot of) the existing protected snapshots
+// are infact older than the expiration duration, we try
+// to retain atleast some of the protected snapshots (controlled
+// by retention factor), so that we don't start protected
+// contiguous snapshots (in which case we would not be protected
+// snapshots that are far apart for the rollback to be effective
+// enough)
```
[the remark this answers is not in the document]
Thejas-bhat: Done

index/scorch/persister.go
Thejas-bhat: Done

index/scorch/persister.go
Thejas-bhat: Done

index/scorch/persister.go
Thejas-bhat: Ah my bad

Thejas-bhat:
I think that the probability of hitting a rollback state that's older than `numSnapshotsToKeep * rollbackSamplingInterval` is low when we have the samplingInterval is of high value. Let's say 10 mins and numSnapshotsToKeep is 3. So, chances of rollback state older than 30 min would be very low in realistic situations. However, if the sampling interval is like 2 min or something, then the chances of hitting a state older than 6 mins need not be that low. So, it mainly depends on the sampling interval and number of snapshots to protect which is supposed to be configurable thing according to the application using bleve.

So, given this relative/application dependent situation, I think it's better to retain some of those older snapshots, because if we have negligible indexing traffic for greater than `numSnapshotsToKeep * rollbackSamplingInterval` duration and also if the application is having this type of traffic a lot in its index lifetime, then we would end up storing contiguous snapshots (the old way) more often than the proposed method (which is supposed to be better). Now, since rollback can happen at any point of time and if we store contiguous snapshots most of the times, we would be likely to have a poor rollback behaviour.

abhinavdangeti:
> So, given this relative/application dependent situation, I think it's better to retain some of those older snapshots, because if we have negligible indexing traffic for greater than numSnapshotsToKeep * rollbackSamplingInterval duration and also if the application is having this type of traffic a lot in its index lifetime, then we would end up storing contiguous snapshots (the old way) more often than the proposed method (which is supposed to be better). Now, since rollback can happen at any point of time and if we store contiguous snapshots most of the times, we would be likely to have a poor rollback behaviour.

By older snapshots - if you mean those with epochs older than `numSnapshotsToKeep * rollbackSamplingInterval` then I just worry this could cause un-predictable behavior. My earlier argument (at least for Couchbase) is - if there's not been a lot of traffic the chance that we'll end up in a rollback situation is going to be quite low. So, it should simply be up to the application to decide what to set for `numSnapshotsToKeep` and `rollbackSamplingInterval` per their usage pattern and our job is to only retain snapshots whose epochs fall within these constraints.

abhinavdangeti (approved):

PR #1785, opened 2023-01-27
perf: Optimize `sync.Pool` usage by storing pointers instead of non-pointer values

iamrajiv:
## Description

By storing the pointer to the slice in the `sync.Pool`, the function is able to reuse the same memory without allocating additional memory.


iamrajiv:
@abhinavdangeti Can you review it once?

abhinavdangeti (approved):

abhinavdangeti:
@iamrajiv Thanks for the contribution.

PR #1788, opened 2023-02-16
Display 0 bytesRead in search responses

metonymic-smokey:
Currently, when firing multiple queries with "score" : "none", the first query response displays the non-zero bytesRead as part of the JSON response. Subsequent queries do not display it as part of the response since the bytesRead is 0.
This PR fixes it such that even if the bytesRead is 0, it displays it as part of the response.

### DIFF
--- a/search.go
+++ b/search.go
@@ -490,7 +490,7 @@ type SearchResult struct {
 	Request   *SearchRequest                 `json:"request"`
 	Hits      search.DocumentMatchCollection `json:"hits"`
 	Total     uint64                         `json:"total_hits"`
-	BytesRead uint64                         `json:"bytesRead,omitempty"`
+	BytesRead uint64                         `json:"bytesRead"`
 	MaxScore  float64                        `json:"max_score"`
 	Took      time.Duration                  `json:"took"`
 	Facets    search.FacetResults            `json:"facets"`


abhinavdangeti:
@metonymic-smokey and why is the bytesRead 0 when `"score":"none"` and non-zero only in the first response?

iamrajiv (changes requested):
While it is true that removing the `omitempty`option from the `json` tag of the `BytesRead`
field would ensure that the field is included in the JSON response, even if it's zero, using a `MarshalJSON`function like this is a good practice. This function creates an `Alias` type with the same field names and types as the original `SearchResult`
struct, and marshals a new anonymous struct that includes a pointer to the `Alias` object. This way, you can customize the way the JSON output is generated, such as including or omitting certain fields or modifying their values. Maybe we can have custom `MarshalJSON` functions in utility packages that we can use in our code.

```go
func (sr *SearchResult) MarshalJSON() ([]byte, error) {
    type Alias SearchResult
    return json.Marshal(&struct {
        *Alias
        BytesRead uint64 `json:"bytesRead,omitempty"`
    }{
        Alias:     (*Alias)(sr),
        BytesRead: sr.BytesRead,
    })
}
```

metonymic-smokey:
> @metonymic-smokey and why is the bytesRead 0 when `"score":"none"` and non-zero only in the first response?

Regarding queries where score is "none":
If the segment in a snapshot is already mmaped, bytes read will not be incremented for it. Since it is mmaped for the first query(and not for subsequent queries), bytesRead is incremented only for the first query.
Also, in the postings list iterator() function, since includeFreqNorm and includeLocs are false, bytesRead will not be incremented there either.

This behaviour is on expected lines, as far as I remember. Thanks!

metonymic-smokey:
> While it is true that removing the `omitempty`option from the `json` tag of the `BytesRead` field would ensure that the field is included in the JSON response, even if it's zero, using a `MarshalJSON`function like this is a good practice. This function creates an `Alias` type with the same field names and types as the original `SearchResult` struct, and marshals a new anonymous struct that includes a pointer to the `Alias` object. This way, you can customize the way the JSON output is generated, such as including or omitting certain fields or modifying their values. Maybe we can have custom `MarshalJSON` functions in utility packages that we can use in our code.
> 
> ```go
> func (sr *SearchResult) MarshalJSON() ([]byte, error) {
>     type Alias SearchResult
>     return json.Marshal(&struct {
>         *Alias
>         BytesRead uint64 `json:"bytesRead,omitempty"`
>     }{
>         Alias:     (*Alias)(sr),
>         BytesRead: sr.BytesRead,
>     })
> }
> ```

@iamrajiv thanks for the suggestion, but I don't think this requires custom JSON marshalling/unmarshalling.

Thejas-bhat (approved):

abhinavdangeti (approved):

PR #1829, opened 2023-05-30
recording the cost for different types of queries

Thejas-bhat:
- Adds a new callback function SearchIncrementalCostCallbackFn which is used to track incremental cost updates for the query. It can also be used to track costs for different types of queries as well.

- Renames "bytesRead" to "search_cost"

### DIFF
--- a/index/scorch/snapshot_index_tfr.go
+++ b/index/scorch/snapshot_index_tfr.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-
 	"reflect"
 	"sync/atomic"
 
@@ -55,16 +54,6 @@ type IndexSnapshotTermFieldReader struct {
 
 func (i *IndexSnapshotTermFieldReader) incrementBytesRead(val uint64) {
 	i.bytesRead += val
-}
-
-func aggregateBytesRead(ctx context.Context, bytes uint64) {
-	queryType, ok := ctx.Value(search.QueryTypeKey).(string)
-	if ok {
-		aggCallbackFn := ctx.Value(search.SearchCostAggregatorKey)
-		if aggCallbackFn != nil {
-			aggCallbackFn.(search.SearchCostAggregatorCallbackFn)("add", queryType, bytes)
-		}
-	}
 }
 
 func (i *IndexSnapshotTermFieldReader) Size() int {
@@ -217,7 +206,7 @@ func (i *IndexSnapshotTermFieldReader) Close() error {
 		}
 
 		// todo: should this is be per hit?
-		aggregateBytesRead(i.ctx, i.bytesRead)
+		search.RecordSearchCost(i.ctx, "add", i.bytesRead)
 	}
 
 	if i.snapshot != nil {
--- a/index/scorch/snapshot_index_tfr.go
+++ b/index/scorch/snapshot_index_tfr.go
@@ -205,7 +205,7 @@ func (i *IndexSnapshotTermFieldReader) Close() error {
 			statsCallbackFn.(search.SearchIOStatsCallbackFunc)(i.bytesRead)
 		}
 
-		search.RecordSearchCost(i.ctx, "add", i.bytesRead)
+		search.RecordSearchCost(i.ctx, search.AddM, i.bytesRead)
 	}
 
 	if i.snapshot != nil {
--- a/index/scorch/snapshot_index_tfr.go
+++ b/index/scorch/snapshot_index_tfr.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+
 	"reflect"
 	"sync/atomic"
 
@@ -112,10 +113,10 @@ func (i *IndexSnapshotTermFieldReader) Next(preAlloced *index.TermFieldDoc) (*in
 			// this is because there are chances of having a series of loadChunk calls,
 			// and they have to be added together before sending the bytesRead at this point
 			// upstream.
-			if delta := i.iterators[i.segmentOffset].BytesRead() - prevBytesRead; delta > 0 {
-				i.incrementBytesRead(delta)
+			bytesRead := i.iterators[i.segmentOffset].BytesRead()
+			if bytesRead > prevBytesRead {
+				i.incrementBytesRead(bytesRead - prevBytesRead)
 			}
-
 			return rv, nil
 		}
 		i.segmentOffset++
--- a/index/scorch/snapshot_index_tfr.go
+++ b/index/scorch/snapshot_index_tfr.go
@@ -56,6 +56,16 @@ func (i *IndexSnapshotTermFieldReader) incrementBytesRead(val uint64) {
 	i.bytesRead += val
 }
 
+func aggregateBytesRead(ctx context.Context, bytes uint64) {
+	queryType, ok := ctx.Value(search.QueryTypeKey).(string)
+	if ok {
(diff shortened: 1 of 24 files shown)


index/scorch/snapshot_index_tfr.go
```diff
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+
```
metonymic-smokey, line 21: Nit: unnecessary newline maybe?

Thejas-bhat: done. thanks

search/util.go
```diff
@@ -67,3 +69,32 @@ func MergeFieldTermLocations(dest []FieldTermLocation, matches []*DocumentMatch)
 
 	return dest
 }
+
+const SearchIOStatsCallbackKey = "_search_io_stats_callback_key"
+
+type SearchIOStatsCallbackFunc func(uint64)
+
+// The callback signature is (message, queryType, cost) which allows
+// the caller to act on a particular query type and what its the associated
```
moshaad7, line 78: Can we concise "query type and what its the associated cost of an operation." 
to
"query type and it's associated cost of operation"

search/util.go
```diff
@@ -67,3 +69,32 @@ func MergeFieldTermLocations(dest []FieldTermLocation, matches []*DocumentMatch)
...
+// cost of an operation. "add" indicates to increment the cost for the query
```
moshaad7, line 79: Values of message argument (like "add" and "done" ) aren't defined by bleve, it is upto the user of bleve.

If you have specified "add" and "done" as examples of message argument value, then we can frame it that way

search/util.go
Thejas-bhat: will do, thanks

search/util.go
Thejas-bhat: hmm, the cost callback according to me includes very basic set of messages, which are add, done and maybe abort, since its mainly enabling the caller to track the costs incrementally and act upon that along with a "done" type of a message that will help the caller to act upon the tracked bytes until now. perhaps an abort invocation can be done when there is a ctx.Done() in the collector phase (which would happen when the search's context is cancelled i believe) in that case, the caller can handle separately perhaps.

moshaad7: Got your point. 
Bleve will invoke the callback with these messages "add", "done", "abort" etc ... 
And it is the responsibility of the author of the callback to properly handle these messages.

In this case, we can frame our comment like this:
"Implementation of SearchIncrementalCostCallbackFn should handle the following messages
    - add: "description ..."
    - abort:  description ...
    - done:  description ...
"

We can even assign IDs to these messages and export from bleve

type IncCostMessage uint

const (
	AddM = Service(1 << iota) // description
	AbortM // description
	DoneM // description
)

type SearchIncrementalCostCallbackFn func(IncCostMessage, string, uint64)

Thoughts ?

Thejas-bhat: yeah i've renamed the messages and query types, thanks

index_impl.go
```diff
@@ -495,11 +495,13 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 			err = serr
 		}
 		if sr != nil {
-			sr.BytesRead = totalBytesRead
+			sr.Cost = totalSearchCost
 		}
```
metonymic-smokey, line 499: Maybe it's just me who feels this way, but I think the previous "BytesRead" was better than Cost. Cost is essentially being added for serverless and I think the concept of serverless should be restricted to cbft. BytesRead is a general enough name. Thoughts?

search/util.go
```diff
@@ -67,3 +69,52 @@ func MergeFieldTermLocations(dest []FieldTermLocation, matches []*DocumentMatch)
...
 	return dest
 }
+
+const SearchIOStatsCallbackKey = "_search_io_stats_callback_key"
+
+type SearchIOStatsCallbackFunc func(uint64)
+
+// Implementation of SearchIncrementalCostCallbackFn should handle the following messages
+//   - add: increment the cost of a search operation
+//     (which can specific to a query type as well)
```
metonymic-smokey, line 79: Nit: "can be specific..."

search/util.go
```diff
@@ -67,3 +69,52 @@ func MergeFieldTermLocations(dest []FieldTermLocation, matches []*DocumentMatch)
...
+//   - abort: query was aborted due to a cancel of search's context (for eg),
+//     which can be handled differently as well
+//   - done: indicates that a search was complete and the tracked cost can be
+//     handled safely by the implementation.
+type SearchIncrementalCostCallbackFn func(SearchIncrementalCostCallbackMsg,
```
metonymic-smokey, line 84: Nit: I think we should do away with the notion of "cost" in names within bleve.

Thejas-bhat: i feel that the notion of "cost" isn't really specific to cbft, since it makes sense to associate each search with a "cost" which corresponds to whats the bytes read from the mmaped file

search/util.go
Thejas-bhat: done, thanks

index_impl.go
Thejas-bhat: the reason why i changed from bytesRead to cost is because i thought the name cost seems generic enough such that if in future the scale of the value changed (KiB or MiB etc.) or if we decided to return the CPU cycles this search corresponded to rather than bytes read from the mmaped file, the name "cost" would still hold true in that case.

metonymic-smokey (approved):

metonymic-smokey (approved):

index_impl.go
moshaad7: Agree  @metonymic-smokey

search.go
```diff
@@ -486,14 +486,14 @@ func (ss *SearchStatus) Merge(other *SearchStatus) {
 // A SearchResult describes the results of executing
 // a SearchRequest.
 type SearchResult struct {
```
abhinavdangeti, line 488: Would you add commentary above this construct explaining what each of the response attribute stands for. This would then include what our definition for `Cost` is - the user can benefit from this with godocs.

search.go
```diff
@@ -486,14 +486,14 @@ func (ss *SearchStatus) Merge(other *SearchStatus) {
...
-	Total     uint64                         `json:"total_hits"`
-	BytesRead uint64                         `json:"bytesRead"`
-	MaxScore  float64                        `json:"max_score"`
-	Took      time.Duration                  `json:"took"`
-	Facets    search.FacetResults            `json:"facets"`
+	Status   *SearchStatus                  `json:"status"`
+	Request  *SearchRequest                 `json:"request"`
+	Hits     search.DocumentMatchCollection `json:"hits"`
+	Total    uint64                         `json:"total_hits"`
+	Cost     uint64                         `json:"search_cost"`
```
abhinavdangeti, line 493: Perhaps the json bit should simply be `cost` - given that this is already part of `SearchResult`.

abhinavdangeti (approved):
Couple comments, looks good otherwise.

abhinavdangeti (approved):

index_impl.go
```diff
@@ -582,7 +584,9 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 		if err != nil {
 			return nil, err
 		}
-		totalBytesRead += storedFieldsBytes
+		totalSearchCost += storedFieldsBytes
+
+		search.RecordSearchCost(ctx, search.AddM, storedFieldsBytes)
```
abhinavdangeti, line 589: Make this one call as opposed to per hit.

PR #1832, opened 2023-06-14
MB-57394: Consolidate documentMappingForPath & closestDocMapping

abhinavdangeti:
+ This is to half the number of times we iterate through the document mappings while mapping a document.


moshaad7 (approved):
+1

metonymic-smokey (approved):

PR #1836, opened 2023-07-06
Avoid double counting when sorting by field for faceted query

metonymic-smokey:
When a query sorts by the same field as the facets, the number of search results per facet doubles. This is because there is double counting of the documents by the collector on that field for both sorting and faceting.

This PR deduplicates the fields in the collector to avoid this.

### DIFF
--- a/search/collector/topn_test.go
+++ b/search/collector/topn_test.go
@@ -19,7 +19,9 @@ import (
 	"context"
 	"testing"
 
+	"github.com/blevesearch/bleve/v2/index/scorch"
 	"github.com/blevesearch/bleve/v2/search"
+	"github.com/blevesearch/bleve/v2/search/facet"
 	index "github.com/blevesearch/bleve_index_api"
 )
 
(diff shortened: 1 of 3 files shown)


Thejas-bhat (approved):

abhinavdangeti (changes requested):
Let's add a unit test test for this - the issue should be be easily reproduced.
Also create an MB for this and let's use it in the commit header.

metonymic-smokey:
Verification re the unit test:
Without the changes, the unit test fails with the following message:
```
aditiahuja@Aditis-MacBook-Pro collector % sudo go test --run=TestSetFacetsBuilder
--- FAIL: TestSetFacetsBuilder (0.00s)
    topn_test.go:768: expected fields in collector: [locations], observed: [locations locations]
FAIL
exit status 1
FAIL    github.com/blevesearch/bleve/v2/search/collector        0.558s

```

search/collector/topn_test.go
```diff
@@ -724,6 +727,48 @@ func TestCollectorChaining(t *testing.T) {
...
+	}
+	err = i.Open()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return i
+}
+
+func TestSetFacetsBuilder(t *testing.T) {
```
abhinavdangeti, line 749: Good test case, would you add an full index and search test case in search_test.go as well ..

- create an index
- add a few documents with doc values etc.
- search request with sort and facets
- make sure facet counts are accurate

abhinavdangeti (approved):
Well done.

metonymic-smokey:
Verification without change:
```
--- FAIL: TestSortedFacetedQuery (0.56s)
    search_test.go:114: expected 2, got 4
    search_test.go:114: expected 1, got 2
FAIL
exit status 1
FAIL    github.com/blevesearch/bleve/v2 1.799s
```

PR #1847, opened 2023-07-20
MB-31210: Support Fuzzy MatchPhrase, MultiPhrase and Phrase queries

CascadingRadium:
## Jira

[MB-31210](https://issues.couchbase.com/browse/MB-31210)

## Description
Added fuzziness support for MatchPhrase queries.

### DIFF
--- a/search/query/multi_phrase.go
+++ b/search/query/multi_phrase.go
@@ -30,7 +30,6 @@ type MultiPhraseQuery struct {
 	Field     string     `json:"field,omitempty"`
 	BoostVal  *Boost     `json:"boost,omitempty"`
 	Fuzziness int        `json:"fuzziness"`
-	Prefix    int        `json:"prefix_length"`
 }
 
 // NewMultiPhraseQuery creates a new Query for finding
@@ -53,10 +52,6 @@ func (q *MultiPhraseQuery) SetFuzziness(f int) {
 	q.Fuzziness = f
 }
 
-func (q *MultiPhraseQuery) SetPrefix(p int) {
-	q.Prefix = p
-}
-
 func (q *MultiPhraseQuery) SetBoost(b float64) {
 	boost := Boost(b)
 	q.BoostVal = &boost
@@ -67,7 +62,7 @@ func (q *MultiPhraseQuery) Boost() float64 {
 }
 
 func (q *MultiPhraseQuery) Searcher(ctx context.Context, i index.IndexReader, m mapping.IndexMapping, options search.SearcherOptions) (search.Searcher, error) {
-	return searcher.NewMultiPhraseSearcher(ctx, i, q.Terms, q.Prefix, q.Fuzziness, q.Field, q.BoostVal.Value(), options)
+	return searcher.NewMultiPhraseSearcher(ctx, i, q.Terms, q.Fuzziness, q.Field, q.BoostVal.Value(), options)
 }
 
 func (q *MultiPhraseQuery) Validate() error {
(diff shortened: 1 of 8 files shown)


search/searcher/search_phrase_test.go
```diff
@@ -130,7 +130,7 @@ func TestMultiPhraseSearch(t *testing.T) {
 		if err != nil {
 			t.Error(err)
 		}
-		searcher, err := NewMultiPhraseSearcher(nil, reader, test.phrase, "desc", soptions)
+		searcher, err := NewMultiPhraseSearcher(nil, reader, test.phrase, "desc", soptions, 0, 0, 1.0)
```
metonymic-smokey, line 133: I think adding a new UT to demonstrate fuzziness would be good.

CascadingRadium: UT?

search/query/multi_phrase.go
```diff
@@ -26,9 +26,11 @@ import (
...
 
 type MultiPhraseQuery struct {
-	Terms    [][]string `json:"terms"`
-	Field    string     `json:"field,omitempty"`
-	BoostVal *Boost     `json:"boost,omitempty"`
+	Terms     [][]string `json:"terms"`
+	Field     string     `json:"field,omitempty"`
+	BoostVal  *Boost     `json:"boost,omitempty"`
+	Fuzziness int        `json:"fuzziness"`
+	Prefix    int        `json:"prefix_length"`
```
abhinavdangeti, line 33: I'm not sure prefix makes much sense with multi phrase, because the order of terms with multi phrase query does not matter. Also this is more in line with `TermQuery` so let's keep the definition of the MultiPhraseQuery like it was before and apply these new attributes only for a `MatchPhraseQuery`.

CascadingRadium: But a match phrase query / phrase query is internally a multi-phrase query, so it must also be supported.
A prefix essentially is applied to the fuzzy search of each term in the Phrase, so it works out like so, each word will be fuzzily matched to a word in the index, provided that a prefix for the word is present.
The unit test file also reflects the same.

abhinavdangeti (changes requested):
Recommending a code change here.

search/query/multi_phrase.go
CascadingRadium: basically yeah, lets say i have a match phrase for 
"Apple banana" it becomes a multi phrase query for
[["apple"]["banana"]], and for both of these terms, we have a fuzzy search being done,
So FuzzySearch ("apple", Fuzziness, prefix) is done, so both fuzziness+prefix must be passed from the top level searcher.

If i have MultiPhraseQuery of:
[[ "apple"   ],[ "banana" ] ]
      "red"           "yellow"

Q1 = DisjunctionQuery of FuzzyQueries over "apple" and "red"
Q2 = DisjunctionQuery of FuzzyQueries over "banana" and "yellow"
And result Q = ConjQuery(Q1,Q2). So i have to send prefix via the top searcher or hard code it to 0 while allowing for a user defined fuzziness

CascadingRadium: In Elastic search, Fuzzy Match phrase is not supported out of the box, but can be implemented using some workaround as suggested in this post:
https://stackoverflow.com/questions/38816955/elasticsearch-fuzzy-phrases

Basically use a span query that allows fine-tuning - **for each word the user can give a fuzziness and a prefix length** there as in this image 
image

The below image shows a match provided a per word fuzziness and prefix
image


So what i can suggest is - We allow this restricted match phrase query which keeps the fuzziness and prefix same throughout, and then add a functionality to implement span queries for more fine tuning.

CascadingRadium: Done

abhinavdangeti (approved):
Looks fine to me. 
@CascadingRadium Let's wait for one more review possibly.

PR #1849, opened 2023-07-24
MB-57926 - Searcher for similarity search 

metonymic-smokey:


index_impl.go
```diff
@@ -482,7 +483,16 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 	ctx = context.WithValue(ctx, search.SearchIOStatsCallbackKey,
 		search.SearchIOStatsCallbackFunc(sendBytesRead))
 
-	searcher, err := req.Query.Searcher(ctx, indexReader, i.m, search.SearcherOptions{
+	// Using a conjunction query to get intersection of results from similarity query
+	// and the original query
+	searchQuery := req.Query
+	if req.Similarity != nil {
+		similarityQuery := query.NewSimilarityQuery(req.Similarity.Vector)
+		similarityQuery.SetFieldVal(req.Similarity.Field)
```
Thejas-bhat, line 491: I'm thinking if it makes sense to have this as a set of fields instead, and what you're trying to do is apply the similarity search over each of those fields. So, you'd be getting the similar vectors from all these indexed fields and do a conjunction of these sub-similarity searchers and the user's non-similarity query.

metonymic-smokey: As per discussion, the user can enter multiple fields, not just one.

metonymic-smokey: @abhinavdangeti do you think it's a valid use case where the user enters multiple fields against 1 vector?
@Thejas-bhat mentioned that a user could have multiple fields with dense vector content and might want to query the same vector across them and hence, the need for multiple fields.

search.go
```diff
@@ -282,9 +282,16 @@ type SearchRequest struct {
 	SearchAfter      []string          `json:"search_after"`
 	SearchBefore     []string          `json:"search_before"`
 
+	Similarity *SimilarityRequest `json:"similarity"`
```
metonymic-smokey, line 285: Is Similarity an appropriate name?

search/searcher/search_similarity.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// TODO What does this mean?
+func (s *SimilaritySearcher) Min() int {
+	return 0
+}
+
+func (s *SimilaritySearcher) Next(ctx *search.SearchContext) (*search.DocumentMatch, error) {
+	// how to get doc matches here?
+
```
metonymic-smokey, line 58: look at code around optimizing conjunctions -> should not push similarity query lower in the hierarchy.

search.go
abhinavdangeti: Let's use kNN everywhere - KNNSearcher, kNNMatch etc.

search/scorer/scorer_similarity.go
```diff
@@ -57,17 +61,35 @@ func (sqs *SimilarityQueryScorer) Score(ctx *search.SearchContext,
...
 func (sqs *SimilarityQueryScorer) Weight() float64 {
-	return 1
+	return sqs.queryBoost * sqs.queryBoost
+}
+
+func (sqs *SimilarityQueryScorer) SetQueryNorm(qnorm float64) {
+	sqs.queryNorm = qnorm
+
+	// update the query weight
+	sqs.queryWeight = sqs.queryBoost
```
Thejas-bhat, line 94: the queryWeight for a child searcher is normalized weights (with respect to other child searchers in a disjunction searcher, for eg). so it's better to do a 

`sqs.queryWeight = sqs.queryBoost * sqs.queryNorm`

mapping/index.go
```diff
@@ -447,6 +448,21 @@ func (im *IndexMappingImpl) FieldAnalyzer(field string) string {
...
+	for _, v := range im.TypeMapping {
+		for field, property := range v.Properties {
+			for _, v1 := range property.Fields {
+				if field == path {
+					// Return field mapping if the name matches the path param.
+					return v1
+				}
+			}
+		}
+	}
```
Thejas-bhat, line 462: i think you might need to check the default mapping for the path as well (if the type mappings don't have the path in it). just like how AnalyzerNameForPath(paht string) does

mapping/mapping.go
```diff
@@ -55,4 +55,6 @@ type IndexMapping interface {
 
 	AnalyzerNameForPath(path string) string
 	AnalyzerNamed(name string) analysis.Analyzer
+
+	FieldMappingForPath(path string) *FieldMapping
```
Thejas-bhat, line 59: perhaps its better to not return a pointer to the fieldMapping. mainly because i think this interface deals with reading of the content in the index mapping, and passing a pointer to the fieldMapping could lead to users having the option to modify the config of the field mapping at instances when they aren't allowed to do so (for example query time).

abhinavdangeti:
Closing in favor of https://github.com/blevesearch/bleve/pull/1869

PR #1858, opened 2023-07-26
MB-27696: Allow a unique ID to be set for a Search Request

CascadingRadium:
## Jira

[MB-27696](https://issues.couchbase.com/browse/MB-27696)

## Description
Add a unique ID for a bleve Search Request which will be echo'd back in the Search Result. Would be useful in a parallel search query scenario to differentiate between the queries.


search.go
```diff
@@ -269,6 +269,7 @@ func (h *HighlightRequest) AddField(field string) {
 //
 // A special field named "*" can be used to return all fields.
 type SearchRequest struct {
+	ClientContextID  string            `json:"client_context_id"`
```
abhinavdangeti, line 272: Allow for `omitempty`

search.go
```diff
@@ -426,11 +433,12 @@ func NewSearchRequest(q query.Query) *SearchRequest {
 // By default results are ordered by score, descending.
 func NewSearchRequestOptions(q query.Query, size, from int, explain bool) *SearchRequest {
 	return &SearchRequest{
-		Query:   q,
-		Size:    size,
-		From:    from,
-		Explain: explain,
-		Sort:    search.SortOrder{&search.SortScore{Desc: true}},
+		ClientContextID: search.NewUUID(),
```
abhinavdangeti, line 436: Let's not do this by default here.

In bleve's purview it should remain something that's exclusively set by the client. So for couchbase, cbft will be responsible for generating a UUID if and only if the user does not already provide one and some config is enabled requiring us to generate for every request.

abhinavdangeti (changes requested):

abhinavdangeti (approved):

abhinavdangeti (approved):

Thejas-bhat (approved):

PR #1859, opened 2023-07-28
MB-57973: Validate custom date time parser layout

CascadingRadium:
## Jira

[MB-57973](https://issues.couchbase.com/browse/MB-57973)

## Description
Currently user does not receive any feedback if the custom date time parser input by him is valid or not. These custom parsers must only contain magic numbers/ constants specified in https://pkg.go.dev/time#pkg-constants else it is not valid even though no error is returned buy the time package. The input validation here ensures only valid magic numbers are used.


analysis/datetime/flexible/flexible.go
```diff
@@ -44,6 +45,52 @@ func (p *DateTimeParser) ParseDateTime(input string) (time.Time, error) {
 	return time.Time{}, analysis.ErrInvalidDateTime
 }
 
+func validateLayout(layout string) bool {
+	validMagicNumbers := map[string]bool{
```
abhinavdangeti, line 49: map[string]struct{}{

abhinavdangeti: Also let's initialize this construct globally as opposed to setting up a map on every call to validateLayout.

analysis/datetime/flexible/flexible.go
```diff
@@ -44,6 +45,52 @@ func (p *DateTimeParser) ParseDateTime(input string) (time.Time, error) {
...
+		"MST":     true,
+		"Z0700":   true, // prints Z for UTC
+		"Z070000": true,
+		"Z07":     true,
+		"0700":    true,
+		"070000":  true,
+		"07":      true,
+		"":        true,
+	}
+	re := regexp.MustCompile("[- :T,\\.<>;\\?!`~@#$%\\^&\\*|\\(\\){}\\[\\]/\\\\]")
```
abhinavdangeti, line 80: Same for the regexp, move the initialization and all the common stuff to outside this method.

analysis/datetime/flexible/flexible.go
```diff
@@ -53,6 +100,11 @@ func DateTimeParserConstructor(config map[string]interface{}, cache *registry.Ca
 	for _, layout := range layouts {
 		layoutStr, ok := layout.(string)
 		if ok {
+			if !validateLayout(layoutStr) {
+				return nil, fmt.Errorf("invalid datetime parser layout: %s,"+
+					" please refer to https://pkg.go.dev/time#pkg-constants for valid"+
+					" constants to use", layoutStr)
```
abhinavdangeti, line 106: .. please refer to .. for supported layouts*.

abhinavdangeti (changes requested):

analysis/datetime/flexible/flexible.go
```diff
@@ -44,6 +77,24 @@ func (p *DateTimeParser) ParseDateTime(input string) (time.Time, error) {
...
 }
 
+func stripLayout(layout string) string {
+	for _, k := range layoutConstants {
+		layout = strings.ReplaceAll(layout, k, "")
+	}
+	return layout
+}
+
+func validateLayout(layout string) bool {
```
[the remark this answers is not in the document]
CascadingRadium: ack

abhinavdangeti (approved):

abhinavdangeti (changes requested):
Sorry - missed one comment in previous review. Would you add a couple of unit tests exercising the new code. Add a full stack unit test rather than just the datetime parsing, so perhaps in search_test.go.

abhinavdangeti (dismissed):

abhinavdangeti:
Let's drop this check, because of the pickle we can get into during an online upgrade if the layout defined was incorrect.

abhinavdangeti (commented):
Re-considering this change because the failure in the upgrade paths (if the user was using invalid layouts before) is similar to a situation when the user created and used an analyzer component called `xy` and then upgraded to a version where we introduced a stock component `xy` - we'd see a similar validation/unmarshal error.

abhinavdangeti (approved):

abhinavdangeti:
Closed in favor of https://github.com/blevesearch/bleve/pull/1877

PR #1860, opened 2023-07-31
MB-58034: Custom Date Time parsers not being applied on query object

CascadingRadium:
## Jira

[MB-58034](https://issues.couchbase.com/browse/MB-58034)

## Description

- Add a new query DateRangeStringQuery, which takes in the start and end bounds as strings and then parses them using either the top level QueryDateTimeParser in the top-level bleve package or the DateTimeParser set in the DateRangeStringQuery struct, if present.
-  ParseQuery, which analyzes a marshalled json to identify the query to run, will now default to DateRangeStringQuery instead, which is better since it already has the dates in string format and unmarshalling to DateRangeQuery will always trigger parsing of the date string with Only QueryDateTimeParser, leaving no room for user config.


search/query/date_range.go
```diff
@@ -30,10 +30,10 @@ import (
 	index "github.com/blevesearch/bleve_index_api"
 )
 
-// QueryDateTimeParser controls the default query date time parser
+// QueryDateTimeParser controls the default query date time parser in User Path.
```
abhinavdangeti, line 33: Not a fan of the "User path" wording - I understand what you are intending it for, but I think the name of the attribute is self-explanatory enough. let's drop the "User Path" verbiage.

search/query/date_range.go
```diff
@@ -84,19 +85,74 @@ func (t *BleveQueryTime) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func DateRangeUnmarshal(input []byte, obj *DateRangeQuery) error {
```
abhinavdangeti, line 88: Don't see a need for this method to be exported.

Also perhaps we should change the signature to this method to  ..

    dateRangeQuery, err := UnmarshalDateRange(input)

Separately can you help clarify the need for this validation here?

search/query/date_range.go
```diff
@@ -135,16 +191,33 @@ func (q *DateRangeQuery) Field() string {
...
-	if err != nil {
-		return nil, err
-	}
-
 	field := q.FieldVal
 	if q.FieldVal == "" {
 		field = m.DefaultSearchField()
 	}
-
+	if q.rawStart != "" || q.rawEnd != "" {
```
abhinavdangeti, line 198: Hmm, should probably do this string job only when index mapping for this field has store enabled and the query requested for this field ?

abhinavdangeti (changes requested):

search/query/date_range.go
abhinavdangeti: Also one rule of thumb - do not add anything "couchbase" specific here. Keep things generic enough.

search/query/date_range.go
CascadingRadium: ack i will not export
signature is the same for all json.Unmarshall calls in parse.go.
So i have kept it consistent

I am basically avoiding the usage of UnmarshalJSON() which is called when json.Unmarshal is called, for a date range query object in the ParseQuery path. By using a custom unmarshaller i can avoid entering the User Path here. Thus i can keep the start and end as raw strings and NOT time objects. 

When the searcher func is called, i have access to the index mapping and can get the layout to parse the raw start and end strings with.

search/query/date_range.go
CascadingRadium: the object here is a query object right? like this check is just done to check if the unmarshaller used was my custom one (which is the only place where raw start and raw end is actually set) or the default one used in User Path. these 2 params are not exported so bleve only users cant see this

search/query/date_range.go
```diff
@@ -117,6 +193,32 @@ func NewDateRangeInclusiveQuery(start, end time.Time, startInclusive, endInclusi
...
+	return NewDateRangeRawInclusiveQuery(start, end, nil, nil)
+}
+
+// NewDateRangeInclusiveQuery creates a new Query for ranges
+// of date values.
+// Date strings are parsed using the DateTimeParser configured for the
+// queried field in the index mapping.
+// Either, but not both endpoints can be nil.
+// startInclusive and endInclusive control inclusion of the endpoints.
+func NewDateRangeRawInclusiveQuery(start, end string, startInclusive, endInclusive *bool) *DateRangeQuery {
```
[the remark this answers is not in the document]
CascadingRadium: *bool was the existing method parameter in the code for NewDateRange and NewDateRangeInclusive. So i have maintained the same here as well.

Thejas-bhat (approved):

Thejas-bhat:
the changes look good to me, however i'd strongly recommend waiting for others' review as well.

search/query/date_range_string.go
```diff
@@ -0,0 +1,177 @@
+//  Copyright (c) 2014 Couchbase, Inc.
```
Thejas-bhat, line 1: oh missed this, as suggested in other PRs as well. this must be 2023.

mapping/index.go
```diff
@@ -17,6 +17,7 @@ package mapping
 import (
 	"encoding/json"
 	"fmt"
+
 	index "github.com/blevesearch/bleve_index_api"
```
abhinavdangeti, line 22: Drop this empty line between these imports.

search/query/date_range_string.go
```diff
@@ -0,0 +1,177 @@
...
+func (q *DateRangeStringQuery) SetDateTimeParser(d string) {
+	q.DateTimeParser = d
+}
+
+func (q *DateRangeStringQuery) DateTimeParserName() string {
+	return q.DateTimeParser
+}
+
+func (q *DateRangeStringQuery) Searcher(ctx context.Context, i index.IndexReader, m mapping.IndexMapping, options search.SearcherOptions) (search.Searcher, error) {
+
```
abhinavdangeti, line 99: Drop empty line.

abhinavdangeti (commented):
Looking good. Couple minor nits, we can merge this after those are addressed.

abhinavdangeti (approved):

PR #1862, opened 2023-08-01
MB-57972: Add new layout to dateTimeOptional

CascadingRadium:
## Jira

[MB-57972](https://issues.couchbase.com/browse/MB-57972)

## Description
Currently in Couchbase sample datasets:
1. gamesim-sample: No date fields are there
2. travel-sample: A reviews.date field is there with the layout YYYY:MM:DD HH:MM:SS Offset
3. beer-sample: A date field is there with layout YYYY:MM:DD HH:MM:SS
 
A default mapping (uncustomized) will use dateTimeOptional which has layouts to parse beer-sample but not travel-sample, because of this the user needs to set a custom date time parser to do a date-range query on travel sample. To avoid this and enchance UX, add YYYY:MM:DD HH:MM:SS Offset layout to dateTimeOptional

### DIFF
--- a/analysis/datetime/flexible/flexible_test.go
+++ b/analysis/datetime/flexible/flexible_test.go
@@ -69,7 +69,7 @@ func TestFlexibleDateTimeParser(t *testing.T) {
 
 	rfc3339NoTimezone := "2006-01-02T15:04:05"
 	rfc3339NoTimezoneNoT := "2006-01-02 15:04:05"
-	rfc3339Offest := "2006-01-02 15:04:05 -0700"
+	rfc3339Offset := "2006-01-02 15:04:05 -0700"
 	rfc3339NoTime := "2006-01-02"
 
 	dateOptionalTimeParser := New(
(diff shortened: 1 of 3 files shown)


analysis/datetime/optional/optional.go
```diff
@@ -26,13 +26,15 @@ const Name = "dateTimeOptional"
 
 const rfc3339NoTimezone = "2006-01-02T15:04:05"
 const rfc3339NoTimezoneNoT = "2006-01-02 15:04:05"
+const rfc3339Offest = "2006-01-02 15:04:05 -0700"
```
abhinavdangeti, line 29: Offset*

analysis/datetime/flexible/flexible_test.go
```diff
@@ -64,6 +69,7 @@ func TestFlexibleDateTimeParser(t *testing.T) {
 
 	rfc3339NoTimezone := "2006-01-02T15:04:05"
 	rfc3339NoTimezoneNoT := "2006-01-02 15:04:05"
+	rfc3339Offest := "2006-01-02 15:04:05 -0700"
```
abhinavdangeti, line 72: Offset*

abhinavdangeti (changes requested):

analysis/datetime/optional/optional.go
CascadingRadium: ack

analysis/datetime/flexible/flexible_test.go
CascadingRadium: ack

abhinavdangeti (approved):

PR #1864, opened 2023-08-08
MB-54131: Geoshape query decode optimization

Likith101:
- Added a buffer pool to be shared by all documents

Related PRs
 - https://github.com/blevesearch/geo/pull/14


search/searcher/search_geoshape.go
```diff
@@ -70,6 +70,20 @@ func buildRelationFilterOnShapes(ctx context.Context, dvReader index.DocValueRea
 	var dvShapeValue []byte
 	var startReading, finishReading bool
 	var reader *bytes.Reader
+
+	bufPool := make([][]byte, 11)
+	bufPool[0] = make([]byte, 8192*3)
```
Thejas-bhat, line 75: since the buffer sizes follow a step like fashion, maybe defining a constant with respect to the biggest buffer size and allocating a buf, buf/2, buf/4 etc. amount of space looks more readable

search/searcher/search_geoshape.go
```diff
@@ -70,6 +70,20 @@ func buildRelationFilterOnShapes(ctx context.Context, dvReader index.DocValueRea
 	var dvShapeValue []byte
 	var startReading, finishReading bool
 	var reader *bytes.Reader
+
+	bufPool := make([][]byte, 11)
```
moshaad7, line 74: We are creating a buffer pool per GeoShapeSearcher, 
But a Complex Query might end up creating multiple GeoShapeSearchers per query, 
In that case we will create that many buffer pools.

Example:
disjunct of a geoShape Query with a {conjunct of a geoShape Query AND match query } 
will end up creating two GeoShapeSearchers.

We should think about a way to create pool Per-Query and not Per-GeoShapeSearchers

moshaad7: In future (may be phase-2), 
We can create a concurrent safe Dynamic Buffer pool per Bleve Import.

Pool Structure:
Pool: 
    buffers of size1
    buffers of size2
    ....
    
list of buffers of a specific size can be dynamic and can grow/shrink based on the exponential weighted moving average of usage metric of it.

search/searcher/search_geoshape.go
```diff
@@ -70,6 +70,20 @@ func buildRelationFilterOnShapes(ctx context.Context, dvReader index.DocValueRea
...
+	bufPool[0] = make([]byte, 8192*3)
+	bufPool[1] = make([]byte, 4096*3)
+	bufPool[2] = make([]byte, 2048*3)
+	bufPool[3] = make([]byte, 1024*3)
+	bufPool[4] = make([]byte, 512*3)
+	bufPool[5] = make([]byte, 256*3)
+	bufPool[6] = make([]byte, 128*3)
```
moshaad7, line 81: As discussed, you can lazily initialise the buffer of a particular size only when there is a demand of it.

search/searcher/search_geoshape.go
Thejas-bhat: a suggestion over here is perhaps use the `ctx context.Context` that's been passed to the function. What I mean is if in index_impl.go where you're defining the single searcher (which internally can have multiple child searchers which are geo searchers) you pass set a callback function which returns a reference to the buffer pool, then the child searchers when they want to get a certain sized buffer from the pool can consult the common pool to grab the required buffer size. Furthermore, things are not really parallel in terms of Next() calls etc. so I don't think there can be a racy situation keeping that mind (please do double check it out though). As an example, you can take a look as to how `search.SearchIOStatsCallbackKey` is set and used throughout the queries etc. as a reference and then use something similar for the buffer pool stuff.

Likith101: For now I have left the creation of the buffer pool in the searcher, I will move it to the context once I test out if the calls are parallel and figure out how to implement it.

index_impl.go
```diff
@@ -482,6 +483,14 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 	ctx = context.WithValue(ctx, search.SearchIOStatsCallbackKey,
 		search.SearchIOStatsCallbackFunc(sendBytesRead))
 
+	bufPool := s2.NewGeoBufferPool(24 * 1024, 24)
```
moshaad7, line 486: MinBufPoolSize and MaxBufPoolSize as arguments instead of raw numbers

index_impl.go
```diff
@@ -482,6 +483,14 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
+	getBufferPool := func() *s2.GeoBufferPool {
+		return bufPool
```
moshaad7, line 488: For queries which doesn't have geoshape component in it, we shouldn't call NewGeoBufferPool.
To avoid that, 

```
var geoBufPool *s2.GeoBufferPool

getBufferPool := func() *s2.GeoBufferPool {
  if geoBufPool == nil {
    geoBufPool = s2.NewGeoBufferPool(MinBufPoolSize, MaxBufPoolSize)
  }
  return geoBufPool
}
```

index_impl.go
```diff
@@ -482,6 +483,18 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
+	getBufferPool := func() *s2.GeoBufferPool {
+		if bufPool == nil {
+			bufPool = s2.NewGeoBufferPool(search.MaxBufPoolSize, search.MinBufPoolSize)
+		}
+
+		return bufPool
+	}
+
+	ctx = context.WithValue(ctx, search.GeoBufferPoolCallbackKey,
+		search.GeoBufferPoolCallbackFunc(getBufferPool))
```
abhinavdangeti, line 496: Why declare this here and send that in as context to retrieved by the geoshape searcher?

It seems we're not really sharing this across searchers either, so why not simply declare this where we need it?

search/util.go
```diff
@@ -118,3 +122,13 @@ func RecordSearchCost(ctx context.Context,
...
 	}
 }
+
+const GeoBufferPoolCallbackKey = "_geo_buffer_pool_callback_key"
+
+// Assigning the size of the largest buffer in the pool to 24KB and
+// the smallest buffer to 24 bytes. The pools are used to read a
+// sequence of vertices which are always 24 bytes each.
+const MaxBufPoolSize = 24 * 1024
+const MinBufPoolSize = 24
```
abhinavdangeti, line 132: Update these names to better reflect what they're intended for - so minGeoBufPoolSize and maxGeoBufPoolSize.

abhinavdangeti (changes requested):
Update go.mod to use v0.1.18 for blevesearch/geo.

index_impl.go
Likith101: The buffer is declared in SearchInContext and attached to the context so that any searcher with the geo component can access this function. Only the first searcher will trigger the creation of the bufPool while the other calls will simply get the reference to this pool. I have tested this path using a conjunction query with 2 geoshape components and it seems to be working as intended

go.sum
```diff
@@ -92,8 +92,10 @@ golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a/go.mod h1:oPkhp1MJrh7nUepCBc
 golang.org/x/sys v0.5.0 h1:MUK/U/4lj1t1oPg0HfuXDN/Z1wv31ZJ/YcPiGccS4DU=
 golang.org/x/sys v0.5.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/text v0.3.0/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
+golang.org/x/text v0.3.2/go.mod h1:bEr9sfX3Q8Zfm5fL9x+3itogRgK3+ptLWKqgva+5dAk=
```
abhinavdangeti, line 95: Line 95 & 98 below seem unrelated.

PR #1870, opened 2023-08-17
MB-58134: Skip parsing date time fields with timestamps 

CascadingRadium:
# Jira

[MB-58134](https://issues.couchbase.com/browse/MB-58134)

## Description
A user can set one of four values as the DateFormat for a DatetimeFieldMapping
- "unix_micro"
- "unix_sec"
- "unix_milli"
- "unix_nano"

This indicates that the field has a UNIX epoch timestamp in microseconds/seconds/milliseconds/nanoseconds.
Such fields should not be parsed with a date time parser since they already contain the timestamp.
User can perform date range queries on these fields like normal.


analysis/type.go
```diff
@@ -99,6 +100,56 @@ func (a *DefaultAnalyzer) Analyze(input []byte) TokenStream {
...
+const UnixMilliSecs = "unix_milli"
+const UnixMicroSecs = "unix_micro"
+const UnixNanoSecs = "unix_nano"
+
+type TimestampBounds struct {
+	Min int64
+	Max int64
+}
+
+var UnixTimestampFormats = map[string]*TimestampBounds{
```
Thejas-bhat, line 113: i think its better to not use pointer/reference type over here, mainly since these formats are supposed to be constant and overriding of the values within these formats should be forbidden.

analysis/type.go
```diff
@@ -99,6 +100,56 @@ func (a *DefaultAnalyzer) Analyze(input []byte) TokenStream {
...
+		Min: math.MinInt64 / 1000,
+		Max: math.MaxInt64 / 1000,
+	},
+	UnixNanoSecs: {
+		Min: math.MinInt64,
+		Max: math.MaxInt64,
+	},
+}
+
+func convertTimestamp(timestamp int64, format string) int64 {
```
Thejas-bhat, line 132: i'm thinking perhaps using a switch case each of which modify a common return value timestamp and having a single return timestamp at the end looks better.

analysis/type.go
```diff
@@ -99,6 +100,56 @@ func (a *DefaultAnalyzer) Analyze(input []byte) TokenStream {
...
+	}
+	return timestamp
+}
+
+// ValidateAndConvertTimestamp validates the timestamp against the bounds and
+// converts it to the nanoseconds if valid.
+func ValidateAndConvertTimestamp(timestamp int64, bounds *TimestampBounds, format string) (int64, error) {
+	if timestamp > bounds.Min && timestamp < bounds.Max {
+		return convertTimestamp(timestamp, format), nil
+	} else {
```
Thejas-bhat, line 148: this else condition is not needed, might as well return the 0, error at the end of the function. it looks more readable that way i think

mapping/document.go
```diff
@@ -436,6 +440,22 @@ func (dm *DocumentMapping) processProperty(property interface{}, path []string,
...
 		}
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		if subDocMapping != nil {
+			// index by explicit mapping in case of datetime
+			for _, fieldMapping := range subDocMapping.Fields {
+				if fieldMapping.Type == "datetime" {
+					// This number is a UNIX timestamp
+					// hence must parse as a datetime
+					bounds, isUnixFormat := analysis.UnixTimestampFormats[fieldMapping.DateFormat]
+					if isUnixFormat {
```
Thejas-bhat, line 450: isUnixFormat -> validFormat maybe? also the common go pattern is to just use "ok" i think, you can use that as well over here.

mapping/field.go
```diff
@@ -259,6 +273,23 @@ func (fm *FieldMapping) processFloat64(propertyValFloat float64, pathString stri
...
 
+func (fm *FieldMapping) processTimestamp(unixTimestamp int64, pathString string, path []string, indexes []uint64, context *walkContext) {
+	fieldName := getFieldName(pathString, path, fm)
+	if fm.Type == "datetime" {
+		options := fm.Options()
+		field, err := document.NewDateTimeFieldWithTimestamp(fieldName, indexes, unixTimestamp, options)
+		if err == nil {
+			context.doc.AddField(field)
+		} else {
+			logger.Printf("could not build date %v", err)
```
Thejas-bhat, line 284: please make sure that these debug logs are cleaned up in your final patch

Thejas-bhat (commented):
please add some unit tests along these code paths as well.

analysis/datetime/timestamp/microseconds/microseconds.go
```diff
@@ -0,0 +1,52 @@
+//  Copyright (c) 2014 Couchbase, Inc.
```
metonymic-smokey, line 1: Nit: pls change 2014 to 2023 in the license header.

mapping/index.go
```diff
@@ -17,6 +17,7 @@ package mapping
 import (
 	"encoding/json"
 	"fmt"
+
 	index "github.com/blevesearch/bleve_index_api"
```
abhinavdangeti, line 22: Drop this new line between the imports.

abhinavdangeti (commented):
Looking good, minor nits.

abhinavdangeti (dismissed):

abhinavdangeti (dismissed):

CascadingRadium:
merge conflict again 😅

Thejas-bhat (approved):

PR #1871, opened 2023-08-18
Replacing deprecated io/ioutil functions

LimJiAn:
I think it would be better to change the io/ioutil package to the os package.
Please check the [link](https://go.dev/doc/go1.16#ioutil) and code and thank you.

Bleve search is using go 1.19 so I think it is better not to use io/ioutil package.

### DIFF
--- a/config.go
+++ b/config.go
@@ -16,7 +16,7 @@ package bleve
 
 import (
 	"expvar"
-	"io/ioutil"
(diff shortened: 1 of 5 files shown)


abhinavdangeti (approved):
Thanks for this contribution @LimJiAn .

PR #1874, opened 2023-08-28
go fmt

moshaad7:


geo/parse.go
```diff
@@ -24,15 +24,24 @@ import (
 // interpret it is as geo point.  Supported formats:
 // Container:
 // slice length 2 (GeoJSON)
-//  first element lon, second element lat
+//
```
abhinavdangeti, line 27: Why the new empty line here, think you should undo this file's change and simply remove that white space before the text in the comment on line 27, 29, 31, 33 before running `go fmt ./...`

query.go
```diff
@@ -48,7 +48,9 @@ func NewConjunctionQuery(conjuncts ...query.Query) *query.ConjunctionQuery {
 // NewDateRangeQuery creates a new Query for ranges
 // of date values.
 // Date strings are parsed using the DateTimeParser configured in the
-//  top-level config.QueryDateTimeParser
```
abhinavdangeti, line 51: ditto

query.go
```diff
@@ -57,7 +59,9 @@ func NewDateRangeQuery(start, end time.Time) *query.DateRangeQuery {
 // NewDateRangeInclusiveQuery creates a new Query for ranges
 // of date values.
 // Date strings are parsed using the DateTimeParser configured in the
-//  top-level config.QueryDateTimeParser
```
abhinavdangeti, line 60: ditto

query.go
moshaad7: This is for the benefit of [gopls](https://pkg.go.dev/golang.org/x/tools/gopls)

query.go
moshaad7: This is for the benefit of [gopls](https://pkg.go.dev/golang.org/x/tools/gopls)

abhinavdangeti (approved):

PR #1878, opened 2023-09-06
MB-58388: Use custom datetime parsers in date range facet queries

CascadingRadium:
# Jira

[MB-58388](https://issues.couchbase.com/browse/MB-58388)

## Description
 
Related to Date range facet queries:

- The "start" and "end" fields of the Facet Result struct are now formatted with the layout that successfully parsed them in the Facet Request struct. Earlier we always used RFC3339Nano by default which led to disparity between the corresponding fields in the facet request and result. 
   Ex - Facet Request Date Range:

          { 
              start : "2014-09-10"
              end : "2015-08-11"
          }

Previously the Facet result had 

          { 
              start : "2014-09-10T00:00:00Z"
              end : "2015-08-11T00:00:00Z"
              Count: 100
          }

Now it has 

          { 
              start : "2014-09-10"
              end : "2015-08-11"
              Count: 100
          }
- Allow a custom date time parser to be added to the date range struct of a facet request, that will override the default date time parser (dateTimeOptional) while parsing the start and end date strings.
- Add error message when the date time parser fails to parse the date string. Earlier if the date time parser failed to parse the start and end strings of the date range in the facet request, no error was thrown and the facet request would be for start and end being equal to 0 time, so facet result was always nil.
- Fix marshalling error for dateRangeQuery struct. If the query had no end / endString, which is the case when upper bound is unspecified, the marshalled JSON always had the 0 time (time.Time{}) as the value for end, now it is omitted from the marshalled JSON.


index_impl.go
```diff
@@ -530,10 +530,23 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 			} else if facetRequest.DateTimeRanges != nil {
 				// build date range facet
 				facetBuilder := facet.NewDateTimeFacetBuilder(facetRequest.Field, facetRequest.Size)
-				dateTimeParser := i.m.DateTimeParserNamed("")
+				dateTimeParserName := defaultDateTimeParser
 				for _, dr := range facetRequest.DateTimeRanges {
-					start, end := dr.ParseDates(dateTimeParser)
-					facetBuilder.AddRange(dr.Name, start, end)
+					if dr.DateTimeParser != "" {
+						dateTimeParserName = dr.DateTimeParser
```
abhinavdangeti, line 536: Hmm could there be a situation here, where one of the facet range has a datetime parser defined while the other doesn't. In this situation, we should fall back to using the defaultDateTimeParser for the 2nd facet as opposed to previous facet's parser.

This code will need correction to ..

```
dateTimeParsersName := defaultDateTimeParser
if dr.DateTimeParser != "" {
    dateTimeParserName = dr.DateTimeParser
}
```

abhinavdangeti (changes requested):

index_impl.go
CascadingRadium: ok yea my bad

index_impl.go
```diff
@@ -530,10 +530,23 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
 				for _, dr := range facetRequest.DateTimeRanges {
-					start, end := dr.ParseDates(dateTimeParser)
-					facetBuilder.AddRange(dr.Name, start, end)
+					dateTimeParserName := defaultDateTimeParser
+					if dr.DateTimeParser != "" {
+						dateTimeParserName = dr.DateTimeParser
+					}
+					dateTimeParser := i.m.DateTimeParserNamed(dateTimeParserName)
+					if dateTimeParser == nil {
+						return nil, fmt.Errorf("no date time parser named `%s` registered", dateTimeParserName)
```
abhinavdangeti, line 540: Wondering if the better thing to do here would skip this data point as opposed to failing the entire facets computations. This is applicable to 544 and 547 as well. Thoughts?

search.go
```diff
@@ -46,36 +46,44 @@ func init() {
...
+func (dr *dateTimeRange) ParseDates(dateTimeParser analysis.DateTimeParser) (start, end time.Time, startLayout, endLayout string, err error) {
 	start = dr.Start
+	startLayout = time.RFC3339Nano
 	if dr.Start.IsZero() && dr.startString != nil {
-		s, _, err := dateTimeParser.ParseDateTime(*dr.startString)
-		if err == nil {
-			start = s
+		s, layout, parseError := dateTimeParser.ParseDateTime(*dr.startString)
+		if parseError != nil {
+			return start, end, startLayout, endLayout, fmt.Errorf("error parsing start date '%s' for date range name '%s': %v", *dr.startString, dr.Name, parseError)
```
abhinavdangeti, line 63: Ditto, if the date time fails the parser, think it better we remove this datetime from the equation than failing.

search.go
```diff
@@ -46,36 +46,44 @@ func init() {
...
 	}
 	end = dr.End
+	endLayout = time.RFC3339Nano
 	if dr.End.IsZero() && dr.endString != nil {
-		e, _, err := dateTimeParser.ParseDateTime(*dr.endString)
-		if err == nil {
-			end = e
+		e, layout, parseError := dateTimeParser.ParseDateTime(*dr.endString)
+		if parseError != nil {
+			return start, end, startLayout, endLayout, fmt.Errorf("error parsing end date '%s' for date range name '%s': %v", *dr.endString, dr.Name, parseError)
```
abhinavdangeti, line 73: ditto

index_impl.go
CascadingRadium: but then the user will not be notified why there are no facet results right? Previous to this commit, the behaviour for ParseDates error was ambiguous, if i ran a query with malformed entries (datetime parser fails) on bleve directly the results were indeed dropped, i.e facet result did not have the corresponding date range. But running via couchbase, an error was thrown "start and end cannot be nil" for the same query. This was because the validate facet request() was called first, and then the query was executed. (validate facet request() throws that error on dateTimeParser failure).

So if validation() throws an error shouldn't running the actual query throw the same error too? The main goal also was to fix the error message thrown.

abhinavdangeti: Not always "no facet results" though. It could be a situation where some documents have malformed content while many/most others don't. This is entirely up to the user. I worry if we take this away - the change in behavior may not be taken lightly by some.

CascadingRadium: but this parseDates function is to parse the start and end sate strings of the FacetQuery object, it is unrelated to document match contents.

abhinavdangeti: Ah I see. My mistake. This looks ok then 👍🏼

abhinavdangeti (approved):

PR #1880, opened 2023-09-08
Support overrideable json Marshaler/Unmarshaler

abhinavdangeti:


Thejas-bhat (approved):

search_test.go
```diff
@@ -3228,4 +3228,4 @@ func TestDateRangeTimestampQueries(t *testing.T) {
 			}
 		}
 	}
-}
\ No newline at end of file
+}
```
moshaad7, line 3231: unrelated change

index/upsidedown/upsidedown.go
```diff
@@ -135,22 +135,22 @@ func (udc *UpsideDownCouch) loadSchema(kvreader store.KVReader) (err error) {
 }
```
moshaad7, line 136: This file have unrelated changes coming from go fmt, 
we have another PR raised for "go fmt" changes (https://github.com/blevesearch/bleve/pull/1874/) based off of master.
We can rebase this PR on top of that.

abhinavdangeti: This is just a go fmt and given that it'll be backported as is - let's leave this as is for now and rebase #1874 instead.

PR #1887, opened 2023-10-10
support hooks in analysis pipeline

moshaad7:


analysis/type.go
```diff
@@ -73,7 +73,89 @@ type TokenFilter interface {
...
-	Analyze([]byte) TokenStream
+	Type() string
+	// return value of this method depends on the type of analyzer
+	Analyze([]byte) interface{}
+}
+
+const (
+	TokensAnalyzerType     = "token"
+	HookTokensAnalyzerType = "hook_token"
+	VectorAnalyzerType     = "vector"
```
moshaad7, line 84: move vector related stuff to a separate file with "vector" build tag

analysis/type.go
```diff
@@ -73,7 +73,89 @@ type TokenFilter interface {
...
+	Tokens TokenStream
+	Err    error
+}
+type VectorAnalyzer []float64
+type HookVectorAnalyzer struct {
+	Vector []float64
+	Err    error
+}
+
+func AnalyzeForTokens(analyzer Analyzer, input []byte) (TokenStream, error) {
```
moshaad7, line 102: add comment:
// A utility function, helpful for analyzing an input to generate TokenStream ( and error, if any )

Previously, Analyze() method of an analyzed to return TokenStream.
But as per the change in this PR, Analyze() method will now return a value of type interface{} , validating and using which can be done based on analyzer.Type().
For the benefit of  users of old Analyzer interface, this utiity will come handly , to migrate to new Analyzer interface.

analysis/type.go
```diff
@@ -73,7 +73,89 @@ type TokenFilter interface {
...
+	analyzerType := analyzer.Type()
+	if analyzerType != TokensAnalyzerType &&
+		analyzerType != HookTokensAnalyzerType {
+		return nil, fmt.Errorf("cannot analyze text with analyzer of type: %s",
```
moshaad7, line 106: alternate error msg: "given analyzer is not compatible to be used as a token analyzer"

Issue #1873, opened 2023-08-25
github.com/blevesearch/bleve/v2/http/search.go bug?


hktalent:
github.com/blevesearch/bleve/v2/http/search.go bug?
No ctx is passed in, when the network times out, or the client cancels the request, the query is not terminated gracefully
https://github.com/blevesearch/bleve/blob/6c3999222a52e2281f01fb64181096efc2ef1a98/http/search.go#L84

change to :
```
	// execute the query
	searchResponse, err := index.SearchInContext(req.Context(),&searchRequest)
```

Likith101:
Added the option of giving the timeout as part of the request which will then be passed on to the search function. The case of client cancelling a request does not make much sense in a http handler function and so hasn't been added.

PR #1898, opened 2023-11-08
#1873: Added timeout option in the Search Handler

Likith101:
- Added timeout value, which the user can pass through request parameters
 - Create context with the timeout value if present
 - Change the call to SearchInContext

### DIFF
--- a/http/search.go
+++ b/http/search.go
@@ -15,10 +15,13 @@
 package http
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
+	"time"
 
 	"github.com/blevesearch/bleve/v2"
 	"github.com/blevesearch/bleve/v2/search/query"
@@ -80,8 +83,22 @@ func (h *SearchHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		}
 	}
 
+	// check for timeout and create context
+	var ctx context.Context
+	timeoutStr := req.FormValue("timeout")
+	if timeoutStr == "" {
+		ctx = context.Background()
+	} else {
+		timeout, err := strconv.Atoi(timeoutStr)
+		if err != nil {
+			showError(w, req, fmt.Sprintf("error parsing timeout value: %v", err), 400)
+			return
+		}
+		ctx, _ = context.WithTimeout(context.Background(), time.Duration(timeout)*time.Millisecond)
+	}
+
 	// execute the query
-	searchResponse, err := index.Search(&searchRequest)
+	searchResponse, err := index.SearchInContext(ctx, &searchRequest)
 	if err != nil {
 		showError(w, req, fmt.Sprintf("error executing query: %v", err), 500)
 		return


http/search.go
```diff
@@ -80,8 +83,22 @@ func (h *SearchHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		}
 	}
 
+	// check for timeout and create context
+	var ctx context.Context
+	timeoutStr := req.FormValue("timeout")
+	if timeoutStr == "" {
+		ctx = context.Background()
+	} else {
+		timeout, err := strconv.Atoi(timeoutStr)
```
moshaad7, line 92: Instead of this, you can use the ParseDuration function from golang's time package.

```

// ParseDuration parses a duration string.
// A duration string is a possibly signed sequence of
// decimal numbers, each with optional fraction and a unit suffix,
// such as "300ms", "-1.5h" or "2h45m".
// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
func ParseDuration(s string) (Duration, error) {
```

http/search.go
```diff
@@ -80,8 +83,22 @@ func (h *SearchHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		if err != nil {
+			showError(w, req, fmt.Sprintf("error parsing timeout value: %v", err), 400)
+			return
+		}
+		ctx, _ = context.WithTimeout(context.Background(), time.Duration(timeout)*time.Millisecond)
+	}
+
 	// execute the query
-	searchResponse, err := index.Search(&searchRequest)
+	searchResponse, err := index.SearchInContext(ctx, &searchRequest)
```
moshaad7, line 101: SearchInContext implementation of indexImpl and indexAliasImpl will have to adapt to this change.
They will now multiplex on ctx.Done() and result channel

Something like this:
```
select {
  case <-ctx.Done():
          fmt.Printf("timeout")

          return nil, timeoutError
  case res := <-resChan: // A goroutine will carry out the search and will update this channel with the result
           fmt.Printf("success")
           return res, nil
  }

```

moshaad7: Also, To test the correctness of the implementation.
We should write a test case, where we can set a very small timeout value (few ms) and see if SearchInContext is stopped on that value or not.


All the goroutines spawned by SearchInContext ( and their children) should obey the timeout (stop the work).

moshaad7: ---
I see (*TopNCollector)Collect is already taking care context deadline.
```
if hc.total%CheckDoneEvery == 0 {
	select {
	case <-ctx.Done():
		search.RecordSearchCost(ctx, search.AbortM, 0)
		return ctx.Err()
	default:
	}
}
```
So, maybe SearchInContext impls don't need any change.

metonymic-smokey:
@Likith101 small tip: On github, pls use "fixes #1873 " or the like in the comment body since that auto closes issues when the fix is merged.

http/search.go
Likith101: Hey Shaad, I tested the implementation outside of the ServeHTTP function with a test case resembling the same implementation of using SearchInContext with a timeout. It does work as intended. But I was wondering which file this test case should go into

http/search.go
Likith101: Will use

http/search.go
Likith101: There are already existing test cases for search in context where the context is created using the timeout value

abhinavdangeti (approved):

PR #1899, opened 2023-11-08
Adding KNN scorer explanation

metonymic-smokey:
This PR adds a score Explanation to the KNN scorer, along with a unit test.
Also, contains a minor refactor of scoring related code.

### DIFF
--- a/search/scorer/scorer_knn.go
+++ b/search/scorer/scorer_knn.go
@@ -83,8 +83,8 @@ func (sqs *KNNQueryScorer) Score(ctx *search.SearchContext,
 		}
 
 		if sqs.options.Explain {
-			childrenExplanations := make([]*search.Explanation, 1)
-			childrenExplanations[0] = &search.Explanation{
+			childExplanations := make([]*search.Explanation, 1)
+			childExplanations[0] = &search.Explanation{
 				Value: score,
 				Message: fmt.Sprintf("vector(field(%s:%s) with similarity_metric(%s)=%e",
 					sqs.queryField, knnMatch.ID, sqs.similarityMetric, score),
@@ -93,7 +93,7 @@ func (sqs *KNNQueryScorer) Score(ctx *search.SearchContext,
 				Value: score,
 				Message: fmt.Sprintf("fieldWeight(%s in doc %s), score of:",
 					sqs.queryField, knnMatch.ID),
-				Children: childrenExplanations,
+				Children: childExplanations,
 			}
 		}
 
--- a/search/scorer/scorer_knn.go
+++ b/search/scorer/scorer_knn.go
@@ -18,6 +18,7 @@
 package scorer
 
 import (
+	"fmt"
 	"math"
 	"reflect"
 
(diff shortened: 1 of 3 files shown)


search/scorer/scorer_knn.go
```diff
@@ -33,28 +34,25 @@ func init() {
...
-	queryBoost       float64
-	queryNorm        float64
-	docTerm          uint64
-	docTotal         uint64
-	options          search.SearcherOptions
-	includeScore     bool
-	similarityMetric string
+	queryVector            []float32
+	queryField             string
+	queryWeight            float64
```
moshaad7, line 39: How large can these value be queryWeight, queryBoost, queryNorm ?

Instead of float64, can we use something smaller ?

moshaad7: nvm, this was unrelated to the change.
But, we can still think about it :)

metonymic-smokey: Yep, mainly used these since term scorer too used the same.

abhinavdangeti:
p.s. the check failures aren't related to this PR.
```
--- FAIL: TestBytesWritten (0.30s)
Error:     index_test.go:309: expected bytes written is 37767, got 10
--- FAIL: TestBytesRead (0.36s)
Error:     index_test.go:405: expected bytes read for query string 32349, got 18138
--- FAIL: TestBytesReadStored (0.28s)
Error:     index_test.go:584: expected the bytes read stat to be around 25928, got 7999
```

It's possible @Thejas-bhat is already looking into these. But, if not yet - @metonymic-smokey would you work with Thejas in determining the cause here or the need to update the unit test expectations for bytes read/written with index sections.

metonymic-smokey:
@abhinavdangeti pls review this PR instead.
Also added a UT for score = 0 with euclidean distance, since you last reviewed.

metonymic-smokey:
So in the case where we set the max score to a large float if score = 0 for euclidean distance, this is what the explanation ends up looking like:
```
"value": 1.7976931348623157e+308,
  "message": "fieldWeight(desc in doc one), score of:",
  "children": [
    {
      "value": 1.7976931348623157e+308,
      "message": "vector(field(desc:one) with similarity_metric(l2_norm)=179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368.000"
    }
  ]
} 

I'm concerned about the large number present in the child message - can't truncate the first part so not sure what to do here. Any thoughts?
Once that's decided, I'll modify the UT to adhere to that.
```

search/scorer/scorer_knn.go
```diff
@@ -72,19 +70,50 @@ func (sqs *KNNQueryScorer) Score(ctx *search.SearchContext,
...
+					sqs.queryField, knnMatch.ID),
+				Children: childrenExplanations,
+			}
 		}
 
 		// if the query weight isn't 1, multiply
 		if sqs.queryWeight != 1.0 && score != maxKNNScore {
 			score = score * sqs.queryWeight
+			if sqs.options.Explain {
+				childExplanations := make([]*search.Explanation, 2)
```
abhinavdangeti, line 104: It seems we're double allocating unnecessarily above and below, this should be within an if and else.

abhinavdangeti: @metonymic-smokey address this separate please.

abhinavdangeti (approved):
The improvement can be addre

PR #1900, opened 2023-11-08
MB-59507: Move to latests of zapx and go-faiss@v1.0.2

abhinavdangeti:


abhinavdangeti:
Ignoring the checks' failures for now - will be addressed eventually.

PR #1903, opened 2023-11-10
MB-59421: add validation for vector field aliases

moshaad7:
vector field Aliases (fields with same name and type as vector) must have same value for dimensions and similarity, respectively.

### DIFF
--- a/mapping/mapping_vectors.go
+++ b/mapping/mapping_vectors.go
@@ -114,6 +114,7 @@ func validateVectorFieldAlias(field *FieldMapping, parentName string,
 	field.SkipFreqNorm = true
 
 	// # If alias is present, validate the field options as per the alias
+	// note: reading from a nil map is safe
 	if fieldAlias, ok := fieldAliasCtx[field.Name]; ok {
 		if field.Dims != fieldAlias.Dims {
 			return fmt.Errorf("field: '%s', invalid alias "+
@@ -143,7 +144,9 @@ func validateVectorFieldAlias(field *FieldMapping, parentName string,
 			reflect.ValueOf(index.SupportedSimilarityMetrics).MapKeys())
 	}
 
-	fieldAliasCtx[field.Name] = field
+	if fieldAliasCtx != nil { // writing to a nil map is unsafe
+		fieldAliasCtx[field.Name] = field
+	}
 
 	return nil
 }
--- a/mapping/mapping_vectors.go
+++ b/mapping/mapping_vectors.go
@@ -26,6 +26,12 @@ import (
 	index "github.com/blevesearch/bleve_index_api"
 )
 
+// Min and Max allowed dimensions for a vector field
+const (
+	MinVectorDims = 1
+	MaxVectorDims = 2048
+)
+
 func NewVectorFieldMapping() *FieldMapping {
 	return &FieldMapping{
 		Type:         "vector",
--- a/mapping/mapping_vectors.go
+++ b/mapping/mapping_vectors.go
@@ -133,9 +133,10 @@ func validateVectorFieldAlias(field *FieldMapping, parentName string,
 
 	// # Validate field options
 
-	if field.Dims <= 0 || field.Dims > 2048 {
+	if field.Dims < index.MinVectorDims || field.Dims > index.MaxVectorDims {
 		return fmt.Errorf("field: '%s', invalid vector dimension: %d,"+
-			" value should be in range (%d, %d)", field.Name, field.Dims, 0, 2048)
+			" value should be in range (%d, %d)", field.Name, field.Dims,
+			index.MinVectorDims, index.MaxVectorDims)
 	}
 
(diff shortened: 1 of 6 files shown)


mapping/mapping_vectors.go
```diff
@@ -98,22 +108,37 @@ func validateVectorField(field *FieldMapping) error {
...
-}
+	// # Validate field options
 
-func validateFieldType(fieldType string) error {
-	switch fieldType {
-	case "text", "datetime", "number", "boolean", "geopoint", "geoshape",
-		"IP", "vector":
-	default:
-		return fmt.Errorf("unknown field type: '%s'", fieldType)
+	if field.Dims <= 0 || field.Dims > 2048 {
```
moshaad7, line 130: move 2048 to a global constant

Thejas-bhat:
i believe i follow the changes your trying to make over here. basically, the argument `fieldAliasCtx map[string]*FieldMapping` is passed all over document mappings that are there and then while validating a vector field, there is a check happening as to whether the same name is present in the map (due to aliasing) and then use the config of that entry to validate the provided similarity value etc. for this field. (please correct me if i'm wrong over here)

there is a cache object being passed in the documentMapping.Validate(cache) [basically the signature before the PR] which is also maintained throughout the parsing of the document mapping. Now, my suggestion over here is if this cache object can be leveraged to do the validations. 

roughly speaking, my thinking revolves around possiblity of introducing a new field in the registry.Cache say "VectorCache" which maps field.Name -> &vecFieldConfig{simlarity, dims ... } 

and during parsing perhaps what can be done is

```
if field.Type == "vector" {
	err := cache.ValidateVecConfig(field.Name, {field.Similarity, field.Dims})
	if err != nil {
		return err
	}
}
```

and inside the cache, you check that if there is an entry corresponding to the field.Name in the VectorCache, and then validate the passed config value with what's there in the cache entry. if the entry is not present already, just set the value. 

the main reason why i'm suggesting this is because, i feel like the changes would be isolated to the VectorCache component and avoid change all the function signatures over here.

please let me know what you all think about this.

mapping/mapping_vectors.go
```diff
@@ -98,22 +108,37 @@ func validateVectorField(field *FieldMapping) error {
 	field.DocValues = false
```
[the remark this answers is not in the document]
moshaad7: on index is applicable for vector field

mapping/mapping_vectors.go
moshaad7: done

moshaad7:
> i believe i follow the changes your trying to make over here. basically, the argument `fieldAliasCtx map[string]*FieldMapping` is passed all over document mappings that are there and then while validating a vector field, there is a check happening as to whether the same name is present in the map (due to aliasing) and then use the config of that entry to validate the provided similarity value etc. for this field. (please correct me if i'm wrong over here)
> 
> there is a cache object being passed in the documentMapping.Validate(cache) [basically the signature before the PR] which is also maintained throughout the parsing of the document mapping. Now, my suggestion over here is if this cache object can be leveraged to do the validations.
> 
> roughly speaking, my thinking revolves around possiblity of introducing a new field in the registry.Cache say "VectorCache" which maps field.Name -> &vecFieldConfig{simlarity, dims ... }
> 
> and during parsing perhaps what can be done is
> 
> ```
> if field.Type == "vector" {
> 	err := cache.ValidateVecConfig(field.Name, {field.Similarity, field.Dims})
> 	if err != nil {
> 		return err
> 	}
> }
> ```
> 
> and inside the cache, you check that if there is an entry corresponding to the field.Name in the VectorCache, and then validate the passed config value with what's there in the cache entry. if the entry is not present already, just set the value.
> 
> the main reason why i'm suggesting this is because, i feel like the changes would be isolated to the VectorCache component and avoid change all the function signatures over here.
> 
> please let me know what you all think about this.

@Thejas-bhat , 
It won't be appropriate to use this cache object.
As, this cache is meant to load the relevant analysis pipeline objects.

For example, If in the index mapping, we specify "en" analyzer.
Then, during the Validation phase, while traversing the mapping, we will encounter the use of "en" and we call the AnalyzerNamed method of the cache. This will go to the analyzer registries and construct an instance of type Analyzer which can support "en" analyzer features.

And that way, after the completion of Validation, we will populate our cache with all the relevant analysis related objects (analyzers, tokenizers, .... )

Now, 
If you look at the` IndexMapping interface `

```
type IndexMapping interface {
        ...
	DateTimeParserNamed(name string) analysis.DateTimeParser
	AnalyzerNamed(name string) analysis.Analyzer
}
```

During index and Search operations, we will call AnalyzerNamed method of the indexMapping to get the cached Analyzer object.

----
That's the whole use of this Registry.Cache.

I don't think fieldAliasCtx should go into this, because, we don't want to keep this object after Validation is over. 

@abhinavdangeti  also agreed to this.

PR #1905, opened 2023-11-10
Added missing nil check

metonymic-smokey:
In case a document is not found, a nil check is required since `doc.StoredFieldsBytes()` then results in a panic.

### DIFF
--- a/index_impl.go
+++ b/index_impl.go
@@ -496,7 +496,6 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 	ctx = context.WithValue(ctx, search.GeoBufferPoolCallbackKey,
 		search.GeoBufferPoolCallbackFunc(getBufferPool))
 
-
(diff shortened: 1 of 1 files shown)


index_impl.go
```diff
@@ -663,6 +662,10 @@ func LoadAndHighlightFields(hit *search.DocumentMatch, req *SearchRequest,
 	var totalStoredFieldsBytes uint64
 	if len(req.Fields) > 0 || highlighter != nil {
 		doc, err := r.Document(hit.ID)
+		// Case if doc doesn't exist
+		if err == nil && doc == nil {
+			return fmt.Errorf("document %s not found", hit.ID), 0
+		}
```
abhinavdangeti, line 668: Rather than this code change, moving line 669 to 670.5 would suffice yes?

abhinavdangeti (changes requested):

abhinavdangeti (approved):

PR #1907, opened 2023-11-10
Adding KNN Operator option 

metonymic-smokey:
This PR allows the user to specify an operator which will be used to determine if KNN and the filter query form a conjunction/disjunction query.


abhinavdangeti:
@metonymic-smokey , let's point everything knn related to the `unstable` branch - so we can merge later without having to deal with any conflicts.

metonymic-smokey:
Was just doing that, having 'master' as a base was a typo of sorts.

PR #1908, opened 2023-11-10
Add KNN Operator

metonymic-smokey:
This PR allows a user to specify a KNN Operator to determine if KNN hits are conjuncted/disjuncted with filter query hits.

### DIFF
--- a/search_knn.go
+++ b/search_knn.go
@@ -26,6 +26,8 @@ import (
 	"github.com/blevesearch/bleve/v2/search/query"
 )
 
+type knnOperator string
+
 type SearchRequest struct {
 	Query            query.Query       `json:"query"`
 	Size             int               `json:"size"`
@@ -40,7 +42,8 @@ type SearchRequest struct {
 	SearchAfter      []string          `json:"search_after"`
 	SearchBefore     []string          `json:"search_before"`
 
-	KNN []*KNNRequest `json:"knn"`
+	KNN         []*KNNRequest `json:"knn"`
+	KNNOperator knnOperator   `json:"knn_operator"`
 
 	sortFunc func(sort.Interface)
 }
@@ -62,6 +65,10 @@ func (r *SearchRequest) AddKNN(field string, vector []float32, k int64, boost fl
 	})
 }
 
+func (r *SearchRequest) AddKNNOperator(operator knnOperator) {
+	r.KNNOperator = operator
+}
+
 // UnmarshalJSON deserializes a JSON representation of
 // a SearchRequest
 func (r *SearchRequest) UnmarshalJSON(input []byte) error {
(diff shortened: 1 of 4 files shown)


search_knn.go
```diff
@@ -142,24 +154,33 @@ func copySearchRequest(req *SearchRequest) *SearchRequest {
...
 				knnQuery.SetBoost(knn.Boost.Value())
-				disjuncts = append(disjuncts, knnQuery)
+				subQueries = append(subQueries, knnQuery)
 			}
 		}
-		return query.NewDisjunctionQuery(disjuncts)
+		if req.KNNOperator == knnOperatorAnd {
+			return query.NewConjunctionQuery(subQueries)
+		}
+		return query.NewDisjunctionQuery(subQueries)
```
metonymic-smokey, line 183: Add cases here if a user enters an operator other than these.
Add UTs?

search_knn.go
```diff
@@ -142,24 +154,36 @@ func copySearchRequest(req *SearchRequest) *SearchRequest {
...
 				knnQuery := query.NewKNNQuery(knn.Vector)
 				knnQuery.SetFieldVal(knn.Field)
 				knnQuery.SetK(knn.K)
 				knnQuery.SetBoost(knn.Boost.Value())
-				disjuncts = append(disjuncts, knnQuery)
+				subQueries = append(subQueries, knnQuery)
 			}
 		}
-		return query.NewDisjunctionQuery(disjuncts)
+		if req.KNNOperator == knnOperatorAnd {
```
Thejas-bhat, line 180: my suggestion over here would be do some string processing like trim and lowercase conversion such that there is a little flexibility for the user to enter the operator

metonymic-smokey:
Screenshot of wrong operator with CB Server:
image

No such issue with blank operator:
image

search_knn.go
metonymic-smokey: Didn't add that since match query operator doesn't seem to have it either - 
https://github.com/blevesearch/bleve/blob/master/search/query/match.go#L64-L70

WDYT?

Thejas-bhat: okay cool

abhinavdangeti (commented):
Looks good, but add some unit tests please.

abhinavdangeti:
Yo, you've clearly not tested your unit test code -
```
# github.com/blevesearch/bleve/v2 [github.com/blevesearch/bleve/v2.test]
./search_knn_test.go:92:15: undefined: getQueryVec
```

search_knn_test.go
```diff
@@ -0,0 +1,147 @@
...
+	for i := 0; i < len(dataset); i++ {
+		batch.Index(strconv.Itoa(i), dataset[i])
+	}
+
+	err = index.Batch(batch)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	termQuery := query.NewTermQuery("world")
```
abhinavdangeti, line 89: Where is `world` even existing in your documents!?

metonymic-smokey: Fixing this.
However, I don't think the specific term matters here since all we're checking is if the operator is being applied right.

abhinavdangeti (approved):

PR #1909, opened 2023-11-15
MB-59447: support nested/chunked vectors

moshaad7:
Enable user to chunk vectors ([]float32) into nested_vectors ([][]float32).
This is helpful to capture more intent from text or to use multiple vector generation models for the same input.

While mapping a document, user submitted nested_vectors are flattened and submitted to bleve_index_api.Index impl (Scorch).
Scorch will then submit it to the segment plugin (Zapx).

Zapx is expected to realise that this flattened_vector have length more than Dims.
zapx should read all the subVecs out of the flattened_vector and index them individually.

We are not passing the nested_vector directly to zap. This is done to keep the API contract intact.

related changes:
    https://github.com/blevesearch/zapx/pull/185

### DIFF
--- a/mapping/mapping_vectors.go
+++ b/mapping/mapping_vectors.go
@@ -37,49 +37,97 @@ func NewVectorFieldMapping() *FieldMapping {
 	}
 }
 
+// validate and process a flat vector
+func processFlatVector(vecI interface{}, dims int) ([]float32, bool) {
+	vecV := reflect.ValueOf(vecI)
+	if !vecV.IsValid() || vecV.Kind() != reflect.Slice || vecV.Len() != dims {
+		return nil, false
+	}
+
+	rv := make([]float32, dims)
+	for i := 0; i < vecV.Len(); i++ {
+		item := vecV.Index(i)
+		if !item.CanInterface() {
+			return nil, false
+		}
+		itemI := item.Interface()
+		itemFloat, ok := util.ExtractNumericValFloat32(itemI)
+		if !ok {
+			return nil, false
+		}
+		rv[i] = itemFloat
+	}
+
+	return rv, true
+}
+
+// validate and process a vector
+// max supported depth of nesting is 2 ([][]float32)
+func processVector(vecI interface{}, dims int) ([]float32, bool) {
+	vecV := reflect.ValueOf(vecI)
+	if !vecV.IsValid() || vecV.Kind() != reflect.Slice || vecV.Len() == 0 {
+		return nil, false
+	}
+
+	// Let's examine the first element (head) of the vector.
+	// If head is a slice, then vector is nested, otherwise flat.
+	headI := vecV.Index(0).Interface()
+	headV := reflect.ValueOf(headI)
+	if !headV.IsValid() {
+		return nil, false
+	}
+	if headV.Kind() != reflect.Slice { // vector is flat
+		return processFlatVector(vecI, dims)
+	}
+
+	// # process nested vector
+
+	// pre-allocate memory for the flattened vector
+	// so that we can use copy() later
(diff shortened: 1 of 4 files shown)


mapping/mapping_vectors.go
```diff
@@ -36,45 +36,97 @@ func NewVectorFieldMapping() *FieldMapping {
...
-		options := fm.Options()
-		field := document.NewVectorFieldWithIndexingOptions(fieldName,
-			indexes, vector, fm.Dims, fm.Similarity, options)
-		context.doc.AddField(field)
+		flatVector, ok := processFlatVector(subVecI, dims)
+		if !ok {
+			return nil, false
+		}
+
+		copy(rv[i*dims:(i+1)*dims], flatVector)
```
Thejas-bhat, line 108: more of a question here, is there a difference (perf wise) between the one you've used and the following way of populating the float array:

`rv = append(rv, flatVector...)`

which is functionally same, since the copy statement is also technically appending to the rv slice.

i know that if you use the append way, it won't dynamically allocate the memory since you've already assigned it before the loop, but just curious as to whether one is better than the other. please let me know if this makes sense

mapping/mapping_vectors.go
```diff
@@ -37,45 +37,97 @@ func NewVectorFieldMapping() *FieldMapping {
...
 
-func (fm *FieldMapping) processVector(propertyMightBeVector interface{},
-	pathString string, path []string, indexes []uint64, context *walkContext) {
-	propertyVal := reflect.ValueOf(propertyMightBeVector)
-	if !propertyVal.IsValid() {
-		return
+// validate and process a flat vector
+func processFlatVector(vecI interface{}, dims int) ([]float32, bool) {
+	vecV := reflect.ValueOf(vecI)
+	if !vecV.IsValid() || vecV.Kind() != reflect.Slice || vecV.Len() != dims {
```
moshaad7, line 43: If we mandate caller of processFlatVector to make sure that vecI is valid, is slice and len equal dims.
Then, we can avoid checking it here.

mapping/mapping_vectors.go
moshaad7: You are right that using copy() or append() can help us achieve the same functional goal.

With pre-allocating the space for resultant slice, we can ensure that append() doesn't perform any allocation while copying the elements.
But even then, it is slower than copy(). 
I am not sure of exact reason, but I suspect the book-keeping operations may be making append() slower than copy() in such case.

Here is my benchmarking test to compare these two:
https://gist.github.com/moshaad7/f459123d336a1a1375621fd5c19578a2

mapping/mapping_vectors.go
```diff
@@ -36,45 +36,97 @@ func NewVectorFieldMapping() *FieldMapping {
...
 }
 
-func (fm *FieldMapping) processVector(propertyMightBeVector interface{},
-	pathString string, path []string, indexes []uint64, context *walkContext) {
-	propertyVal := reflect.ValueOf(propertyMightBeVector)
-	if !propertyVal.IsValid() {
-		return
+// validate and process a flat vector
+func processFlatVector(vecI interface{}, dims int) ([]float32, bool) {
+	vecV := reflect.ValueOf(vecI)
```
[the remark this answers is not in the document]
moshaad7: done

mapping/mapping_vectors.go
moshaad7: done

abhinavdangeti (changes requested):
This looks good. Let merge this only after getting the zapx commit in and updating the go.mod within this PR.

abhinavdangeti (approved):

PR #1911, opened 2023-11-16
MB-59692: Upgrade to bleve_index_api@v1.1.3

abhinavdangeti:
Use re-situated SimilarityMetrics constants.


mapping/mapping_vectors.go
```diff
@@ -87,7 +88,7 @@ func validateVectorField(field *FieldMapping) error {
 	}
 
 	if field.Similarity == "" {
```
Thejas-bhat, line 90: extended my suggestion of what i had in the zap PR, if we can modify the type of VectorField in the field_vector.go such that the metric value over there is an int, then what can be done is just pass a index.SupportedMetrics[fm.Similarity] and the zap side of things just does a simple access of metric field from the VectorField object and nothing more than that. 

My main reasoning behind this is that we can keep the zap side of things super simple and since the VectorField is not directly visible to the user he won't notice there being an int flag (confusing him) while keeping the fieldMapping side which the user directly interacts with as a string (a user friendly way of specifying the metric)

abhinavdangeti: Fair suggestion, but this'll need the interface change again. Let's leave it generic enough as it is now - more descriptive and not tied with the go-faiss enums. We can reconsider it later.

Thejas-bhat (approved):

PR #1915, opened 2023-11-21
Optimizations for KNN Search

metonymic-smokey:
Adds an optimization phase specific to KNN Searchers - 
For each field of each segment, read the vector index only once, across searchers.

### DIFF
--- a/index/scorch/optimize_knn.go
+++ b/index/scorch/optimize_knn.go
@@ -1,3 +1,17 @@
+//  Copyright (c) 2023 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 //go:build vectors
 // +build vectors
 
@@ -10,14 +24,14 @@ import (
 	segment_api "github.com/blevesearch/scorch_segment_api/v2"
 )
 
-type OptimizeVRDisjunction struct {
+type OptimizeVR struct {
 	snapshot *IndexSnapshot
 
 	// map of the same field --> vrs
 	vrs map[string][]*IndexSnapshotVectorReader
 }
 
-func (o *OptimizeVRDisjunction) Finish() (index.Optimized, error) {
+func (o *OptimizeVR) Finish() error {
 
 	// for each field, get the faiss index --> invoke the zap func.
 
@@ -34,7 +48,7 @@ func (o *OptimizeVRDisjunction) Finish() (index.Optimized, error) {
 				// reading just once per field per segment.
 				faissIndex, err := sv.ReadVectorIndex(field)
 				if err != nil {
-					return nil, err
+					return err
 				}
 				defer faissIndex.Close()
 
@@ -44,7 +58,7 @@ func (o *OptimizeVRDisjunction) Finish() (index.Optimized, error) {
 					pl, err := sv.SearchSimilarVectors(faissIndex, vr.field,
 						vr.queryVector, vr.k, seg.deleted)
 					if err != nil {
-						return nil, err
+						return err
 					}
 					vr.postings[i] = pl
 					vr.iterators[i] = pl.Iterator(vr.iterators[i])
@@ -53,19 +67,19 @@ func (o *OptimizeVRDisjunction) Finish() (index.Optimized, error) {
 		}
 	}
 
-	return nil, nil
+	return nil
 }
 
-func (s *IndexSnapshotVectorReader) Optimize(kind string,
-	octx index.OptimizableContext) (index.OptimizableContext, error) {
+func (s *IndexSnapshotVectorReader) VectorOptimize(
+	octx index.VectorOptimizableContext) (index.VectorOptimizableContext, error) {
 
 	if octx == nil {
-		octx = &OptimizeVRDisjunction{snapshot: s.snapshot,
+		octx = &OptimizeVR{snapshot: s.snapshot,
 			vrs: make(map[string][]*IndexSnapshotVectorReader),
 		}
 	}
 
-	o, ok := octx.(*OptimizeVRDisjunction)
+	o, ok := octx.(*OptimizeVR)
 	if !ok {
 		return octx, nil
 	}
--- a/index/scorch/optimize_knn.go
+++ b/index/scorch/optimize_knn.go
@@ -32,12 +32,9 @@ type OptimizeVR struct {
 }
 
 func (o *OptimizeVR) Finish() error {
-
 	// for each field, get the faiss index --> invoke the zap func.
-
 	// for each VR, populate postings list and iterators
 	// by passing the obtained FAISS index and getting similar vectors.
-
 	// defer close index - just once.
 
 	for i, seg := range o.snapshot.segment {
@@ -46,7 +43,7 @@ func (o *OptimizeVR) Finish() error {
 			// for each VR belonging to that field
 			if sv, ok := seg.segment.(segment_api.VectorSegment); ok {
 				// reading just once per field per segment.
-				faissIndex, err := sv.ReadVectorIndex(field)
+				faissIndex, err := sv.GetVectorIndex(field)
 				if err != nil {
 					return err
 				}
--- a/index/scorch/optimize_knn.go
+++ b/index/scorch/optimize_knn.go
@@ -47,7 +47,11 @@ func (o *OptimizeVR) Finish() error {
 				if err != nil {
 					return err
 				}
-				defer faissIndex.Close()
+				defer func(vecIndex segment_api.VectorIndex) {
+					go func(vecIndex segment_api.VectorIndex) {
+						vecIndex.Close()
+					}(vecIndex)
+				}(faissIndex)
 
 				for _, vr := range vrs {
 					// for each VR, populate postings list and iterators
--- a/index/scorch/optimize_knn.go
+++ b/index/scorch/optimize_knn.go
@@ -43,21 +43,22 @@ func (o *OptimizeVR) Finish() error {
 			// for each VR belonging to that field
 			if sv, ok := seg.segment.(segment_api.VectorSegment); ok {
 				// reading just once per field per segment.
-				vecIndex, err := sv.GetVectorIndex(field)
+				searchVectorIndex, closeVectorIndex, err := sv.InterpretVectorIndex(field)
+
+				defer func(closeVectorIndex segment_api.CloseVectorIndex) {
+					go func(closeVectorIndex segment_api.CloseVectorIndex) {
+						closeVectorIndex()
+					}(closeVectorIndex)
+				}(closeVectorIndex)
+
 				if err != nil {
 					return err
 				}
-				defer func(vecIndex segment_api.VectorIndex) {
-					go func(vecIndex segment_api.VectorIndex) {
-						vecIndex.Close()
-					}(vecIndex)
-				}(vecIndex)
 
 				for _, vr := range vrs {
 					// for each VR, populate postings list and iterators
 					// by passing the obtained vector index and getting similar vectors.
-					pl, err := sv.SearchSimilarVectors(vecIndex, vr.field,
-						vr.vector, vr.k, seg.deleted)
+					pl, err := searchVectorIndex(vr.field, vr.vector, vr.k, seg.deleted)
 					if err != nil {
 						return err
 					}
--- a/index/scorch/optimize_knn.go
+++ b/index/scorch/optimize_knn.go
@@ -1,0 +1,80 @@
+//go:build vectors
+// +build vectors
+
+package scorch
+
+import (
+	"fmt"
+
+	index "github.com/blevesearch/bleve_index_api"
+	segment_api "github.com/blevesearch/scorch_segment_api/v2"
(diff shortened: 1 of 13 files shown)


search/searcher/search_knn.go
```diff
@@ -52,7 +52,7 @@ func NewKNNSearcher(ctx context.Context, i index.IndexReader, m mapping.IndexMap
 		}
 
 		knnScorer := scorer.NewKNNQueryScorer(vector, field, boost,
-			vectorReader.Count(), count, options, similarityMetric)
+			10, count, options, similarityMetric)
```
metonymic-smokey, line 55: Update this once scorer explanation is merged.

index/scorch/optimize_knn.go
```diff
@@ -0,0 +1,80 @@
+//go:build vectors
```
metonymic-smokey, line 1: TODO Add 2023 license.

search/searcher/search_disjunction.go
```diff
@@ -46,6 +46,18 @@ func optionsDisjunctionOptimizable(options search.SearcherOptions) bool {
...
 	qsearchers []search.Searcher, min float64, options search.SearcherOptions,
 	limit bool) (search.Searcher, error) {
+
+	// The KNN Searcher optimization is a necessary pre-req for KNN Searchers, not an
+	// optional optimization like for, say term searchers.
+	// It's an optimization in the sense that it's more optimal to batch calls
+	// Hence, this optimization should be invoked for KNN searchers irrespective
+	// of whether there are other searchers in the disjunction or not.
+
+	err := optimizeKNNSearcher(ctx, "", indexReader, qsearchers, options)
```
metonymic-smokey, line 56: TODO Need to double check nested queries of the type, to avoid double optimization:
{
  disjunction: {
     disjunct1: disjunction: {term,KNN}
     disjunct2: KNN
 }
}

search/query/disjunction.go
```diff
@@ -64,6 +64,7 @@ func (q *DisjunctionQuery) Searcher(ctx context.Context, i index.IndexReader, m
 	options search.SearcherOptions) (search.Searcher, error) {
 	ss := make([]search.Searcher, 0, len(q.Disjuncts))
 	for _, disjunct := range q.Disjuncts {
+		// at this point, VR has already read the vector index
```
metonymic-smokey, line 67: Lose this comment?

abhinavdangeti (changes requested):

search_knn.go
```diff
@@ -168,7 +168,8 @@ var (
 )
 
 func queryWithKNN(req *SearchRequest) (query.Query, error) {
-	if len(req.KNN) > 0 {
+	_, ok := req.Query.(*query.KNNQuery)
```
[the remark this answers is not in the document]
metonymic-smokey: We do support KNNQueries when using bleve as a library. Example code snippet:
```
nestedKNNQuery := query.NewKNNQuery(getQueryVec("hilly region worldwide"))
nestedKNNQuery.SetFieldVal("vector")
nestedKNNQuery.SetK(3)
```
This results in a KNN query which then needs to be added to the KNN field of the search request(done in queryWithKNN()).

index/scorch/snapshot_vector_index.go
```diff
@@ -42,16 +43,7 @@ func (is *IndexSnapshot) VectorReader(ctx context.Context, vector []float32,
...
-		if sv, ok := seg.segment.(segment_api.VectorSegment); ok {
-			pl, err := sv.SimilarVectors(field, vector, k, seg.deleted)
-			if err != nil {
-				return nil, err
-			}
-			rv.postings[i] = pl
-			rv.iterators[i] = pl.Iterator(rv.iterators[i])
-		}
-	}
+	// Do this as part of optimizations' Finish() func.
```
Thejas-bhat, line 46: i think its better to leave this code snippet as is and protect it under an "optimize" flag via the context, since using VectorReader as a component in other parts of the code (if needed) or when this struct is used for unit testing the struct would be complete with all the necessary information.

search/searcher/optimize_knn.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func optimizeKNNSearcher(ctx context.Context, optimizationKind string,
+	indexReader index.IndexReader, qsearchers []search.Searcher,
+	options search.SearcherOptions) error {
+	var octx index.VectorOptimizableContext
+
+	for _, searcher := range qsearchers {
+		// Only applicable to KNN Searchers.
+		if _, ok := searcher.(*KNNSearcher); !ok {
```
Thejas-bhat, line 34: since the implementation check of VectorOptimizable is applicable only for knn searchers, i don't think this check is necessary - because the check at 38 passes only if its a knn searcher.

metonymic-smokey:
Reviewers, just found a bug today:
In case score is set to "none", the optimization for KNN searchers doesn't kick in because we replace the current searchers with a consolidated term searcher - https://github.com/blevesearch/bleve/blob/c8e3daf115aee56897e841aee9ddc9dfa78ba5d6/search/searcher/search_disjunction.go#L101-L102.

Working on fixing this so that both term and KNN searchers can be optimized.

index/scorch/snapshot_vector_index.go
metonymic-smokey: and pass the vectorIndex as part of the context?

Thejas-bhat: ah i see, yeah i forgot that the search interface was changed. sounds good to me.

metonymic-smokey:
So, as part of exploring this, I realised that setting score = "none" basically invalidates a KNN query since hits with score = 0 are dropped as shown here - 
https://github.com/blevesearch/bleve/blob/unstable/search_knn.go#L280-L287

index/scorch/optimize_knn.go
```diff
@@ -0,0 +1,91 @@
...
+		// for each field, get the faiss index --> invoke the zap func.
+		for field, vrs := range o.vrs {
+			// for each VR belonging to that field
+			if sv, ok := seg.segment.(segment_api.VectorSegment); ok {
+				// reading just once per field per segment.
+				faissIndex, err := sv.GetVectorIndex(field)
+				if err != nil {
+					return err
+				}
+				defer faissIndex.Close()
```
Thejas-bhat, line 50: so in the profiles, i saw that the Close() is like a blocking call in the SimilarVectors() API (it now blocks Finish() API) and adds to the latency. here is my reasoning - the close() basically frees the memory allocated on the C heap (since the faissIndex is resident on the C heap memory) and there is no dependency of it to the goruntime environment. 

so i'm thinking if its better to make it asynchronous so that the closing of the index doesn't block the Finish() API. 

please let me know if this makes sense to you all.

metonymic-smokey: yep, good point, done it in the latest commit.

metonymic-smokey:
Update: made some slight changes to approach considering that we won't be dropping kNN hits when score = "none"/0 any more.

search_knn.go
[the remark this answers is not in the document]
metonymic-smokey: But if we are going to work with KNNs using existing code for conj/disj *queries*, we need to support kNN queries since conjuncts/disjuncts are of type Query.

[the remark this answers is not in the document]
metonymic-smokey: As per offline conversation, this condition is necessary only when the query is a KNN query, which we don't support since the users should populate the KNN field instead of query.

search/searcher/search_disjunction.go
```diff
@@ -46,16 +46,33 @@ func optionsDisjunctionOptimizable(options search.SearcherOptions) bool {
...
 	if len(qsearchers) > 1 && min <= 1 &&
 		optionsDisjunctionOptimizable(options) {
 		rv, err := optimizeCompositeSearcher(ctx, "disjunction:unadorned",
 			indexReader, qsearchers, options)
-		if err != nil || rv != nil {
+		if err != nil {
 			return rv, err
 		}
+		if rv != nil && optimizedKNNSearchers != nil && len(optimizedKNNSearchers) != 0 {
+			qsearchers = make([]search.Searcher, 1+len(optimizedKNNSearchers))
```
abhinavdangeti, line 72: Why `1+`? Also if this is necessary suffix `+1` instead.

search/searcher/util_no_knn.go
```diff
@@ -60,3 +60,39 @@ func (s *ConjunctionSearcher) computeQueryNorm() {
 		searcher.SetQueryNorm(s.queryNorm)
 	}
 }
+
+func optimizeCompositeSearcher(ctx context.Context, optimizationKind string,
```
abhinavdangeti, line 64: Valid catch below, import `context` in this file.

index/scorch/optimize_knn.go
```diff
@@ -0,0 +1,95 @@
...
+	// by passing the obtained FAISS index and getting similar vectors.
+	// defer close index - just once.
+
+	for i, seg := range o.snapshot.segment {
+		// for each field, get the faiss index --> invoke the zap func.
+		for field, vrs := range o.vrs {
+			// for each VR belonging to that field
+			if sv, ok := seg.segment.(segment_api.VectorSegment); ok {
+				// reading just once per field per segment.
+				faissIndex, err := sv.GetVectorIndex(field)
```
abhinavdangeti, line 46: Rename this to `vecIndex`, let's avoid the faiss name usage here as much as possible.

abhinavdangeti (changes requested):
@metonymic-smokey this approach is actually quite sub-optimal for single kNN. I've addressed the issue with your https://github.com/blevesearch/zapx/pull/184 proposal with these 2 changes:

- https://github.com/blevesearch/scorch_segment_api/pull/33
- https://github.com/blevesearch/zapx/pull/190

Please update your code here based on the new interface.

index/scorch/snapshot_vector_index.go
```diff
@@ -29,10 +29,11 @@ func (is *IndexSnapshot) VectorReader(ctx context.Context, vector []float32,
...
 	rv := &IndexSnapshotVectorReader{
-		vector:   vector,
-		field:    field,
-		k:        k,
-		snapshot: is,
+		vector:      vector,
+		field:       field,
+		k:           k,
+		snapshot:    is,
+		queryVector: vector,
```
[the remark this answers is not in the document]
metonymic-smokey: yes, my bad, wanted to rename it but I think, I'll just let it remain as is in this PR.

search/searcher/search_disjunction.go
metonymic-smokey: I'm assuming the latter point is just for coding style :)

index/scorch/optimize_knn.go
```diff
@@ -0,0 +1,96 @@
...
+			// for each VR belonging to that field
+			if sv, ok := seg.segment.(segment_api.VectorSegment); ok {
+				// reading just once per field per segment.
+				searchVectorIndex, closeVectorIndex, err := sv.InterpretVectorIndex(field)
+
+				defer func(closeVectorIndex segment_api.CloseVectorIndex) {
+					go func(closeVectorIndex segment_api.CloseVectorIndex) {
+						closeVectorIndex()
+					}(closeVectorIndex)
+				}(closeVectorIndex)
```
abhinavdangeti, line 52: Don't think this is necessary. Refactor to -
```
defer func() {
    go closeVectorIndex()
}()
```

search/searcher/optimize_knn.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/blevesearch/bleve/v2/search"
+	index "github.com/blevesearch/bleve_index_api"
+)
+
+func optimizeKNNSearcher(ctx context.Context, optimizationKind string,
+	indexReader index.IndexReader, qsearchers []search.Searcher,
+	options search.SearcherOptions) ([]search.Searcher, error) {
+	var octx index.VectorOptimizableContext
+	var err error
+	knnSearchers := make([]search.Searcher, 0)
```
abhinavdangeti, line 32: This make is quite unnecessary again, append allocs below.

search/searcher/search_conjunction.go
```diff
@@ -68,9 +75,14 @@ func NewConjunctionSearcher(ctx context.Context, indexReader index.IndexReader,
 		options.Score == "none" && !options.IncludeTermVectors {
 		rv, err := optimizeCompositeSearcher(ctx, "conjunction:unadorned",
 			indexReader, searchers, options)
-		if err != nil || rv != nil {
+		if err != nil {
 			return rv, err
 		}
+		if rv != nil && optimizedKNNSearchers != nil && len(optimizedKNNSearchers) != 0 {
+			qsearchers = make([]search.Searcher, 0, len(optimizedKNNSearchers)+1)
```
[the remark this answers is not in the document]
abhinavdangeti: Yes you can, append is an alloc. 

How many allocs are we now doing in this method !?!?!
- Line 61
- Line 82

Where the hell are we even using qsearchers !?

search/searcher/search_disjunction.go
```diff
@@ -46,16 +46,33 @@ func optionsDisjunctionOptimizable(options search.SearcherOptions) bool {
...
 	if len(qsearchers) > 1 && min <= 1 &&
 		optionsDisjunctionOptimizable(options) {
 		rv, err := optimizeCompositeSearcher(ctx, "disjunction:unadorned",
 			indexReader, qsearchers, options)
-		if err != nil || rv != nil {
+		if err != nil {
 			return rv, err
 		}
+		if rv != nil && optimizedKNNSearchers != nil && len(optimizedKNNSearchers) != 0 {
+			qsearchers = make([]search.Searcher, 0, len(optimizedKNNSearchers)+1)
```
abhinavdangeti, line 72: Not good.

abhinavdangeti (commented):
I'mma just dump this PR, because it's NOT sitting on the bleve repo and there's way too many things to fix. I'll be addressing things with https://github.com/blevesearch/bleve/pull/1924

PR #1920, opened 2023-11-22
integrate knn operator

CascadingRadium:
- Support usage of the knn operator "and" while merging knn results across multiple indexes of an alias.
- Drop hits at the merge phase when any score in a hit's score breakdown is made zero to accomodate for the top K knn results

### DIFF
--- a/search_knn.go
+++ b/search_knn.go
@@ -272,14 +272,10 @@ func mergeKNN(req *SearchRequest, sr *SearchResult) {
 			}
 		}
 	}
-	operator := 0
-	if _, ok := req.Query.(*query.ConjunctionQuery); ok {
-		operator = 1
-	}
 	nonZeroScoreHits := make([]*search.DocumentMatch, 0, len(sr.Hits))
 	maxScore := 0.0
 	for _, hit := range sr.Hits {
-		newScore := recomputeTotalScore(operator, hit)
+		newScore := recomputeTotalScore(req.KNNOperator, hit)
 		if newScore > 0 {
 			hit.Score = newScore
 			if newScore > maxScore {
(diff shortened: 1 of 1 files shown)


search_knn.go
```diff
@@ -293,18 +289,21 @@ func mergeKNN(req *SearchRequest, sr *SearchResult) {
...
 		}
 		totalScore += score
 	}
-	if operator == 0 {
-		coord := float64(numNonZero) / float64(len(hit.ScoreBreakdown))
+	if operator == knnOperatorOr || operator == "" {
+		coord := numNonZero / numTotal
 		totalScore = totalScore * coord
+	} else if operator == knnOperatorAnd && numNonZero != numTotal {
+		totalScore = 0
```
abhinavdangeti, line 306: Why should totalScore be 0 in this case?

search_knn.go
```diff
@@ -293,18 +289,21 @@ func mergeKNN(req *SearchRequest, sr *SearchResult) {
...
 	for _, score := range hit.ScoreBreakdown {
 		if score != 0 {
 			numNonZero += 1
 		}
 		totalScore += score
 	}
-	if operator == 0 {
-		coord := float64(numNonZero) / float64(len(hit.ScoreBreakdown))
+	if operator == knnOperatorOr || operator == "" {
+		coord := numNonZero / numTotal
```
abhinavdangeti, line 303: Can't numTotal be zero in case the request carried "score":"none" ?

search_knn.go
CascadingRadium: that is because some of the scores of the subsearchers of the conjunction searcher have become zero.

for example - you have match + knn query(k=2) with operator(and)

with 2 partitions, you can get 4 results; their score breakdown can be like this
D1 - [1,2]
D2 - [2,3]
D3 - [3,4]
D4 - [4,5] (index 0 = match score and index 1 = knn score)

so you choose top 2 = D3 and D4 in the heap. => numNonZero = numTotal = 2
the other two D1 and D2 will be made [1,0] and [2,0] (following the logic if disjunction (operator or) was used) -> now D1 and D2 essentially satisfy only match query which breaks conjunction clause set earlier in operator. So score is zero => remove them from hits

search_knn.go
abhinavdangeti: Suppose this is applicable only to knn scoring, so scoreBreakdown is expected to have at least one score even if score:none. This area will need revisiting later.

abhinavdangeti (approved):

PR #1921, opened 2023-11-27
change score breakdown from array to map

CascadingRadium:
looking at https://issues.couchbase.com/browse/MB-59821
it seems that having score breakdown as a slice would be inoptimal in other use cases of disjunction searchers such as in Prefix and Wildcard queries. 
Here the number of searchers involved in a single disjunction searcher may be very large and can affect cpu execution time.
This approach changes the datatype of scorebreakdown to a map


search_knn.go
```diff
@@ -225,7 +225,7 @@ func adjustRequestSizeForKNN(req *SearchRequest, numIndexPartitions int) int {
 
 // heap impl
 type scoreHeap struct {
-	scoreBreakdown []*[]float64
+	scoreBreakdown []*map[int]float64
```
abhinavdangeti, line 228: Does this need to be an array of maps?

Also the map does not need to be a pointer, just `[]map[int]float64` should do.

search/search.go
```diff
@@ -181,7 +181,7 @@ type DocumentMatch struct {
 	// the same as the number of sub-queries in the query.
 	// the order of the scores will match the order of the sub-queries
 	// in the query.
-	ScoreBreakdown []float64 `json:"score_breakdown,omitempty"`
+	ScoreBreakdown map[int]float64 `json:"score_breakdown,omitempty"`
```
abhinavdangeti, line 184: Update the commentary above on what the key-values in this map represent.

abhinavdangeti (changes requested):

abhinavdangeti (commented):
@CascadingRadium I thought we spoke of determining the score breakdown and shipping it _only_ when a knn request comes in ? If you are doing that already, I can't quite seem to spot it - would you point me to it.

CascadingRadium:
yea but i wanted to check if the regressions go away with just the optimization in this PR. if the results arent so good, i will add that check, in a different PR.

abhinavdangeti:
Per conversation w/ @CascadingRadium - new patch coming here addressing above comments.

search/query/conjunction.go
```diff
@@ -26,9 +26,10 @@ import (
 )
 
 type ConjunctionQuery struct {
-	Conjuncts       []Query `json:"conjuncts"`
-	BoostVal        *Boost  `json:"boost,omitempty"`
-	queryStringMode bool
+	Conjuncts              []Query `json:"conjuncts"`
+	BoostVal               *Boost  `json:"boost,omitempty"`
+	RetrieveScoreBreakdown bool    `json:"retrieve_score_breakdown,omitempty"`
```
abhinavdangeti, line 31: Yo, I don't like the feel of this at all.

abhinavdangeti (changes requested):
Our conversation was this -
* Ok to do a score breakdown when a compound query is involved.
* I wanted us to put in an optimization where we do optimize a compound query with kNN to simply a kNN when possible -
    * `match_none` OR single `kNN`
    * `match_all` AND single `kNN`
* It is ok to do a score breakdown for multi kNN

abhinavdangeti (approved):

PR #1923, opened 2023-12-01
Knn collector

CascadingRadium:
Uses a 2 + M fixed size heap approach to optimally solve the multi-segment knn results merging problem. 
The naive approach of just keeping a min-heap of size N = (size+from) which is based on the total score was flawed because the heap elements(total score) were subject to constant change, therefore we might eject out a value from the top N prematurely as that total score might end up to be zero later on when some other segment is read. The other approach was to keep a heap of size X where X is the number of search results, which i promptly discarded as the space complexity contract will be broken

The current solution - 
Time complexity = O(Nlog(max(size+from, maxK))
Space Complexity = O(max(maxK, size+from)) 
All collectors have the similar time and space complexities, i have increased it by a little bit - max of all k values.

Please let me know if there is any error in the logic

### DIFF
--- a/search/collector/knn.go
+++ b/search/collector/knn.go
@@ -1,0 +1,116 @@
+//  Copyright (c) 2014 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"github.com/blevesearch/bleve/v2/search"
+)
+
+type collectStoreKNN struct {
+	internalHeaps  []collectorStore
+	sizes          []int64
+	totalScoreHeap collectorStore
+	ejectedDocs    []*search.DocumentMatch
+	scoreCorrector func(docMatch *search.DocumentMatch)
+}
+
+// size is the from+size used for the tf-idf and total score heaps.
+func newStoreKNN(totalScoreHeap collectorStore, sizes []int64, internalHeaps []collectorStore,
+	scoreCorrector func(docMatch *search.DocumentMatch)) *collectStoreKNN {
+	return &collectStoreKNN{
+		internalHeaps:  internalHeaps,
+		totalScoreHeap: totalScoreHeap,
+		sizes:          sizes,
+		ejectedDocs:    make([]*search.DocumentMatch, len(internalHeaps)),
+		scoreCorrector: scoreCorrector,
+	}
+}
+
+func (c *collectStoreKNN) AddNotExceedingSize(doc *search.DocumentMatch, size int) []*search.DocumentMatch {
+	numDocsEjected := 0
+
+	for heapIdx := 1; heapIdx < len(c.internalHeaps); heapIdx++ {
+		if doc.ScoreBreakdown[heapIdx] == 0 {
+			continue
+		}
+		ejectedDoc := c.internalHeaps[heapIdx].AddNotExceedingSize(doc, int(c.sizes[heapIdx]))
+		if ejectedDoc != nil {
+			delete(ejectedDoc[0].ScoreBreakdown, heapIdx)
+			c.ejectedDocs[numDocsEjected] = ejectedDoc[0]
+			numDocsEjected++
+		}
+	}
+	if doc.ScoreBreakdown[0] != 0 {
+		tfIdfEject := c.internalHeaps[0].AddNotExceedingSize(doc, int(c.sizes[0]))
+		if tfIdfEject != nil && len(tfIdfEject[0].ScoreBreakdown) == 1 {
+			delete(tfIdfEject[0].ScoreBreakdown, 0)
+			c.ejectedDocs[numDocsEjected] = tfIdfEject[0]
+			numDocsEjected++
+		}
+	}
+
+	for idx := 0; idx < numDocsEjected; idx++ {
+		c.scoreCorrector(c.ejectedDocs[idx])
+	}
+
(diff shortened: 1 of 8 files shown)


abhinavdangeti:
@CascadingRadium just a rebase here to allow for unit tests to pass. Please checkout brand new `knnCollector` from your here to continue any work (apologies for the force push).

search/collector/knn.go
```diff
@@ -0,0 +1,116 @@
+//  Copyright (c) 2014 Couchbase, Inc.
```
metonymic-smokey, line 1: Nit: fix license header.

metonymic-smokey (changes requested):
@CascadingRadium pls check this case:
The search request has 1 term query and 2 KNN queries, with the operator being the default "or"(disjunction). Set score = "none".
When I run the query on this branch, I get the following result:
`search res are 5 matches, showing 1 through 0, took 3.245ms`

Running it on unstable branch gets me the following result:
`search res are No matches ` which is a known bug on unstable.

The expectation is that the UT produces the same results as when score = "none" isn
t set, but with the results having a score of 0.

Pls let me know if you want to take a look at the UT which I used to reproduce this.

cc @abhinavdangeti

search/collector/knn.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	for idx := 0; idx < numDocsEjected; idx++ {
+		c.scoreCorrector(c.ejectedDocs[idx])
+	}
+
+	c.totalScoreHeap.Add(doc)
+	rv := c.totalScoreHeap.PopWhile(func(doc *search.DocumentMatch) bool {
+		return doc.Score == 0
+	})
```
metonymic-smokey, line 71: Could you pls explain why you are removing from the heap when score = 0?

search/collector/knn.go
```diff
@@ -0,0 +1,116 @@
...
+			continue
+		}
+		ejectedDoc := c.internalHeaps[heapIdx].AddNotExceedingSize(doc, int(c.sizes[heapIdx]))
+		if ejectedDoc != nil {
+			delete(ejectedDoc[0].ScoreBreakdown, heapIdx)
+			c.ejectedDocs[numDocsEjected] = ejectedDoc[0]
+			numDocsEjected++
+		}
+	}
+	if doc.ScoreBreakdown[0] != 0 {
```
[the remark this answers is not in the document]
metonymic-smokey: IMO, the collector should handle score = "none" too, can't have a separate approach just for that.

search/collector/knn.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+	}
+
+	for idx := 0; idx < numDocsEjected; idx++ {
+		c.scoreCorrector(c.ejectedDocs[idx])
+	}
+
+	c.totalScoreHeap.Add(doc)
+	rv := c.totalScoreHeap.PopWhile(func(doc *search.DocumentMatch) bool {
+		return doc.Score == 0
```
metonymic-smokey, line 70: Also, why are we removing docs with score = 0 over here if we're also doing it in mergeKNNResults()? 
https://github.com/blevesearch/bleve/blob/unstable/search_knn.go#L284-L292

search/collector/knn.go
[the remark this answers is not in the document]
CascadingRadium: ah ok

search/collector/knn.go
CascadingRadium: knn collector is for merging knn results across multiple segments of a single index.
The mergeKNNResults() is called only in index alias impl -> merging knn results across multiple indexes in an alias. 2 phase approach, you first merge all knn results across all segments in each index and then from each index you merge

search/collector/topn.go
```diff
@@ -418,6 +426,68 @@ func (hc *TopNCollector) Total() uint64 {
...
+	// of D*K - K = K*(D-1) hits to 0, which reduces the total score of the document
+	// which inturn messes up the pagination - since the total score is used for pagination
+	//
+	// this problem is a constrained equivalent to the problem of maintaining top N elements from an infinite
+	// stream of objects where the stream can contain duplicates, (wherein we need to
+	// update the value of the duplicate object received in the stream), with the additional
+	// constraint that the space complexity of the solution should be O(N) and not infinite
+	//
+	// idea:-
+	// keep a min-heap for each KNN query 	- this is because we need to maintain top K elements for each KNN query
```
Thejas-bhat, line 451: i'm a little confused, it's a max heap you're constructing right? because i see below that 
```
if i.Score < j.Score {
	return 1
}
return -1
```
 
and the heap comparator logic indicates it returns 1 if i.Score < j.Score (which is opposite) of the logic in sort.go

search/collector/topn.go
```diff
@@ -418,6 +426,68 @@ func (hc *TopNCollector) Total() uint64 {
...
+
+	internalHeaps := make([]collectorStore, len(kArray))
+	// index 0 of kArray is the size+from for the TF-IDF searcher
+	// index 1 of kArray is the K for the KNN searcher at index 0 of SearchRequest.KNN slice
+	// index 2 of kArray is the K for the KNN searcher at index 1 of SearchRequest.KNN slice
+	for knnIdx, k := range kArray {
+		// TODO - Check if the datatype of k can be made into an int instead of int64
+		idx := knnIdx
+		internalHeaps[knnIdx] = getOptimalCollectorStore(int(k), 0, func(i, j *search.DocumentMatch) int {
+			if i.ScoreBreakdown[idx] < j.ScoreBreakdown[idx] {
```
Thejas-bhat, line 474: nit: why not reuse the knnIdx over here?

search_knn.go
```diff
@@ -194,6 +195,21 @@ func queryWithKNN(req *SearchRequest) (query.Query, error) {
 	return req.Query, nil
 }
 
+func setStoreForKNN(req *SearchRequest, coll *collector.TopNCollector, size int) {
+	if len(req.KNN) > 0 {
+		kArray := make([]int64, 1+len(req.KNN))
```
Thejas-bhat, line 200: i think in the codebase the convention is +1, rather than 1+. Just to keep things similar, could you please change it to +1?

Thejas-bhat (commented):
still trying to understand the PR, i think i'm getting what's the core logic being followed over here. however, i'll review with any edge case i can think of.

PR #1928, opened 2023-12-05
Fix scoring for match all query

metonymic-smokey:
Currently, match all queries used to return 1.0 as the score even when score = "none".
This PR fixes that by ensuring they return 0.0 in such a case and adding a UT for verification.


abhinavdangeti (approved):

abhinavdangeti (changes requested):
Hey @metonymic-smokey, the change looks good but let's make this change over the `unstable` branch, to reduce the possibility of conflicts when we'll be merging `unstable` into master.

abhinavdangeti (approved):

PR #1929, opened 2023-12-05
Minimising conjunctions/disjunctions with KNN queries

metonymic-smokey:
In certain cases, like the ones below, the KNN+filter searchers can be reduced to a single searcher:
* match_none OR single-kNN
* match_all AND single-kNN
They need to read the vector index and search similar vectors, without being executed as conjunction/disjunction queries.

### DIFF
--- a/search/query/conjunction.go
+++ b/search/query/conjunction.go
@@ -15,14 +15,9 @@
 package query
 
 import (
-	"context"
 	"encoding/json"
 
-	"github.com/blevesearch/bleve/v2/mapping"
-	"github.com/blevesearch/bleve/v2/search"
-	"github.com/blevesearch/bleve/v2/search/searcher"
 	"github.com/blevesearch/bleve/v2/util"
-	index "github.com/blevesearch/bleve_index_api"
 )
 
 type ConjunctionQuery struct {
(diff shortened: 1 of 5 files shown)


search/query/conjunction.go
```diff
@@ -59,33 +54,6 @@ func (q *ConjunctionQuery) AddQuery(aq ...Query) {
 	}
 }
 
-func (q *ConjunctionQuery) Searcher(ctx context.Context, i index.IndexReader, m mapping.IndexMapping, options search.SearcherOptions) (search.Searcher, error) {
```
abhinavdangeti, line 62: Moving this method out of where they are feels wrong. Let's retain this where it is add support for a new method that does only the required estimation for us in util_knn.go and util_no_knn.go.

Same for disjunction.go.

search/query/util_knn.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	if len(ss) < 1 {
+		return searcher.NewMatchNoneSearcher(i)
+	}
+	nctx := context.WithValue(ctx, search.IncludeScoreBreakdownKey, q.retrieveScoreBreakdown)
+
+	// Here, check if it's effectively a single KNN searcher
+	// If so, return new KNN Searcher
+	if knnSearcherCount == 1 && matchAllSearcherCount == 1 {
```
abhinavdangeti, line 63: Oh this is not complete. What if there're other searchers alongside 1 match_all and 1 knn. This also needs to be applicable ONLY when the operator is `and` here and for disjunction an `or` with match_none.

search/query/util_knn.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if len(ss) < 1 {
+		return searcher.NewMatchNoneSearcher(i)
+	}
+
+	nctx := context.WithValue(ctx, search.IncludeScoreBreakdownKey, q.retrieveScoreBreakdown)
+
+	// Here, check if it's effectively a single KNN searcher
+	// If so, return new KNN Searcher
+	if knnSearcherCount == 1 && matchNoneSearcherCount == 1 {
```
abhinavdangeti, line 112: This is incomplete. See comment above.

abhinavdangeti (changes requested):

abhinavdangeti:
I feel there're several more corner cases to consider here, in case a user puts in ..
* multiple match_all
* multiple match_none
* a combination of both (depends on the kind of compound query)

search/query/util_knn.go
metonymic-smokey: @abhinavdangeti re the second point:
A conjunction query will be formed only when the KNNOperator is "and". Hence, i don't see why there needs to be an explicit check for the operator.
https://github.com/blevesearch/bleve/blob/unstable/search_knn.go#L182-L185

metonymic-smokey:
> I feel there're several more corner cases to consider here, in case a user puts in ..
> 
> * multiple match_all
> * multiple match_none
> * a combination of both (depends on the kind of compound query)

@abhinavdangeti  If the user wants to combine queries like the cases you cited, they'll have to use a conjunction/disjunction.
Eg. if the user combines a match_none and match_all in a disjunction(say disj1), that's equivalent to a match_all but when forming the overall query with KNN included, it will form a disjunction query disj2:
disj2: {
    disjuncts: {disj1, KNN query}
}
Hence, this won't leverage this optimization and will proceed using normal disjunction query paths.

abhinavdangeti:
@metonymic-smokey Per our conversation, let's not go ahead with this for now. We can revisit this optimization differently once we address the correct merging of hits.

CascadingRadium:
macha is this code required now like with new aproach we not putting knn into mix with tht conj and disj searcher

PR #1930, opened 2023-12-05
`explain` in SearchRequest to toggle including request as part of response

abhinavdangeti:


metonymic-smokey (approved):

abhinavdangeti:
Turns out the SDK team doesn't really care about the request bit in the response payload, we're in the clear on this now.

metonymic-smokey (approved):

PR #1910, opened 2023-11-16
MB-59102: Merging KNN Results

CascadingRadium:
# Jira

[MB-59102](https://issues.couchbase.com/browse/MB-59102)

## Description

When performing a MultiSearch across an alias representing a partitioned index, we need special logic to merge the KNN results together into the final search result.

PR #1936, opened 2023-12-12
MB-59102: Merging kNN hits from a distributed index with exact search hits

CascadingRadium:
- Reverts commit https://github.com/blevesearch/bleve/pull/1910, which was an earlier attempt to address this issue.
- Implements the PreSearch Construct in Bleve alias search, enabling a preliminary query to collect metadata from all alias indexes before executing the main search query in MultiSearch. PreSearch gathers KNN results from all alias indexes, selecting the top K results. This facilitates the main Bleve Query to operate within the context of documents that matched the KNN query, ensuring seamless functionality of existing Bleve constructs such as Faceting, Sorting, Pagination, SearchAfter, and SearchBefore.
- Introduces the KNN Collector construct to merge and obtain accurate Top K results from multiple Zap Segments' KNN results.
- Enhances KNN Unit Tests for greater generality.
- Addresses an issue where errors generated within the Top N Document handler were being discarded.
- Resolves an issue where document matches failing to meet the SearchAfter clause weren't being returned to the pool.

### DIFF
--- a/index_alias_impl.go
+++ b/index_alias_impl.go
@@ -161,18 +161,49 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
 	if len(i.indexes) < 1 {
 		return nil, ErrorAliasEmpty
 	}
+	if _, ok := ctx.Value(search.PreSearchKey).(bool); ok {
+		// this alias is an index in a larger alias
+		// we need to do a metadata search and NOT
+		// a real search
+		return MetadataSearch(ctx, req, i.indexes...)
+	}
+
+	// at this point we know we are doing a real search
+	// either after a presearch is done, or directly
+	// on the alias
+
+	// check if request has metadata which would indicate that the
+	// request has already been presearched and we can skip the
+	// presearch step now, we call an optional function to
+	// redistribute the metadata to the individual indexes
+	// if necessary
+	var metadata []map[string]interface{}
+	if req.Metadata != nil {
+		if requestHasKNN(req) {
+			metadata = make([]map[string]interface{}, len(i.indexes))
+			for i := 0; i < len(metadata); i++ {
+				metadata[i] = make(map[string]interface{})
+			}
+			redistributeKNNMetadata(req, metadata)
+		}
+	}
 
 	// short circuit the simple case
 	if len(i.indexes) == 1 {
+		if metadata != nil {
+			req.Metadata = metadata[0]
+		}
 		return i.indexes[0].SearchInContext(ctx, req)
 	}
 
 	// at this stage we know we have multiple indexes
 	// check if metadata needs to be gathered from all indexes
 	// before executing the query
-	var metadata []map[string]interface{}
 	var err error
-	if PreSearchRequired(ctx, req) {
+	// only perform presearch if
+	//  - the request does not already have metadata
+	//  - the request requires presearch
+	if req.Metadata == nil && PreSearchRequired(ctx, req) {
 		metadata, err = PreSearch(ctx, req, i.indexes...)
 		if err != nil {
 			return nil, err
@@ -480,7 +511,7 @@ func PreSearch(ctx context.Context, req *SearchRequest, indexes ...Index) ([]map
 
 func tagHitsWithIndexNum(sr *SearchResult, indexNum int) {
 	for _, hit := range sr.Hits {
-		hit.IndexId = indexNum
+		hit.IndexId = append(hit.IndexId, indexNum)
 	}
 }
 
@@ -600,7 +631,7 @@ func MultiSearch(ctx context.Context, req *SearchRequest, metadata []map[string]
 		if metadata != nil {
 			md = metadata[idx]
 		}
-		searchChildIndex(in, createChildSearchRequest(req, md))
+		go searchChildIndex(in, createChildSearchRequest(req, md))
 	}
 
 	// on another go routine, close after finished
--- a/index_alias_impl.go
+++ b/index_alias_impl.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/blevesearch/bleve/v2/mapping"
 	"github.com/blevesearch/bleve/v2/search"
+	"github.com/blevesearch/bleve/v2/search/query"
 	index "github.com/blevesearch/bleve_index_api"
 )
 
@@ -165,7 +166,20 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
 	if len(i.indexes) == 1 {
 		return i.indexes[0].SearchInContext(ctx, req)
 	}
-	return MultiSearch(ctx, req, i.indexes...)
+
+	// at this stage we know we have multiple indexes
+	// check if metadata needs to be gathered from all indexes
+	// before executing the query
+	var metadata []map[string]interface{}
+	var err error
+	if PreSearchRequired(ctx, req) {
+		metadata, err = PreSearch(ctx, req, i.indexes...)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return MultiSearch(ctx, req, metadata, i.indexes...)
 }
 
 func (i *indexAliasImpl) Fields() ([]string, error) {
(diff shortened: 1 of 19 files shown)


abhinavdangeti:
Thank you @CascadingRadium . I'll share my review by tomorrow. 
Meanwhile - @Thejas-bhat , @metonymic-smokey , @moshaad7 , @Likith101  would you please start reviewing this.

abhinavdangeti:
@CascadingRadium I tried this change E2E with couchbase. Things don't work on 2 accounts -

- one, if I don't specify knn_operator, I see `unknown knn operator: ` - this is not right, it needs to default to `or` when unspecified
- two, I go ahead and put in knn_operator to "or", a kNN search over that SIFT dataset in combination with match_none, now returns nothing (it seems the response is just empty)

Non-vector indexes are affected as well, so this needs a bit more testing from your end.
Can you look into what's going on? Should be straightforward to reproduce.

metonymic-smokey:
@abhinavdangeti Re point 1, I believe this PR is the cause(https://github.com/blevesearch/bleve/pull/1935) since this doesn't allow an empty string as the KNN operator. If it is, it should be reverted.

abhinavdangeti:
Thanks @metonymic-smokey , here's the PR adjusting that - https://github.com/blevesearch/bleve/pull/1937

CascadingRadium:
there's a separate PR in cbft that I will raise today. the marshal unmatshal in logic there needs to be updated

index_alias_impl.go
```diff
@@ -428,19 +442,136 @@ func (i *indexAliasImpl) Swap(in, out []Index) {
 // the actual final results.
 // Perhaps that part needs to be optional,
 // could be slower in remote usages.
-func createChildSearchRequest(req *SearchRequest) *SearchRequest {
-	return copySearchRequest(req)
+func createChildSearchRequest(req *SearchRequest, metadata map[string]interface{}) *SearchRequest {
```
abhinavdangeti, line 445: Let's rename metadata to `presearchData` or something.

index_alias_impl.go
```diff
@@ -428,19 +442,136 @@ func (i *indexAliasImpl) Swap(in, out []Index) {
...
+	return metadata, nil
+}
+
+func tagHitsWithIndexNum(sr *SearchResult, indexNum int) {
+	for _, hit := range sr.Hits {
+		hit.IndexId = indexNum
+	}
+}
+
+func mergeMetadata(req *SearchRequest, res *SearchResult, numIndexes int) ([]map[string]interface{}, error) {
```
abhinavdangeti, line 487: mergePresearchData*

index_alias_impl.go
```diff
@@ -428,19 +442,136 @@ func (i *indexAliasImpl) Swap(in, out []Index) {
...
+	for i := 0; i < len(mergedOut); i++ {
+		mergedOut[i] = make(map[string]interface{})
+	}
+	if requestHasKNN(req) {
+		mergeKNNDocumentMatches(req, res.Hits, mergedOut)
+	}
+	return mergedOut, nil
+}
+
+func MetadataSearch(ctx context.Context, req *SearchRequest, indexes ...Index) (*SearchResult, error) {
```
abhinavdangeti, line 498: Presearch(*

index_alias_impl.go
```diff
@@ -428,19 +442,136 @@ func (i *indexAliasImpl) Swap(in, out []Index) {
...
+		}
+	}
+
+	return sr, nil
 }
 
 // MultiSearch executes a SearchRequest across multiple Index objects,
 // then merges the results.  The indexes must honor any ctx deadline.
-func MultiSearch(ctx context.Context, req *SearchRequest, indexes ...Index) (*SearchResult, error) {
+func MultiSearch(ctx context.Context, req *SearchRequest, metadata []map[string]interface{}, indexes ...Index) (*SearchResult, error) {
```
abhinavdangeti, line 574: ditto.

index_alias_impl.go
```diff
@@ -468,8 +595,12 @@ func MultiSearch(ctx context.Context, req *SearchRequest, indexes ...Index) (*Se
 	}
 
 	waitGroup.Add(len(indexes))
-	for _, in := range indexes {
-		go searchChildIndex(in, createChildSearchRequest(req))
+	for idx, in := range indexes {
+		var md map[string]interface{}
+		if metadata != nil {
```
abhinavdangeti, line 600: Replace all occurrences of metadata.

index_alias_impl.go
```diff
@@ -468,8 +595,12 @@ func MultiSearch(ctx context.Context, req *SearchRequest, indexes ...Index) (*Se
...
+			md = metadata[idx]
+		}
+		searchChildIndex(in, createChildSearchRequest(req, md))
```
abhinavdangeti, line 603: Should be a non blocking call.

search/util.go
```diff
@@ -134,3 +134,7 @@ const MaxGeoBufPoolSize = 24 * 1024
 const MinGeoBufPoolSize = 24
 
 type GeoBufferPoolCallbackFunc func() *s2.GeoBufferPool
+
+const KnnMetadataKey = "_knn_metadata_key"
```
abhinavdangeti, line 138: `KnnHitsKey` perhaps?

search_knn.go
```diff
@@ -46,6 +48,18 @@ type SearchRequest struct {
...
+	// in the second phase of any 2-phase search, to provide additional
+	// context to the second phase. This is useful in the case of index
+	// aliases where the first phase will gather the metadata from all
+	// the indexes in the alias, and the second phase will use that
+	// metadata to perform the actual search.
+	// The currently accepted map configuration is:
+	//
+	// "_knn_metadata_key": []*search.DocumentMatch
+
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
```
abhinavdangeti, line 61: `PresearchData` or something better.

search_no_knn.go
```diff
@@ -60,25 +63,38 @@ type SearchRequest struct {
...
+	// in the second phase of any 2-phase search, to provide additional
+	// context to the second phase. This is useful in the case of index
+	// aliases where the first phase will gather the metadata from all
+	// the indexes in the alias, and the second phase will use that
+	// metadata to perform the actual search.
+	// The currently accepted map configuration is:
+	//
+	// "_knn_metadata_key": []*search.DocumentMatch
+
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
```
abhinavdangeti, line 76: Is this needed in a non-knn setting?

abhinavdangeti (changes requested):

abhinavdangeti:
> there's a separate PR in cbft that I will raise today. the marshal unmatshal in logic there needs to be updated

I don't see this still. Are any gRPC changes required in cbft to support this?

index_impl.go
```diff
@@ -460,16 +498,31 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
+					}
+				}
+				skipKnnCollector = true
+			}
+		}
 	}
-	defer func() {
-		if cerr := indexReader.Close(); err == nil && cerr != nil {
-			err = cerr
+	if !skipKnnCollector && requestHasKNN(req) {
```
Thejas-bhat, line 518: a question over here, is this block hit when you run the KNN query on a single bleve index?

search/collector/topn.go
```diff
@@ -215,6 +222,15 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
 			}
 		}
 
+		skipped, err = hc.adjustDocumentMatch(searchContext, reader, next)
```
Thejas-bhat, line 225: let me see if i got this right, with regards to the pagination stuff, so essentially the score of the knn hits got from the presearch/runknnCollector are adjusted based on the tf-idf score in the second phase, and the topn collector which maintains a skip+size amount of doc matches finalizes the results at the end where the total score of the knn hits can spread across the range of 0-skip+size hits that are there. 

when the user tries to get the 2nd page we just slice out that portion of the 0-skip+size hits that are there and now the retrieved second page doesn't necessarily have k nearest neighbors but some number less than that (which i believe is the expected behaviour since the pagination etc. is based on the total final score of a doc match) 

is the above thinking correct?

Thejas-bhat (commented):
still trying to understand the details of the implementation, although i get the core logic behind it. will keep at it periodically and let you know any concerns i have

index_impl.go
CascadingRadium: this block gets hit only when 
- It is not presearch which means 
     - It is in phase 2 of a 2 phase search or 
     - It is a simple query with no 2 phase requirement
- If it is in phase 2 of a 2 phase search then Request will have the PresearchData/Metadata so you just use that and dont search for knn

search/collector/topn.go
CascadingRadium: say the phase 1 gives 4 docs D1 D2 D3 D4 with scores 1 2 3 4

say tfidf alone gives 3 docs D1 D5 D6 with scores 4 2 3 

say it is page 1 
from = 0 
size = 2

youll increase the score of D1 to be 5 and put D1 D2 D3 D4 D5 D6 into a min heap of size 2 and youll get the top 2 = D1 with 5 and D4 with 4

in page 2 
from = 2 
size = 2

youll now keep a heap of size 4 and do the above same step for page 1 basically
youll get D1,5 D4,4 D3,3 D6,3 in it (docId,score style this is)

you dump the first 2 since from is 2
youll output D3 and D6 which is page 2
so ye pagination is total score thing not based on knn anol

CascadingRadium:
thanks bro @Thejas-bhat for reviewing 🙏🏻

index_alias_impl.go
CascadingRadium: missed this sory

search_no_knn.go
CascadingRadium: its generic to support extensibility in the future.
i can use this for sending in merged synonym definitions for example

index_impl.go
```diff
@@ -460,16 +498,31 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
+		for k, v := range req.PreSearchData {
+			switch k {
+			case search.KnnPreSearchDataKey:
+				if v != nil {
+					knnHits, ok = v.([]*search.DocumentMatch)
+					if !ok {
+						return nil, fmt.Errorf("knn preSearchData must be of type []*search.DocumentMatch")
+					}
+				}
+				skipKnnCollector = true
```
metonymic-smokey, line 514: SearchInContext is run for all queries.
So shouldn't skipKNNCollector be true only if the KNNPreSearchKey is true and the user is allowed to run KNN queries?
If the latter is a valid condition, then there ought to be separate files which determine if skipKNNCollector should be true/false --> should always. be false for file with the !vectors tag.

Does this make sense?

CascadingRadium: no like the idea is in the incoming presearchdata in the request i can give a null value for KnnPreSearchDataKey which means that the index (in the alias) did not provide any valid knn hits. So it must just do basic tfidf and skip the knn collector search.

CascadingRadium: with the !vectors tag, the collector will never run because requestHasKNN() will always return false

search/collector/knn.go
```diff
@@ -0,0 +1,257 @@
...
+		if err != nil {
+			return nil, err
+		}
+		rv[i] = doc
+		i++
+	}
+	return rv, nil
+}
+
+func MakeKNNDocMatchHandler(ctx *search.SearchContext) (search.DocumentMatchHandler, error) {
```
metonymic-smokey, line 93: Again, same point as earlier:
Do we want to limit this DocMatchHandler() and the related KNN collector functions to files with vectors tag so they're used only for KNN?

index_alias_impl.go
```diff
@@ -428,19 +473,136 @@ func (i *indexAliasImpl) Swap(in, out []Index) {
...
+	sr.Took = searchDuration
+
+	// fix up errors
+	if len(indexErrors) > 0 {
+		if sr.Status.Errors == nil {
+			sr.Status.Errors = make(map[string]error)
+		}
+		for indexName, indexErr := range indexErrors {
+			sr.Status.Errors[indexName] = indexErr
+			sr.Status.Total++
```
metonymic-smokey, line 595: Why are we incrementing Total only if errors are greater than 0?

metonymic-smokey: As far as I see, on line 554, we aren't incrementing Total in 'sr', so am I missing something?

metonymic-smokey: Oh never mind, realised my oversight.
sr.Merge() does increment Total.

index_alias_impl.go
```diff
@@ -160,12 +161,56 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
 	if len(i.indexes) < 1 {
 		return nil, ErrorAliasEmpty
 	}
+	if _, ok := ctx.Value(search.PreSearchKey).(bool); ok {
+		// this alias is an index in a larger alias
```
abhinavdangeti, line 165: This comment didn't make much sense to me, reword.

search_knn.go
```diff
@@ -167,146 +191,175 @@ var (
...
+		distributeKNNHit(docMatch, indexNumToDocMatchList)
+	}
+	for i := 0; i < len(mergeOut); i++ {
+		mergeOut[i][search.KnnPreSearchDataKey] = indexNumToDocMatchList[i]
+	}
 }
 
-func (s *scoreHeap) Swap(i, j int) {
-	s.scoreMaps[i], s.scoreMaps[j] = s.scoreMaps[j], s.scoreMaps[i]
+func distributeKNNHit(docMatch *search.DocumentMatch, indexNumToDocMatchList map[int][]*search.DocumentMatch) {
```
abhinavdangeti, line 328: Add commentary to what this api is doing.

search_knn.go
```diff
@@ -167,146 +191,175 @@ var (
...
-	return x
+func addKnnToDummyRequest(dummyReq *SearchRequest, realReq *SearchRequest) {
+	dummyReq.KNN = realReq.KNN
+	dummyReq.KNNOperator = knnOperatorOr
 }
 
-func mergeKNN(req *SearchRequest, sr *SearchResult) {
-	maxHeap := &scoreHeap{
-		scoreMaps: make([]map[int]float64, 0),
+func redistributeKNNPreSearchData(req *SearchRequest, mergedOut []map[string]interface{}) error {
```
abhinavdangeti, line 343: Add commentary to what this api is doing.

search_knn.go
```diff
@@ -46,6 +48,19 @@ type SearchRequest struct {
...
+	// PreSearchData will be a  map that will be used
+	// in the second phase of any 2-phase search, to provide additional
+	// context to the second phase. This is useful in the case of index
+	// aliases where the first phase will gather the PreSearchData from all
+	// the indexes in the alias, and the second phase will use that
+	// PreSearchData to perform the actual search.
+	// The currently accepted map configuration is:
+	//
+	// "_knn_pre_search_data_key": []*search.DocumentMatch
+	// "_synonym_pre_search_data_key":		[]*synonym.SynonymDefinition
```
abhinavdangeti, line 60: Drop all references to synonyms in this commit.

abhinavdangeti (approved):
Let's go with this. Well done @CascadingRadium .
Keep an eye out for any optimizations we can fish out in this area for the future.

PR #1941, opened 2023-12-15
MB-60112 - adding a supported version check for vector search on query path

Thejas-bhat:


index/scorch/snapshot_index_vr.go
```diff
@@ -28,6 +28,8 @@ import (
 	segment_api "github.com/blevesearch/scorch_segment_api/v2"
 )
 
+const VectorSeachSupportedSegmentVersion = 16
```
metonymic-smokey, line 31: Do you think this should be exported from zapx instead?
Something like this - https://github.com/blevesearch/zapx/blob/86b116167ea9755282d9a830f5d159cd0b3b857a/build.go#L27 .

Maybe it could be titled MinVectorSearchSupportedVersion since we may release newer zapx versions, all of which will support vector search and this will just be the minimum.

Just my $0.02. What do you think?

Thejas-bhat: Actually I think scorch_segment_api is the best place for it

abhinavdangeti: Neither zapx nor scorch_segment_api feel like the right place for this.
Let's leave this here and get the check in place for now. 

I think this naming and the check should be ok when we introduce later zapx versions (whenever that is) also.

abhinavdangeti (approved):

abhinavdangeti:
Only a rebase here to get the unit tests to run, which hadn't for some reason.

PR #1944, opened 2023-12-19
Code Cleanup + Optimizations

CascadingRadium:
- Cleanup up unused code paths since the merging of https://github.com/blevesearch/bleve/pull/1936 changed the overall approach, resulting in a lot of dead code.
- Optimized the KNN collector to not update the allHits map for every hit and instead just use its internal heaps when the Final() is called.
- Typo Fix
- Optimized the OptimizeVR's Finish() method to concurrently search across segments.
 
Results of Optimization

Dataset -  SIFT 1 Million dataset 
Index - 1 partition index 

Before this commit - Query Execution time - 2.25 seconds
With this commit - Query Execution time - 1.70 seconds
Percentage Reduction - **24.4%**


abhinavdangeti:
@CascadingRadium let's have you walk the team through this real quick on tomorrow's call. We can move ahead with review and merging after.

index/scorch/optimize_knn.go
```diff
@@ -36,35 +37,48 @@ func (o *OptimizeVR) Finish() error {
...
-					// by passing the obtained vector index and getting similar vectors.
-					pl, err := searchVectorIndex(vr.field, vr.vector, vr.k, seg.deleted)
+		if sv, ok := seg.segment.(segment_api.VectorSegment); ok {
+			wg.Add(1)
+			go func(index int, segment segment_api.VectorSegment, origSeg *SegmentSnapshot) {
+				defer wg.Done()
+				for field, vrs := range o.vrs {
+					searchVectorIndex, closeVectorIndex, err := segment.InterpretVectorIndex(field)
+					defer func() {
+						if closeVectorIndex != nil {
```
abhinavdangeti, line 51: Remove this nil check and move this defer to 58.5 below, because if err == nil, neither searchVectorIndex nor closeVectorIndex will ever be nil.

index/scorch/optimize_knn.go
```diff
@@ -36,35 +37,48 @@ func (o *OptimizeVR) Finish() error {
 	// for each VR, populate postings list and iterators
 	// by passing the obtained vector index and getting similar vectors.
 	// defer close index - just once.
-
+	errors := make([]error, len(o.snapshot.segment))
```
abhinavdangeti, line 40: Don't see why this is necessary - you can just declare this as ..

```
var errors []error
```

And then where ever you encounter an err, append to that list (which will alloc).

Replace line 79-81 below with ..
```
if len(errors) > 0 {
    return errors[0]
}
```

abhinavdangeti (changes requested):

abhinavdangeti (approved):

search/collector/topn.go
```diff
@@ -39,6 +39,9 @@ type collectorStore interface {
 	AddNotExceedingSize(doc *search.DocumentMatch, size int) *search.DocumentMatch
 
 	Final(skip int, fixup collectorFixup) (search.DocumentMatchCollection, error)
+
+	// Provide access the internal heap implementation
+	Internal() search.DocumentMatchCollection
```
abhinavdangeti, line 44: `collectStoreList` needs to deploy this as well?

abhinavdangeti (changes requested):

search/collector/topn.go
CascadingRadium: collectStoreHeap and collectStoreHeap are both impl of the collector interface and represent a general-purpose K-size heap. collectStoreList just skips usage of golang's built-in heap since its faster for K < 10. So hence both the collectStoreList and collectStoreHeap should impl the Internal() api that just provides access to the internal slice being used for the heap.

abhinavdangeti: Right, but I don't see where you're implementing the Internal() api for collectStoreList.

CascadingRadium: oh ok i got confused between collectStoreSlice and collectStoreList 
i see that only collectStoreSlice and collectStoreHeap is used and collectStoreList is unused / deprecated. Hence i had forgot to implement for collectStoreList.

index_alias_impl.go
```diff
@@ -205,14 +205,34 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
...
+		if preSearchResult.Status.Failed > 0 {
+			return preSearchResult, nil
+		} else {
+			// if there are no errors, then use the presearch data
+			// to execute the query
+			preSearchData, err = mergePreSearchData(req, preSearchResult, len(i.indexes))
+			if err != nil {
+				return nil, err
+			}
+		}
```
abhinavdangeti, line 227: All this code (preSearchResult.status check and mergePreSearchData) was better positioned where it was. Let's move this back to where it was and just do the preSearchDuration calculation here.

abhinavdangeti: I understand why you're doing this though - so this is ok.

abhinavdangeti (approved):

PR #1946, opened 2023-12-20
MB-60207 fix facets merge

moshaad7:


search/facets_builder.go
```diff
@@ -321,7 +321,11 @@ func (fr *FacetResult) Merge(other *FacetResult) {
 	fr.Total += other.Total
 	fr.Missing += other.Missing
 	fr.Other += other.Other
-	if fr.Terms != nil && other.Terms != nil {
+	if other.Terms != nil {
+		if fr.Terms == nil {
+			fr.Terms = other.Terms
```
metonymic-smokey, line 326: So from what I understand, if there are no term facets, you can simply assign Other's term range facets and early exit.
If fr.numeric/date range facets are nil, why don't we do the same there?
just trying to understand.

metonymic-smokey: actually, nvm, I realised why.

search/facets_builder.go
```diff
@@ -321,7 +321,11 @@ func (fr *FacetResult) Merge(other *FacetResult) {
 	fr.Total += other.Total
 	fr.Missing += other.Missing
 	fr.Other += other.Other
-	if fr.Terms != nil && other.Terms != nil {
+	if other.Terms != nil {
```
metonymic-smokey, line 324: QQ: Why aren't we doing this for numeric/date range facets?
The key idea being we want to merge facet results if even one of them is not nil. So I think that should apply to other facet types too?
Pls let me know if i missed something, thanks

moshaad7: Added it for numeric and date ranges too.

The reason I didn't add these before was because.
NumericRanges and DateRanges have omitempty json tag. 

Meaning, when marshalling, they will never be marshalled as "null"

During unmarshalling, they will always be empty slice and not nil.


But, In future, if we decide to add customer Marshal Unmarshal Logic to these, then Merge should take care of it.

metonymic-smokey (approved):

Thejas-bhat (approved):

PR #1948, opened 2023-12-21
MB-60081: Fix Score Explanation

CascadingRadium:
- The score explanation is now accurate. It's determined by establishing the final score and its explanation only when the KNN hits are finalized. This happens after a PreSearch for an index alias or at the end of the KNN collector phase for a single index.

### DIFF
--- a/search/collector/knn.go
+++ b/search/collector/knn.go
@@ -83,9 +83,11 @@ func (c *collectStoreKNN) Final(fixup collectorFixup) (search.DocumentMatchColle
 	rv := make(search.DocumentMatchCollection, size)
 	i := 0
 	for doc := range c.allHits {
-		err := fixup(doc)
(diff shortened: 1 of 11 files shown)


Thejas-bhat (commented):
from what i understand, the explanation is being corrected at the finalizeResults level by adding another level of explanation of total score.
i see that in the unit test file you're doing a query.Explain = true but i dont see any specific unit test checks happening specifically in the verifyResult() function of the TestSimilaritySearchMultipleSegments (for eg). 
is it possible to have the unit testing in that verifyResult() function and wherever relevant? there is a unit test for score explanation in the scorer_knn_test file however i dont think that unit testing happens at the same level of logic as the propose code changes

Thejas-bhat (approved):

Issue #1956, opened 2024-01-08
[Bug] Spanish analyzer not normalizing all accented words.


svera:
Hi,
I've been working with the spanish analyzer in order to index documents in spanish, when I found and issue with some accented words which are not normalised as they should be. This is easily reproducible using the [Bleve text analysis wizard](https://bleveanalysis.couchbase.com/analysis), choosing the `es` analyzer and putting `fría` or `guía` in the _Text to analyze_ input box. The words are kept as they are, however, using plural forms of these words (`guías` and `frías`) works as expected. 

Other accented words such as `tentación` or `comeré` are correctly stemmed to `coleccion` and `comer`, respectively.

svera:
I've found where the issue comes from. Looking at [bleve](https://github.com/blevesearch/bleve/tree/v2.3.8)/[analysis](https://github.com/blevesearch/bleve/tree/v2.3.8/analysis)/[lang](https://github.com/blevesearch/bleve/tree/v2.3.8/analysis/lang)/[es](https://github.com/blevesearch/bleve/tree/v2.3.8/analysis/lang/es)
/light_stemmer_es.go, the normalization of accented letters only happens if the input is larger than 5 characters, something that neither `guía` nor `fría` comply.
The solution would be to always execute the accented characters normalization, by moving it to a separate file just like it is done in the german analyzer, or maybe use the `asciifolding` filter.

svera:
PR opened addressing this issue: https://github.com/blevesearch/bleve/pull/1957

abhinavdangeti:
Thanks for raising the pull request @svera . The team will review soon.

PR #1957, opened 2024-01-08
Fixed spanish accents normalization

svera:
Normalization of accented letters only happens if the input is larger than 5 characters, something that, for example, neither `guía` nor `fría` comply.
The solution would be to always execute the accented characters normalization, by moving it to a separate file just like it is done in the german analyzer.

Fixes: https://github.com/blevesearch/bleve/issues/1956


abhinavdangeti (approved):
Looks good @svera , let's wait on one more review possibly.

PR #1958, opened 2024-01-09
MB-60241: Add consistency checks for redistribution of KNN hits

CascadingRadium:
In a distributed indexing setup, the child indexes linked to an alias are labeled as `[IP]:[port]/[indexName]`. During a two-phase KNN search, if a node is taken out and re-added (by means of a rebalance operation) after the first phase but before the start of the second phase, the reference to the child index in the alias remains unchanged (with the same IP address, port, and indexName). However, the actual indexes present on that node will differ from those available during the first phase. This discrepancy causes inaccurate results due to inconsistencies in the indexId stack.

This method substitutes the indexId with the specific indexName, ensuring redistribution even if the index arrangement undergoes changes. Yet, if the index isn’t found within that alias, an error will be triggered.

### DIFF
--- a/error.go
+++ b/error.go
@@ -26,7 +26,7 @@ const (
 	ErrorUnknownIndexType
 	ErrorEmptyID
 	ErrorIndexReadInconsistency
-	ErrorSearchInconsistency
+	ErrorTwoPhaseSearchInconsistency
 )
 
 // Error represents a more strongly typed bleve error for detecting
@@ -38,15 +38,15 @@ func (e Error) Error() string {
 }
 
 var errorMessages = map[Error]string{
-	ErrorIndexPathExists:        "cannot create new index, path already exists",
-	ErrorIndexPathDoesNotExist:  "cannot open index, path does not exist",
-	ErrorIndexMetaMissing:       "cannot open index, metadata missing",
-	ErrorIndexMetaCorrupt:       "cannot open index, metadata corrupt",
-	ErrorIndexClosed:            "index is closed",
-	ErrorAliasMulti:             "cannot perform single index operation on multiple index alias",
-	ErrorAliasEmpty:             "cannot perform operation on empty alias",
-	ErrorUnknownIndexType:       "unknown index type",
-	ErrorEmptyID:                "document ID cannot be empty",
-	ErrorIndexReadInconsistency: "index read inconsistency detected",
-	ErrorSearchInconsistency:    "topology change caused two phase search to fail",
+	ErrorIndexPathExists:             "cannot create new index, path already exists",
+	ErrorIndexPathDoesNotExist:       "cannot open index, path does not exist",
+	ErrorIndexMetaMissing:            "cannot open index, metadata missing",
+	ErrorIndexMetaCorrupt:            "cannot open index, metadata corrupt",
+	ErrorIndexClosed:                 "index is closed",
+	ErrorAliasMulti:                  "cannot perform single index operation on multiple index alias",
+	ErrorAliasEmpty:                  "cannot perform operation on empty alias",
+	ErrorUnknownIndexType:            "unknown index type",
+	ErrorEmptyID:                     "document ID cannot be empty",
+	ErrorIndexReadInconsistency:      "index read inconsistency detected",
+	ErrorTwoPhaseSearchInconsistency: "2-phase search failed, likely due to an overlapping topology change",
 }
--- a/error.go
+++ b/error.go
@@ -26,7 +26,7 @@ const (
 	ErrorUnknownIndexType
 	ErrorEmptyID
 	ErrorIndexReadInconsistency
-	ErrorPreSearchFailed
+	ErrorSearchInconsistency
 )
 
 // Error represents a more strongly typed bleve error for detecting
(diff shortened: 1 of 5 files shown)


search_knn.go
```diff
@@ -349,10 +342,35 @@ func mergeKNNDocumentMatches(req *SearchRequest, knnHits []*search.DocumentMatch
...
+		top := hit.IndexNames[stackTopIdx]
+		hit.IndexNames = hit.IndexNames[:stackTopIdx]
+		segregatedKnnHits[top] = append(segregatedKnnHits[top], hit)
+	}
+	return segregatedKnnHits
+}
+
+func checkIndexNamesConsistency(knnHits []*search.DocumentMatch, indexes []Index) error {
+	// create a set of all the index names of this alias
+	indexNames := make(map[string]bool)
```
moshaad7, line 358: In my opinion, It would be better to use map[string}struct{} instead of map[string]bool

search_knn.go
```diff
@@ -349,10 +342,35 @@ func mergeKNNDocumentMatches(req *SearchRequest, knnHits []*search.DocumentMatch
...
+	for _, index := range indexes {
+		indexNames[index.Name()] = true
+	}
+
+	// a topology change / inconsistency is detected when, for a given KNN hit,
```
moshaad7, line 363: I don't think we need this commentary. 
Code is simple enough and self explanatory .

search_knn.go
```diff
@@ -349,10 +342,26 @@ func mergeKNNDocumentMatches(req *SearchRequest, knnHits []*search.DocumentMatch
 // hits, then the top K hits need to be distributed to I1 and I2,
 // so that the preSearchData for I1 contains the top K hits from I1 and
 // the preSearchData for I2 contains the top K hits from I2.
-func distributeKNNHit(docMatch *search.DocumentMatch, indexNumToDocMatchList map[int][]*search.DocumentMatch) {
-	top := docMatch.IndexId[len(docMatch.IndexId)-1]
-	docMatch.IndexId = docMatch.IndexId[:len(docMatch.IndexId)-1]
-	indexNumToDocMatchList[top] = append(indexNumToDocMatchList[top], docMatch)
+func validateAndDistributeKNNHits(knnHits []*search.DocumentMatch, indexes []Index) (map[string][]*search.DocumentMatch, error) {
+	// create a set of all the index names of this alias
+	indexNames := make(map[string]interface{})
```
moshaad7, line 347: make(map[string]struct{})

error.go
```diff
@@ -48,5 +48,5 @@ var errorMessages = map[Error]string{
 	ErrorUnknownIndexType:       "unknown index type",
 	ErrorEmptyID:                "document ID cannot be empty",
 	ErrorIndexReadInconsistency: "index read inconsistency detected",
-	ErrorPreSearchFailed:        "index pre-search failed",
+	ErrorPreSearchFailed:        "index pre-search failed, inconsistency in results detected, due to a topology change/rebalance. Please retry the search",
```
abhinavdangeti, line 51: Drop `Please retry the search` from error message. Also it seems you're using the same error message that's used else where - which is a bad idea. Let's declare and use a new one -

`ErrorSearchInconsistency: "Topology change caused 2-phase search to fail"`

abhinavdangeti (changes requested):

abhinavdangeti:
@CascadingRadium Is there an optimization in place where we skip the 2nd phase in a distributed search when it's pure-kNN, i.e query is match_none ?

search/search.go
```diff
@@ -183,12 +183,12 @@ type DocumentMatch struct {
...
-	// indicated the index id of the index that this match came from
-	// used in vector search.
-	// it is a stack of index ids, the top of the stack is the index id
+	// to indicate the name of the index that this match came from.
+	// used in knn search.
+	// it is a stack of index names, the top of the stack is the name
 	// of the index that this match came from
 	// of the current alias view, used in alias of aliases scenario
-	IndexId []int `json:"index_id,omitempty"`
+	IndexNames []string `json:"index_names,omitempty"`
```
abhinavdangeti, line 191: I worry about the noise this is going to generate to the end user.

CascadingRadium:
yea the optimization to skip phase 2 is done i was finalizing it, i will post the PR tomorrow. had gotten side tracked with this one

search/search.go
CascadingRadium: this stack will not be output to the end user, since when the doc match reaches the index in phase 2, this slice will always be empty, it is similar to score breakdown, both score breakdown and indexNames will be used between the 2 phases only, the final result will be sans them

abhinavdangeti: That's good!

abhinavdangeti:
Sounds good - just the one open comment on the error messaging here then.

search_knn.go
```diff
@@ -351,10 +354,10 @@ func validateAndDistributeKNNHits(knnHits []*search.DocumentMatch, indexes []Ind
 	segregatedKnnHits := make(map[string][]*search.DocumentMatch)
 	for _, hit := range knnHits {
 		if len(hit.IndexNames) == 0 {
-			return nil, ErrorPreSearchFailed
+			return nil, ErrorSearchInconsistency
```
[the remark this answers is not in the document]
CascadingRadium: ErrorPreSearchFailed is not used anywhere so i just removed it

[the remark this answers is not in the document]
CascadingRadium: no 
its an `or` condition, 
if `len(hit.IndexNames) == 0` then topology change occurred that replaced an alias with another
the presearch / phase 1 would have been completed by this point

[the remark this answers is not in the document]
abhinavdangeti: My comment is more around the cause for these 2 error situations being different - which warrant different error messaging - do you disagree?

How about you add some commentary around when `len(hit.IndexNames) is 0` and when line:359 below can occur. Perhaps that'll clarify things.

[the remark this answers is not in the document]
CascadingRadium: ok yea the stack wont be empty in any topology change scene. 
but since i am peeking the stack top just to be safe im checking if stack is empty just to avoid panik

abhinavdangeti (approved):

PR #1961, opened 2024-01-11
Optionally skip multisearch after presearch

CascadingRadium:
If a search request has KNN requests in it, and
- the main query is a match_none 
- the facets request is nil 

the final search result is constructed with just the presearch results, and multi search is skipped.

### DIFF
--- a/index_alias_impl.go
+++ b/index_alias_impl.go
@@ -218,20 +218,22 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
 		// so that the errors are not lost
 		if preSearchResult.Status.Failed > 0 {
 			return preSearchResult, nil
+		}
+
+		// if there are no errors, then merge the data in the presearch result
+		preSearchResult = mergePreSearchResult(req, preSearchResult, i.indexes)
+		if requestSatisfiedByPreSearch(req) {
+			sr = finalizeSearchResult(req, preSearchResult)
+			// no need to run the 2nd phase MultiSearch(..)
 		} else {
-			// if there are no errors, then merge the data in the presearch result
-			preSearchResult = mergePreSearchResult(req, preSearchResult, i.indexes)
-			if requestSatisfiedByPreSearch(req) {
-				sr = finalizeSearchResult(req, preSearchResult)
-			} else {
-				preSearchData, err = constructPreSearchData(req, preSearchResult, i.indexes)
-				if err != nil {
-					return nil, err
-				}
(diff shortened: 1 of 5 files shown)


index_alias_impl.go
```diff
@@ -218,18 +219,25 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
 		if preSearchResult.Status.Failed > 0 {
 			return preSearchResult, nil
 		} else {
```
abhinavdangeti, line 228: You can drop this else line.

search_knn.go
```diff
@@ -266,6 +296,14 @@ func (i *indexImpl) runKnnCollector(ctx context.Context, req *SearchRequest, rea
 	if !preSearch {
 		knnHits = finalizeKNNResults(req, knnHits, len(req.KNN))
 	}
+	// at this point, irrespective of whether it is a presearch or not,
+	// the knn hits are populated with Sort and Fields.
+	// it must be ensured downstream that the Sort and Fields are not
+	// re-evaluated, for these hits.
```
abhinavdangeti, line 302: Can you point me to where this is being handled?

abhinavdangeti (commented):
Pushed 1 commit addressing a small refactor. An in-line open question to @CascadingRadium .

search_knn.go
CascadingRadium: to avoid putting sort values to knn hits 
- in index_impl.go -> `if len(req.Fields) > 0 && hit.Fields == nil {` i have added
  so in multisearch, if i encounter a hit with where hit.Fields is already present, it 
  could only mean it was a knn hit whose hit.Fields was calculated in presearch itself, 
  so its a null op 
to avoid putting the hit.Sort into doc match again
- topn.go -> i have made some changes
   - if a doc is common, ie between fts and knn, i would call prepareDocMatch with 
     `knnDoc` as false, so the normal code path will be followed, once that doc match 
     has been 'adjusted' to increase the score
   - any left over docs, pure knn hits from presearch will be 'prepared' with `knnDoc`
     as true, where i call the special dvVisitor (not the one in hc) that only calls 
     UpdateVisitor for the facetsBuilder, and i exit out before i call Sort.Value(d) as 
     it was already called for d in presearch, so basically the doc value workflow for 
     only knn docs are split with sort's DV values being evaluated at knn collector / 
     presearch and facet DV values being calculated in topn, which allows for 
     multisearch skippability when facets are not there in request.

abhinavdangeti: Ack.

abhinavdangeti (approved):

PR #1964, opened 2024-01-15
MB-60370: Reset search result if presearch fails

CascadingRadium:
Ensure the integrity of Presearch operation by resetting the presearch result in an alias when encountering the following conditions, preventing inaccuracies in the final result:
- Any child index fails during the operation.
- Any child alias provides partial results.

### DIFF
--- a/index_alias_impl.go
+++ b/index_alias_impl.go
@@ -641,8 +641,15 @@ func preSearchDataSearch(ctx context.Context, req *SearchRequest, indexes ...Ind
 		}
 	}
 
-	// fix up errors
-	if len(indexErrors) > 0 {
+	// in presearch partial results are not allowed as it can lead to
+	// the real search giving incorrect results, and hence the search
+	// result is reset.
+	// discard partial hits if some child index has failed or
+	// if some child alias has returned partial results.
+	if len(indexErrors) > 0 || sr.Status.Failed > 0 {
+		sr = &SearchResult{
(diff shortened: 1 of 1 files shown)


metonymic-smokey:
Looks good to me, will take another look tomorrow just to be doubly sure.

index_alias_impl.go
```diff
@@ -641,8 +641,15 @@ func preSearchDataSearch(ctx context.Context, req *SearchRequest, indexes ...Ind
...
-	if len(indexErrors) > 0 {
+	// in presearch partial results are not allowed as it can lead to
+	// the real search giving incorrect results, and hence the search
+	// result is reset.
+	// discard partial hits if some child index has failed or
+	// if some child alias has returned partial results.
+	if len(indexErrors) > 0 || sr.Status.Failed > 0 {
+		sr = &SearchResult{
+			Status: sr.Status,
+		}
```
abhinavdangeti, line 652: What part of the SearchResult were re propagating here that causes the issue?

CascadingRadium: sr.Hits
By calling sr.Merge(asr) earlier, the hits were being appended.
so if any partition errored out, partial results was there, with the errors being 
displayed in the search status

abhinavdangeti: Ack.

abhinavdangeti (approved):

PR #1967, opened 2024-01-18
MB-60400: Add a semaphore for concurrently reading vector indexes from zap segments

CascadingRadium:
- Calling InterpretVectorIndex() with BleveMaxK = 10000 results in 200KB of memory usage per call. With a million documents and around 1000 Zap segments (a scenario that occurs while the index is still being built and the merger routine is still running), parallel calls consume 200MB, making it necessary to limit the concurrency involved.
- Created a global variable named BleveMaxKNNConcurrency with a value of 10, representing the maximum number of concurrent goroutines allowed. This ensures that 10 segments can be read concurrently at once.

### DIFF
--- a/index/scorch/optimize_knn.go
+++ b/index/scorch/optimize_knn.go
@@ -32,6 +32,8 @@ type OptimizeVR struct {
 	vrs map[string][]*IndexSnapshotVectorReader
 }
 
+var BleveMaxKNNConcurrency = 10
+
 func (o *OptimizeVR) Finish() error {
 	// for each field, get the vector index --> invoke the zap func.
 	// for each VR, populate postings list and iterators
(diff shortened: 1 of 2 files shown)


metonymic-smokey:
As an alternative, setting the number of OpenMP threads fixes the high CPU issue too - there was an explosion in the number of threads without it.

CascadingRadium:
yaav how u do that da

CascadingRadium:
ahh sike idea da i can use tht semaphore like this

```
	// Define the maximum number of concurrent goroutines
	maxGoroutines := 10
	semaphore := make(chan struct{}, maxGoroutines)
	for i, seg := range o.snapshot.segment {
		wg.Add(1)
		semaphore <- struct{}{} // Acquire a semaphore slot
		go func(index int, segment segment_api.VectorSegment, origSeg *SegmentSnapshot) {
			// Defer releasing the semaphore slot when the goroutine completes
                        // blah blah code
			defer func() {
				<-semaphore // Release the semaphore slot
			}()
			processSegment(index, segment, origSeg)
		}(i, seg, &o.snapshot)
	}
```

metonymic-smokey:
> ahh sike idea da i can use tht semaphore like this
> 
> ```
> 	// Define the maximum number of concurrent goroutines
> 	maxGoroutines := 10
> 	semaphore := make(chan struct{}, maxGoroutines)
> 	for i, seg := range o.snapshot.segment {
> 		wg.Add(1)
> 		semaphore <- struct{}{} // Acquire a semaphore slot
> 		go func(index int, segment segment_api.VectorSegment, origSeg *SegmentSnapshot) {
> 			// Defer releasing the semaphore slot when the goroutine completes
>                         // blah blah code
> 			defer func() {
> 				<-semaphore // Release the semaphore slot
> 			}()
> 			processSegment(index, segment, origSeg)
> 		}(i, seg, &o.snapshot)
> 	}
>         wg.Wait()
> ```
> 
> is this good idea da like u have only max 10 go routine running like this

So if I understand correctly, there could be multiple bleve indexes(due to multiple FTS indexes) being queried concurrently and each of them can have up to 10 goroutines spawned at a given moment?

metonymic-smokey:
> yaav how u do that da

Just a 1 hacky liner C++ function which sets num threads to 1 when searching - seemed to fix the high CPU issue and the thread count too stayed low.

abhinavdangeti (approved):

PR #1971, opened 2024-01-25
MB-60945 - Add scorch event for pre merge check

metonymic-smokey:
This PR adds an event to check whether a merge can proceed based on criteria set by the caller(eg. cbft). If not, it's aborted and the merger loop continues its next iteration. 
Also includes minor refactor to the function signature of the event callback registry to accommodate the new event type.

### DIFF
--- a/index/scorch/event.go
+++ b/index/scorch/event.go
@@ -22,11 +22,7 @@ var RegistryAsyncErrorCallbacks = map[string]func(error, string){}
 
 // RegistryEventCallbacks should be treated as read-only after
 // process init()'ialization.
-var RegistryEventCallbacks = map[string]func(Event){}
-
-// RegistryContinueEventCallback should be treated as read-only after
-// process init()'ialization.
-var RegistryContinueEventCallback = map[string]func(Event) bool{}
+var RegistryEventCallbacks = map[string]func(Event) bool{}
 
 // Event represents the information provided in an OnEvent() callback.
 type Event struct {
--- a/index/scorch/event.go
+++ b/index/scorch/event.go
@@ -67,6 +67,6 @@ var EventKindMergeTaskIntroductionStart = EventKind(7)
 // the introduction of merged segment from a single merge task.
 var EventKindMergeTaskIntroduction = EventKind(8)
 
-// Fired before the merge begins to check if there's sufficient memory to
-// proceed with the merge.
-var EventKindPreMergeMemoryCheck = EventKind(9)
+// EventKindPreMergeCheck is fired before the merge begins to check if
+// the caller should proceed with the merge.
+var EventKindPreMergeCheck = EventKind(9)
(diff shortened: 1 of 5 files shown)


metonymic-smokey:
TODO Will investigate other metrics for tracking memory used during a merge.

metonymic-smokey:
Potential downsides:
1. Since there is no estimate for the memory used by a merge, it's possible that merges which wouldn't take up very high memory too would be unnecessarily blocked.

index/scorch/scorch.go
```diff
@@ -190,6 +195,14 @@ func (s *Scorch) fireAsyncError(err error) {
 	atomic.AddUint64(&s.stats.TotOnErrors, 1)
 }
 
+func (s *Scorch) continueEvent(kind EventKind) bool {
```
abhinavdangeti, line 198: Should name this better - perhaps `skipEvent`

index/scorch/scorch.go
```diff
@@ -69,8 +69,9 @@ type Scorch struct {
 	rootBolt                 *bolt.DB
 	asyncTasks               sync.WaitGroup
 
-	onEvent      func(event Event)
-	onAsyncError func(err error, path string)
+	onEvent               func(event Event)
+	onAsyncError          func(err error, path string)
+	continueEventCallback func(event Event) bool
```
abhinavdangeti, line 74: skipEventCallback*

index/scorch/event.go
```diff
@@ -24,6 +24,10 @@ var RegistryAsyncErrorCallbacks = map[string]func(error, string){}
 // process init()'ialization.
 var RegistryEventCallbacks = map[string]func(Event){}
 
+// RegistryContinueEventCallback should be treated as read-only after
+// process init()'ialization.
+var RegistryContinueEventCallback = map[string]func(Event) bool{}
```
abhinavdangeti, line 29: Can we consolidate this map and the above into one - it's ok to change the return type for all the existing events if you must.

abhinavdangeti (changes requested):

abhinavdangeti:
@metonymic-smokey have you checked why the unit tests are failing?

index/scorch/event.go
```diff
@@ -22,7 +22,7 @@ var RegistryAsyncErrorCallbacks = map[string]func(error, string){}
 
 // RegistryEventCallbacks should be treated as read-only after
 // process init()'ialization.
-var RegistryEventCallbacks = map[string]func(Event){}
+var RegistryEventCallbacks = map[string]func(Event) bool{}
```
abhinavdangeti, line 25: Would you add some commentary above to explain when the expected return values must be true for the function above.

abhinavdangeti (commented):
Minor comment, looks ok to me otherwise.

index/scorch/event.go
metonymic-smokey: The return values are determined by the callbacks. Without a callback, they will fall back to true.

abhinavdangeti (approved):

PR #1974, opened 2024-02-01
MB-60367: App herder check on seacher creation path

Thejas-bhat:
- invoking an app_herder check during the searcher creation time so that if the memory is spiking we don't do the expensive job of loading the vector index from buffer (which can be big) and searching it. 
- the approach is to leverage the segmentSnapshot struct to get what's the size of the vec index so that the future searcher creations will get limited when there is memory pressure as per app_herder decision
- the app_herder check to mark the start of the operation to be limited (which is the searcher creation) is invoked at the VectorOptimize() level and the check marking the end of it is invoked at the defer of the Finish() call

### DIFF
--- a/index/scorch/optimize_knn.go
+++ b/index/scorch/optimize_knn.go
@@ -28,9 +28,9 @@ import (
 )
 
 type OptimizeVR struct {
-	ctx      context.Context
-	snapshot *IndexSnapshot
-
+	ctx       context.Context
+	snapshot  *IndexSnapshot
+	totalCost uint64
 	// maps field to vector readers
 	vrs map[string][]*IndexSnapshotVectorReader
 }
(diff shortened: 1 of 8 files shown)


index/scorch/optimize_knn.go
```diff
@@ -56,22 +59,45 @@ func (o *OptimizeVR) Finish() error {
...
-					searchVectorIndex, closeVectorIndex, err := segment.InterpretVectorIndex(field)
+					vectorIndexInterpretImpl, err := segment.InterpretVectorIndex(field)
 					if err != nil {
 						errorsM.Lock()
 						errors = append(errors, err)
 						errorsM.Unlock()
 						return
 					}
+
+					vectorIndexSize := vectorIndexInterpretImpl.Size()
```
Thejas-bhat, line 70: this code block should be moved to the beginning of the function - the idea is to do this check once for every searcher creation not for every segment

PR #1976, opened 2024-02-02
MB-60367: App herder check on seacher creation path

abhinavdangeti:
- invoking an app_herder check during the searcher creation time so that if the memory is spiking we don't do the expensive job of loading the vector index from buffer (which can be big) and searching it.
- the approach is to leverage the segmentSnapshot struct to get what's the size of the vec index so that the future searcher creations will get limited when there is memory pressure as per app_herder decision
- the app_herder check to mark the start of the operation to be limited (which is the searcher creation) is invoked at the VectorOptimize() level and the check marking the end of it is invoked at the defer of the Finish() call


index/scorch/optimize_knn.go
```diff
@@ -57,22 +69,26 @@ func (o *OptimizeVR) Finish() error {
...
 					if err != nil {
 						errorsM.Lock()
 						errors = append(errors, err)
 						errorsM.Unlock()
 						return
 					}
+
+					// update the vector index size as a meta value in the segment snapshot
+					vectorIndexSize := vecIndex.Size()
+					seg.cachedMeta.updateMeta(field, vectorIndexSize)
```
abhinavdangeti, line 82: I'm not sure I follow this. Isn't it here we want to prevent a Search() if the app_herder deems it so. Or if i am interpreting this right - you want subsequent access to these segments to be stopped on this account?

abhinavdangeti (commented):
@Thejas-bhat can you comment on my questions here please.

index/scorch/optimize_knn.go
Thejas-bhat: Basically the thing is that if I put the check at this point, then I _must_ read the index (which is expensive) get the vector index and its size and then subject it to the app_herder check. It also comes with the drawback that the locks in the app_herder code would serialize the deployed goroutines over here which parallelizes the things. Another drawback is that now the app_herder check would be deployed for each segment.

So, in order to avoid too many redundant app_herder calls (its sufficient to invoke it once for every searcher creation since that's what we are trying to limit) and also to prevent the reading of the index from the buffer altogether in case of app_herder rejection, I pushed the check to one level above (i.e. at the VectorOptimize() function which is invoked once for every searcher).

In order to achieve this - the approach I have taken is that the very first thread which hits those segments and loads index now just records the size of the index as a meta data in the segment snapshot - and the reason why I used this struct is because one the context is established that we are storing something specific to a segment and the other reason is I followed the technique of how cachedDocs is populated and used for doc value retrieval (and i think that codepath is hit when the docvalue option is not enabled in the type mappings). The subsequent threads 
which corresponding to different queries which create a different OptimizeVR object altogether now consult the underlying segment snapshot which is now populated with the vector index size and then perform an app_herder check hence making sure that we are invoking this check only once for every searcher and not len(segments) times as earlier.

In the Finish() function we basically invoke the end callback as a defer logic.

index/scorch/snapshot_segment.go
```diff
@@ -282,3 +284,30 @@ func (c *cachedDocs) visitDoc(localDocNum uint64,
 
 	c.m.Unlock()
 }
+
+// the purpose of the cachedMeta is to simply allow the user of this type to record
```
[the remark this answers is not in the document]
Thejas-bhat: My bad, I meant it as this "struct". But I kept this generic enough for future use cases and if needed

index/scorch/optimize_knn.go
abhinavdangeti: Ok, looks like it was what I thought. Let me do one more review in some time.

search/util.go
```diff
@@ -140,3 +140,9 @@ const KnnPreSearchDataKey = "_knn_pre_search_data_key"
 const PreSearchKey = "_presearch_key"
 
 type ScoreExplCorrectionCallbackFunc func(queryMatch *DocumentMatch, knnMatch *DocumentMatch) (float64, *Explanation)
+
+type SearchSearcherStartCallbackFn func(size uint64) error
+type SearchSearcherEndCallbackFn func(size uint64) error
+
+const SearchSearcherStartCallbackKey = "_search_searcher_start_callback_key"
+const SearchSearcherEndCallbackKey = "_search_searcher_end_callback_key"
```
abhinavdangeti, line 148: It seems we're applying all of these only for kNN at the moment - do you see it being used for anything else in the future that we keep the name as generic as this?

Thejas-bhat: yeah so this was like a future proof time of a thing - if we want to do the same for text search paths, we can use these keys and invoke at appropriate paths over there. this is assuming that in the future we decide that its in our interest to invoke the app_herder checks at those paths after a fair bit of system and volume testing.

PR #1981, opened 2024-02-08
Minor optimizations

CascadingRadium:
- Avoid creating an unnecessary copy of a bitmap 
- Added capacities for some maps
- Fixed bug where a loop variable was used inside a closure

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -155,7 +155,11 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 		if root.segment[i].deleted == nil {
 			newss.deleted = delta
 		} else {
-			newss.deleted = roaring.Or(root.segment[i].deleted, delta)
+			if delta != nil && delta.IsEmpty() {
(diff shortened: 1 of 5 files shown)


abhinavdangeti (approved):

CascadingRadium:
To verify the validity of this fix here's the test i ran :
- The code path is only hit when you introduce a new segment with fully new doc IDs and no mutations with one of the existing segments having some deleted documents. Since we reuse the segment snapshot's old deleted bitmap instead of creating a copy of it.
- In my test i disabled merger and created 2 segments each with 2 documents and ran a query that gave 3 results. Now i deleted one doc in one of the segments and ran the same query which gave 2 docs.
- Now I created two more segments with full fresh documents -> no mutations, with 2 documents per segment. 
- Now i have 4 segments, A B C D, A has a document which is deleted. The same query returned same 2 results always.
- Now i do a force merge -> where I convert all 4 segments to 1 segment. 
- So by theory the old segment Base is not there anymore and hence the old snapshot is gone as well.
- The query still gives the same 2 results

PR #1983, opened 2024-02-11
Fix potential stack overflow in mapping.Validate

mjholub:
I was experiencing an issue where the call to Validate would cause infinite recursion, so I've added some checks to ensure that doesn't happen

### DIFF
--- a/mapping/index.go
+++ b/mapping/index.go
@@ -166,25 +166,34 @@ func NewIndexMapping() *IndexMappingImpl {
 // Validate will walk the entire structure ensuring the following
 // explicitly named and default analyzers can be built
 func (im *IndexMappingImpl) Validate() error {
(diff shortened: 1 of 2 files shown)


abhinavdangeti:
@mjholub thank you for the contribution, the team will review soon.

mapping/index.go
```diff
@@ -166,25 +166,34 @@ func NewIndexMapping() *IndexMappingImpl {
...
+		for _, docMapping := range mapping.TypeMapping {
+			err = docMapping.Validate(mapping.cache)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
 	}
-	return nil
+	return recursiveValidate(im)
```
abhinavdangeti, line 196: @mjholub I don't see where exactly you're recursively validating the mapping here.

abhinavdangeti:
Closing on account of inactivity.

PR #1994, opened 2024-03-11
MB-60914: Supporting field specific metrics

Likith101:
- These metrics are calculated based on analysis results initially and directly from the SegmentBase after a merge
 - This map will be stored in segment Introduction and segment snapshot
 - After merge, field level stats are collected from segment base after checking if it implements the FieldStatsReporter interface defined in scorch_segment_api
 - This map is also persisted in bolt to avoid recomputation under the key "stats" as part of the persisted part of the segment snapshot
 - FieldStats have been added to the StatsMap function of scorch. This will give an aggregated value of the stats with respect to the segments but specific to the field name and the stat using the stat name field:<field name>:<stat name>. For example, field:vector:num_vectors

### DIFF
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -30,7 +30,7 @@ type segmentIntroduction struct {
 	obsoletes map[uint64]*roaring.Bitmap
 	ids       []string
 	internal  map[string][]byte
-	stats     map[string]map[string]int
+	stats     *fieldStats
 
 	applied           chan error
 	persisted         chan error
@@ -427,7 +427,7 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 	if nextMerge.new != nil &&
 		nextMerge.new.Count() > newSegmentDeleted.GetCardinality() {
 
-		stats := initFieldStats()
+		stats := newFieldStats()
 		if fsr, ok := nextMerge.new.(segment.FieldStatsReporter); ok {
 			fsr.UpdateFieldStats(stats)
 		}
--- a/index/scorch/introducer.go
+++ b/index/scorch/introducer.go
@@ -30,6 +30,7 @@ type segmentIntroduction struct {
 	obsoletes map[uint64]*roaring.Bitmap
 	ids       []string
 	internal  map[string][]byte
+	stats     map[string]map[string]int
 
 	applied           chan error
 	persisted         chan error
@@ -146,6 +147,7 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 		newss := &SegmentSnapshot{
 			id:         root.segment[i].id,
 			segment:    root.segment[i].segment,
+			stats:      root.segment[i].stats,
 			cachedDocs: root.segment[i].cachedDocs,
 			cachedMeta: root.segment[i].cachedMeta,
 			creator:    root.segment[i].creator,
@@ -193,6 +195,7 @@ func (s *Scorch) introduceSegment(next *segmentIntroduction) error {
 		newSegmentSnapshot := &SegmentSnapshot{
 			id:         next.id,
 			segment:    next.data, // take ownership of next.data's ref-count
+			stats:      next.stats,
 			cachedDocs: &cachedDocs{cache: nil},
 			cachedMeta: &cachedMeta{meta: nil},
 			creator:    "introduceSegment",
@@ -281,6 +284,7 @@ func (s *Scorch) introducePersist(persist *persistIntroduction) {
 				id:         segmentSnapshot.id,
 				segment:    replacement,
 				deleted:    segmentSnapshot.deleted,
+				stats:      segmentSnapshot.stats,
 				cachedDocs: segmentSnapshot.cachedDocs,
 				cachedMeta: segmentSnapshot.cachedMeta,
 				creator:    "introducePersist",
@@ -381,6 +385,7 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 				id:         root.segment[i].id,
 				segment:    root.segment[i].segment,
 				deleted:    root.segment[i].deleted,
+				stats:      root.segment[i].stats,
 				cachedDocs: root.segment[i].cachedDocs,
 				cachedMeta: root.segment[i].cachedMeta,
 				creator:    root.segment[i].creator,
@@ -402,7 +407,6 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 			}
 		}
 	}
-
 	// before the newMerge introduction, need to clean the newly
 	// merged segment wrt the current root segments, hence
 	// applying the obsolete segment contents to newly merged segment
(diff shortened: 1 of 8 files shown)


index/scorch/scorch.go
```diff
@@ -438,11 +439,24 @@ func (s *Scorch) Batch(batch *index.Batch) (err error) {
...
+
+		stats["num_vectors"] = make(map[string]int)
+		for _, result := range analysisResults {
+			result.VisitFields(func(field index.Field) {
+				if string(field.EncodedFieldType()) == "v" {
+					if _, exists := stats["num_vectors"][field.Name()]; !exists {
+						stats["num_vectors"][field.Name()] = 0
+					}
+
+					stats["num_vectors"][field.Name()] += 1
```
Thejas-bhat, line 451: does this include the nested vectors as well? like i see you're increasing the value by 1 rather than say length/dims over here.

metonymic-smokey: I'd discussed this with @Likith101 earlier and IIRC he said yes.

index/scorch/scorch.go
```diff
@@ -641,6 +656,31 @@ func (s *Scorch) StatsMap() map[string]interface{} {
...
+
+					fieldStats[statName][fieldName] += val
+				}
+			}
+		}
+	}
+
+	for statName, stats := range fieldStats {
+		for fieldName, val := range stats {
+			m["field:"+fieldName+":"+statName] = val
```
metonymic-smokey, line 681: I understand that this is a per-field stat but I'm not sure adding the "field" prefix here is the best idea?

index/scorch/persister.go
```diff
@@ -873,6 +887,17 @@ func (s *Scorch) loadSegment(segmentBucket *bolt.Bucket) (*SegmentSnapshot, erro
 			rv.deleted = deletedBitmap
 		}
 	}
+	statBytes := segmentBucket.Get(boltStatsKey)
+	if statBytes != nil {
+		var stats map[string]map[string]int
```
abhinavdangeti, line 892: Using a struct would be a better idea keeping future in mind.

index/scorch/scorch.go
abhinavdangeti: If there's way we can avoid to individually increment the count for the metric and instead do it once - that _must_ be preferred.

index/scorch/scorch.go
```diff
@@ -462,14 +476,15 @@ func (s *Scorch) Batch(batch *index.Batch) (err error) {
 }
 
 func (s *Scorch) prepareSegment(newSegment segment.Segment, ids []string,
-	internalOps map[string][]byte, persistedCallback index.BatchCallback) error {
+	internalOps map[string][]byte, persistedCallback index.BatchCallback, stats map[string]map[string]int) error {
```
abhinavdangeti, line 479: Same comment as above - let's replace this with json-interpretable struct.

index/scorch/scorch.go
```diff
@@ -765,3 +805,12 @@ func parseToInteger(i interface{}) (int, error) {
 		return 0, fmt.Errorf("expects int or float64 value")
 	}
 }
+
+func initFieldStats() map[string]map[string]int {
+
+	statsMap := make(map[string]map[string]int)
```
abhinavdangeti, line 811: ditto and drop empty lines above and below.

abhinavdangeti (changes requested):

index/scorch/scorch.go
Likith101: Updated this so that the flow goes to the zap layer to populate the stats.

index/scorch/scorch.go
Likith101: I have discussed this with Abhi and we thought this was a good idea. Open to suggestions though

index/scorch/persister.go
Likith101: Changed to use a struct that implements an interface defined in scorch_segment_api. Zap layer uses the functions exported by this interface to add its stats.

index/scorch/scorch.go
```diff
@@ -765,3 +787,53 @@ func parseToInteger(i interface{}) (int, error) {
 		return 0, fmt.Errorf("expects int or float64 value")
 	}
 }
+
+type fieldStats struct {
+	statMap map[string]map[string]uint64
```
abhinavdangeti, line 792: Would this map access not need a lock, or are you controlling the access to it such that you'd never call read until it's completely updated?

index/scorch/scorch.go
```diff
@@ -765,3 +787,53 @@ func parseToInteger(i interface{}) (int, error) {
...
+}
+
+func (fs *fieldStats) AddStat(statName, fieldName string, value uint64) {
```
abhinavdangeti, line 795: API name suggestions in https://github.com/blevesearch/scorch_segment_api/pull/38

index/scorch/scorch.go
abhinavdangeti: Add commentary around what this map's content is.

index/scorch/scorch.go
```diff
@@ -765,3 +787,53 @@ func parseToInteger(i interface{}) (int, error) {
...
+	}
+
+	for statName, statMap := range statMap {
+		for fieldName, val := range statMap {
+			fs.AppendStat(statName, fieldName, val)
+		}
+	}
+}
+
+func (fs *fieldStats) AppendStat(statName, fieldName string, value uint64) {
```
abhinavdangeti, line 818: I see no need for this exported API in the interface.

abhinavdangeti (changes requested):

index/scorch/scorch.go
Likith101: The map shouldn't require a lock. Data is written into the map immediately after making the zap segment and remains unchanged throughout the segment's lifetime (As of now, adding additional stats may change this behaviour). There isn't a code path that accesses the map before it is written into or while it is being populated, so I believe this should be fine.

I will address the commentary though

index/scorch/scorch.go
Likith101: Removed it completely

index/scorch/scorch.go
```diff
@@ -765,3 +787,54 @@ func parseToInteger(i interface{}) (int, error) {
...
+// Add the data into the map after checking if the statname is valid
+func (fs *fieldStats) Store(statName, fieldName string, value uint64) {
+	if _, exists := segment.Stats[statName]; exists {
+		fs.statMap[statName][fieldName] = value
+	}
+}
+
+// Combine the given stats map with the existing map
+func (fs *fieldStats) Aggregate(stats segment.FieldStats) {
+
```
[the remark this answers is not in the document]
Likith101: Done

index/scorch/introducer.go
```diff
@@ -402,7 +407,6 @@ func (s *Scorch) introduceMerge(nextMerge *segmentMerge) {
 			}
 		}
 	}
-
```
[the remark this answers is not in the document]
Likith101: I have no clue where this new line is supposed to be but everything looks readable to me tbh.

abhinavdangeti (approved):

PR #1996, opened 2024-03-14
MB-60844: MapReduce for PreSearch

CascadingRadium:
- Instead of Merging all the PreSearch Results in one shot at the main coordinator node of an alias tree, merge them incrementally at each level of the tree instead, which would balance the reduction process across all the indexes in a distributed Bleve index, leading to a more even memory distribution.


CascadingRadium:
https://github.com/blevesearch/bleve/pull/1999

PR #1999, opened 2024-03-19
MB-60844: MapReduce for PreSearch

CascadingRadium:
- Instead of Merging all the PreSearch Results in one shot at the main coordinator node of an alias tree, merge them incrementally at each level of the tree instead, which would balance the reduction process across all the indexes in a distributed Bleve index, leading to a more even memory distribution.

### DIFF
--- a/index_alias_impl.go
+++ b/index_alias_impl.go
@@ -640,7 +640,9 @@ func preSearchDataSearch(ctx context.Context, req *SearchRequest, indexes ...Ind
 				// create a new presearch result processor
 				presearchResultProcessor = CreatePreSearchResultProcessor(req)
 			}
-			presearchResultProcessor.Add(asr.Result, asr.Name)
+			if presearchResultProcessor != nil {
+				presearchResultProcessor.Add(asr.Result, asr.Name)
+			}
 			if sr == nil {
 				// first result
 				sr = &SearchResult{
@@ -678,7 +680,7 @@ func preSearchDataSearch(ctx context.Context, req *SearchRequest, indexes ...Ind
 			sr.Status.Total++
 			sr.Status.Failed++
 		}
-	} else {
+	} else if presearchResultProcessor != nil {
 		presearchResultProcessor.Finalize(sr)
 	}
 	return sr, nil
--- a/index_alias_impl.go
+++ b/index_alias_impl.go
@@ -163,16 +163,16 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
 		return nil, ErrorAliasEmpty
 	}
 	if _, ok := ctx.Value(search.PreSearchKey).(bool); ok {
-		// since presearchKey is set, it means that the request
-		// is being executed as part of a presearch, which
+		// since preSearchKey is set, it means that the request
+		// is being executed as part of a preSearch, which
 		// indicates that this index alias is set as an Index
-		// in another alias, so we need to do a presearch search
+		// in another alias, so we need to do a preSearch search
 		// and NOT a real search
 		return preSearchDataSearch(ctx, req, i.indexes...)
 	}
 
 	// at this point we know we are doing a real search
-	// either after a presearch is done, or directly
+	// either after a preSearch is done, or directly
 	// on the alias
 
 	// check if request has preSearchData which would indicate that the
@@ -239,7 +239,7 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
 		preSearchDuration = time.Since(searchStart)
 	}
 
-	// check if search result was generated as part of presearch itself
+	// check if search result was generated as part of preSearch itself
 	if sr == nil {
 		sr, err = MultiSearch(ctx, req, preSearchData, i.indexes...)
 		if err != nil {
@@ -537,7 +537,7 @@ func preSearch(ctx context.Context, req *SearchRequest, indexes ...Index) (*Sear
 	return preSearchDataSearch(newCtx, dummyRequest, indexes...)
 }
 
-// if the request is satisfied by just the presearch result,
+// if the request is satisfied by just the preSearch result,
 // finalize the result and return it directly without
 // performing multi search
 func finalizeSearchResult(req *SearchRequest, preSearchResult *SearchResult) *SearchResult {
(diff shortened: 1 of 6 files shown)


index_alias_impl.go
```diff
@@ -866,3 +856,45 @@ func (f *indexAliasImplFieldDict) Close() error {
...
+	var knnProcessor *KnnPreSearchResultProcessor
+	if requestHasKNN(req) {
+		knnProcessor = NewKnnPreSearchResultProcessor(req)
+	}
+	return &PreSearchResultProcessor{
+		knnProcessor: knnProcessor,
+	}
+}
+
+func (p *PreSearchResultProcessor) AddPreSearchResult(sr *SearchResult, indexName string) {
```
abhinavdangeti, line 890: Rename this to Add as well.

index_alias_impl.go
```diff
@@ -866,3 +856,45 @@ func (f *indexAliasImplFieldDict) Close() error {
...
+	if p.knnProcessor != nil {
+		p.knnProcessor.Add(sr, indexName)
+	}
+}
+
+func (p *PreSearchResultProcessor) SetProcessedData(sr *SearchResult) {
```
abhinavdangeti, line 896: And this to Finalize

search/scorer/scorer_knn.go
```diff
@@ -71,7 +71,7 @@ func NewKNNQueryScorer(queryVector []float32, queryField string, queryBoost floa
 
 // Score used when the knnMatch.Score = 0 ->
 // the query and indexed vector are exactly the same.
-const maxKNNScore = math.MaxFloat64
+const maxKNNScore = math.MaxFloat32
```
abhinavdangeti, line 74: There a reason you changed this?

index_alias_impl.go
```diff
@@ -866,3 +856,45 @@ func (f *indexAliasImplFieldDict) Close() error {
...
+type KnnPreSearchResultProcessor struct {
+	Add      func(*SearchResult, string)
+	Finalize func(*SearchResult)
+}
+
+type PreSearchResultProcessor struct {
+	knnProcessor *KnnPreSearchResultProcessor
+}
+
+func CreatePreSearchResultProcessor(req *SearchRequest) *PreSearchResultProcessor {
```
abhinavdangeti, line 880: Why can't return type here simply be KnnPreSearchResultProcessor? I don't really see the need for 2 structs above. We can drop `PreSearchResultProcessor`.

abhinavdangeti (changes requested):

index_alias_impl.go
CascadingRadium: the aim is to allow `Presearch` to be generic construct that can be reused for other purposes. Lets say i have another logic for merging synonym results from multiple partitions, i would add the `SynonymPreSearchResultProcessor` as another struct in the PreSearchResult Processor, so any presearch result containing either `knn` or `synonym` data can be processed independently. By just keeping it as `KnnPreSearchResultProcessor` in `PreSearchDataSearch`, i would have to refactor it later on for synonym search.

search/scorer/scorer_knn.go
CascadingRadium: with grpc i encountered a marshalling error where returning math.MaxFloat64, it got resolved however by setting it to MaxFloat32. Looks like grpc doesnt work too well with marshalling and unmarshalling a MaxFloat64 value

index_alias_impl.go
CascadingRadium: ack

index_alias_impl.go
CascadingRadium: ack

search/scorer/scorer_knn.go
abhinavdangeti: Ack.

index_alias_impl.go
abhinavdangeti: I understand your rationale but we can adapt this or add a separate processor for synonyms later, we'll also want to refactor the name - `preSearchDataSearch` at that point as well, which is quite distasteful.

For now let's reduce the number of redirects, keep it simple.

CascadingRadium: have made it an interface and one struct implementing it now ... cannot use it directly as the index_alias_impl.go file is not in the vectors tag so it cannot read the knn specific code like the knn store ...

index_alias_impl.go
```diff
@@ -866,3 +856,41 @@ func (f *indexAliasImplFieldDict) Close() error {
...
+	Add(*SearchResult, string)
+	// Update the final search result with the finalized
+	// data from the processor
+	Finalize(*SearchResult)
+}
+
+type KnnPreSearchResultProcessor struct {
+	add      func(sr *SearchResult, indexName string)
+	finalize func(sr *SearchResult)
+}
```
abhinavdangeti, line 881: Again, this feels like we're unnecessarily overcomplicating things.

Drop the `add` and `finalize` function pointers and declare what you need within it's member methods, rather than this redirects below.

abhinavdangeti: Ok i take this back - it seems it's because of the the knnCollector in a file within the go build `vectors` flag.

index_alias_impl.go
```diff
@@ -866,3 +856,41 @@ func (f *indexAliasImplFieldDict) Close() error {
...
+
+func (k *KnnPreSearchResultProcessor) Add(sr *SearchResult, indexName string) {
+	k.add(sr, indexName)
```
abhinavdangeti, line 884: For the no vectors support use case -
```
if k.add != nil {
    k.add(..)
}
```

index_alias_impl.go
```diff
@@ -866,3 +856,41 @@ func (f *indexAliasImplFieldDict) Close() error {
...
+}
+
+func (k *KnnPreSearchResultProcessor) Finalize(sr *SearchResult) {
+	k.finalize(sr)
```
abhinavdangeti, line 888: For the no vectors support use case -
```
if k.finalize != nil {
    k.finalize(..)
}
```

CascadingRadium: done

index_alias_impl.go
CascadingRadium: done

abhinavdangeti (approved):

PR #2002, opened 2024-03-25
MB-61134: Consider segment fileSize (when with vectors) during merge plan computation

moshaad7:
Existing merge policy use a segment's total and live docs count as the measure for determining the merge tasks. 
A doc may contain multiple vector fields ( even chunked vectors ) , of varying vector dimensions, this usually means that index size will be greater than docs size by orders of magnitude.

Thus, docs count as merge policy measure, can easily lead us to formation of huge segments. That's a problem during merge and search time. 

Thus, this PR intends to utilize segment fileSize as an additional limiting check, to ensure that we don't end up creating huge segments.

example:
2M docs, each with a vector field of 2048 dimensions
size := 2000000 * 2048 * 4 bytes = 16GB


index/scorch/snapshot_segment.go
```diff
@@ -66,6 +66,53 @@ func (s *SegmentSnapshot) LiveSize() int64 {
...
+	var deletedDocsCount uint64
+	if s.deleted != nil {
+		deletedDocsCount = s.deleted.GetCardinality()
+	}
+
+	if deletedDocsCount == 0 {
+		return fullSize
+	}
+
+	// # Live Vector Size Estimation
```
Thejas-bhat, line 103: i have a doubt here, from what i understand over here the assumption is that each deleted document has a vector and the vector size is distributed equally (equal distribution of vectors over all docs in the segment). 

in a segment having 3 fields 1 text and 2 vectors - if the workload pattern (upserts) was such that the deleted documents in this segment corresponding to only the text field content and the two vector field contents' documents weren't invalidated (because the updates of vectors is not a frequent thing lets say), then we could perhaps end up making incorrect decision over here. 

instead, maybe what can be done is 
1. track the min and max docIDs across all the vector fields.
2. compare this range with the deleted bitmap's range (which also has min and max APIs).
3. make the decision of how much is the live vector size of this segment based on the overlap of these ranges. for eg, if there is no overlap in the two ranges then the live vector size of the segment wouldn't be affected.

index/scorch/mergeplan/merge_plan.go
```diff
@@ -37,6 +37,12 @@ type Segment interface {
 	// Size of the live data of the segment; i.e., FullSize() minus
 	// any logical deletions.
 	LiveSize() int64
+
+	HasVector() bool
+
+	FullVectorsByteSize() uint64
```
abhinavdangeti, line 43: Better naming, perhaps `IndexedVectorsBytes()`

Also looking at the code below, it seems LiveVectorsBytesSize() is the only needed exported API.

index/scorch/mergeplan/merge_plan.go
```diff
@@ -37,6 +37,12 @@ type Segment interface {
...
+
+	LiveVectorsByteSize() uint64
```
abhinavdangeti, line 45: LiveVectorsBytes*

index/scorch/mergeplan/merge_plan.go
```diff
@@ -76,6 +82,11 @@ type MergePlanOptions struct {
 	// planner’s predicted sizes.
 	MaxSegmentSize int64
 
+	// Max size of segment having vector content produced after merging.
+	// Will be respected based on estimated size of vector content, thus
+	// is a best effort soft limit.
+	MaxSegmentVectorByteSize uint64
```
abhinavdangeti, line 88: MaxSegmentVectorBytes*

index/scorch/snapshot_segment.go
```diff
@@ -66,6 +66,53 @@ func (s *SegmentSnapshot) LiveSize() int64 {
...
 
+func (s *SegmentSnapshot) HasVector() bool {
+	if fieldVecs, ok := s.stats.Fetch()["num_vecs"]; ok {
+		return len(fieldVecs) > 0
+	}
+
+	return false
+}
+
+func (s *SegmentSnapshot) FullVectorsByteSize() uint64 {
```
abhinavdangeti, line 77: No need for this exported API.

index/scorch/snapshot_segment.go
abhinavdangeti: >in a segment having 3 fields 1 text and 2 vectors - if the workload pattern (upserts) was such that the deleted documents in this segment corresponding to only the text field content and the two vector field contents' documents weren't invalidated (because the updates of vectors is not a frequent thing lets say), then we could perhaps end up making incorrect decision over here.

@Thejas-bhat raises a good point, am I misremembering in thinking - per our last conversation we decide not to do the live size but just use the actual size in making decisions.

abhinavdangeti (changes requested):

index/scorch/snapshot_segment.go
moshaad7: @abhinavdangeti  In our last conversation, we decided to not use liveSize , because the solution that I proposed needed scorch to know dimensions of vectors for each field. 

But, with this new approach, where we don't need to know dimensions. 
I think we should use LiveSize as an eligibility criteria for segments to be merged

moshaad7: @Thejas-bhat  @abhinavdangeti  

> then we could perhaps end up making incorrect decision over here.
Can you elaborate on this line 

My notion of a mutation is that , even if editing is done only in text field, we mark entire content of the doc in old segments as invalid, am I right here ? 

As part of a mutation, KV will send us the new doc ( same ID,  unedited + edited content both ) , right ?

index/scorch/mergeplan/merge_plan.go
moshaad7: Done

index/scorch/mergeplan/merge_plan.go
moshaad7: Done

index/scorch/snapshot_segment.go
moshaad7: Done, 
Thanks

index/scorch/mergeplan/merge_plan.go
moshaad7: Done, 
Thanks

index/scorch/snapshot_segment.go
abhinavdangeti: You're right Shaad - but I think, @Thejas-bhat 's point is that these docs that were updated or deleted never had any vector content in them to begin with.

abhinavdangeti:
@moshaad7 would you update the PR description above to explain everything the code here intends for.
The code in itself is looking mostly good - I'd love to hear if you've gotten this perf tested to see if we get the same gains as that updated merge policy got us a couple weeks ago.

abhinavdangeti:
@moshaad7 I think you talked about this but for the record - you'll be getting us details on how much did we see file size grow to say with 25-50million documents (with vector content) - yes?

abhinavdangeti (approved):

PR #2003, opened 2024-03-25
MB-60971: Avoiding unnecessary persister notifs from merger

Thejas-bhat:
- There are chances that the merger doesn't see any eligible segments to be merged in the current iteration which causes the tasks list to be empty. In this situation, merger which didn't update the root snapshot would notify the persister. 
- Now, if the persister was napping at this point of time, and assuming there were mutations coming into the system (so the root snapshot would be updated by the introducer) would lead to the persister to be awoken and start flushing out the in-memory segments to disk. 
- Perhaps a better behaviour would be to not let the merger notify the persister when there is no change in the root snapshot. This would help the persister to perform healthier in memory merging of the segments before persisting out to disk thereby helping in controlling the number of IO ops scorch would do. 

Some numbers on local testing (~4.18M dataset with lorem ipsum content)

With patch
```
  "num_bytes_used_ram": 186097560,
  "test_bucket:test_bucket._default.test:num_file_merge_ops": 7,
  "test_bucket:test_bucket._default.test:num_mutations_to_index": 0,
  "test_bucket:test_bucket._default.test:num_persister_nap_merger_break": 4,
  "test_bucket:test_bucket._default.test:num_persister_nap_pause_completed": 80,
```

Without patch
```
  "num_bytes_used_ram": 265234328,
  "test_bucket:test_bucket._default.test:num_file_merge_ops": 10,
  "test_bucket:test_bucket._default.test:num_mutations_to_index": 0,
  "test_bucket:test_bucket._default.test:num_persister_nap_merger_break": 69,
  "test_bucket:test_bucket._default.test:num_persister_nap_pause_completed": 45,
```

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -300,6 +300,7 @@ func (s *Scorch) planMergeAtSnapshot(ctx context.Context,
 	go cw.listen()
 
 	if len(resultMergePlan.Tasks) == 0 {
+		atomic.AddUint64(&s.stats.TotFileMergePlanZeroTasks, 1)
 		return errNoOpMerge
 	}
 
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -77,6 +77,21 @@ OUTER:
 				// lets get started
 				err := s.planMergeAtSnapshot(ctrlMsg.ctx, ctrlMsg.options,
 					ourSnapshot)
+
+				// if there were no segments merged because there were no tasks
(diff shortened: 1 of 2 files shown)


abhinavdangeti (dismissed):

abhinavdangeti (approved):

abhinavdangeti:
This commit has been reverted with https://github.com/blevesearch/bleve/pull/2005

PR #2009, opened 2024-04-03
Allow CopyTo to skip zap segment files that are not found

abhinavdangeti:
+ This could occur when a long running merge operation races with a potentially long running CopyTo operation (file transfer rebalance).


abhinavdangeti:
So this commit of simply allowing this error to slide could cause data loss - where a merger has dropped a segment and introduced a new one. What we should instead be looking at is to raise the ref count of all the segments prior to the operation so the merger does not drop any that are needed here.

cc @Thejas-bhat

PR #2019, opened 2024-04-22
Fix search cost unit test

CascadingRadium:
- Fix the BytesRead and BytesStored unit tests to pass when using the `vectors` tag. The following commands can be used to validate that all UTs now pass.
    - go test -tags=vectors -race ./...
    - go test -race ./...

### DIFF
--- a/index_test.go
+++ b/index_test.go
@@ -403,7 +403,7 @@ func TestBytesRead(t *testing.T) {
 	prevBytesRead, _ := stats["num_bytes_read_at_query_time"].(uint64)
 
 	expectedBytesRead := uint64(21639)
-	if BleveFeatureVectorSearch {
+	if supportForVectorSearch {
 		expectedBytesRead = 22049
 	}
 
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -60,6 +61,10 @@ func newConfiguration() *configuration {
 // Config contains library level configuration
 var Config *configuration
 
+const (
+	BleveVectorSearch = iota
```
abhinavdangeti, line 65: I understand what you're trying to do here, but could you look into an other approach that doesn't change the config construct. This is applicable for testing only and the scope of the change must be just that.

CascadingRadium: a global variable in bleve package ?
searchKnn.go -> const BleveVectorSearchFeature = true
searchNoKnn.go -> const BleveVectorSearchFeature = false

abhinavdangeti (changes requested):
My recommendation here is we simply drop these cost checks or the entire unit tests even and not over-engineer anything here.

Numbers such as `21639` or whatever don't mean anything in the tests's context.

config.go
abhinavdangeti: Ya sounds better.

CascadingRadium:
the test has been failing for many months now ...
mainly its because the SectionVectorIndex is locked under the 
vectors build tag so without the tag only 1 section is there

i am not sure if there is a better way to validate the cost 
other than a hard coded value...

abhinavdangeti (approved):

PR #2026, opened 2024-05-02
MB-61742: Interpreting vectors w/ spurious float32s from base64 encodings

abhinavdangeti:
+ Handle the situation when interpreted vectors have NaN or Inf when decoded from base64 encoded strings.
+ Also, in the vector_base64 path - it seems we were unnecessarily casting a []float32 into an interface{} to again be interpreted as a []float32 which we can avoid.


Likith101 (dismissed):

abhinavdangeti:
This has been reviewed earlier. A couple more cleanup fixes followed.

PR #2031, opened 2024-05-06
MB-61716: Corrected Field Mapping for Path Logic

Likith101:
The previous logic here only checked fields one level deep which means any nested fields were not being considered. I replaced these with existing functions to get back the correct field mapping.

### DIFF
--- a/mapping/index.go
+++ b/mapping/index.go
@@ -437,24 +437,16 @@ func (im *IndexMappingImpl) FieldAnalyzer(field string) string {
 func (im *IndexMappingImpl) FieldMappingForPath(path string) FieldMapping {
 	if im.TypeMapping != nil {
 		for _, v := range im.TypeMapping {
-			for field, property := range v.Properties {
-				for _, v1 := range property.Fields {
-					if field == path {
-						// Return field mapping if the name matches the path param.
-						return *v1
-					}
-				}
+			fm := v.fieldDescribedByPath(path)
+			if fm != nil {
+				return *fm
 			}
 		}
 	}
 
-	for field, property := range im.DefaultMapping.Properties {
-		for _, v1 := range property.Fields {
-			if field == path {
-				// Return field mapping if the name matches the path param.
-				return *v1
-			}
-		}
+	fm := im.DefaultMapping.fieldDescribedByPath(path)
+	if fm != nil {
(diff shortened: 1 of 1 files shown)


mapping/index.go
```diff
@@ -437,24 +437,16 @@ func (im *IndexMappingImpl) FieldAnalyzer(field string) string {
...
-			for field, property := range v.Properties {
-				for _, v1 := range property.Fields {
-					if field == path {
-						// Return field mapping if the name matches the path param.
-						return *v1
-					}
-				}
+			fm := v.fieldDescribedByPath(path)
+			if fm != nil {
+				return *fm
```
CascadingRadium, line 442: can't the pointer just be returned
this will create a copy of the whole struct for every query
the only place where FieldMappingForPath is used is below
to get the similarity metric

```
	fieldMapping := m.FieldMappingForPath(q.VectorField)
	similarityMetric := fieldMapping.Similarity
	if similarityMetric == "" {
		similarityMetric = index.DefaultSimilarityMetric
	}
```

maybe we can make `FieldMappingForPath` return a `*FieldMapping`
but that would make the `im` configurable (which is why i think the copy of the struct was returned in the first place, as `im` is supposed to be read only in the query path)

could you perhaps write a wrapper like 
`KNNSimilarityMetricForPath` instead, which just returns the 
Similarity metric (string) directly similar to `AnalyzerNameForPath`
thanks

CascadingRadium (changes requested):

mapping/index.go
Thejas-bhat: i think these points were brought up in the initial PR, but the conclusion over there was to make this read-only and keep the IndexMapping interface generic (by avoiding the KNNSimilarityMetricForPath API). 

to be honest, i'll let others weigh in whether it makes sense re-evaluate and have the suggested API. 

in my opinion i think its fine to have KNNSimilarityMetricForPath API because it would make the return value lighter - even if its specific to vector indexing (although that same reasoning can be applicable for AnalyzerNameForPath with respect to text indexing). 

other than this, the changes in the PR looks good to me 👍

abhinavdangeti (approved):

CascadingRadium (approved):

PR #2032, opened 2024-05-13
MB-61421: Prevent deletion of segments scheduled for copy

CascadingRadium:
- Use a modified index reader, CopyReader, to mark segments in the Scorch root for online copy/backup operations. This prevents their deletion by the asynchronous cleanup routine during the copy/backup process, thereby mitigating the race condition between the merger/persistor and the copy/backup routine.

### DIFF
--- a/index.go
+++ b/index.go
@@ -309,14 +309,6 @@ func NewBuilder(path string, mapping mapping.IndexMapping, config map[string]int
 	return newBuilder(path, mapping, config)
 }
 
-// IndexCopyable is an index which supports an online copy operation
-// of the index.
-type IndexCopyable interface {
-	// CopyTo creates a fully functional copy of the index at the
-	// specified destination directory implementation.
-	CopyTo(d index.Directory) error
-}
-
 // FileSystemDirectory is the default implementation for the
 // index.Directory interface.
 type FileSystemDirectory string
--- a/index.go
+++ b/index.go
@@ -309,6 +309,14 @@ func NewBuilder(path string, mapping mapping.IndexMapping, config map[string]int
 	return newBuilder(path, mapping, config)
 }
 
+// IndexCopyable is an index which supports an online copy operation
+// of the index.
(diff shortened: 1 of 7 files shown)


index.go
```diff
@@ -309,14 +309,6 @@ func NewBuilder(path string, mapping mapping.IndexMapping, config map[string]int
 	return newBuilder(path, mapping, config)
 }
 
-// IndexCopyable is an index which supports an online copy operation
-// of the index.
-type IndexCopyable interface {
-	// CopyTo creates a fully functional copy of the index at the
-	// specified destination directory implementation.
-	CopyTo(d index.Directory) error
-}
```
abhinavdangeti, line 318: Bring this and the supporting changes back to here - let's not use https://github.com/blevesearch/bleve_index_api/pull/48.

index/scorch/scorch.go
```diff
@@ -77,6 +77,12 @@ type Scorch struct {
 	segPlugin SegmentPlugin
 
 	spatialPlugin index.SpatialAnalyzerPlugin
+
+	// keeps track of segments scheduled for backup. Each segment's filename maps to
+	// the count of backup schedules. Segments with non-zero counts are protected
+	// from removal by the cleanup operation. Counts decrement upon successful backup,
+	// allowing removal of segments with zero or absent counts.
+	backupScheduled map[string]uint
```
abhinavdangeti, line 85: Naming is not consistent - if you're using the word "copy" for the interface, let's use the same here.

abhinavdangeti (changes requested):

index/scorch/scorch.go
abhinavdangeti: Would you add a comment above that this needs to be accessed within lock - perhaps best move it to line 58 too , indicating access needs rootLock.

abhinavdangeti (approved):

PR #2035, opened 2024-05-17
Make CopyTo use copyLock instead of the rootLock

CascadingRadium:
- Acquiring a write lock over the rootLock within copyTo might be detrimental since it can cause the CloseCopyReader() to be a blocking operation over the whole scorch environment. We can benefit by having a separate lock for CopyTo to use which would make the CopyReader and its close method mutually exclusive with only the cleanup routine and nothing else, as copyScheduled is only used in those areas.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -1132,6 +1132,7 @@ func (s *Scorch) removeOldZapFiles() error {
 	}
 
 	s.rootLock.RLock()
+	s.copyLock.RLock()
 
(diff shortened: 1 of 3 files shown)


abhinavdangeti:
@CascadingRadium Would you expedite a toybuild with this change and re-run those tests.

abhinavdangeti:
@CascadingRadium per our conversation let's not go ahead with this one, for it's unnecessary.

PR #2043, opened 2024-06-12
MB-62124: Dealing with bolt panics while reading

Likith101:
- Added checks for cases where bolt buckets are nil


abhinavdangeti:
@Likith101 I doubt these would address the situation - you mentioned you were able to reproduce the original issue (however rare), were you able to reproduce with instrumented code?

abhinavdangeti (approved):

PR #2049, opened 2024-07-08
MB-61889: support search with params

moshaad7:
- update knn search request syntax, to enable users to supply search-time parameters,
  using which they can control the latency v/s recall tradeoff.
- Supported parameters are
  + ivf_nprobe_pct
  + ivf_max_codes_pct
- These parameters will be applied to those segments which have faissIVFIndex

- In future, to support other faiss Index classes (like hnsw), the list of
  supported knn search params can be extended.


abhinavdangeti (changes requested):
Would you update the commit description documenting the new settings we're introducing with this commit please.

abhinavdangeti (approved):

PR #2050, opened 2024-07-09
Updating SingleSegmentMergePlanOptions with MaxSegmentFileSize

Thejas-bhat:
There was a small oversight introduced in force merge paths, where `MaxSegmentFileSize` option was missing in the `SingleSegmentMergePlanOptions`. This lead to none of the segments being eligible for a force merge. (https://github.com/blevesearch/bleve/blob/0672efa5cc7e79a9534b11517a9b8808bd8a277a/index/scorch/mergeplan/merge_plan.go#L189)

Consequently I was observing that the force merge was a no op when I triggered the API. Currently, this value has been set to 1TB.

### DIFF
--- a/index/scorch/mergeplan/merge_plan.go
+++ b/index/scorch/mergeplan/merge_plan.go
@@ -150,6 +150,7 @@ var DefaultMergePlanOptions = MergePlanOptions{
 var SingleSegmentMergePlanOptions = MergePlanOptions{
 	MaxSegmentsPerTier:   1,
 	MaxSegmentSize:       1 << 30,
+	MaxSegmentFileSize:   1 << 40,
 	TierGrowth:           1.0,
 	SegmentsPerMergeTask: 10,
 	FloorSegmentSize:     1 << 30,


metonymic-smokey (approved):

CascadingRadium (approved):

index/scorch/mergeplan/merge_plan.go
```diff
@@ -150,6 +150,7 @@ var DefaultMergePlanOptions = MergePlanOptions{
 var SingleSegmentMergePlanOptions = MergePlanOptions{
 	MaxSegmentsPerTier:   1,
 	MaxSegmentSize:       1 << 30,
+	MaxSegmentFileSize:   1 << 40,
```
abhinavdangeti, line 153: This won't cause an overflow with int64 would it?

Thejas-bhat: nope it doesn't

abhinavdangeti (approved):

PR #2051, opened 2024-07-09
 MB-62354: Support Cosine Similarity

CascadingRadium:
- Normalize the query vector if the Similarity Metric is `cosine`
- Add unit test cases for Cosine Similarity


abhinavdangeti (changes requested):
Bring the normalization on the index paths to here when we interpret the vector.

abhinavdangeti (approved):

PR #2052, opened 2024-07-16
Fix Size Calculation

CascadingRadium:
- Fix Size() calculation to account for dynamic entries of the struct, to ensure that accurate information is passed on to the memory throttler

### DIFF
--- a/document/field_boolean.go
+++ b/document/field_boolean.go
@@ -43,10 +43,15 @@ type BooleanField struct {
 }
 
 func (b *BooleanField) Size() int {
+	var freqSize int
+	if b.frequencies != nil {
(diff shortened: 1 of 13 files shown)


abhinavdangeti (changes requested):
@CascadingRadium would you throw in a detailed description and a much better commit header on what's exactly being addressed here and why.

CascadingRadium:
I noticed a few structs had incorrect Size values due to some of the its fields where not being accounted for...i fixed them to account such fields

abhinavdangeti (dismissed):

abhinavdangeti (approved):

PR #2054, opened 2024-07-18
MB-61985: Add a Scorch event for Index()

CascadingRadium:
- Add a scorch event for notifying listeners, if available, that a document is about to get added to the index, using the Index() API.
- This event can be useful for halting indexing before the Index() operation is carried out - since Index() converts an interface into a Document object using the IndexMapping's MapDocument, which consumes some extra memory.
- Shifted the EventKindBatchIntroductionStart event to before the documents were analyzed as document analysis consumes extra memory.

### DIFF
--- a/index/scorch/event.go
+++ b/index/scorch/event.go
@@ -67,3 +67,10 @@ var EventKindMergeTaskIntroduction = EventKind(8)
 // EventKindPreMergeCheck is fired before the merge begins to check if
 // the caller should proceed with the merge.
 var EventKindPreMergeCheck = EventKind(9)
+
+// EventKindIndexStart is fired when Index() is invoked which
+// creates a new Document object from an interface using the index mapping.
+var EventKindIndexStart = EventKind(10)
+
+// EventKindIndex is fired when Index() completes.
+var EventKindIndex = EventKind(11)
--- a/index/scorch/event.go
+++ b/index/scorch/event.go
@@ -71,6 +71,3 @@ var EventKindPreMergeCheck = EventKind(9)
 // EventKindIndexStart is fired when Index() is invoked which
 // creates a new Document object from an interface using the index mapping.
 var EventKindIndexStart = EventKind(10)
-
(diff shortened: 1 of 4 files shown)


index/scorch/event.go
```diff
@@ -67,3 +67,10 @@ var EventKindMergeTaskIntroduction = EventKind(8)
 // EventKindPreMergeCheck is fired before the merge begins to check if
 // the caller should proceed with the merge.
 var EventKindPreMergeCheck = EventKind(9)
+
+// EventKindIndexStart is fired when Index() is invoked which
+// creates a new Document object from an interface using the index mapping.
+var EventKindIndexStart = EventKind(10)
+
+// EventKindIndex is fired when Index() completes.
```
abhinavdangeti, line 75: Then this should be called OnEnd, no?

abhinavdangeti (changes requested):

index/scorch/event.go
CascadingRadium: this is following the same naming convention as  `EventKindBatchIntroductionStart` and `EventKindBatchIntroduction` (for end) 
in cbft there is no handler for either `EventKindBatchIntroduction` or `EventKindIndex` so they are just dummy events added for API completeness

CascadingRadium: actually never mind i have refactored the code now

abhinavdangeti (approved):

index/scorch/scorch.go
```diff
@@ -376,6 +375,8 @@ func (s *Scorch) Delete(id string) error {
 func (s *Scorch) Batch(batch *index.Batch) (err error) {
 	start := time.Now()
 
+	// notify handlers that we're about to introduce a segment
```
Thejas-bhat, line 378: nit: notify handlers that we're about to index a batch of data, sounds better perhaps

(apart from this looks good to me)

abhinavdangeti (approved):

PR #2055, opened 2024-07-23
MB-62400: Allow vector and numeric aliases for the same field

Likith101:
- Removed logic that prevents numeric field mappings to be executed after a vector field has been detected
 - Added checks to walk document on non vector field types preserving previous logic

### DIFF
--- a/mapping/document.go
+++ b/mapping/document.go
@@ -534,33 +534,33 @@ func (dm *DocumentMapping) processProperty(property interface{}, path []string, 
 			dm.walkDocument(property, path, indexes, context)
 		}
 	case reflect.Map, reflect.Slice:
-		var isPropertyVector bool
-		var isPropertyVectorInitialized bool
-		if subDocMapping != nil {
+		walkDocument := false
+		if subDocMapping != nil && len(subDocMapping.Fields) != 0 {
 			for _, fieldMapping := range subDocMapping.Fields {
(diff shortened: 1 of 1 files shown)


CascadingRadium (approved):

mapping/document.go
```diff
@@ -534,33 +534,33 @@ func (dm *DocumentMapping) processProperty(property interface{}, path []string,
...
 				case "vector":
-					processed := fieldMapping.processVector(property, pathString, path,
+					fieldMapping.processVector(property, pathString, path,
 						indexes, context)
-					if !isPropertyVectorInitialized {
-						isPropertyVector = processed
-						isPropertyVectorInitialized = true
-					} else {
-						isPropertyVector = isPropertyVector && processed
-					}
```
abhinavdangeti, line 550: @Likith101 I believe @CascadingRadium  added this and my memory on this is somewhat hazy - it seems we did this to avoid having to process a vector field multiple times, I like your code proposal here which is simple and easy to follow - just want to make sure we're not inadvertently dropping any optimization here.

Likith101: I collaborated with @CascadingRadium on this and we made sure to preserve any existing behaviour that was present before vectors and only remove the double processing for vector fields.

abhinavdangeti (approved):

PR #2060, opened 2024-08-02
MB-19243: Detect fuzziness automatically based on term length

CascadingRadium:
- The following queries can now automatically detect fuzziness based on the length of the terms:
  - Match Query
  - Fuzzy Query
  - Match-Phrase Query
  - Multi-Phrase Query
  - Phrase Query
- In these queries, each term (whether in a multi-term query like Match or Phrase, or in a single-term query like Fuzzy can have its own edit distance based on its length. The edit distance is calculated as follows:
  - For terms with 1 or 2 characters: edit distance = 0 (exact match)
  - For terms with 3, 4, or 5 characters: edit distance = 1 (fuzzy match)
  - For terms with more than 5 characters: edit distance = 2 (fuzzy match)
- This feature can be enabled using the `<query>.SetAutoFuzziness(<bool>)` API.
- Additionally, we've extended the functionality to query JSON parsing. You can specify fuzziness as either "auto" or a static value in the JSON query. Both formats are valid:
1. With auto fuzziness:
```
{
  "match" : "lorem",
  "field" : "bleve"
  "fuzziness" : "auto"
}
```
2. With static fuzziness:
```
{
  "match" : "lorem",
  "field" : "bleve"
  "fuzziness" : 2
}
```
When unmarshalled, the query will correctly apply the chosen fuzziness method.
- Fixed a bug where the code incorrectly returned an error message saying `fuzziness exceeds maximum` when using a fuzzy searcher with `fuzziness = 0`. Instead, a term searcher is now returned in this case.

### DIFF
--- a/search/searcher/search_fuzzy.go
+++ b/search/searcher/search_fuzzy.go
@@ -35,6 +35,19 @@ func NewFuzzySearcher(ctx context.Context, indexReader index.IndexReader, term s
 	if fuzziness < 0 {
 		return nil, fmt.Errorf("invalid fuzziness, negative")
 	}
+	if fuzziness == 0 {
+		// no fuzziness, just do a term search
+		// check if the call is made from a phrase searcher
+		// and if so, add the term to the fuzzy term matches
+		// since the fuzzy candidate terms are not collected
+		// for a term search, and the only candidate term is
+		// the term itself
+		fuzzyTermMatches := ctx.Value(search.FuzzyMatchPhraseKey)
+		if fuzzyTermMatches != nil {
+			fuzzyTermMatches.(map[string][]string)[term] = []string{term}
+		}
+		return NewTermSearcher(ctx, indexReader, term, field, boost, options)
+	}
 
 	// Note: we don't byte slice the term for a prefix because of runes.
 	prefixTerm := ""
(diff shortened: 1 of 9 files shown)


abhinavdangeti (dismissed):

CascadingRadium:
force pushed a rebase please review again 
thanks

search/searcher/search_fuzzy.go
```diff
@@ -71,6 +84,21 @@ func NewFuzzySearcher(ctx context.Context, indexReader index.IndexReader, term s
 		boost, options, true)
 }
 
+func getAutoFuzziness(term string) int {
+	termLength := len(term)
+	if termLength > 5 {
+		return 2
+	} else if termLength > 2 {
+		return 1
+	}
```
abhinavdangeti, line 93: Can you add some commentary around here on why we're fixing these at these numbers?
I think it'd be better to add global exported vars for these constants allowing library users to tweak them if necessary.

abhinavdangeti (changes requested):

search/searcher/search_fuzzy.go
CascadingRadium: done

abhinavdangeti (approved):

PR #2063, opened 2024-08-12
MB-62230 - Pre-Filtering Support for kNN

metonymic-smokey:
1. Extended KNN queries to support filter queries if specified.
2. Created a new collector, eligible collector, and a new document match handler for the filter queries as part of kNN.
3. Run the filter queries prior to each kNN query and pass the results when constructing the kNN queries.
4. Added unit tests for filtered kNN cases.
5. Updated the documentation with an example of a filtered kNN request.


search/collector/eligible.go
```diff
@@ -0,0 +1,152 @@
+package collector
```
metonymic-smokey, line 1: TODO: Add license header.

search/query/knn.go
```diff
@@ -35,7 +35,10 @@ type KNNQuery struct {
 	BoostVal    *Boost    `json:"boost,omitempty"`
 
 	// see KNNRequest.Params for description
-	Params json.RawMessage `json:"params"`
+	Params           json.RawMessage `json:"params"`
+	FilterQuery      Query           `json:"filter,omitempty"`
+	FilterResults    []index.IndexInternalID
+	RequireFiltering bool
```
abhinavdangeti, line 41: Don't quite understand the need for this.

search/query/knn.go
```diff
@@ -35,7 +35,10 @@ type KNNQuery struct {
 	BoostVal    *Boost    `json:"boost,omitempty"`
 
 	// see KNNRequest.Params for description
-	Params json.RawMessage `json:"params"`
+	Params           json.RawMessage `json:"params"`
+	FilterQuery      Query           `json:"filter,omitempty"`
+	FilterResults    []index.IndexInternalID
```
abhinavdangeti, line 40: Does this need exporting?

search_knn.go
```diff
@@ -86,16 +86,19 @@ type KNNRequest struct {
 	//  - ivf_max_codes_pct : float // percentage of total vectors to visit to do a query (across all clusters)
 	//
 	// Consult go-faiss to know all supported search params
-	Params json.RawMessage `json:"params"`
+	Params      json.RawMessage `json:"params"`
+	FilterQuery query.Query     `JSON:"filter,omitempty"`
```
abhinavdangeti, line 90: Introduce a linebreak above and add some commentary for this.

index/scorch/optimize_knn.go
```diff
@@ -89,9 +92,24 @@ func (o *OptimizeVR) Finish() error {
 					vectorIndexSize := vecIndex.Size()
 					origSeg.cachedMeta.updateMeta(field, vectorIndexSize)
 					for _, vr := range vrs {
+						eligibleVectorInternalIDs := vr.EligibleDocIDs()
+						// Only the eligible documents belonging to this segment
+						// will get filtered out.
+						// There is no way to determine which doc belongs to which segment
+						eligibleVectorInternalIDs.And(snapshotGlobalDocNums[index])
+
+						eligibleLocalDocNums := roaring.NewBitmap()
```
abhinavdangeti, line 101: Do you have any obvious reasons or see gains in making this a bitmap and then doing all the allocations down the line to convert it back into an acceptable format?

index/scorch/snapshot_index.go
```diff
@@ -471,16 +471,35 @@ func (is *IndexSnapshot) Document(id string) (rv index.Document, err error) {
 	return rvd, nil
 }
 
+func (is *IndexSnapshot) localDocNumFromGlobal(segmentIndex int, docNum uint64) uint64 {
```
abhinavdangeti, line 474: Can you add some commentary around what this does and for the pre-existing API `segmentIndexAndLocalDocNumFromGlobal` as well please.

index/scorch/snapshot_index_vr.go
```diff
@@ -50,7 +51,19 @@ type IndexSnapshotVectorReader struct {
 	currID        index.IndexInternalID
 	ctx           context.Context
 
-	searchParams json.RawMessage
+	searchParams     json.RawMessage
+	eligibleDocIDs   []index.IndexInternalID
+	requireFiltering bool
+}
+
+func (i *IndexSnapshotVectorReader) EligibleDocIDs() *roaring.Bitmap {
```
abhinavdangeti, line 59: Don't quite see the need for this to be exported.

index/scorch/snapshot_index_vr.go
```diff
@@ -50,7 +51,19 @@ type IndexSnapshotVectorReader struct {
...
+	res := roaring.NewBitmap()
```
abhinavdangeti, line 60: Need to re-think whether this really needs to be a roaring bitmap.

abhinavdangeti (changes requested):
Made one pass, will require a few more after changes to API signatures.

search/query/knn.go
metonymic-smokey: @abhinavdangeti this is used downstream to indicate whether a KNN query requires filtering(is a filtered query) or not(does not have a filter query).

search/query/knn.go
metonymic-smokey: I've changed the approach to not export these now.

index/scorch/snapshot_index_vr.go
metonymic-smokey: my bad, fixed it.

search/query/knn.go
metonymic-smokey: Removed it in my recent commits, since I've made some changes to the overall approach.

index/scorch/snapshot_index_vr.go
metonymic-smokey: Having this as a roaring bitmap simplifies the AND'ing with the global doc nums in each segment.

metonymic-smokey: done, pls take a look.

docs/vectors.md
```diff
@@ -80,6 +80,26 @@ if err != nil {
 fmt.Println(searchResult.Hits)
 ```
 
+## Querying with Filters
```
abhinavdangeti, line 83: Appreciate you adding this, would you throw in this at the end ..

```
## Querying with filters (v2.4.3+)
```

metonymic-smokey: good point, thanks!

Thejas-bhat (approved):

abhinavdangeti (changes requested):
@metonymic-smokey no force pushing here please, please checkout latest from branch `pre-filter` before pushing commits up. There're unit tests failing that needs looking into -
```
-- FAIL: TestSimilaritySearchPartitionedIndex (0.31s)
    search_knn_test.go:876: test case #0 failed: 0 hits returned
--- FAIL: TestSimilaritySearchMultipleSegments (0.27s)
    search_knn_test.go:876: test case #0 failed: 0 hits returned
```

search_knn.go
```diff
@@ -188,13 +188,15 @@ func (r *SearchRequest) UnmarshalJSON(input []byte) error {
...
-		knnReq.FilterQuery, err = query.ParseQuery(temp.KNN[i].FilterQuery)
+	r.KNN = make([]*KNNRequest, len(temp.KNN))
+	for i, knnReq := range temp.KNN {
+		r.KNN[i] = &KNNRequest{}
+		r.KNN[i].Field = temp.KNN[i].Field
+		r.KNN[i].Vector = temp.KNN[i].Vector
+		r.KNN[i].VectorBase64 = temp.KNN[i].VectorBase64
+		r.KNN[i].K = temp.KNN[i].K
+		r.KNN[i].Boost = temp.KNN[i].Boost
+		r.KNN[i].FilterQuery, err = query.ParseQuery(knnReq.FilterQuery)
```
abhinavdangeti: Maybe parse this if and only if knnReq.FilterQuery is provided, otherwise I believe ParseQuery would interpret this as a `match_none` - which is not ideal, yes?

abhinavdangeti (changes requested):

search_knn.go
abhinavdangeti: Ah you've already addressed this in a subsequent commit I see.

abhinavdangeti (approved):

PR #2065, opened 2024-08-28
Optimize boolean query handling with early exit on nil conditions

CascadingRadium:
- Early exit in boolean query if must/should/mustNot query list is nil.


abhinavdangeti (changes requested):
Commit headers _must_ indicate what the code pull request does/intends-to-achieve.

abhinavdangeti (approved):

PR #2066, opened 2024-08-29
Introducing config for zap layer

Thejas-bhat:


index/scorch/merge.go
```diff
@@ -372,8 +372,8 @@ func (s *Scorch) planMergeAtSnapshot(ctx context.Context,
 
 			atomic.AddUint64(&s.stats.TotFileMergeZapBeg, 1)
 			prevBytesReadTotal := cumulateBytesRead(segmentsToMerge)
-			newDocNums, _, err := s.segPlugin.Merge(segmentsToMerge, docsToDrop, path,
-				cw.cancelCh, s)
+			newDocNums, _, err := s.segPlugin.MergeEx(segmentsToMerge, docsToDrop, path,
```
abhinavdangeti: Naming will need updates in parlance with comments on https://github.com/blevesearch/zapx/pull/256.

CascadingRadium (dismissed):

index/scorch/segment_plugin.go
```diff
@@ -46,10 +46,14 @@ type SegmentPlugin interface {
 	// New takes a set of Documents and turns them into a new Segment
 	New(results []index.Document) (segment.Segment, uint64, error)
 
+	NewUsing(results []index.Document, config map[string]interface{}) (segment.Segment, uint64, error)
```
capemox, line 49: Rather than define new functions for the interface, can create a new interface like `SegmentPluginWithConfig` with the new methods. You don't need to change the existing zap versions, and you can use comma-ok idioms to fallback on the existing implementation.

Thejas-bhat: I think having a new interface with just the *Using APIs would probably force us to maintain a separate plugin and possibly a new registry for the same. Also, the config is supposed to be passed around and useful for the existing paths in scorch where the zap APIs are called - so we could've just updated existing APIs that were there but just to keep the backward compatibility for users of the zap package independently I don't suppose we could do that. The segment plugin definitely needs to be refactor'd though since its messy, and we can talk about it as part of a separate github issue or so?

PR #2071, opened 2024-09-11
MB-62230: Added params in KNN request when parsing a query

metonymic-smokey:
This PR adds support for search-time params in the query parser and custom JSON unmarshaling logic.

### DIFF
--- a/search_knn.go
+++ b/search_knn.go
@@ -128,6 +128,7 @@ func (r *SearchRequest) UnmarshalJSON(input []byte) error {
 		VectorBase64 string          `json:"vector_base64"`
 		K            int64           `json:"k"`
 		Boost        *query.Boost    `json:"boost,omitempty"`
+		Params       json.RawMessage `json:"params"`
 		FilterQuery  json.RawMessage `JSON:"filter,omitempty"`
 	}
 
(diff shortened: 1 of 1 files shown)


search_knn.go
```diff
@@ -387,7 +389,7 @@ func (i *indexImpl) runKnnCollector(ctx context.Context, req *SearchRequest, rea
 		}
 
 		// Applies to all supported types of queries.
-		filterSearcher, _ := filterQ.Searcher(ctx, reader, i.m, search.SearcherOptions{
+		filterSearcher, _ := filterQ.(query.Query).Searcher(ctx, reader, i.m, search.SearcherOptions{
```
abhinavdangeti, line 392: If this type assertion were to fail, wouldn't we be dereferencing a `nil` pointer.

search_knn.go
```diff
@@ -196,6 +197,7 @@ func (r *SearchRequest) UnmarshalJSON(input []byte) error {
 		r.KNN[i].VectorBase64 = temp.KNN[i].VectorBase64
 		r.KNN[i].K = temp.KNN[i].K
 		r.KNN[i].Boost = temp.KNN[i].Boost
+		r.KNN[i].Params = temp.KNN[i].Params
```
abhinavdangeti, line 200: Was this a regression from the previous change?

abhinavdangeti (changes requested):

search_knn.go
metonymic-smokey: yes, was _just_ pushing a commit up for this - gimme a min

search_knn.go
metonymic-smokey: yes, fixed in a recent commit.

abhinavdangeti (approved):

PR #2076, opened 2024-09-19
MB-62230 - Avoiding unnecessary computations for un-filtered kNN.

metonymic-smokey:
Currently, there are a bunch of additional computations done on every kNN query that are necessary only for pre-filtered kNN. 
This PR adds checks to perform these only when at least one kNN query, out of one or more in the search request, is a filtered kNN query.

Also added some commentary.


abhinavdangeti:
@metonymic-smokey Let's also add some unit test coverage for pre filtering here, seperately.

abhinavdangeti (approved):

PR #2085, opened 2024-10-09
MB-63742 - Fix populating the filter hits map

metonymic-smokey:
Populates the initialised filter hits map using an index.


abhinavdangeti:
Nice fix @metonymic-smokey :)

PR #2087, opened 2024-10-09
MB-63736: Yet another shortcoming from an earlier commit

abhinavdangeti:
Regression introduced with https://github.com/blevesearch/bleve/pull/2084


abhinavdangeti:
Getting this commit in immediately to address a regression introduced by yours truly.

PR #2093, opened 2024-11-02
feat: remove sloppy math for sin, cos and asin

Gusted:
Remove the LUT implementation for `sin`, `cos` and `asin`. For `sin` and `cos` they are not faster than the hardware provided `sin` and `cos` instructions on AMD64 and ARM64. For `asin` LUT does seem to be a bit faster but hardware and LUT implementation are both in the same order of magnitude for performance and `asin` LUT unconditionally takes 1.5MiB of memory which seems a bit excessive.

If the usage of `asin` is really performance sensitive (I'm not that familiar with the geo library) it might be better to lazy-load this LUT to avoid users of the bleve library that don't have a use of the geo library to inherit this 1.5MiB of memory.


abhinavdangeti:
Thank you for this contribution @Gusted .

abhinavdangeti (approved):

PR #2097, opened 2024-11-08
feat: expose all available terms

scottlepp:
The lookup contains all possible facets.  We would like to expose this to use this as a way to expose all available facets for showing filters.

You can see an example [here of what we are doing](https://github.com/grafana/grafana/pull/96130).  We used [reflection here](https://github.com/grafana/grafana/blob/166c0b8d8508ea70b77e98dc240f510af5287f37/pkg/storage/unified/resource/index.go#L521) for now, but would like to avoid that.

### DIFF
--- a/search/facets_builder.go
+++ b/search/facets_builder.go
@@ -133,6 +133,17 @@ func (tf *TermFacets) Terms() []*TermFacet {
 	return tf.termFacets
 }
 
+func (tf *TermFacets) AvailableTerms() []*TermFacet {
+	if tf == nil {
+		return []*TermFacet{}
+	}
+	terms := make([]*TermFacet, 0, len(tf.termLookup))
+	for _, term := range tf.termLookup {
+		terms = append(terms, term)
+	}
+	return terms
+}
+
 func (tf *TermFacets) TrimToTopN(n int) {
 	tf.termFacets = tf.termFacets[:n]
 }
--- a/search/facets_builder_test.go
+++ b/search/facets_builder_test.go
@@ -422,3 +422,28 @@ func TestDateFacetResultsMerge(t *testing.T) {
 		t.Errorf("expected %#v, got %#v", expectedFrs, frs1)
 	}
 }
+
+func TestAvailableTerms(t *testing.T) {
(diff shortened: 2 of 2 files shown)


abhinavdangeti:
Thanks for this contribution @scottlepp , the team will review soon.

search/facets_builder.go
```diff
@@ -133,6 +133,17 @@ func (tf *TermFacets) Terms() []*TermFacet {
 	return tf.termFacets
 }
 
+func (tf *TermFacets) AvailableTerms() []*TermFacet {
```
abhinavdangeti, line 136: Would you add a supporting unit test for this please.

scottlepp: hi @abhinavdangeti added a test

tylerkovacs: @scottlepp Could you accomplish what you need with the existing `Terms()` func on `TermFacets`?

scottlepp: The `Terms()` func returns `termFacets`, which is empty.  However you can see the `termsLookup` contains the terms.

image

tylerkovacs: Just remembered that there are some other methods that change `termFacets`. `Fixup()` and `TrimToTopN()` can both remove (or entirely truncate) the `termFacets` slice.

I'm not that familiar with how these functions are called but, if I had to guess, it might only happen when the size of the facet request is very small (or zero).

```
bleve.NewFacetRequest("foo", 0)
```

As an aside, there doesn't appear to be a validation on the size of the facet request so perhaps there's a good reason to allow zero (I can't think of one, but again I'm only passingly familiar with the bleve code base).

abhinavdangeti:
@CascadingRadium would you take a look into the situation here and see if this is necessary.

search/facets_builder.go
scottlepp: @tylerkovacs you're right.  the limit was defaulting to 0.

abhinavdangeti:
Thanks @tylerkovacs for helping out on this one!

PR #2098, opened 2024-11-11
MB-62230 - Pre-filtering Optimisation

metonymic-smokey:
This PR - 
1. Avoids creating document matches for the pre-filter phase, when IDs suffice. 
2. Re-uses document matches by adding them to the doc match pool after each hit.

### DIFF
--- a/search/collector/eligible.go
+++ b/search/collector/eligible.go
@@ -23,28 +23,40 @@ import (
 	index "github.com/blevesearch/bleve_index_api"
 )
 
+type eligibleStore struct {
+	ids []index.IndexInternalID
+}
+
+func (s *eligibleStore) AddID(doc *search.DocumentMatch) *search.DocumentMatch {
+	copyOfID := make([]byte, len(doc.IndexInternalID))
+	copy(copyOfID, doc.IndexInternalID)
+	s.ids = append(s.ids, copyOfID)
+	return doc
+}
+
 type EligibleCollector struct {
 	size    int
 	total   uint64
 	took    time.Duration
 	results search.DocumentMatchCollection
 
-	store collectorStore
+	store *eligibleStore
 }
 
 func NewEligibleCollector(size int) *EligibleCollector {
 	return newEligibleCollector(size)
 }
 
+func getEligibleCollectorStore() *eligibleStore {
+	return &eligibleStore{
+		ids: make([]index.IndexInternalID, 0),
+	}
+}
+
 func newEligibleCollector(size int) *EligibleCollector {
 	// No sort order & skip always 0 since this is only to filter eligible docs.
 	ec := &EligibleCollector{size: size}
-
-	// comparator is a dummy here
-	ec.store = getOptimalCollectorStore(size, 0, func(i, j *search.DocumentMatch) int {
-		return 0
-	})
-
+	ec.store = getEligibleCollectorStore()
 	return ec
 }
 
@@ -56,7 +68,8 @@ func makeEligibleDocumentMatchHandler(ctx *search.SearchContext) (search.Documen
 			}
 
 			// No elements removed from the store here.
-			_ = ec.store.Add(d)
+			doc := ec.store.AddID(d)
+			ctx.DocumentMatchPool.Put(doc)
 			return nil
 		}, nil
 	}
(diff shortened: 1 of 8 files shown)


CascadingRadium (approved):

search/collector/eligible.go
```diff
@@ -23,28 +23,40 @@ import (
...
+}
+
 type EligibleCollector struct {
 	size    int
 	total   uint64
 	took    time.Duration
 	results search.DocumentMatchCollection
 
-	store collectorStore
+	store *eligibleStore
```
abhinavdangeti, line 43: This need not be a pointer looking at this contents.

search/collector/eligible.go
```diff
@@ -23,28 +23,40 @@ import (
...
 )
 
+type eligibleStore struct {
+	ids []index.IndexInternalID
+}
+
+func (s *eligibleStore) AddDocumentMatch(doc *search.DocumentMatch) *search.DocumentMatch {
+	copyOfID := make([]byte, len(doc.IndexInternalID))
+	copy(copyOfID, doc.IndexInternalID)
+	s.ids = append(s.ids, copyOfID)
```
abhinavdangeti, line 33: Is the copy above necessary? Wouldn't append allocate and copy implicitly?

search/collector/eligible.go
```diff
@@ -23,28 +23,40 @@ import (
 	index "github.com/blevesearch/bleve_index_api"
 )
 
+type eligibleStore struct {
+	ids []index.IndexInternalID
+}
+
+func (s *eligibleStore) AddDocumentMatch(doc *search.DocumentMatch) *search.DocumentMatch {
```
abhinavdangeti, line 30: Does this need to be exported?
Also the original `AddID` sound more accurate.

search/collector/eligible.go
```diff
@@ -23,28 +23,40 @@ import (
...
 
-	store collectorStore
+	store *eligibleStore
 }
 
 func NewEligibleCollector(size int) *EligibleCollector {
 	return newEligibleCollector(size)
 }
 
+func getEligibleCollectorStore(size int) *eligibleStore {
```
abhinavdangeti, line 50: Bad naming, and I don't really see the need for this method, use it's contents directly in line 59 below.

search/collector/eligible.go
abhinavdangeti: Also, this feels over done with the type routing, simply rename to `ids` (in-line with the method you're using to obtain this list) and make it of type `[]index.IndexInternalID`

search/collector/eligible.go
```diff
@@ -23,28 +23,40 @@ import (
 	index "github.com/blevesearch/bleve_index_api"
 )
 
+type eligibleStore struct {
```
abhinavdangeti, line 26: Don't see the need for this type really.

search/collector/eligible.go
```diff
@@ -23,28 +23,40 @@ import (
...
 func newEligibleCollector(size int) *EligibleCollector {
 	// No sort order & skip always 0 since this is only to filter eligible docs.
 	ec := &EligibleCollector{size: size}
-
-	// comparator is a dummy here
-	ec.store = getOptimalCollectorStore(size, 0, func(i, j *search.DocumentMatch) int {
-		return 0
-	})
-
+	ec.store = getEligibleCollectorStore(size)
```
abhinavdangeti, line 59:
```
    ec := &EligibleCollector{
        size: size,
        ids: make([]index.IndexInternalID, size),
    }
```

abhinavdangeti (changes requested):
Some more comments (all around naming and readability) that override some earlier ones.

CascadingRadium (approved):

search/collector/eligible.go
```diff
@@ -55,8 +51,10 @@ func makeEligibleDocumentMatchHandler(ctx *search.SearchContext) (search.Documen
 				return nil
 			}
 
-			// No elements removed from the store here.
-			_ = ec.store.Add(d)
+			copyOfID := make([]byte, len(d.IndexInternalID))
+			copy(copyOfID, d.IndexInternalID)
```
abhinavdangeti, line 55: My previous question on this copy operation is still open - is this necessary?

search/collector/eligible.go
metonymic-smokey: slices are reference types. This means they need to be copied explicitly to avoid changing the underlying slice.

search/collector/eligible.go
```diff
@@ -55,8 +51,10 @@ func makeEligibleDocumentMatchHandler(ctx *search.SearchContext) (search.Documen
...
+			ec.ids = append(ec.ids, copyOfID)
+			ctx.DocumentMatchPool.Put(d)
```
abhinavdangeti, line 57: @metonymic-smokey real quick why is this `DocumentMatchPool.Put(..)` necessary?

metonymic-smokey: This is to avoid hitting into the DocMatchPoolTooSmall issue - by putting it back in the pool, the pool will always have a doc match and won't have to allocate more doc matches.

abhinavdangeti: Cheers.

abhinavdangeti (approved):

PR #2103, opened 2024-11-21
Move PartialMatch field to Explanation struct and update related logic

CascadingRadium:
- Previously, the `PartialMatch` field was returned for every hit, but this caused confusion in complex queries involving disjunctions and match queries. As a result, we moved `PartialMatch` to the score explanation, where each subquery's explanation will include its own `PartialMatch`. This field is set only if the query uses the DisjunctionSearcher or scorer.

### DIFF
--- a/search/explanation.go
+++ b/search/explanation.go
@@ -30,9 +30,10 @@ func init() {
 }
 
 type Explanation struct {
-	Value    float64        `json:"value"`
-	Message  string         `json:"message"`
-	Children []*Explanation `json:"children,omitempty"`
+	Value        float64        `json:"value"`
+	Message      string         `json:"message"`
+	PartialMatch bool           `json:"partial_match,omitempty"`
+	Children     []*Explanation `json:"children,omitempty"`
 }
 
 func (expl *Explanation) String() string {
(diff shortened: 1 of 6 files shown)


abhinavdangeti:
So just so we're clear here - partial_match will be set if an explicit disjunction were used and all its components did not match yes?
And also separately, if a match query resolves to a disjunction - and if some of the components did not match, we'll now be setting partial_match - correct?

CascadingRadium:
- A match query will be a disjunction query over all the terms, here if the document has all 
  the terms then partial_match will be set to true else it will be set to false.
- For all other queries partial_match will never show up in the search result as its `omitempty` and will be simply excluded from the search result.

abhinavdangeti:
Ok, then I disagree with your proposal here - we should be returning a partial_match when a disjunction's component doesn't match as well.

CascadingRadium:
like in a disjunction query ?

abhinavdangeti:
> like when the top-most query is a disjunction query ?

Ya.

abhinavdangeti (changes requested):
Per last conversation.

CascadingRadium:
Actually, this approach has the limitation with disjunction + match queries
like so 

```
{
  "disjuncts":[
    {
      "match":"termA termB",
    },
    {
      "match":"SONY PS5",
      "prefix_length":0,
      "fuzziness":1
    },
  ]
}
```

where a single PartialMatch in the searchResult would lead to confusion regarding which component was partially matched (matchQuery1 / matchQuery2 / outerDisjuncts)

i will propose a new solution where the partialMatch is reflected in the score explanation, where each subQuery's expl will have the partialMatch attribute

abhinavdangeti (approved):

PR #2106, opened 2024-11-26
MB-57888: Support for downtime mitigation upon updates to index mapping

Likith101:
- Added new apis for index update
- Added logic to determine whether index mappings can be updated and what specifically changed between mappings
- Added logic to store and retrieve said information within bolt 
- Added checks to prevent deleted data from being referenced within queries till the actual data on the segment is removed during the merge process

Related:
- https://github.com/blevesearch/bleve_index_api/pull/64
- https://github.com/blevesearch/scorch_segment_api/pull/61
- https://github.com/blevesearch/zapx/pull/318


index_update.go
```diff
@@ -0,0 +1,625 @@
...
+			}
+			if upd.StoreDynamic {
+				return nil, fmt.Errorf("Mapping cannot be removed when store dynamic is true")
+			}
+			if upd.DocValuesDynamic {
+				return nil, fmt.Errorf("Mapping cannot be removed when docvalues dynamic is true")
+			}
+		}
+	}
+	return fieldInfo, nil
```
Thejas-bhat, line 123: @Likith101, let me know if the following understanding is correct.

To me it looks like we're doing multiple iterations over the two pairs of type mappings to first validate, then accumulate all the fields' information in both the type mappings as part of fieldInfo and then remove the ones which are identical.

This approach seems to have redundant work in it imo, like isn't it possible to do all this in a single iteration with a custom callback that defines the checking of the field mapping's info between the orig and updated ones and populate a fieldInfo map that's maintained in this function? 

Please let me know if the suggested thought is valid or not, thanks

Likith101: The multiple iterations of processing the two mappings were necessary. At each step, we eliminate a lot of possible edge cases and handling them all at once in the end is a huge task.

For example, the very first iteration, we simply check for mappings or fields that are present in the updated mapping which are not present in the original. This was done so that any future functions do not have to worry about the edge case where a mapping in updated does not have a corresponding one in the original.

Another case is where after adding every single field info, we validate with all of the existing ones to check if there are any corner cases like same path or same field alias while having different deleted information being hit. While this seems redundant, doing it all at the end was near impossible.

Code complexity wise, I don't believe we are doing too much redundant work overall. We do two passes of each mapping and then if there are n fields in updated mapping, we do n<sup>2</sup> checks with a map along with n proper field comparisons.

index/scorch/optimize_knn.go
```diff
@@ -85,8 +85,16 @@ func (o *OptimizeVR) Finish() error {
 					wg.Done()
 				}()
 				for field, vrs := range o.vrs {
-					vecIndex, err := segment.InterpretVectorIndex(field,
-						o.requiresFiltering, origSeg.deleted)
+					var vecIndex segment_api.VectorIndex
+					var err error
+					// Noop if the vector field or its index data is supposed to be deleted
+					if info, ok := o.snapshot.updatedFields[field]; ok && (info.Deleted || info.Index) {
+						vecIndex, err = segment.InterpretVectorIndex("",
```
Thejas-bhat, line 92: we can just short circuit the downstream operations by returning here right? like we know that an empty string field name is never going to exist in the index files and hence a backing faiss index won't be formed. so let's not perform calls to zap and keep it simple like

```
if info, ok := o.snapshot.updatedFields[field]; ok && (info.Deleted || info.Index) {
     return saying field doesn't exist
}
vecIndex, err := segment.InterpretVectorIndex(field,
		o.requiresFiltering, origSeg.deleted)
```

so we can just do a no-op or return an error saying "field doesn't exist"

index_update.go
Thejas-bhat: hmm i see. cool let me try understanding this code a little better as well, thanks

index_update.go
```diff
@@ -0,0 +1,625 @@
...
+		return fmt.Errorf(("docvalues dynamic cannot be changed"))
+	}
+
+	return nil
+}
+
+// Ensures updated document mapping does not contain new
+// field mappings or document mappings
+func checkUpdatedMapping(ori, upd *mapping.DocumentMapping) error {
+
```
Thejas-bhat, line 159: small nit: could you please remove these new lines (i see a bunch of them) just so that we are consistent with the other functions' structure

CascadingRadium:
hi @Likith101 can you please rebase over master and fix conflicts

index_impl.go
```diff
@@ -169,6 +170,21 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
 	storeConfig["error_if_exists"] = false
 	for rck, rcv := range runtimeConfig {
 		storeConfig[rck] = rcv
+		if rck == "updated_mapping" {
```
CascadingRadium, line 173: `updated_mapping` is a cross-reference term; it is set within cbft.
set a bleve global variable 
`const UpdatedMappingKey = "updated_mapping"`
and reference `bleve.UpdatedMappingKey` in cbft

CascadingRadium (commented):
first pass will continue

index_impl.go
```diff
@@ -169,6 +170,21 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
...
+			if val, ok := rcv.(string); ok {
```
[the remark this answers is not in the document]
Likith101: This seems a bit redundant. I believe the extra indent makes it easier to understand.

index_impl.go
```diff
@@ -129,7 +127,7 @@ func newIndexUsing(path string, mapping mapping.IndexMapping, indexType string,
 	if err != nil {
 		return nil, err
 	}
-	err = rv.i.SetInternal(mappingInternalKey, mappingBytes)
+	err = rv.i.SetInternal(scorch.MappingInternalKey, mappingBytes)
```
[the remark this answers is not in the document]
Likith101: Could you point out where in upside_down we are using this key? I am unable to find it.

index_update.go
```diff
@@ -0,0 +1,616 @@
...
+		} else {
+			return fmt.Errorf("analyser %s newly added to an existing field", name)
+		}
+	}
+
+	return nil
+}
+
+// Compares all date time parsers currently in use
+// Date time parsers in custom analysis but not in use are not compared
```
CascadingRadium, line 327: there is no comparison for `customDateTimeParsers`? refactor to how it is above for analyzers. Here you're checking for equality for built-in datetime parsers also which is not required.

```
oriDateTimeParsers := make(map[string]interface{})
updDateTimeParsers := make(map[string]interface{})

extractDateTimeParsers := func(paths map[string]PathInfo, customParsers map[string]interface{}, parsers map[string]interface{}, defaultParser string) {
	for _, info := range paths {
		for _, fInfo := range info.fieldMapInfo {
			if fInfo.fieldMapping.Type == "datetime" {
				dateTimeParserName := fInfo.fieldMapping.DateFormat
				if dateTimeParserName == "" {
					dateTimeParserName = defaultParser
				}
				fInfo.datetimeParser = dateTimeParserName
				if val, ok := customParsers[dateTimeParserName]; ok {
					parsers[dateTimeParserName] = val
				}
			}
		}
	}
}

extractDateTimeParsers(oriPaths, ori.CustomAnalysis.DateTimeParsers, oriDateTimeParsers, ori.DefaultDateTimeParser)
extractDateTimeParsers(updPaths, upd.CustomAnalysis.DateTimeParsers, updDateTimeParsers, upd.DefaultDateTimeParser)

```

index_update.go
```diff
@@ -0,0 +1,616 @@
...
+			return fmt.Errorf("datetime parser %s added to an existing field", name)
+		}
+	}
+
+	return nil
+}
+
+// Compares all synonym sources
+// Synonym sources currently not in use are also compared
+func compareSynonymSources(oriPaths, updPaths map[string]*pathInfo, ori, upd *mapping.IndexMappingImpl) error {
```
CascadingRadium, line 380: why checking for this differently to `analyzers` and `datetimeparsers`?
If you add / remove / modify any synonym source **NOT** being used by any field mapping, it should not rebuild but restart.

```
func compareSynonymSources( oriPaths, updPaths map[string]PathInfo, ori, upd *mapping.IndexMappingImpl) error {
	oriSynonymSources := make(map[string]interface{})
	updSynonymSources := make(map[string]interface{})

	extractSynonymSources := func(paths map[string]PathInfo, customSources map[string]interface{}, sources map[string]interface{}, getSynonymSourceName func(path string) string) {
		for path, info := range paths {
			for _, fInfo := range info.fieldMapInfo {
				if fInfo.fieldMapping.Type == "text" {
					synonymSourceName := getSynonymSourceName(path)
					fInfo.synonymSource = synonymSourceName
					if val, ok := customSources[synonymSourceName]; ok {
						sources[synonymSourceName] = val
					}
				}
			}
		}
	}

	extractSynonymSources(oriPaths, ori.CustomAnalysis.SynonymSources, oriSynonymSources, ori.SynonymSourceForPath)
	extractSynonymSources(updPaths, upd.CustomAnalysis.SynonymSources, updSynonymSources, upd.SynonymSourceForPath)

	for name, synUpd := range upd.CustomAnalysis.SynonymSources {
		if synOri, ok := ori.CustomAnalysis.SynonymSources[name]; ok {
			if !reflect.DeepEqual(synUpd, synOri) {
				return fmt.Errorf("synonym source %s changed while being used by fields", name)
			}
		} else {
			return fmt.Errorf("synonym source %s added to an existing field", name)
		}
	}

	return nil
}
```

CascadingRadium (commented):
second pass

index_update.go
```diff
@@ -0,0 +1,616 @@
...
+	// and consolidate according to the document paths
+	for name, oriDMapping := range ori.TypeMapping {
+		addPathInfo(oriPaths, "", oriDMapping, ori, nil, name)
+	}
+	addPathInfo(oriPaths, "", ori.DefaultMapping, ori, nil, "")
+
+	// Go through each mapping present in the updated
+	// and consolidate according to the document paths
+	for name, updDMapping := range upd.TypeMapping {
+		addPathInfo(updPaths, "", updDMapping, ori, nil, name)
```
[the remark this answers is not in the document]
CascadingRadium: resolved

CascadingRadium:
theres one case which is not implemented.
consider the following scenario

1. OriginalMapping
```
   "analysis": {
    "analyzers": {
     "custAna": {
      "token_filters": [
       "custFilt"
      ],
      "type": "custom",
      "tokenizer": "unicode"
     }
    },
    "token_filters": {
     "custFilt": {
      "min": 3,
      "max": 255,
      "type": "length"
     }
    },
   "types": {
    "abc": {
     "enabled": true,
     "dynamic": true,
     "default_analyzer": "custAna"
    }
   }
   
```

2. UpdatedMapping
```
   "analysis": {
    "analyzers": {
     "custAna": {
      "token_filters": [
       "custFilt"
      ],
      "type": "custom",
      "tokenizer": "unicode"
     }
    },
    "token_filters": {
     "custFilt": {
      "min": 10,
      "max": 255,
      "type": "length"
     }
    },
   "types": {
    "abc": {
     "enabled": true,
     "dynamic": true,
     "default_analyzer": "custAna"
    }
   }
```

Here assume the `custAna` analyzer is used in a type_mapping. Now, although the custom analyzer has not changed and the type_mapping is the same, the custom analysis component used within the custom analyzer has changed. This check is not there and it can wrongly flag the update as successful.

index_update.go
```diff
@@ -0,0 +1,616 @@
...
+}
+
+// Compares two field mappings against each other, checking for changes in index, store, doc values
+// and complete deletiion of the mapping while noting that the changes made are doable based on
+// other values like includeInAll and dynamic
+// first return argument gives an empty fieldInfo if no changes detected
+// second return argument gives a flag indicating whether any changes, if detected, are doable or if
+// update is impossible
+// third argument is an error explaining exactly why the change is not possible
+func compareFieldMapping(original, updated *mapping.FieldMapping) (*index.UpdateFieldInfo, bool, error) {
```
CascadingRadium, line 495: so the `compareFieldMapping` returns a `bool` flag that i feel is redundant.
the only usage of this updated is in L601
```
	if updated {
		if ori.dynamic {
			return fmt.Errorf("updated field is under a dynamic property")
		}
	}
```
you can just remove the bool return value and just return the  `UpdateFieldInfo` and `error`. You can just do the check in L601 instead
```
	if (newInfo.Deleted || newInfo.Index || newInfo.Store || newInfo.DocValues) && ori.dynamic {
	     return fmt.Errorf("updated field is under a dynamic property")
	}
```

index_update.go
```diff
@@ -0,0 +1,616 @@
...
+	if ori.TypeField != upd.TypeField &&
+		(len(ori.TypeMapping) != 0 || len(upd.TypeMapping) != 0) {
+		return fmt.Errorf("type field cannot be changed when type mappings are present")
+	}
+
+	if ori.DefaultType != upd.DefaultType {
+		return fmt.Errorf("default type cannot be changed")
+	}
+
+	if ori.DefaultField != upd.DefaultField {
```
CascadingRadium, line 137: L137 - L139 can be removed as this check is not required as unpdating `DefaultField` should not cause an index rebuild

index_update.go
```diff
@@ -0,0 +1,616 @@
...
+// third argument is an error explaining exactly why the change is not possible
+func compareFieldMapping(original, updated *mapping.FieldMapping) (*index.UpdateFieldInfo, bool, error) {
+	rv := &index.UpdateFieldInfo{}
+
+	if updated == nil {
+		if original != nil && !original.IncludeInAll {
+			rv.Deleted = true
+			return rv, true, nil
+		} else if original == nil {
+			return nil, false, nil
```
CascadingRadium, line 503: shouldn't this be 
```
return rv, false, nil
```
as downstream there is no check for the `UpdateFieldInfo` returned by this function being nil, which can cause a panic - in L607 for example

CascadingRadium: actually seeing the code again in L239
```
	for _, fMap := range mp.Fields {
		fieldMapInfo := &fieldMapInfo{
			fieldMapping: fMap,
			rootName:     rootName,
			parent:       pInfo,
		}
		pInfo.fieldMapInfo = append(pInfo.fieldMapInfo, fieldMapInfo)
	}
```
so there can be no chance of original being nil as in L425 we are iterating through `ori.fieldMapInfo` which if it has a valid value, Must have a valid non-nil `fieldMapping` as per the L239

```
	if upd == nil {
		for _, oriFMapInfo := range ori.fieldMapInfo {
			info, updated, err = compareFieldMapping(oriFMapInfo.fieldMapping, nil)
			if err != nil {
				return err
			}
```

index_update.go
```diff
@@ -0,0 +1,616 @@
...
+			}
+			// Compare analyser, datetime parser and synonym source before comparing
+			// the field mapping as it might not have this information
+			if updAnalyser != "" && oriFMapInfo.analyzer != updAnalyser {
+				return fmt.Errorf("analyser has been changed for a text field")
+			}
+			if updDatetimeParser != "" && oriFMapInfo.datetimeParser != updDatetimeParser {
+				return fmt.Errorf("datetime parser has been changed for a text field")
+			}
+			if updSynonymSource != "" && oriFMapInfo.synonymSource != updSynonymSource {
```
CascadingRadium, line 465: remove L465 - L467 as re previous comment, also remove `synonymSource` from struct in L41

CascadingRadium (dismissed):

index/scorch/scorch.go
```diff
@@ -36,6 +36,8 @@ const Version uint8 = 2
 
 var ErrClosed = fmt.Errorf("scorch closed")
 
+var MappingInternalKey = []byte("_mapping")
```
abhinavdangeti, line 39: Should make this a `const` - and I don't think this is the right place for this as @CascadingRadium commented earlier.

index_impl.go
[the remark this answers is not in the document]
abhinavdangeti: I agree with @CascadingRadium that we should not relocate this to within `index/scorch`.
That said, relocating this into `bleve_index_api` doesn't feel right either.

To avoid circular dependencies, let's move this alongside all other keys as well into a a key.go file within `index` - they all must be `const` and exported for their use everywhere.

abhinavdangeti (changes requested):

index/scorch/optimize_knn.go
```diff
@@ -101,6 +101,11 @@ func (o *OptimizeVR) Finish() error {
 					wg.Done()
 				}()
 				for _, field := range fields {
+					// Noop if the vector field or its index data is supposed to be deleted
+					if info, ok := o.snapshot.updatedFields[field]; ok && (info.Deleted || info.Index) {
```
abhinavdangeti, line 105: As far as readability goes, I have to say this looks a bit off-putting, perhaps commentary here on why `|| info.Index` will help.

index/scorch/persister.go
```diff
@@ -728,6 +740,7 @@ var (
 	boltMetaDataSegmentVersionKey = []byte("version")
 	boltMetaDataTimeStamp         = []byte("timeStamp")
 	boltStatsKey                  = []byte("stats")
+  boltUpdatedFieldsKey          = []byte("fields")
```
abhinavdangeti, line 743: Fix formatting here - think I'm responsible here while resolving conflicts.

index/scorch/snapshot_index.go
```diff
@@ -81,6 +81,8 @@ type IndexSnapshot struct {
 
 	m2        sync.Mutex                                 // Protects the fields that follow.
 	fieldTFRs map[string][]*IndexSnapshotTermFieldReader // keyed by field, recycled TFR's
+
+	updatedFields map[string]*index.UpdateFieldInfo
```
abhinavdangeti, line 85: Let's add a lot of commentary around here indicating what this is for (for the future), and what each of the attributes of `UpdateFieldInfo` stand for.

index/scorch/snapshot_segment.go
```diff
@@ -146,6 +147,29 @@ func (s *SegmentSnapshot) Size() (rv int) {
 	return
 }
 
+// Merge given updated field information with existing and pass it on to the segment base
+func (s *SegmentSnapshot) UpdateFieldsInfo(updatedFields map[string]*index.UpdateFieldInfo) {
+
```
abhinavdangeti, line 152: Drop empty line.

index_impl.go
```diff
@@ -202,7 +218,7 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
 		}
 	}()
 
-	mappingBytes, err := indexReader.GetInternal(mappingInternalKey)
+	mappingBytes, err := indexReader.GetInternal(scorch.MappingInternalKey)
```
abhinavdangeti, line 221: Relocate MappingInternalKey and other keys into a separate keys.go within `index/`.

abhinavdangeti (changes requested):
Looking mostly ok to me @Likith101 - I've a few requests for some more commentary and the positioning the mapping_key and we should be good here.

abhinavdangeti:
@Likith101 Would you add a new `docs/index_update.md` with a title "Ability to reduce downtime during index mapping updates" (in the same format as the other docs there) indicating the status quo and capturing how to leverage this feature for our community users.

I'll leave it be your decision if you want to make that a separate commit, but it'll need to be available before we produce v2.5.0.

index_impl.go
Thejas-bhat: @abhinavdangeti, adding to what you've suggested here - creating a new file called keys.go under means that we'd need to define the package name for the file. Perhaps we can leverage the existing util package that exists at the bleve level? or maybe we can create a new util package within index/ (although that sounds redundant because its something applicable to all types of indexes)

Likith101: The only other keys are boltdb related keys from index/scorch/persister.go. There are other keys within the search package as well. Which of these keys should I move to keys.go.

Also, we would need a package for keys.go. Would it be within /index or do I need to make a new directory as well so that it would be /index/keys/keys.go? If it is the former, then what would the package name be?

docs/index_update.md
```diff
@@ -0,0 +1,29 @@
+# Ability to reduce downtime during index mapping updates
+
+* *v2.5.0* (and after) will come with support to delete certain fields or parts of the fields without requiring a full rebuild of the index
```
CascadingRadium, line 3: `delete or modify any field mapping in the index mapping`

docs/index_update.md
```diff
@@ -0,0 +1,29 @@
...
+* We do this by storing which portions of the field has to be deleted within zap and then lazily executing the deletion during subsequent merging of the segments
+
+## Usage
+
+While opening an index, if an updated mapping is provided as a string under the key `updated_mapping` within the `runtimeConfig` parameter of `openIndexUsing`, then we open the index and try to update it to use the new mapping provided.
```
CascadingRadium, line 8: `OpenUsing` as it is a `bleve` package exported method.
also simplify the english
```
If `updated_mapping` is provided as a string in the `runtimeConfig` parameter of `OpenUsing`, the index is opened and updated with the new mapping.
```

docs/index_update.md
```diff
@@ -0,0 +1,29 @@
...
+
+On failure, we still return a usable index with an error explaining why the update failed.
```
CascadingRadium, line 10: the old index with old mapping is returned, `usable` index is misleading 
```
If the update fails, the index is opened with the existing mapping, and an error is returned explaining why the update was unsuccessful.
```

docs/index_update.md
```diff
@@ -0,0 +1,29 @@
...
+
+## What can be deleted and what can't be deleted?
+
+* Non updatable changes
+    * Any additional fields or enabled document mappings in the new index mapping
+    * Any changes to IncludeInAll, type, IncludeTermvECTORS AND SkipFreqNorm
```
CascadingRadium, line 16: Fix `IncludeTermvECTORS` typo

docs/index_update.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Usage
+
+While opening an index, if an updated mapping is provided as a string under the key `updated_mapping` within the `runtimeConfig` parameter of `openIndexUsing`, then we open the index and try to update it to use the new mapping provided.
+
+On failure, we still return a usable index with an error explaining why the update failed.
+
+## What can be deleted and what can't be deleted?
+
+* Non updatable changes
```
CascadingRadium, line 14: I would suggest replacing `Non updatable changes` section with a `Updatable changes` section, as this would make it clear to the end-user about what functionality is supported, rather than what is not. 
This would make it clear to the user that any index mapping change which is not present in the `Updatable changes` section is by default not supported

docs/index_update.md
```diff
@@ -0,0 +1,29 @@
...
+    * Text fields with a different analyser or date time fields with a different date time format
+    * Vector and VectorBase64 fields changing dims, similarity or vectorIndexOptimizedFor
+    * Any changes when field is part of `_all`
+    * Full field deletions when it is covered by any dynamic setting (Index, Store or DocValues Dynamic)
+    * Any changes to dynamic settings at the top level or any enabled document mapping
+    * If multiple fields sharing the same field name either from different type mappings or aliases are present, then any non compatible changes across all of these fields
+* Updatable changes provided non of the other contitions are hit
+    * Index, DocValues, Store of a field changing from true to false
+    * Document mapping being disabled or completely removed
+
```
CascadingRadium, line 27: also add a Sample code here demonstrating the feature, from a Bleve standpoint, using the exported `bleve` package-level APIs

index_impl.go
abhinavdangeti: Good point @Thejas-bhat . @Likith101 let's do that - move into util/keys.go all the keys that are common the `scorch` and `upside_down`.

docs/index_update.md
abhinavdangeti: Correct^, `openIndexUsing` is _not_ an exported API - so needs no mention here.

abhinavdangeti: Throw in an example here, showing that usecase.

abhinavdangeti (changes requested):
@Likith101 It seems this is introducing multiple unit test failures that need your attention - _please_ make it a practice to run `go test ./...` locally with and without the `vectors` tag before proposing each patch.

abhinavdangeti:
.. and `go test ./... -race`.

abhinavdangeti:
Removed all traces of supporting code needed here with https://github.com/blevesearch/bleve/pull/2171, will need new PRs and linking those to this one.

util/keys.go
```diff
@@ -0,0 +1,32 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+var (
```
abhinavdangeti, line 17: Should change these to `const`s.

abhinavdangeti (changes requested):
A vector unit test requires attention ..
```
--- FAIL: TestIndexUpdateVector (0.07s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30 pc=0x100381494]

goroutine 14502 [running]:
testing.tRunner.func1.2({0x1006bc4c0, 0x100b1ce80})
	/opt/homebrew/Cellar/go/1.25.0/libexec/src/testing/testing.go:1872 +0x190
testing.tRunner.func1()
	/opt/homebrew/Cellar/go/1.25.0/libexec/src/testing/testing.go:1875 +0x31c
panic({0x1006bc4c0?, 0x100b1ce80?})
	/opt/homebrew/Cellar/go/1.25.0/libexec/src/runtime/panic.go:783 +0x120
github.com/blevesearch/bleve/v2/index/scorch.(*IndexSnapshotVectorReader).Next(0x1400039ad00, 0x1?)
	/Users/abhinav.dangeti/Documents/go/src/github.com/blevesearch/bleve/index/scorch/snapshot_index_vr.go:86 +0x74
github.com/blevesearch/bleve/v2/search/searcher.(*KNNSearcher).Next(0x140007488c0, 0x14001b8cb40)
	/Users/abhinav.dangeti/Documents/go/src/github.com/blevesearch/bleve/search/searcher/search_knn.go:118 +0x94
github.com/blevesearch/bleve/v2/search/searcher.(*DisjunctionSliceSearcher).initSearchers(0x140000d66c0, 0x14001b8cb40)
	/Users/abhinav.dangeti/Documents/go/src/github.com/blevesearch/bleve/search/searcher/search_disjunction_slice.go:151 +0x9c
github.com/blevesearch/bleve/v2/search/searcher.(*DisjunctionSliceSearcher).Next(0x140000d66c0, 0x14001b8cb40)
	/Users/abhinav.dangeti/Documents/go/src/github.com/blevesearch/bleve/search/searcher/search_disjunction_slice.go:215 +0x30
github.com/blevesearch/bleve/v2/search/collector.(*KNNCollector).Collect(0x14000647a00, {0x10078e588, 0x100bde840}, {0x100793100, 0x140000d66c0}, {0x100794d68, 0x1400012e4d0})
	/Users/abhinav.dangeti/Documents/go/src/github.com/blevesearch/bleve/search/collector/knn.go:181 +0x264
github.com/blevesearch/bleve/v2.(*indexImpl).runKnnCollector(0x140001d2930, {0x10078e588, 0x100bde840}, 0x14000432f00, {0x100794d68, 0x1400012e4d0}, 0x0)
	/Users/abhinav.dangeti/Documents/go/src/github.com/blevesearch/bleve/search_knn.go:412 +0x354
github.com/blevesearch/bleve/v2.(*indexImpl).SearchInContext(0x140001d2930, {0x10078e588, 0x100bde840}, 0x14000432f00)
	/Users/abhinav.dangeti/Documents/go/src/github.com/blevesearch/bleve/index_impl.go:700 +0x6e0
github.com/blevesearch/bleve/v2.(*indexImpl).Search(0x0?, 0x48?)
	/Users/abhinav.dangeti/Documents/go/src/github.com/blevesearch/bleve/index_impl.go:480 +0x30
github.com/blevesearch/bleve/v2.TestIndexUpdateVector(0x14000910e00)
	/Users/abhinav.dangeti/Documents/go/src/github.com/blevesearch/bleve/search_knn_test.go:1904 +0x1c40
testing.tRunner(0x14000910e00, 0x1007843a8)
	/opt/homebrew/Cellar/go/1.25.0/libexec/src/testing/testing.go:1934 +0xc8
created by testing.(*T).Run in goroutine 1
	/opt/homebrew/Cellar/go/1.25.0/libexec/src/testing/testing.go:1997 +0x364
FAIL	github.com/blevesearch/bleve/v2	48.307s
```

docs/index_update.md
```diff
@@ -0,0 +1,41 @@
+# Ability to reduce downtime during index mapping updates
+
+* *v2.5.0* (and after) will come with support to delete or modify any field mapping in the index mapping without requiring a full rebuild of the index
```
abhinavdangeti: v2.5.4

index/scorch/scorch.go
```diff
@@ -217,9 +220,11 @@ func (s *Scorch) fireAsyncError(err error) {
 }
 
 func (s *Scorch) Open() error {
-	err := s.openBolt()
-	if err != nil {
-		return err
+	if !s.loadedBolt {
```
abhinavdangeti: Can a data race occur over this?

util/keys.go
Likith101: Byte arrays cannot be constant. These variables were always var, so I left it the same way I found them.

They have to be byte arrays for their usage within bolt.

index/scorch/scorch.go
Likith101: No, this is only used during the update path where bolt needs to be loaded before the index is started. The order of execution is always sequential, so I believe there is no risk of a data race here.

abhinavdangeti: Perhaps this check needs to be situated in `OpenMeta()` also then for readability and for safety as it's an exported API.

util/keys.go
abhinavdangeti: Ack.

index/scorch/scorch.go
```diff
@@ -940,3 +945,100 @@ func (s *Scorch) CopyReader() index.CopyReader {
...
+// Updates bolt db with the given field info. Existing field info already in bolt
+// will be merged before persisting. The index mapping is also overwritted both
+// in bolt as well as the index snapshot
+func (s *Scorch) UpdateFields(fieldInfo map[string]*index.UpdateFieldInfo, mappingBytes []byte) error {
+	err := s.updateBolt(fieldInfo, mappingBytes)
+	if err != nil {
+		return err
+	}
+	// Pass the update field info to all snapshots and segment bases
+	s.root.m.Lock()
```
abhinavdangeti: Let's not use `m` outside an IndexSnapshot member method - it doesn't seem like we call the IndexSnapshot method `UpdateFieldsInfo` outside lock context, so let's move the acquiring and releasing lock functionality to within it.

CascadingRadium (dismissed):

abhinavdangeti (dismissed):
@Likith101 one check required for `scoring_model` as well allowing any change between states - `"", "tf-idf", "bm25" to be changeable without index rebuild. Please weigh in on whether that change is required here or in cbft.

index/scorch/scorch.go
```diff
@@ -74,6 +75,7 @@ type Scorch struct {
 	merges                   chan *segmentMerge
 	introducerNotifier       chan *epochWatcher
 	persisterNotifier        chan *epochWatcher
+	loadedBolt               bool
```
Thejas-bhat: given that `s.rootbolt` is pointed to the right instance within this function, it would suffice to use that as a nil check whether there's a bolt instance right? or do you see any scenario where that's not a sufficient check/might be brittle?

index_impl.go
```diff
@@ -185,15 +203,32 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
...
-	if err != nil {
-		return nil, err
-	}
-	defer func(rv *indexImpl) {
-		if !rv.open {
-			rv.i.Close()
+
+	var ui index.UpdateIndex
+	if um != nil {
+		var ok bool
```
Thejas-bhat, line 209: i don't think you need to define this, the := notation will pass through and the 211 check will be on a valid var.

Likith101: There is a check on line 264 that will fail if I define this inside the if statement

index/scorch/scorch.go
Likith101: Agreed, I removed the variable and adjusted so that s.rootbolt is used everywhere instead

index_update.go
```diff
@@ -153,6 +153,13 @@ func compareMappings(ori, upd *mapping.IndexMappingImpl) error {
 		return fmt.Errorf("default datetime parser cannot be changed if index dynamic is true")
 	}
 
+	if ori.ScoringModel != upd.ScoringModel {
+		if ori.ScoringModel != "" && ori.ScoringModel != "tf-idf" && ori.ScoringModel != "bm25" ||
```
abhinavdangeti: There're constants in bleve_index_api for `tf-idf` and `bm25` which you should use here.

Also why check `ori.ScoringModel` here, just checking `upd.ScoringModel` should suffice yes?

index_update.go
```diff
@@ -154,9 +154,9 @@ func compareMappings(ori, upd *mapping.IndexMappingImpl) error {
 	}
 
 	if ori.ScoringModel != upd.ScoringModel {
-		if ori.ScoringModel != "" && ori.ScoringModel != "tf-idf" && ori.ScoringModel != "bm25" ||
-			upd.ScoringModel != "" && upd.ScoringModel != "tf-idf" && upd.ScoringModel != "bm25" {
-			return fmt.Errorf("scoring model can only be changed between \"\", tf-idf and bm25")
+		if ori.ScoringModel != "" && ori.ScoringModel != index.TFIDFScoring && ori.ScoringModel != index.BM25Scoring ||
```
abhinavdangeti, line 158: As we discussed would you add detailed commentary around here on why we add these checks.

CascadingRadium (approved):

abhinavdangeti (approved):

PR #2113, opened 2024-12-06
MB-58901: Introduce support for BM25 scoring

Thejas-bhat:
Introducing support for BM25 scoring 

Key stats necessary for the scoring
- fieldLength - the number of terms in a field within a doc.
- avgDocLength - the average of terms in a field across all the docs in the index.
- totalDocs - total number of docs in an index.

Introduces a mechanism to maintain consistent scoring in a situation where the index is partitioned as a `bleve.IndexAlias`. This is achieved using the existing preSearch mechanism where the first phase of the entire search involves fetching the above mentioned stats, aggregating them and redistributing back to the bleve indexes which would use them while calculating the score for a hit. In order to enable this global scoring mechanism, the user needs to set the `context` argument of the SearchInContext with:
`ctx = context.WithValue(ctx, search.SearchTypeKey, search.GlobalScoring)`

Implementation wise, the user needs to explicitly mention BM25 as the scoring mechanism at `indexMapping.ScoringModel` level to actually use this scoring mechanism. This parameter is a global setting, i.e. when performing a search on multiple fields, all the fields are scored with the same scoring model.
The storage layer exposes an API which returns the number of terms in a field's term dictionary which is used to compute the `avgDocLength`. At the indexing layer, we check if the queried field supports BM25 scoring and if consistent scoring is availed. This is followed by fetching the stats either from the local bleve index or from a context (in the case where we're availing the consistent scoring) to compute the actual score. 

Note: The scoring is highly dependent on the size of an individual bleve index's termDictionary (specific to a field) so there can be some discrepancies especially given that each index is further composed of multiple 'segments'. However in large scale use cases these discrepancies can be quite small and don't affect the order of the doc hits - in which case the user may choose to avoid this altogether.


mapping/index.go
```diff
@@ -201,6 +202,11 @@ func (im *IndexMappingImpl) Validate() error {
 			return err
 		}
 	}
+
+	if _, ok := index.SupportedScoringModels[im.ScoringModel]; !ok {
```
abhinavdangeti, line 206: You'll need a check here for the `empty` case.

search/searcher/search_term.go
```diff
@@ -60,17 +64,82 @@ func NewTermSearcherBytes(ctx context.Context, indexReader index.IndexReader, te
...
 		return nil, err
 	}
-	scorer := scorer.NewTermQueryScorer(term, field, boost, count, reader.Count(), options)
+	if ctx != nil {
+		if similaritModelCallback, ok := ctx.Value(search.
+			GetScoringModelCallbackKey).(search.GetScoringModelCallbackFn); ok {
+			similarityModel := similaritModelCallback()
+			if similarityModel == index.BM25Scoring {
+				// in case of bm25 need to fetch the multipliers as well (perhaps via context's presearch data)
+				count, avgDocLength, err = bm25ScoreMetrics(ctx, field, indexReader)
```
abhinavdangeti, line 134: Hmm it seems we'd be double calculating in the event of `bm25` scoring choice - perhaps an if-and-else is warranted here to either calculate just `tfTDFScoreMetrics` or `bm25ScoreMetrics`.

search/util.go
```diff
@@ -137,10 +137,15 @@ type GeoBufferPoolCallbackFunc func() *s2.GeoBufferPool
...
+const BM25PreSearchDataKey = "_bm25_pre_search_data_key"
 
 const PreSearchKey = "_presearch_key"
 
-type ScoreExplCorrectionCallbackFunc func(queryMatch *DocumentMatch, knnMatch *DocumentMatch) (float64, *Explanation)
+const SearchTypeKey = "_search_type_key"
+const FetchStatsAndSearch = "fetch_stats_and_search"
+
+const SearcherStartCallbackKey = "_searcher_start_callback_key"
+const SearcherEndCallbackKey = "_searcher_end_callback_key"
```
abhinavdangeti, line 148: I realize this is older code, but just to resolve some tech-debt as well here - would you add some commentary to what each key here stands for.

search/util.go
```diff
@@ -162,6 +167,15 @@ func (f FieldTermSynonymMap) MergeWith(fts FieldTermSynonymMap) {
 }
 
 const FieldTermSynonymMapKey = "_field_term_synonym_map_key"
+const BM25MapKey = "_bm25_map_key"
```
abhinavdangeti, line 170: ditto^

search/util.go
```diff
@@ -162,6 +167,15 @@ func (f FieldTermSynonymMap) MergeWith(fts FieldTermSynonymMap) {
...
 
-const SearcherStartCallbackKey = "_searcher_start_callback_key"
-const SearcherEndCallbackKey = "_searcher_end_callback_key"
+type BM25Stats struct {
+	DocCount         float64        `json:"doc_count"`
+	FieldCardinality map[string]int `json:"field_cardinality"`
+}
+
+const GetScoringModelCallbackKey = "_get_scoring_model"
```
abhinavdangeti, line 177: ditto^

Worth situating all these keys together perhaps too?

abhinavdangeti (changes requested):
@Thejas-bhat Make sure to pull the latest from `origin/bm25-refactor` before you push more commits here :)

abhinavdangeti:
@Thejas-bhat a thought regarding the "global scoring" code path for bm25 - what is the default behavior in elastic?

Would you add a couple of GO benchmark test to differentiate between bm25 with and without global scoring and record these numbers within the commit message^ - trying to decide whether to enable "global scoring" by default.

search/searcher/search_term.go
abhinavdangeti: 👍🏼

search/scorer/scorer_term.go
```diff
@@ -114,6 +132,65 @@ func (s *TermQueryScorer) SetQueryNorm(qnorm float64) {
 	}
 }
 
+// multiplies deciding how much does a doc length affect the score and also
+// how much can the term frequency affect the score in BM25 scoring
+var k1 float64 = 1.2
+var b float64 = 0.75
```
abhinavdangeti, line 138: Like we discussed, no point in making these `var`s and keeping them unexported.
Perhaps export them and prefix a `BM25` to indicate what they're applicable to.

search/scorer/scorer_term.go
```diff
@@ -61,19 +62,36 @@ func (s *TermQueryScorer) Size() int {
...
+	} else {
+		rv = 1.0 + math.Log(float64(docTotal)/
+			float64(docTerm+1.0))
+	}
+
+	return rv
+}
+
+func NewTermQueryScorer(queryTerm []byte, queryField string, queryBoost float64, docTotal,
+	docTerm uint64, avgDocLength float64, options search.SearcherOptions) *TermQueryScorer {
```
abhinavdangeti, line 80: I realize `docTerm` is carry-over name, but perhaps we should replace all occurrences of it with something better indicative of what it exactly is.

search/searcher/search_term.go
```diff
@@ -60,17 +64,86 @@ func NewTermSearcherBytes(ctx context.Context, indexReader index.IndexReader, te
...
-		_ = reader.Close()
-		return nil, err
+		return 0, 0, err
+	}
+	// field cardinality metric is not used in the tf-idf scoring algo.
+	fieldCardinality := 0
+	if count == 0 {
+		return 0, 0, nil
+	}
+	return count, float64(fieldCardinality / int(count)), nil
```
abhinavdangeti, line 81: It seems the 2nd parameter is always 0 here, so why declare it even and do a divide by operation and then return it. Change return params to 0 and `avgDocLength` below where this method is referenced will remain at zero.

search/searcher/search_term.go
```diff
@@ -60,17 +64,86 @@ func NewTermSearcherBytes(ctx context.Context, indexReader index.IndexReader, te
...
+	case index.BM25Scoring:
+		count, avgDocLength, err = bm25ScoreMetrics(ctx, field, indexReader)
+		if err != nil {
+			_ = reader.Close()
+			return nil, err
+		}
+	case index.TFIDFScoring:
+		fallthrough
+	default:
+		count, avgDocLength, err = tfTDFScoreMetrics(indexReader)
```
abhinavdangeti, line 140: No need for this 2nd parameter here at all, related to previous comment^.

abhinavdangeti (changes requested):

search/scorer/scorer_term.go
Thejas-bhat: maybe the name is self-explanatory (the alternate names i can think of is docFreq - number of documents having the term). it's just that the lack commentary makes it a little less obvious. shall i just add the commentary alongside docTerm in the struct and leave it as is?

Thejas-bhat:
> @Thejas-bhat a thought regarding the "global scoring" code path for bm25 - what is the default behavior in elastic?
> 
> Would you add a couple of GO benchmark test to differentiate between bm25 with and without global scoring and record these numbers within the commit message^ - trying to decide whether to enable "global scoring" by default.

By default, elastic disables the feature. It'll be a bit difficult to benchmark this at golang unit level over here, because the latency is mainly visible when the index alias has multiple shards and each of which is spread across multiple nodes.

search/scorer/scorer_term.go
```diff
@@ -114,6 +132,60 @@ func (s *TermQueryScorer) SetQueryNorm(qnorm float64) {
...
+		fieldNormVal := 1 - search.BM25_b + (search.BM25_b * fieldLength / s.avgDocLength)
+		fieldNormalizeExplanation := &search.Explanation{
+			Value: fieldNormVal,
+			Message: fmt.Sprintf("fieldNorm(field=%s), b=%f, fieldLength=%f, avgFieldLength=%f)",
+				s.queryField, search.BM25_b, fieldLength, s.avgDocLength),
+		}
+
+		saturationExplanation := &search.Explanation{
+			Value: search.BM25_k1 / (tf + search.BM25_k1*fieldNormVal),
+			Message: fmt.Sprintf("saturation(term:%s), k1=%f/(tf=%f + k1*fieldNorm=%f))",
```
metonymic-smokey, line 162: While I see that you've used BM25 specific terms like 'saturation' here, as you should, I think it would be good to add the scoring mode(tf-idf/BM25) too in the explanation?

search/searcher/search_term.go
```diff
@@ -38,14 +40,16 @@ type TermSearcher struct {
...
+	term string, field string, boost float64, options search.SearcherOptions) (search.Searcher, error) {
 	if isTermQuery(ctx) {
 		ctx = context.WithValue(ctx, search.QueryTypeKey, search.Term)
 	}
 	return NewTermSearcherBytes(ctx, indexReader, []byte(term), field, boost, options)
 }
 
-func NewTermSearcherBytes(ctx context.Context, indexReader index.IndexReader, term []byte, field string, boost float64, options search.SearcherOptions) (search.Searcher, error) {
+func NewTermSearcherBytes(ctx context.Context, indexReader index.IndexReader,
+	term []byte, field string, boost float64, options search.SearcherOptions) (search.Searcher, error) {
```
metonymic-smokey, line 52: Nit: I think this extra line space can be removed.

search/util.go
```diff
@@ -161,7 +192,14 @@ func (f FieldTermSynonymMap) MergeWith(fts FieldTermSynonymMap) {
...
-const FieldTermSynonymMapKey = "_field_term_synonym_map_key"
-
-const SearcherStartCallbackKey = "_searcher_start_callback_key"
-const SearcherEndCallbackKey = "_searcher_end_callback_key"
+// BM25 specific multipliers which affect the scoring of a document.
+//
+// BM25_b - how much does a doc's field length affect the score
+// BM25_k1 - how much can the term frequency affect the score
+var BM25_k1 float64 = 1.2
+var BM25_b float64 = 0.75
```
metonymic-smokey, line 200: We can discuss this separately - how did you arrive at these defaults for b and k1? Could you pls add some commentary around that?

search/scorer/scorer_term.go
abhinavdangeti: Roger, perhaps some commentary on all the function parameters then.

search/scorer/scorer_term.go
Thejas-bhat: done

search/util.go
Thejas-bhat: these are defaults as per elastic search docs. i've added links to the blog and docs as a code comment.

search/searcher/search_term.go
Thejas-bhat: i broke the line mainly because it'll be easy in the future while working on this area (so that all the params are fit within that 80 chars limit and we don't have to scroll much)

index_impl.go
```diff
@@ -485,6 +485,8 @@ func (i *indexImpl) preSearch(ctx context.Context, req *SearchRequest, reader in
 	}
 
 	var fts search.FieldTermSynonymMap
+	var count uint64
+	fieldCardinality := make(map[string]int)
```
abhinavdangeti, line 489: Lets change this to ..
```
var fieldCardinality map[string]int
```

.. and `make` only if `isBM25Enabled(..) == true` in line 499 below.

index_impl.go
```diff
@@ -578,6 +604,14 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 					}
 					skipSynonymCollector = true
 				}
+				skipKNNCollector = true
```
abhinavdangeti, line 607: Is this right? Something we missed?

index_alias_impl.go
```diff
@@ -573,11 +576,20 @@ type asyncSearchResult struct {
 type preSearchFlags struct {
 	knn      bool
 	synonyms bool
+	bm25     bool // needs presearch for this too
 }
 
-// preSearchRequired checks if preSearch is required and returns a boolean flag
-// It only allocates the preSearchFlags struct if necessary
-func preSearchRequired(req *SearchRequest, m mapping.IndexMapping) (*preSearchFlags, error) {
+func isBM25Enabled(m mapping.IndexMapping) bool {
```
abhinavdangeti, line 582: Let's refactor this to `scoringModel(..)` which returns the scoring model to use instead.

index_impl.go
```diff
@@ -605,6 +639,21 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 		ctx = context.WithValue(ctx, search.FieldTermSynonymMapKey, fts)
 	}
 
+	scoringModelCallback := func() string {
+		if isBM25Enabled(i.m) {
```
abhinavdangeti, line 643: Related to other commend, let's update this method to return the scoring model to use.
To establish if it's bm25, you can add the extra check - `scoringModel() == index.BM25Scoring`.

abhinavdangeti (commented):
Minor refactor suggestion, looks good to me otherwise.

index_alias_impl.go
abhinavdangeti: On second thought, perhaps this is fine as is - given that IndexMappingImpl's ScoringModel is exported.

index_impl.go
abhinavdangeti: Perhaps this is also fine as is.

index_test.go
```diff
@@ -350,6 +350,137 @@ func TestBytesWritten(t *testing.T) {
 	cleanupTmpIndexPath(t, tmpIndexPath4)
 }
 
+func TestGlobalScoring(t *testing.T) {
```
abhinavdangeti, line 353: Rename to `TestBM25GlobalScoring`

abhinavdangeti (commented):
Can we also add a unit test comparing bm25 to tf-idf scoring over the same dataset?

index_impl.go
Thejas-bhat: yeah i've removed this, it seems like a merge conflict issue, thanks!

Thejas-bhat:
> Can we also add a unit test comparing bm25 to tf-idf scoring over the same dataset?

I've added a unit test on a sample 10 doc dataset, however i don't think there something concrete we can compare over there apart from tfidf scores being higher than bm25 (due to algo difference). a much better test would be to have a large enough dataset, and two indexes - one with bm25 and the other with tfidf and then compare the relevancy of the hits from each of them (https://www.elastic.co/blog/found-bm-vs-lucene-default-similarity)

abhinavdangeti (approved):

PR #2116, opened 2024-12-17
MB-64604: Fix interpreting scorch config: "fieldTFRCacheThreshold"

abhinavdangeti:
+ Firstly, refactor `FieldTFRCacheThreshold` to `fieldTFRCacheThreshold` for _some_ naming consistency here.
+ This threshold can be used to toggle recycling of TermFieldReaders on/off.
+ Couchbase users will have the ability to provide this setting within a JSON payload, which when interpreted into a map[string]interface{} will need to be interpreted as a `float64`.
+ Should library users set it as `int` we'll honor that setting as well.

### DIFF
--- a/index/scorch/snapshot_index.go
+++ b/index/scorch/snapshot_index.go
@@ -52,15 +52,6 @@ type asynchSegmentResult struct {
 
 var reflectStaticSizeIndexSnapshot int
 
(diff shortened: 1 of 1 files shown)


abhinavdangeti:
Let's use https://github.com/blevesearch/bleve/pull/2117 instead and forward merge `trinity-couchbase` into `master` after.

PR #2127, opened 2025-01-06
MB-64636 - Modified Weight for KNN Scorer

metonymic-smokey:
1. Changed weight of a kNN query to 1 to allow the boost value to kick in when computing query score.
Since the weight of only the kNN scorer is changed, this will not impact how boosting is calculated for other types of queries.

To reduce the kNN score relative to the FTS query score, set boost to <1.

2. Added a unit test which demonstrates boost increasing scores even for pure kNN queries(kNN + match none query).

### DIFF
--- a/search/scorer/scorer_knn.go
+++ b/search/scorer/scorer_knn.go
@@ -128,7 +128,7 @@ func (sqs *KNNQueryScorer) Score(ctx *search.SearchContext,
 }
 
 func (sqs *KNNQueryScorer) Weight() float64 {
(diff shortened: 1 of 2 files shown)


abhinavdangeti:
@metonymic-smokey we'll want to raise a backport PR with this to `7.6.x-couchbase`.

metonymic-smokey:
@abhinavdangeti here's the BP - https://github.com/blevesearch/bleve/pull/2130

abhinavdangeti (approved):

PR #2137, opened 2025-02-05
MB-64766: Decoding geo distances from the sort results

Likith101:


search/collector/topn.go
```diff
@@ -500,7 +502,22 @@ func (hc *TopNCollector) finalizeResults(r index.IndexReader) error {
...
 	})
+	if err != nil {
+		return err
+	}
 
+	// Decode geo sort keys back to its distance values
+	for i, so := range hc.sort {
+		if _, ok := so.(*search.SortGeoDistance); ok {
+			for _, dm := range hc.results {
+				distInt, err := numeric.PrefixCoded(dm.Sort[i]).Int64()
```
abhinavdangeti, line 513: Can we not do this when the sort value is being added to the response object the first time around as opposed to adding it and now looping over the results to change it again?

Likith101: We have to do this after the sorting is done because the sort value is a string and the encoding was done to ensure that you can compare the values using '>' and '<' within the sorting algorithm.

search/collector/topn.go
```diff
@@ -500,7 +502,22 @@ func (hc *TopNCollector) finalizeResults(r index.IndexReader) error {
...
+				if err != nil {
+					return err
+				}
+				dm.Sort[i] = fmt.Sprintf("%v", numeric.Int64ToFloat64(distInt))
```
abhinavdangeti, line 517: Using `strconv`'s `FormatFloat` is a better option than `fmt.Sprintf`.

Would you also throw in some comments on why it is necessary to assert the `[]byte` output to `Int64()` above first and then convert it into Float64 - how are we achieving any more precision with `Int64ToFloat64` ?

search_test.go
```diff
@@ -4375,3 +4375,78 @@ func TestSynonymSearchQueries(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for i, doc := range res.Hits {
+		hitDist, err := strconv.ParseFloat(doc.Sort[0], 64)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if math.Abs(hitDist-docs[i].distance) > 1 {
```
abhinavdangeti, line 4448: Does the distance unit always match the unit used in the query?

Let's adapt this test to use some real world geo points above and establish the exact distance expectation for those points as opposed to this.

search/collector/topn.go
Likith101: The original distance is calculated in float64. After which, it is converted to Int64 by directly reading the bytes. So the conversion back is necessary as the int is not really the distance at this point.

search_test.go
Likith101: The default unit used by the query is meters. The sort portion of the query in the testcase also is in meters so any sort of conversion based on units is not necessary. 

Also, I am a bit confused as to how changing the points to real world points makes any difference. We would still have to do a distance calculation outside of the query and compare it approximately with the results. The approximation is required because when we encode the distance and when we encode the point itself in a morton hash, we lose some precision, so it will not be a hundred percent accurate

abhinavdangeti (approved):

PR #2143, opened 2025-02-20
MB-64883 - Avoid redundant computation of eligible IDs 

metonymic-smokey:
Currently, filter eligible hits, which aren't specific to a segment, are computed redundantly for each segment.  

This PR pre-computes the eligible segment IDs and re-uses these for each segment.

### DIFF
--- a/index/scorch/optimize_knn.go
+++ b/index/scorch/optimize_knn.go
@@ -124,13 +124,12 @@ func (o *OptimizeVR) Finish() error {
 								eligibleVectorInternalIDs.And(snapshotGlobalDocNums[index])
 							}
 
-							eligibleLocalDocNums := make([]uint64,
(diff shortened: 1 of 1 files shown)


abhinavdangeti (dismissed):

metonymic-smokey:
@abhinavdangeti this PR will require - https://github.com/blevesearch/scorch_segment_api/pull/50

abhinavdangeti:
@metonymic-smokey with the type for `eligibleLocalDocNums` changed back to []uint64, I suppose you won't need that segment_api interface change. Have you checked how throughput looks with this?

abhinavdangeti (approved):

metonymic-smokey:
@abhinavdangeti yes, that's why I changed it.

The throughput still isn't too great but as far as I know, no queries are failing unlike when the tests were first run.

PR #2146, opened 2025-02-24
BugFix: Fixed Advance() API implementation in Optimized Composite Searchers

CascadingRadium:
- When using the `Advance()` API on an `IndexSnapshotTermFieldReader`, a special replacement mechanism is applied during a `Rewind`— when the requested document ID is behind the current document ID pointed to by the reader. In such cases, the object itself gets replaced.  
- However, this approach fails when the `IndexSnapshotTermFieldReader` is created by the `optimizeCompositeSearcher` method, which is used to construct optimized `conjunction` or `disjunction` searchers.  
- The issue arises because the `unadornedTermFieldReader` is initialized with the `term` set to `<optimization-type>` and `field` set to `*`. As a result, calling `Advance()` renders the entire `TFR` unusable, as it gets replaced by a dummy `TFR` that no longer functions.  
- This problem occurs specifically when calling `Advance()` on the `unadornedTermFieldReader` with an ID less than the current posting ID (triggering the rewind mechanism).  
- The issue is resolved by resetting the underlying `unadornedPostingsIteratorBitmap`, which effectively achieves the same result as the `TFR` replacement technique without rendering it unusable.

### DIFF
--- a/index/scorch/optimize.go
+++ b/index/scorch/optimize.go
@@ -393,5 +393,7 @@ func (i *IndexSnapshot) unadornedTermFieldReader(
 		includeNorm:        false,
 		includeTermVectors: false,
 		recycle:            false,
+		// signal downstream that this is a special unadorned termFieldReader
+		unadorned: true,
 	}
 }
--- a/index/scorch/snapshot_index_tfr.go
+++ b/index/scorch/snapshot_index_tfr.go
@@ -50,6 +50,7 @@ type IndexSnapshotTermFieldReader struct {
 	recycle            bool
 	bytesRead          uint64
 	ctx                context.Context
+	unadorned          bool
 }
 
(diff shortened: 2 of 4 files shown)


abhinavdangeti:
@CascadingRadium would you add a unit test for the code in question here demonstrating the issue which your commit is meant to address.

CascadingRadium:
Done, The added test `TestUnadornedDisjunctionAdvance` fails on master but passes on this branch.

The reason why this issue was uncaught was that the exiting implementation does not utiize the `Rewind` feature.

Thanks.

abhinavdangeti (dismissed):

abhinavdangeti (approved):

Likith101 (approved):

PR #2152, opened 2025-02-25
Introduce a new stat which indicates a bleve index's convergence

Thejas-bhat:
`index_bgthreads_active` indicates whether the background routines that maintain the index are busy doing some work. This means that the index hasn't converged to a steady state and there could be potential file segment merges or in-memory segment flushes still remaining. 
This stat is beneficial for the application layer to get better insight as to whether the index is doing some background work (which can have implications on the system's resource utilisation) or not.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -658,6 +658,16 @@ func (s *Scorch) StatsMap() map[string]interface{} {
 	m["num_persister_nap_merger_break"] = m["TotPersisterMergerNapBreak"]
 	m["total_compaction_written_bytes"] = m["TotFileMergeWrittenBytes"]
 
+	// the bool stat `index_converging` indicates whether the background routines
+	// which causes the index to reach a steady state are still doing some work.
+	if rootEpoch, ok := m["CurRootEpoch"].(uint64); ok {
+		if lastMergedEpoch, ok := m["LastMergedEpoch"].(uint64); ok {
+			if lastPersistedEpoch, ok := m["LastPersistedEpoch"].(uint64); ok {
+				m["index_converging"] = !(lastMergedEpoch == rootEpoch && lastPersistedEpoch == rootEpoch)
+			}
+		}
+	}
+
 	// calculate the aggregate of all the segment's field stats
 	aggFieldStats := newFieldStats()
 	for _, segmentSnapshot := range indexSnapshot.Segments() {
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -658,12 +658,12 @@ func (s *Scorch) StatsMap() map[string]interface{} {
 	m["num_persister_nap_merger_break"] = m["TotPersisterMergerNapBreak"]
 	m["total_compaction_written_bytes"] = m["TotFileMergeWrittenBytes"]
 
-	// the bool stat `index_converging` indicates whether the background routines
+	// the bool stat `index_active` indicates whether the background routines
 	// which causes the index to reach a steady state are still doing some work.
 	if rootEpoch, ok := m["CurRootEpoch"].(uint64); ok {
 		if lastMergedEpoch, ok := m["LastMergedEpoch"].(uint64); ok {
 			if lastPersistedEpoch, ok := m["LastPersistedEpoch"].(uint64); ok {
-				m["index_converging"] = !(lastMergedEpoch == rootEpoch && lastPersistedEpoch == rootEpoch)
+				m["index_active"] = !(lastMergedEpoch == rootEpoch && lastPersistedEpoch == rootEpoch)
 			}
 		}
 	}
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -658,12 +658,13 @@ func (s *Scorch) StatsMap() map[string]interface{} {
 	m["num_persister_nap_merger_break"] = m["TotPersisterMergerNapBreak"]
 	m["total_compaction_written_bytes"] = m["TotFileMergeWrittenBytes"]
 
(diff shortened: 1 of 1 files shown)


index/scorch/scorch.go
```diff
@@ -658,6 +658,16 @@ func (s *Scorch) StatsMap() map[string]interface{} {
 	m["num_persister_nap_merger_break"] = m["TotPersisterMergerNapBreak"]
 	m["total_compaction_written_bytes"] = m["TotFileMergeWrittenBytes"]
 
+	// the bool stat `index_converging` indicates whether the background routines
+	// which causes the index to reach a steady state are still doing some work.
+	if rootEpoch, ok := m["CurRootEpoch"].(uint64); ok {
+		if lastMergedEpoch, ok := m["LastMergedEpoch"].(uint64); ok {
+			if lastPersistedEpoch, ok := m["LastPersistedEpoch"].(uint64); ok {
+				m["index_converging"] = !(lastMergedEpoch == rootEpoch && lastPersistedEpoch == rootEpoch)
```
CascadingRadium, line 666: maybe call `index_idle` which will be true if index is idle (when `lastMergedEpoch == rootEpoch && lastPersistedEpoch == rootEpoch`)

Thejas-bhat: i see. is there any specific reason why you'd want it that way? from what i gather based on the suggestion, it is just speaking the stat in a different way without adding any extra information to the stat relative what's defined right now (which can be renamed as `index_active` i guess?)

CascadingRadium: ya either `index_active` or `index_idle` ... `index_converging` doesnt make sense

abhinavdangeti: Agree with Rahul here @Thejas-bhat - `index_idle` makes the most sense to me.

abhinavdangeti: Fair enough, let's keep the naming consistent between the two though - so if you choose `index_active` let's go with it.

index/scorch/scorch.go
```diff
@@ -658,6 +658,16 @@ func (s *Scorch) StatsMap() map[string]interface{} {
 	m["num_persister_nap_merger_break"] = m["TotPersisterMergerNapBreak"]
 	m["total_compaction_written_bytes"] = m["TotFileMergeWrittenBytes"]
 
+	// the bool stat `index_active` indicates whether the background routines
```
abhinavdangeti, line 661: @Thejas-bhat after a little more thought, would you change this to something a little less ambiguous - like `index_merger_active` perhaps?

index/scorch/scorch.go
```diff
@@ -658,6 +658,16 @@ func (s *Scorch) StatsMap() map[string]interface{} {
...
+	// which causes the index to reach a steady state are still doing some work.
```
abhinavdangeti, line 662:
```
// which is responsible for* the index to attain a steady state ..
```

index/scorch/scorch.go
```diff
@@ -658,6 +658,16 @@ func (s *Scorch) StatsMap() map[string]interface{} {
...
+	if rootEpoch, ok := m["CurRootEpoch"].(uint64); ok {
+		if lastMergedEpoch, ok := m["LastMergedEpoch"].(uint64); ok {
+			if lastPersistedEpoch, ok := m["LastPersistedEpoch"].(uint64); ok {
+				m["index_active"] = !(lastMergedEpoch == rootEpoch && lastPersistedEpoch == rootEpoch)
```
abhinavdangeti, line 666: Separately, wouldn't this stat incorrectly show as active in the situation should the merge policy prevent the merger from running over the last batch of items?

Thejas-bhat: in the case where there were no merged segments introduced, the LastMergedEpoch would still get updated with the latest root epoch's value. Assuming that the merger doesn't do any active merging of segments for a while (based on merge policy) it would remain equal to the root epoch and the stat value over here would be false (indicating that the index is not doing any activity in the background)

index/scorch/scorch.go
Thejas-bhat: like my thinking behind this stat was to indicate whether the entire index activity is idle because it has reached a steady state. that would be mean that both persister and merger need to be in sync with the root (epoch wise) and be idle, so naming the stat `index_merger_active` would be misleading in my opinion since if the persister's epoch went ahead the stat value would be true even if merger is in sync with root (but even that would eventually end up doing some work some time later).

index/scorch/scorch.go
abhinavdangeti: Ack.

index/scorch/scorch.go
abhinavdangeti: For closure here, believe we've agreed upon `index_bgthreads_active` or of the like.

abhinavdangeti:
@Thejas-bhat I suppose we're getting another patch here with the `bg` name?

abhinavdangeti:
Also, I saw your bool to uint conversion downstream - let's just make it an int here itself.

Thejas-bhat:
> Also, I saw your bool to uint conversion downstream - let's just make it an int here itself.

so the main reason why I kept it as bool here and didn't bother to change it is because its semantically more correct to keep the stat as a bool at the bleve level. 
The boolToUint() conversion is necessary mainly due to how cbft plays around with the stats init, tracks and displays it (which is the application level concern I think and something bleve doesn't have to worry about in my opinion). 
However, if we are okay with maintaining it as an int over here itself I can do that, its just that it won't be immediately understandable if the bleve community user sees it as 1 or 0.

abhinavdangeti (approved):

PR #2157, opened 2025-03-05
MB-65554: Add a `total_mutations_filtered` stat to track mutations that qualify for a type mapping

CascadingRadium:
- This stat tracks the number of mutations processed by the Bleve index that 
  are searchable by the user. It helps monitor the progress of the index-building 
  process.
- The stat is incremented for each mutation processed by the Bleve index that 
  is associated with a valid type mapping.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,14 +5,14 @@ go 1.21
 require (
 	github.com/RoaringBitmap/roaring/v2 v2.4.5
 	github.com/bits-and-blooms/bitset v1.12.0
-	github.com/blevesearch/bleve_index_api v1.2.2
+	github.com/blevesearch/bleve_index_api v1.2.3
 	github.com/blevesearch/geo v0.1.20
 	github.com/blevesearch/go-faiss v1.0.24
 	github.com/blevesearch/go-metrics v0.0.0-20201227073835-cf1acfcdf475
 	github.com/blevesearch/go-porterstemmer v1.0.3
 	github.com/blevesearch/goleveldb v1.0.1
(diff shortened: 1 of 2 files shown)


document/document.go
```diff
@@ -148,3 +149,11 @@ func (d *Document) VisitSynonymFields(visitor index.SynonymFieldVisitor) {
 		}
 	}
 }
+
+func (d *Document) SetSearchable(searchable bool) {
+	d.searchable = searchable
+}
+
+func (d *Document) Searchable() bool {
```
abhinavdangeti, line 157: Let's not call this `Searchable()`, let's change this `IsIndexed()`

document/document.go
```diff
@@ -34,6 +34,7 @@ type Document struct {
 	Fields           []Field `json:"fields"`
 	CompositeFields  []*CompositeField
 	StoredFieldsSize uint64
+	searchable       bool
```
abhinavdangeti, line 37: `isIndexed`

document/document.go
```diff
@@ -148,3 +149,11 @@ func (d *Document) VisitSynonymFields(visitor index.SynonymFieldVisitor) {
 		}
 	}
 }
+
+func (d *Document) SetSearchable(searchable bool) {
```
abhinavdangeti, line 153: There is no occasion where the input parameter is false, so change signature to `SetIndexed()`.

abhinavdangeti (approved):

Thejas-bhat (approved):

PR #2162, opened 2025-03-11
MB-59633: Removed encoded geo shapes from dictionaries

Likith101:


CascadingRadium (dismissed):

Thejas-bhat (dismissed):

Thejas-bhat (dismissed):

abhinavdangeti (changes requested):
@Likith101 Second instance of a pull request from you with unit tests failing. Please do better.

index/scorch/scorch.go
```diff
@@ -728,6 +728,23 @@ func analyze(d index.Document, fn customAnalyzerPluginInitFunc) {
 				d.VisitComposite(func(cf index.CompositeField) {
 					cf.Compose(field.Name(), field.AnalyzedLength(), field.AnalyzedTokenFrequencies())
 				})
+				if f, ok := field.(index.GeoShapeField); ok {
```
abhinavdangeti, line 731: Would you add some commentary here on why this is needed.

abhinavdangeti (approved):

index/scorch/scorch.go
```diff
@@ -728,6 +728,27 @@ func analyze(d index.Document, fn customAnalyzerPluginInitFunc) {
 				d.VisitComposite(func(cf index.CompositeField) {
 					cf.Compose(field.Name(), field.AnalyzedLength(), field.AnalyzedTokenFrequencies())
 				})
+				// Since the encoded geoShape is only necessary within the doc values
+				// of the geoShapeField, it has been removed from the field's term dictionary.
+				// However, '_all' field uses its term dictionary as its docValues, so it
+				// becomes necessary to add the geoShape into the '_all' field's term dictionary
```
CascadingRadium, line 734: doesn't this mean that the benefit is nullified when a geoshape query is run against the `_all` field, and is only applicable when the field itself is queried? 

can't the `extraDocValues` map added in the related zapx PR be modified to add the encodedShape there?

if doc1 has 2 fields (geoShapeFieldA and _all)
extraDocValues[doc1] -> map[_all] = map[geoShapeFieldA] -> encodedShape

Likith101: The _all field has no concept of doc values. There will always be a query hit when a geoshape query is run against this field and it cannot be prevented. The query benefit is only applicable against the geoshape field as you mentioned and the other benefit will be the reduction in index size and build time.

Since the perf results are not out yet, we still are not sure if the query latency has been improved by this change

CascadingRadium (approved):

PR #2169, opened 2025-03-21
MB-65473: Refactor and Optimize Pre-Filtered Vector Search

CascadingRadium:
- Refactor pre-filtered vector search to enhance performance and reduce memory footprint.
- Replace the current bitmap-based approach for calculating segment local document numbers with a more direct method, where the local document numbers are mapped directly to the segment ID during the execution of the eligible collector.
- Requires: 
    - https://github.com/blevesearch/bleve_index_api/pull/63
    - https://github.com/blevesearch/bleve_index_api/pull/66
    - https://github.com/blevesearch/zapx/pull/317
    - https://github.com/blevesearch/go-faiss/pull/41
    - https://github.com/blevesearch/faiss/pull/49


CascadingRadium:
adding a `do not merge` label for now, as the zapx side is yet to be investigated

metonymic-smokey (approved):

abhinavdangeti (commented):
Looks good, let's get some perf results for this and the order of merge should be `faiss` > `go-faiss` > `index_api/segment_api` > `zapx` > `bleve`

search/collector/eligible.go
```diff
@@ -155,3 +165,30 @@ func (ec *EligibleCollector) FacetResults() search.FacetResults {
 	// facet unsupported for pre-filtering in KNN search
 	return nil
 }
+
+type eligibleDocumentSelector struct {
```
Thejas-bhat, line 169: Overall I understand the approach and like how this PR is reducing the creation of bitmaps and operation of docNum conversions. However, I don't really like the fact that we're breaking the collector package (which is agnostic of the backing `IndexReader` implementation) with scorch specific constructs - and we absolutely must avoid it if possible.

What I think is a cleaner approach would be to move the concept of `eligibleDocumentSelector` struct to the scorch level. Here are my thoughts

Have the `IndexSnapshot` implement a `NewEligibleCollector() `API which returns `index.EligibleDocumentSelector`
```
func (is *IndexSnapshot) NewEligibleDocumentSelector() index.EligibleDocumentSelector {	
    return &eligibleDocumentSelector(InScorch){
      eligibleDocNums: map[int][]uint64{},
      indexSnapshot: is,
    }
}
```

and the VectorIndexReader interface would look like this 
```
type VectorIndexReader interface {
	VectorReader(ctx context.Context, vector []float32, field string, k int64,
		searchParams json.RawMessage) (VectorReader, error)

        NewEligibleDocumentSelector() EligibleDocumentSelector
}
```

The indexSnapshot already implements the `VectorIndexReader` interface. So, in the `makeEligibleDocumentMatchHandler` function, we can do a interface check on the indexReader and if implements the `VectorIndexReader` interface, we call `VectorIndexReader.NewEligibleDocumentSelector()` to get the eligibleDocumentSelector instance.

The existing index.EligibleDocumentSelector interface could look like this

```
type EligibleDocumentSelector interface {
	// EligibleDocNums returns a list of document numbers that are eligible within the specified segment.
	// segmentID identifies the segment for which eligible document numbers are retrieved.
	SegmentEligibleDocs(segmentID int) []uint64

        // comments
        AddEligibleDocumentMatch(d docMatch)
}
```

Now on the implementation side which is basically the exact same thing that's here, would also have a reference to the indexSnapshot within it and would implement something like this

```
func (ec *eligibleDocumentSelector(InScorch)) AddEligibleDocumentMatch(d docMatch){
  segIdx, docNum, err := ec.is.SegmentIndexAndLocalDocNum(d.IndexInternalID)
    if err != nil {
      return err
    }
    // Add the local doc number to the list of eligible doc numbers for this segment.
    ec.eligibleDocNums[segIdx] = append(ec.eligibleDocNums[segIdx], docNum)
    return nil
}

func (eds *eligibleDocumentSelector(InScorch)) SegmentEligibleDocs(segmentID int) []uint64 {
	return eds.eligibleDocNums[segmentID]
}
```

Please let me know if this is a valid concern according to you all (cc @abhinavdangeti)

CascadingRadium: ok i will do that, but the problem is that `eligible.go` is not in the `vectors` tag, so i will not extend the `VectorReader` interface; but just extend the normal `IndexReader` interface. Will follow the rest of the suggestions as is: 
- Shift the `eligibleDocumentSelector` struct to `scorch` package
- Create a new extended `IndexReader` interface to have the constructor for the struct
- In `EligibleCollector` i will check if `IndexReader` interface can be casted ot the new extended one from point 2.
- Rest of the code flow remains the same

search_knn.go
```diff
@@ -293,12 +287,6 @@ func createKNNQuery(req *SearchRequest, eligibleDocsMap map[int][]index.IndexInt
 }
 
 func validateKNN(req *SearchRequest) error {
-	if req.KNN != nil &&
```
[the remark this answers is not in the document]
CascadingRadium: if req.KNN is nil then this method is a no-op, this would make sense as a normal FTS request without a knn object is still valid.

search/collector/eligible.go
Thejas-bhat: since this is collector type is pre-filter specific and its only used in search_knn.go, i don't think its compiled without the `vectors` tag anyways. in my opinion, its just better to put the `vectors` tag in this eligible.go file and extend the `VectorIndexReader` interface since that's more appropriate than the `IndexReader` interface.

CascadingRadium: ok

CascadingRadium: finished 
check now

Thejas-bhat (approved):
yeah the collector code looks much better now

Thejas-bhat (dismissed):

abhinavdangeti (approved):

Thejas-bhat (approved):

PR #2170, opened 2025-03-24
Validate scorch routine options when creating or opening a Bleve Index.

CascadingRadium:
- If a malformed setting is used in the persistor/merger options, an async error is thrown leading to a crash loop in a live system.
- This PR adds a validation step to ensure that the options are valid before creating/opening the index.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -160,13 +160,14 @@ func NewScorch(storeName string,
 	if ok {
 		rv.onAsyncError = RegistryAsyncErrorCallbacks[aecbName]
 	}
-	// validate any custom persistor/merger planner options
-	// to prevent an async error in the background routines.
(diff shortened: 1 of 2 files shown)


index/scorch/scorch.go
```diff
@@ -160,6 +160,17 @@ func NewScorch(storeName string,
 	if ok {
 		rv.onAsyncError = RegistryAsyncErrorCallbacks[aecbName]
 	}
+	// validate any custom persistor/merger planner options
```
abhinavdangeti, line 163: Move the comment about merger planner options validation to line 169.5 below.

abhinavdangeti (approved):
Looks great 👍🏼

Thejas-bhat (approved):

Likith101 (approved):

PR #2178, opened 2025-04-14
MB-60971: update the bolt with the correct file segments

Thejas-bhat:
- This is applicable to concurrent persister algorithm (introduced in v2.5.0) - which is not _ON_ by default.
- In cases where the persister creates new file segments, the bolt file is supposed to be updated with the complete set of file segments. 
- However currently, while loading the index from the bolt source, there can be data loss issues because a set of segments associated with the root snapshot weren't tracked in the bolt file and hence wouldn't be loaded from the disk.
- Note that the live index doesn't have issues because the root would track the live persisted segments and point to the corresponding files. 
- Also added some better code comments

### DIFF
--- a/index/scorch/merge.go
+++ b/index/scorch/merge.go
@@ -488,7 +488,11 @@ func closeNewMergedSegments(segs []segment.Segment) error {
 	return nil
 }
 
-func (s *Scorch) mergeSegmentBasesParallel(snapshot *IndexSnapshot, flushableObjs []*flushable) (*IndexSnapshot, []uint64, error) {
+// mergeAndPersistInMemorySegments takes an IndexSnapshot and a list of in-memory segments,
+// which are merged and persisted to disk concurrently. These are then introduced as
+// the new root snapshot in one-shot.
+func (s *Scorch) mergeAndPersistInMemorySegments(snapshot *IndexSnapshot,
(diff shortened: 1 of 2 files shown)


abhinavdangeti:
@Thejas-bhat For couchbase it seems if the node were to go down during this situation, KV would ask us to rollback yes?

Thejas-bhat:
nope i don't think so - this is because bolt separately stores the high seqno. at vBucket level inside the index snapshot bucket of bolt which is tracked correctly. However, the segment snapshot bucket won't be filled with all the file segment's entries. So when the node goes down and comes back up, the seq no.s are correct (so no rollback) however the loadSegment() itself will be applied only to a subset of file segments that _should've_ been there

abhinavdangeti:
I see, and does this affects the situation because of the concurrent persister model or has it been always around? Do we _need_ to backport this asap?

Thejas-bhat:
i dont think we have to backport it because this is a bug that got introduced with the concurrent persister PR (finally a better name for it) that got merged recently in 2.5.0

abhinavdangeti (dismissed):

abhinavdangeti:
Thanks, I've updated the commit description^.

CascadingRadium (approved):

abhinavdangeti (approved):

PR #2180, opened 2025-04-16
MB-66295: avoid redundant calculation of bm25 metrics

Thejas-bhat:
- cache a map which maps the field to its cardinality value so that we don't create the fieldDict everytime
- Furthermore in the case all the threads have a cache miss, use the write lock to make sure that we create the fieldDict only once to minimize garbage. 
- Use a separate mutex construct so that the contention is localised to the metrics gathering and doesn't affect the other threads' TFR generation.
- https://github.com/blevesearch/bleve_index_api/pull/68

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,14 +7,14 @@ toolchain go1.23.5
 require (
 	github.com/RoaringBitmap/roaring/v2 v2.4.5
 	github.com/bits-and-blooms/bitset v1.22.0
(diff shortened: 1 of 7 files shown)


CascadingRadium (dismissed):

abhinavdangeti (approved):
Looks good @Thejas-bhat, nice one.

abhinavdangeti:
Already reviewed by @CascadingRadium , so will move ahead.

PR #2182, opened 2025-04-17
Refactor BM25 context key usage

CascadingRadium:
- Separates the use of `BM25PreSearchDataKey` (for pre-search data) and introduces `BM25StatsKey` for passing BM25 stats via context, similar to the existing `SynonymPreSearchDataKey` vs. `FieldTermSynonymMapKey` pattern. This improves clarity and consistency in key usage.

### DIFF
--- a/index_impl.go
+++ b/index_impl.go
@@ -591,7 +591,7 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 	var fts search.FieldTermSynonymMap
 	var skipSynonymCollector bool
 
-	var bm25Data *search.BM25Stats
+	var bm25Stats *search.BM25Stats
 	var ok bool
 	if req.PreSearchData != nil {
 		for k, v := range req.PreSearchData {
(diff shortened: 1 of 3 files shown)


abhinavdangeti (dismissed):

abhinavdangeti:
Shouldn't these be of type `ContextKey` as well? 

https://github.com/blevesearch/bleve/blob/3850d7fe0c17062ec6d2d3a37965f79fe70a3cb6/search/util.go#L181...L187

Let's add more commentary around these and around `GlobalScoring` as well as part of this change.

CascadingRadium:
no @abhinavdangeti the keys in 
https://github.com/blevesearch/bleve/blob/3850d7fe0c17062ec6d2d3a37965f79fe70a3cb6/search/util.go#L181...L187 
are not context keys, as they are not used in a `Context`. They are merely used to access and write data into the `PresearchData` struct of a `SearchRequest`. The `PresearchData` is a `map[string]interface{}`, and the PresearchKeys are the keys of this map.

abhinavdangeti:
I see, alright then - would you update the commentary around them then, indicative of this.

abhinavdangeti (approved):
Perhaps the type name is sufficient, this is fine.

Likith101 (approved):

PR #2186, opened 2025-04-24
MB-66163: fix checkpointing mechanism under sparse mutation scenarios 

Thejas-bhat:
- the checkpointing mechanism considers snapshots which fall within the `numSnapshotsToKeep * rollbackSamplingInterval` duration window. The snapshots outside this window are purged. 
- Under sparse mutation scenario for eg, when the workload pattern in such that it results in seldomly persisting snapshots that too outside this duration window - we could end up purging ALL those snapshots which makes the system fallback to preserving the latest snapshots
- For a better partial rollback behaviour, we retain `rollbackRetentionFactor` worth of snapshots out of this duration window such that the cost of rebuilding the index from scratch is avoided. 
- This PR fixes the decision logic involved in the scenario pointed and tracks the previous checkpoints while opening back the index as well.

### DIFF
--- a/index/scorch/persister.go
+++ b/index/scorch/persister.go
@@ -846,7 +846,7 @@ var (
 )
 
 func (s *Scorch) loadFromBolt() error {
-	return s.rootBolt.View(func(tx *bolt.Tx) error {
+	err := s.rootBolt.View(func(tx *bolt.Tx) error {
 		snapshots := tx.Bucket(boltSnapshotsBucket)
 		if snapshots == nil {
 			return nil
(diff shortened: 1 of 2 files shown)


index/scorch/persister.go
```diff
@@ -1283,15 +1293,15 @@ func (s *Scorch) removeOldZapFiles() error {
...
 // in such a scenario using the following function
 func getBoundaryCheckPoint(retentionFactor float64,
 	checkPoints []*snapshotMetaData, timeStamp time.Time,
 ) time.Time {
 	if checkPoints != nil {
 		boundary := checkPoints[int(math.Floor(float64(len(checkPoints))*
 			retentionFactor))]
-		if timeStamp.Sub(boundary.timeStamp) < 0 {
+		if timeStamp.Sub(boundary.timeStamp) > 0 {
 			// too less checkPoints would be left.
```
abhinavdangeti, line 1305: I think I follow the code here, but would you rephrase this comment here as well - indicating that we're moving the boundary on account of no checkpoints within?

abhinavdangeti:
@Thejas-bhat Let's backport this one as well to `7.6.x-couchbase` after this PR is merged.

index/scorch/persister.go
Thejas-bhat: done thansk!

abhinavdangeti (approved):

metonymic-smokey (approved):

PR #2190, opened 2025-05-05
Bug Fix: Prevent Indexing of `vector` and `geoshape` fields when `Index` option is `False`

CascadingRadium:
- If a field mapping for a vector or geoshape has the `Index` option set to `false`, the field is still indexed because the current code incorrectly always overrides this option to `true`.

- This behavior has been fixed by ensuring that the Index option is not overridden and the field is only indexed if explicitly allowed by the user-specified Index setting.

### DIFF
--- a/document/field_geoshape.go
+++ b/document/field_geoshape.go
@@ -158,7 +158,9 @@ func NewGeoShapeFieldWithIndexingOptions(name string, arrayPositions []uint64,
 		return nil
 	}
 
-	options = options | DefaultGeoShapeIndexingOptions
+	// docvalues are always enabled for geoshape fields, even if the
+	// indexing options are set to not include docvalues.
+	options = options | index.DocValues
 
 	return &GeoShapeField{
 		shape:             shape,
@@ -189,7 +191,9 @@ func NewGeometryCollectionFieldWithIndexingOptions(name string,
 		return nil
 	}
 
-	options = options | DefaultGeoShapeIndexingOptions
+	// docvalues are always enabled for geoshape fields, even if the
+	// indexing options are set to not include docvalues.
+	options = options | index.DocValues
 
 	return &GeoShapeField{
 		shape:             shape,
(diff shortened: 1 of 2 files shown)


abhinavdangeti (changes requested):
@CascadingRadium Would you add unit tests covering these cases?

abhinavdangeti:
On second thought, I believe this will affect only those who use these APIs (so not cbft) - the use of this API implies the user wants to index these fields. I'm questioning the need for this change now.

CascadingRadium:
No @abhinavdangeti, these APIs (`NewGeoShapeFieldWithIndexingOptions` and `NewVectorFieldWithIndexingOptions` are used in the `MapDocument` path - https://github.com/blevesearch/bleve/blob/8a928049eeb4b8d181176000ccfd56f75ee8b08c/mapping/mapping_vectors.go#L151 
The options passed here are the same as the ones set by the cbft user in the field mapping in the index def. The `index` option is toggleable in the classical editor UI, but only for these two field mapping (vector and geoshape), it does not do anything. For `text`, `numeric` etc. the expected behaviour is followed.

CascadingRadium:
Yes, if we look at the `text` field, the API `NewTextFieldWithIndexingOptions` directly takes the options and uses it as is, without any override.
https://github.com/blevesearch/bleve/blob/8a928049eeb4b8d181176000ccfd56f75ee8b08c/document/field_text.go#L132-L140
The basic API `NewTextField` uses the Default options.
https://github.com/blevesearch/bleve/blob/8a928049eeb4b8d181176000ccfd56f75ee8b08c/document/field_text.go#L128-L130
The GeoShape field has the similar `NewGeoShapeField` API that uses the Default.
https://github.com/blevesearch/bleve/blob/8a928049eeb4b8d181176000ccfd56f75ee8b08c/document/field_geoshape.go#L126-L130

abhinavdangeti (approved):
Thank you @CascadingRadium

Likith101 (approved):

PR #2198, opened 2025-05-22
MB-64766: Adding geo sort values to new array

Likith101:


search/sort.go
```diff
@@ -212,7 +213,11 @@ type SortOrder []SearchSort
 
 func (so SortOrder) Value(doc *DocumentMatch) {
 	for _, soi := range so {
-		doc.Sort = append(doc.Sort, soi.Value(doc))
+		value := soi.Value(doc)
+		doc.Sort = append(doc.Sort, value)
+		if _, ok := soi.(*SortGeoDistance); ok {
```
CascadingRadium, line 218: Currently even numeric and datetime sort values (from SortField) are also returned to the user in an encoded format. 

Would be better to account for those as well.
Check for sort Field and check if sort value is a PrefixCoded string using `ValidPrefixCodedTermBytes`. If invalid means its plain text. 

If valid its either a numeric or datetime. Based on the mode of sort Field you can decode the sort value to numeric and datetime as well

search/sort.go
```diff
@@ -212,7 +213,11 @@ type SortOrder []SearchSort
 
 func (so SortOrder) Value(doc *DocumentMatch) {
 	for _, soi := range so {
-		doc.Sort = append(doc.Sort, soi.Value(doc))
+		value := soi.Value(doc)
+		doc.Sort = append(doc.Sort, value)
```
CascadingRadium, line 217: There should be a one to one mapping to the `Sort` array and the `GeoDistance` array right. Decoded(Sort[i]) = GeoDistance[i]

If i have a sort order like this

[Field Geo Score Geo]

the Sort array will be of length 4,  but the geo distance array will be of size 2 

giving the wrong assumption that Decoded(Sort[1]) = GeoDistance[1]. It is actually Decoded(Sort[3]) = GeoDistance[1]

search/search.go
```diff
@@ -154,6 +154,7 @@ type DocumentMatch struct {
 	Locations       FieldTermLocationMap  `json:"locations,omitempty"`
 	Fragments       FieldFragmentMap      `json:"fragments,omitempty"`
 	Sort            []string              `json:"sort,omitempty"`
+	GeoDistance     []string              `json:"geo_distance,omitempty"`
```
CascadingRadium, line 157: need to recycle this like Sort. Check 
```
func (dm *DocumentMatch) Reset() 
```
also size accounting
```
func (dm *DocumentMatch) Size() int {
```

search/sort.go
```diff
@@ -212,7 +214,15 @@ type SortOrder []SearchSort
...
+		if _, ok := soi.(*SortGeoDistance); ok {
```
CascadingRadium, line 219: i think you can use a generic interface here like this

```

  for _, soi := range so {
      value := soi.Value(doc)
      doc.Sort = append(doc.Sort, value)

      if decoder, ok := soi.(SortValueDecoder); ok {
          doc.DecodedSort = append(doc.DecodedSort, decoder.DecodeValue(value))
      } else {
          doc.DecodedSort = append(doc.DecodedSort, value)
      }
  }

```

add interface

```
type SortValueDecoder interface {
    DecodeValue(value []byte) any
}
```
so that now the sort struct can have the impl like this (the oness that dont have `DecodeValue` like ID and score will not impl method and hence put if cond will fail
```
func (s *SortGeoDistance) DecodeValue(value []byte) any {
//
}

func (s *SortField) DecodeValue(value []byte) any {
   ///
}
```

search/sort.go
```diff
@@ -390,6 +400,25 @@ func (s *SortField) Value(i *DocumentMatch) string {
 	return iTerm
 }
 
+func (s *SortField) decodeValue(value string) string {
+	switch s.Type {
+	default:
```
CascadingRadium, line 405: default should be below all case

search/sort.go
```diff
@@ -757,6 +786,14 @@ func (s *SortGeoDistance) Reverse() {
 	s.Desc = !s.Desc
 }
 
+func decodeGeoValue(value string) string {
```
CascadingRadium, line 789: should be attached to the SortGEoDistance struct not standalone

func (s *SortGeoDistance) DecodeValue(value []byte) any { // }

CascadingRadium (approved):

CascadingRadium (commented):
update PR description and title to match what is done in the PR

CascadingRadium (approved):

PR #2204, opened 2025-06-17
WIP fast merge

Thejas-bhat:


index.go
```diff
@@ -396,3 +396,7 @@ type InsightsIndex interface {
 	// CentroidCardinalities returns the centroids (clusters) from IVF indexes ordered by data density.
 	CentroidCardinalities(field string, limit int, desceding bool) ([]index.CentroidCardinality, error)
 }
+type VectorIndex interface {
```
abhinavdangeti: Thought we agreed on `TrainableIndex` for this?

go.mod
```diff
@@ -43,3 +43,25 @@ require (
...
+
+replace github.com/blevesearch/zapx/v16 => /Users/thejas.orkombu/fts/blevesearch/zapx
+
+replace github.com/blevesearch/scorch_segment_api/v2 => /Users/thejas.orkombu/fts/blevesearch/scorch_segment_api
+
+replace github.com/blevesearch/go-faiss => /Users/thejas.orkombu/fts/blevesearch/go-faiss
+
+replace github.com/blevesearch/bleve_index_api => /Users/thejas.orkombu/fts/blevesearch/bleve_index_api
+
+replace github.com/blevesearch/sear => /Users/thejas.orkombu/fts/blevesearch/sear
```
abhinavdangeti: Why import sear here!? Also drop the replaces and let's adapt instead using the branches for your commits and run go-mod-tidy.

index_alias_impl.go
```diff
@@ -103,6 +103,24 @@ func (i *indexAliasImpl) IndexSynonym(id string, collection string, definition *
...
+
+	err := i.isAliasToSingleIndex()
+	if err != nil {
+		return err
+	}
+
+	if vi, ok := i.indexes[0].(VectorIndex); ok {
+		return vi.Train(batch)
+	}
+	return fmt.Errorf("not a vector index")
```
abhinavdangeti: Should we fail here or return `nil` as not supported over this index?

index_impl.go
```diff
@@ -369,6 +369,20 @@ func (i *indexImpl) IndexSynonym(id string, collection string, definition *Synon
...
+	defer i.mutex.RUnlock()
+
+	if !i.open {
+		return ErrorIndexClosed
+	}
+
+	if vi, ok := i.i.(index.TrainableIndex); ok {
+		return vi.Train(batch.internal)
+	}
+	return fmt.Errorf("not a vector index")
```
abhinavdangeti: Declare an `error` to re-use - if this is necessary to do.

Thejas-bhat: I feel like the error would be necessary in the sense that we'd be doing the wrong operation on an index and the user should be notified - i'll declare a const for it

index.go
```diff
@@ -396,7 +396,7 @@ type InsightsIndex interface {
 	// CentroidCardinalities returns the centroids (clusters) from IVF indexes ordered by data density.
 	CentroidCardinalities(field string, limit int, desceding bool) ([]index.CentroidCardinality, error)
 }
-type VectorIndex interface {
+type TrainableIndex interface {
```
abhinavdangeti, line 400: Line break before.

index/scorch/train_vector.go
```diff
@@ -0,0 +1,275 @@
...
+// this is not a routine that will be running throughout the lifetime of the index. It's purpose
+// is to only train the vector index before the data ingestion starts.
+func (t *vectorTrainer) trainLoop() {
+	defer func() {
+		t.parent.asyncTasks.Done()
+	}()
+	// initialize stuff
+	t.parent.segmentConfig[index.TrainedIndexCallback] = t.getCentroidIndex
+	path := filepath.Join(t.parent.path, index.TrainedIndexFileName)
+	for {
```
abhinavdangeti, line 82: Wouldn't we need a fast exit path in the event this is not necessary - no vector fields or fast merge not enabled?

index/scorch/train_vector.go
```diff
@@ -0,0 +1,275 @@
...
+type trainRequest struct {
+	sample   segment.Segment
+	vecCount int
+	ackCh    chan error
+}
+
+const IndexTrainedWithFastMerge = "fast-merge"
+
+func initTrainer(s *Scorch, config map[string]interface{}) *vectorTrainer {
+	if f, ok := config["vector_index_merge"]; ok {
```
abhinavdangeti: vector_index_fast_merge*

index/scorch/train_vector.go
```diff
@@ -0,0 +1,275 @@
...
+		feature, ok := f.(string)
+		if ok && feature == IndexTrainedWithFastMerge {
```
abhinavdangeti: Let's check for "vector_index_fast_merge":"true"

index/scorch/train_vector.go
Thejas-bhat: yeah we don't hit this path in that case, there will be no trainer loop running

PR #2207, opened 2025-07-17
MB-66396: Index APIs for `TermFrequencies` and `CentroidCardinalities`

abhinavdangeti:
Requires:
- https://github.com/blevesearch/faiss/pull/54
- https://github.com/blevesearch/go-faiss/pull/45
- https://github.com/blevesearch/bleve_index_api/pull/71
- https://github.com/blevesearch/scorch_segment_api/pull/65
- https://github.com/blevesearch/zapx/pull/338


index/scorch/snapshot_index_vr.go
```diff
@@ -163,3 +164,54 @@ func (i *IndexSnapshotVectorReader) Close() error {
 	// TODO Consider if any scope of recycling here.
 	return nil
 }
+
+func (i *IndexSnapshot) CentroidCardinalities(field string, limit int, descending bool) (
```
capemox, line 172: Because this function is in snapshot_index_vr.go, the method wouldn't be included in `IndexSnapshot` if the vectors flag is turned off. So it wouldn't be an implementation of `IndexInsightsReader` anymore. Calls to `TermFrequencies` in index_impl.go would then fail because the code checks if the interface is implemented for the reader. 

Basically, `TermFrequencies` would only work if it is compiled with vectors, is this intended?

abhinavdangeti: Ya it was intentional - because the insights project is meant for AI applications which will use this alongside vector data. This is just a start though, hoping to add more functions to the interface.

index/scorch/snapshot_index.go
```diff
@@ -1161,3 +1162,91 @@ func (is *IndexSnapshot) ThesaurusKeysRegexp(name string,
...
+		})
+	} else {
+		sort.Slice(termFreqList, func(i, j int) bool {
+			if termFreqList[i].Frequency == termFreqList[j].Frequency {
+				// If frequencies are equal, sort by term lexicographically
+				return strings.Compare(termFreqList[i].Term, termFreqList[j].Term) < 0
+			}
+			return termFreqList[i].Frequency < termFreqList[j].Frequency
+		})
+	}
```
CascadingRadium:
```
sort.Slice(termFreqList, func(i, j int) bool {
if termFreqList[i].Count == termFreqList[j].Count {
      // If frequencies are equal, sort by term lexicographically
      return strings.Compare(termFreqList[i].Term, termFreqList[j].Term) < 0
}
if descending {
      return termFreqList[i].Count > termFreqList[j].Count
}
      return termFreqList[i].Count < termFreqList[j].Count
})
```

index/scorch/snapshot_index.go
```diff
@@ -1161,3 +1162,91 @@ func (is *IndexSnapshot) ThesaurusKeysRegexp(name string,
...
 	atomic.AddUint64(&is.parent.stats.TotSynonymSearches, delta)
 }
+
+// TermFrequencies returns the top N terms ordered by the frequencies
+// for a given field across all segments in the index snapshot.
+func (is *IndexSnapshot) TermFrequencies(field string, limit int, descending bool) (
+	termsFreqs []index.TermFreq, err error) {
+	if len(is.segment) == 0 {
+		return nil, fmt.Errorf("no segments available")
+	}
```
CascadingRadium, line 1245: error message not required to be propagated, as empty index means theres no `TermFreqs` so nil, nil is enough

index/scorch/snapshot_index.go
```diff
@@ -1161,3 +1162,91 @@ func (is *IndexSnapshot) ThesaurusKeysRegexp(name string,
...
+			pl, err := dict.PostingsList([]byte(next.Term), segment.deleted, nil)
+			if err != nil {
+				continue // Skip this term if we can't get postings
+			}
+
+			// Aggregate frequency across segments
+			termStr := string(next.Term)
+			termFreqs[termStr] += pl.Count()
+		}
+	}
```
CascadingRadium: not required, the Implementation for this is already present as part of `FieldDict`

```
	// Use FieldDict which aggregates term frequencies across all segments
	fieldDict, err := is.FieldDict(field)
	if err != nil {
		return nil, fmt.Errorf("failed to get field dictionary for field %s: %v", field, err)
	}
	defer fieldDict.Close()

	// Preallocate slice with capacity equal to the number of unique terms
	// in the field dictionary
	termFreqList := make([]index.DictEntry, 0, fieldDict.Cardinality())

	// Iterate through all terms using FieldDict
	for {
		dictEntry, err := fieldDict.Next()
		if err != nil {
			return nil, fmt.Errorf("error iterating field dictionary: %v", err)
		}
		if dictEntry == nil {
			break // End of terms
		}

		termFreqList = append(termFreqList, dictEntry)
	}
```

index/scorch/snapshot_index.go
```diff
@@ -1161,3 +1162,91 @@ func (is *IndexSnapshot) ThesaurusKeysRegexp(name string,
...
+
+	if len(termFreqs) == 0 {
+		return nil, fmt.Errorf("no terms found for field %s", field)
+	}
```
CascadingRadium: can just return `nil, nil`
here - no terms found => not error

index/scorch/snapshot_index.go
```diff
@@ -1161,3 +1162,91 @@ func (is *IndexSnapshot) ThesaurusKeysRegexp(name string,
...
+
+	termFreqList := make([]index.TermFreq, 0, len(termFreqs))
+	for termStr, freq := range termFreqs {
+		termFreqList = append(termFreqList, index.TermFreq{
+			Term:      termStr,
+			Frequency: freq,
+		})
+	}
```
CascadingRadium: not required as per above

index_alias_impl.go
```diff
@@ -1065,3 +1067,159 @@ func (f *indexAliasImplFieldDict) Close() error {
...
+			rv, _ = idx.TermFrequencies(field, limit*5, descending)
+		}
+		asyncResults <- rv
+		waitGroup.Done()
+	}
+
+	waitGroup.Add(len(i.indexes))
+	for _, in := range i.indexes {
+		go searchChildIndex(in, field, limit, descending)
+	}
```
CascadingRadium, line 1182: This will not work over the GRPC layer -> if the child indices are grpc clients over the distributed search nodes.

Better to execute this as part of a presearch, Deploy a real search request similar to how FieldCardinalities is gathered across a global BM25 search - @Thejas-bhat can add more here

index_alias_impl.go
```diff
@@ -1065,3 +1067,159 @@ func (f *indexAliasImplFieldDict) Close() error {
...
+			rv, _ = idx.CentroidCardinalities(field, limit, descending)
+		}
+		asyncResults <- rv
+		waitGroup.Done()
+	}
+
+	waitGroup.Add(len(i.indexes))
+	for _, in := range i.indexes {
+		go searchChildIndex(in, field, limit, descending)
+	}
```
CascadingRadium, line 1267: same point as above, `in Index` will be a gRPC client that will not implement or gather the metadata from the gRPC server, need to execute a presearch only query that gathers the metadata -> Look into FieldCardinalities

index.go
```diff
@@ -386,3 +386,11 @@ type SynonymIndex interface {
 	// IndexSynonym indexes a synonym definition, with the specified id and belonging to the specified collection.
 	IndexSynonym(id string, collection string, definition *SynonymDefinition) error
 }
+
+type InsightsIndex interface {
+	Index
+	// TermFrequencies returns the tokens ordered by frequencies for the field index.
+	TermFrequencies(field string, limit int, descending bool) ([]index.TermFreq, error)
```
CascadingRadium, line 395: []*index.DictEntry

search_knn_test.go
```diff
@@ -1701,3 +1703,163 @@ func TestNumVecsStat(t *testing.T) {
 		}
 	}
 }
+
+func TestIndexInsightsTermFrequencies(t *testing.T) {
```
CascadingRadium, line 1915: `TestIndexInsightsTermFrequencies` should be in `search_test.go`

index/scorch/snapshot_index_vr.go
```diff
@@ -163,3 +164,54 @@ func (i *IndexSnapshotVectorReader) Close() error {
...
+	centroids []index.CentroidCardinality, err error) {
+	if len(i.segment) == 0 {
+		return nil, fmt.Errorf("no segments available")
```
CascadingRadium: should return nil, nil

index/scorch/snapshot_index.go
```diff
@@ -1161,3 +1162,91 @@ func (is *IndexSnapshot) ThesaurusKeysRegexp(name string,
...
+			}
+			return termFreqList[i].Frequency < termFreqList[j].Frequency
+		})
+	}
+
+	if limit > len(termFreqList) {
+		limit = len(termFreqList)
+	}
+
+	return termFreqList[:limit], nil
```
CascadingRadium: avoid unnecessary resclicing

```
if limit >= len(termFreqList) {
    return termFreqList, nil
}
return termFreqList[:limit], nil
```

index/scorch/snapshot_index_vr.go
```diff
@@ -163,3 +164,54 @@ func (i *IndexSnapshotVectorReader) Close() error {
...
+			centroidCardinalities = append(centroidCardinalities, rvCentroids...)
+			if descending {
+				sort.Slice(centroidCardinalities, func(i, j int) bool {
+					return centroidCardinalities[i].Cardinality > centroidCardinalities[j].Cardinality
+				})
+			} else {
+				sort.Slice(centroidCardinalities, func(i, j int) bool {
+					return centroidCardinalities[i].Cardinality < centroidCardinalities[j].Cardinality
+				})
+			}
```
CascadingRadium: this is causing sort ops for every segment - suboptimal

```
func (i *IndexSnapshot) CentroidCardinalities(field string, limit int, descending bool) (
	centroids []index.CentroidCardinality, err error) {
	if limit <= 0 {
		return nil, fmt.Errorf("limit must be positive")
	}

	// Collect all centroids from all vector segments
	allCentroids := make([]index.CentroidCardinality, 0)

	for _, segment := range i.segment {
		if sv, ok := segment.segment.(segment_api.VectorSegment); ok {
			vecIndex, err := sv.InterpretVectorIndex(field,
				false /* does not require filtering */, segment.deleted)
			if err != nil {
				return nil, fmt.Errorf("failed to interpret vector index for field %s in segment: %v", field, err)
			}

			// Pass a large limit to get all centroids from this segment since we'll sort globally
			centroidCardinalities, err := vecIndex.ObtainKCentroidCardinalitiesFromIVFIndex(limit*len(i.segment), descending)
			if err != nil {
				return nil, fmt.Errorf("failed to obtain top k centroid cardinalities for field %s in segment: %v", field, err)
			}

			if len(centroidCardinalities) > 0 {
				allCentroids = append(allCentroids, centroidCardinalities...)
			}
		}
	}

	// If no centroids found, return empty result
	if len(allCentroids) == 0 {
		return []index.CentroidCardinality{}, nil
	}

	// Sort all centroids once at the end
	sort.Slice(allCentroids, func(i, j int) bool {
		if descending {
			return allCentroids[i].Cardinality > allCentroids[j].Cardinality
		}
		return allCentroids[i].Cardinality < allCentroids[j].Cardinality
	})

	// Return top limit results
	if limit >= len(allCentroids) {
		return allCentroids, nil
	}

	return allCentroids[:limit], nil
}
```

search_knn_test.go
abhinavdangeti: Insights are currently not available without `vectors` - but I suppose this is worth revisiting.

index_alias_impl.go
abhinavdangeti: We'll not be supporting this over gRPC - cbft will use rest clients to do the aggregation.

index_alias_impl.go
abhinavdangeti: ditto

index/scorch/snapshot_index.go
abhinavdangeti: Done.

PR #2208, opened 2025-07-18
MB-67212: Supporting search_after/search_before for deep pagination over non textual content

capemox:
Add search after functionality for numeric, datetime, and geo fields in TopNCollector

- Introduced new test cases for searching after numeric, datetime, and geo fields.
- Implemented encoding for pagination on these fields to enhance search capabilities.
- Updated the createSearchAfterDocument function to support encoded sort values.
- Ensured proper handling of search results in the new test cases.

This enhances the search functionality by allowing users to paginate through results based on specific criteria.

### DIFF
--- a/search/collector/topn.go
+++ b/search/collector/topn.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/blevesearch/bleve/v2/numeric"
 	"github.com/blevesearch/bleve/v2/search"
 	"github.com/blevesearch/bleve/v2/size"
 	index "github.com/blevesearch/bleve_index_api"
@@ -117,9 +118,15 @@ func newTopNCollector(size int, skip int, sort search.SortOrder) *TopNCollector 
 	return hc
 }
 
+// Creates a dummy document to compare with for pagination.
 func createSearchAfterDocument(sort search.SortOrder, after []string) *search.DocumentMatch {
+	encodedAfter := make([]string, len(after))
+	for i, ss := range sort {
+		encodedAfter[i] = encodeSearchAfter(ss, after[i])
+	}
+
 	rv := &search.DocumentMatch{
-		Sort: after,
+		Sort: encodedAfter,
 	}
 	for pos, ss := range sort {
 		if ss.RequiresDocID() {
@@ -132,6 +139,52 @@ func createSearchAfterDocument(sort search.SortOrder, after []string) *search.Do
 		}
 	}
 	return rv
+}
+
+// encodeSearchAfter applies prefix-coding to SearchAfter
+// if required to enable pagination on numeric, datetime,
+// and geo fields
+func encodeSearchAfter(ss search.SearchSort, after string) string {
+	switch ss := ss.(type) {
+	case *search.SortGeoDistance:
+		dist, err := strconv.ParseFloat(after, 64)
+		if err != nil {
+			return after
+		}
+
+		distInt64 := numeric.Float64ToInt64(dist)
+		return string(numeric.MustNewPrefixCodedInt64(distInt64, 0))
+	case *search.SortField:
+		switch ss.Type {
+		case search.SortFieldAsNumber:
+			f64, err := strconv.ParseFloat(after, 64)
+			if err != nil {
+				return after
+			}
+
+			i64 := numeric.Float64ToInt64(f64)
+			return string(numeric.MustNewPrefixCodedInt64(i64, 0))
+		case search.SortFieldAsDate:
+			t, err := time.Parse("2006-01-02 15:04:05.999999999 +0000 UTC", after)
+			if err != nil {
+				return after
+			}
+
+			i64 := t.UnixNano()
+			return string(numeric.MustNewPrefixCodedInt64(i64, 0))
+		default:
+			// For SortFieldAsString and SortFieldAuto
+			// NOTE: SortFieldAuto is used if you set Sort with a string
+			// or if the type of the field is not set in the object
+			// in the Sort slice. We cannot perform type inference in
+			// this case, so we return the original string, even if
+			// its actually numeric or date.
+			return after
+		}
+	default:
+		// For SortDocID and SortScore
+		return after
+	}
 }
 
 // Filter document matches based on the SearchAfter field in the SearchRequest.
--- a/search/collector/topn_test.go
+++ b/search/collector/topn_test.go
@@ -17,11 +17,16 @@ package collector
 import (
 	"bytes"
 	"context"
+	"strconv"
 	"testing"
+	"time"
 
+	"github.com/blevesearch/bleve/v2/document"
+	"github.com/blevesearch/bleve/v2/geo"
 	"github.com/blevesearch/bleve/v2/index/scorch"
 	"github.com/blevesearch/bleve/v2/search"
 	"github.com/blevesearch/bleve/v2/search/facet"
+	"github.com/blevesearch/bleve/v2/search/searcher"
 	index "github.com/blevesearch/bleve_index_api"
 )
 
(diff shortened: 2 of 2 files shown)


search/collector/topn.go
```diff
@@ -134,6 +141,52 @@ func createSearchAfterDocument(sort search.SortOrder, after []string) *search.Do
...
 
+// encodeSearchAfter applies prefix-coding to SearchAfter
+// if required to enable pagination on numeric, datetime,
+// and geo fields
+func encodeSearchAfter(ss search.SearchSort, after string) string {
+	switch ss := ss.(type) {
+	case *search.SortGeoDistance:
+		dist, err := strconv.ParseFloat(after, 64)
+		if err != nil {
+			return after
```
CascadingRadium: if the user has explicitly mentioned that the `ith` sort object is a `GeoDistance sort`, then it is expected for the query to have a valid numeric string in the `ith` array position of `searchAfter`. 
So i think the error must not be simply ignored and must actually be returned back to the user by failing the query. 
Add some code to validate the SearchAfter / SearchBefore encodings when validating the searchRequest -> https://github.com/blevesearch/bleve/blob/487fa5f8891dd731978e4cfb74e3a0aa79d0e894/search.go#L285
The validation check will just check for error result that could be thrown and not actually encode anything; the real encoding is done correctly.

search/collector/topn.go
```diff
@@ -134,6 +141,52 @@ func createSearchAfterDocument(sort search.SortOrder, after []string) *search.Do
...
+		case search.SortFieldAsNumber:
+			f64, err := strconv.ParseFloat(after, 64)
+			if err != nil {
+				return after
+			}
+
+			i64 := numeric.Float64ToInt64(f64)
+			return string(numeric.MustNewPrefixCodedInt64(i64, 0))
+		case search.SortFieldAsDate:
+			t, err := time.Parse("2006-01-02 15:04:05.999999999 +0000 UTC", after)
```
CascadingRadium: please use `time.RFC3339Nano` inbuilt golang constant as the pattern layout to use for parsing, as it is standard layout

search/collector/topn.go
```diff
@@ -134,6 +141,52 @@ func createSearchAfterDocument(sort search.SortOrder, after []string) *search.Do
...
+	case *search.SortField:
+		switch ss.Type {
+		case search.SortFieldAsNumber:
+			f64, err := strconv.ParseFloat(after, 64)
+			if err != nil {
+				return after
+			}
+
+			i64 := numeric.Float64ToInt64(f64)
+			return string(numeric.MustNewPrefixCodedInt64(i64, 0))
```
CascadingRadium: this block of code is repeated for `SortGeoDistance`
consider an inline helper method

```
func encodeSearchAfter(ss search.SearchSort, after string) string {
	// Helper to encode float values (e.g., for geo or numeric sorts)
	encodeFloat := func(s string) string {
		f64, err := strconv.ParseFloat(s, 64)
		if err != nil {
			return s
		}
		i64 := numeric.Float64ToInt64(f64)
		return string(numeric.MustNewPrefixCodedInt64(i64, 0))
	}
	// Helper to encode date values using RFC3339Nano
	encodeDate := func(s string) string {
		t, err := time.Parse(time.RFC3339Nano, s)
		if err != nil {
			return s
		}
		i64 := t.UnixNano()
		return string(numeric.MustNewPrefixCodedInt64(i64, 0))
	}
	switch ss := ss.(type) {
	case *search.SortGeoDistance:
		return encodeFloat(after)
	case *search.SortField:
		switch ss.Type {
		case search.SortFieldAsNumber:
			return encodeFloat(after)
		case search.SortFieldAsDate:
			return encodeDate(after)
		default:
			// For SortFieldAsString and SortFieldAuto
			return after
		}
	default:
		// For SortDocID and SortScore
		return after
	}
}
```

search/collector/topn.go
capemox: Yes this is a big issue, will fix

search/collector/topn.go
capemox: https://github.com/blevesearch/bleve/blob/487fa5f8891dd731978e4cfb74e3a0aa79d0e894/search/sort.go#L406-L411
The `decoded_sort` string value follows this format, which does not follow any RFC string. This is because the decoded date is formatted as per time's String() function, which just formats it using the string in my code.

CascadingRadium: change L411 to:
```
return time.Unix(0, i64).UTC().Format(time.RFC3339Nano)
```
this code was added recently and it seems to have slipped past in code review. The decoded sort should be in a standard notation

search.go
```diff
@@ -301,6 +302,11 @@ func (r *SearchRequest) Validate() error {
 		if len(r.SearchAfter) != len(r.Sort) {
 			return fmt.Errorf("search after must have same size as sort order")
 		}
+		
+		err := validatePagination(r.Sort, r.SearchAfter, "search after")
```
CascadingRadium: `r.Sort` and (`r.SearchAfter` or `r.SearchBefore`) are already part of the `r *SearchRequest` object. So the signature can be simplified to 
```
err := r.validateSort()
if err != nil {
	return err
}
```
Can use the values directly

CascadingRadium (commented):
small refactor

search.go
capemox: Made the changes

search.go
```diff
@@ -311,13 +312,54 @@ func (r *SearchRequest) Validate() error {
...
+		pagination = r.SearchBefore
+		afterOrBefore = "search before"
+	} else {
+		return nil
+	}
+
+	for i := range pagination {
+		switch ss := r.Sort[i].(type) {
+		case *search.SortGeoDistance:
+			valid, _ := numeric.ValidPrefixCodedTerm(pagination[i])
```
CascadingRadium: sorry i had not seen this earlier, but the expectation is that the user will enter the decoded sort value here, right? `pagination[i]` is a float number that will get encoded to prefixCodedTerm later on. SearchRequest validation takes place before Request Execution. The validation should check if a valid float exists for SortGeoDistance as well as for SortFieldAsNumber.
```
		_, err := strconv.ParseFloat(pagination[i], 64)
		if err != nil {
			return err
		}
```
for SortFieldAsDate check for valid date
```
		_, err := time.Parse(time.RFC3339Nano, pagination[i])
		if err != nil {
			return err
		}
```
that way any error can be ignored safely when the real encoding takes place in the tail end of search request execution (in `encodeSearchAfter`).

capemox: Yes this makes sense. Since I will end up doing a bulk of the work in `validatePagination` itself, should I shift the logic of encoding the `search after`/`search before` fields inside `validatePagination`? This way I can return any error and reduce redundant checking as well.

CascadingRadium: no, a `validate` should not modify the data, as a Bleve user has a choice not to validate. From cbft side we always validate pre-execution. Even though redundant work is done here, the validation is in itself meant to be an extra overhead.

CascadingRadium (approved):

search/collector/topn.go
```diff
@@ -117,9 +118,15 @@ func newTopNCollector(size int, skip int, sort search.SortOrder) *TopNCollector
 	return hc
 }
 
+// Creates a dummy document to compare with for pagination.
 func createSearchAfterDocument(sort search.SortOrder, after []string) *search.DocumentMatch {
```
abhinavdangeti, line 122: I forget this, but are there no such utilities for `searchBefore`?

capemox: Internally, `SearchBefore` is converted to a `SearchAfter` problem by reversing the sort order, so we don't need to handle both cases separately
https://github.com/blevesearch/bleve/blob/487fa5f8891dd731978e4cfb74e3a0aa79d0e894/index_impl.go#L584-L589

PR #2209, opened 2025-07-21
MB-65860: Introducing support for fileIO Callbacks

Likith101:


index/scorch/merge.go
```diff
@@ -96,10 +97,19 @@ OUTER:
...
-
+				var err error
 				// lets get started
-				err := s.planMergeAtSnapshot(ctrlMsg.ctx, ctrlMsg.options,
-					ourSnapshot)
+				if ctrlMsg.plan == nil {
+					err = s.planMergeAtSnapshot(ctrlMsg.ctx, ctrlMsg.options,
+						ourSnapshot)
+				} else {
+					cw := newCloseChWrapper(s.closeCh, ctrlMsg.ctx)
```
abhinavdangeti: Can you add some commentary around why this change here.

index/scorch/merge.go
```diff
@@ -425,7 +447,8 @@ func (s *Scorch) planMergeAtSnapshot(ctx context.Context,
 		select {
 		case <-s.closeCh:
 			_ = seg.Close()
-			return segment.ErrClosed
+			err = segment.ErrClosed
+			return err
```
abhinavdangeti, line 460: Unnecessary change here.

index/scorch/persister.go
```diff
@@ -1068,6 +1107,136 @@ func (s *Scorch) loadSegment(segmentBucket *bolt.Bucket) (*SegmentSnapshot, erro
 	return rv, nil
 }
 
+// identify all the file callback ids that are in use by boltdb
```
abhinavdangeti: Let's keep the name consistent - if we're calling them file callbacks .. let's rename the file you've added in `util` to `file_callbacks.go`.

Also this commend needs to change to "file callback writer IDs" and update the method name to boltCallbackWriterIDsInUse*.

Also `WriterId` is pretty vague for a bleve user - some thought into what would be better here would be a good move.

index/scorch/persister.go
```diff
@@ -1068,6 +1107,136 @@ func (s *Scorch) loadSegment(segmentBucket *bolt.Bucket) (*SegmentSnapshot, erro
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return idMap, nil
+}
+
+// remove all content in boltdb associated with the callback ids
+// and process the data using the latest writer
```
abhinavdangeti: Ditto about naming consistency here as well.

index/scorch/scorch.go
```diff
@@ -1066,13 +1121,206 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+					}
 				}
 			}
 		}
 		return nil
 	})
 }
+
+// returns the set of writer ids in use by both the segments and boltdb
+func (s *Scorch) WriterIdsInUse() (map[string]struct{}, error) {
```
abhinavdangeti: Ditto.

index.go
```diff
@@ -389,6 +389,12 @@ type SynonymIndex interface {
 	IndexSynonym(id string, collection string, definition *SynonymDefinition) error
 }
 
+type CustomizableIndex interface {
```
abhinavdangeti: Related to comment here - https://github.com/blevesearch/bleve_index_api/pull/72

Let's drop the bleve_index_api commit - and just use a declaration for this interface here. Let's refactor to IndexWithFileWriterCallbacks* or simply IndexWithIOCallbacks

Likith101: So remove the "Index" from CustomizableIndex so that it is applicable to both Scorch and IndexImpl?

abhinavdangeti: Not sure I follow.

index/scorch/merge.go
Likith101: This is so that the error is transmitted to the done channel properly

index_impl.go
```diff
@@ -46,14 +46,16 @@ import (
...
-	stats *IndexStat
+	path       string
+	name       string
+	meta       *indexMeta
+	i          index.Index
+	m          mapping.IndexMapping
+	mutex      sync.RWMutex
+	open       bool
+	stats      *IndexStat
+	fileWriter *util.FileWriter
```
Thejas-bhat: instead of maintaining a pointer to the struct, wouldn't keeping these writer and reader as an interface "FileWriter/Reader" defining some APIs be better? and the New() would return the pointer ofc but its hidden behind the interface. I think having a pointer right at the top level feels a bit wrong

index/scorch/scorch.go
```diff
@@ -1024,26 +1027,78 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+								return err
+							}
+						}
+					}
 				} else if kk[0] != util.BoltMetaDataKey[0] {
 					segmentBucket := snapshot.Bucket(kk)
 					if segmentBucket == nil {
 						return fmt.Errorf("segment key, but bucket missing %x", kk)
 					}
+
```
Thejas-bhat: remove these kind of extra line breaks please

index/scorch/scorch.go
```diff
@@ -1066,13 +1121,222 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
 			}
 		}
 		return nil
 	})
 }
+
+// returns the set of file callback writer ids in use by all of the segments and boltdb
+func (s *Scorch) FileWriterIDsInUse() (map[string]struct{}, error) {
+	s.rootLock.RLock()
+
```
Thejas-bhat: remove line break

index/scorch/scorch.go
```diff
@@ -1066,13 +1121,222 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+	s.rootLock.RUnlock()
+
+	return keyMap, nil
+}
+
+// removes all file callback writer ids in use from all of the segments and boltdb
+// boltdb is updated with the latest callback writer while segments are force
+// merged blockingly until snapshot is persisted with the latest callback writer
+func (s *Scorch) DropFileWriterIDs(ids map[string]struct{}) error {
+
```
Thejas-bhat: remove line break

index/scorch/scorch.go
```diff
@@ -1066,13 +1121,222 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+	err := s.removeBoltFileWriterIDs(ids)
+	if err != nil {
+		return err
+	}
+
+	s.rootLock.Lock()
+
```
Thejas-bhat, line 1348: remove line break

index/scorch/scorch.go
```diff
@@ -1066,13 +1121,222 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+		err = s.waitTillFileCleanup(filePaths)
+		if err != nil {
+			return err
+		}
+
+		// reset rollback snapshot retention
+		s.rootLock.Lock()
+		s.numSnapshotsToKeep = prevNumSnapshotsToKeep
+	}
+
```
Thejas-bhat, line 1408: remove the line break here and add one below the unlock

util/file_callbacks.go
```diff
@@ -0,0 +1,120 @@
...
+	zapv17.ReaderHook = ReaderHook
+}
+
+// FileWriter and FileReader are wrappers around the callback functions provided
+// by the user. They provide a convenient way to apply the callbacks to data
+// being written to or read from a file. They also store the id the callbacks,
+// which can be useful for managing state across multiple reads and writes.
+type FileWriter struct {
+	processor func(data []byte) []byte
+	id        string
```
Thejas-bhat, line 52: i think accepted practice is defining the vars and then the funcs within a struct

index_impl.go
```diff
@@ -479,6 +494,52 @@ func (i *indexImpl) Search(req *SearchRequest) (sr *SearchResult, err error) {
 	return i.SearchInContext(context.Background(), req)
 }
 
+// returns the set of file callback writer ids in use by the index
+func (i *indexImpl) FileWriterIDsInUse() (map[string]struct{}, error) {
+
```
Thejas-bhat: remove line breaks

index_meta.go
```diff
@@ -40,27 +41,67 @@ func newIndexMeta(indexType string, storage string, config map[string]interface{
...
+
 	if im.IndexType == "" {
 		im.IndexType = upsidedown.Name
 	}
-	return &im, nil
+	return &im, fileReader, nil
 }
 
-func (i *indexMeta) Save(path string) (err error) {
+func (i *indexMeta) Save(path string, writer *util.FileWriter) (err error) {
```
Thejas-bhat: i see there's a lot of NewFileWriter() invocations done for every Save() operation which corresponds to an instance of indexMeta. i feel like this generates a lot of redundant code, instead why not just register the util.FileWriter in the indexMeta struct as well? the code will become a lot cleaner imo

index_meta.go
```diff
@@ -27,17 +28,30 @@ import (
 const metaFilename = "index_meta.json"
 
 type indexMeta struct {
-	Storage   string                 `json:"storage"`
-	IndexType string                 `json:"index_type"`
-	Config    map[string]interface{} `json:"config,omitempty"`
+	Storage    string                 `json:"storage"`
+	IndexType  string                 `json:"index_type"`
+	Config     map[string]interface{} `json:"config,omitempty"`
+	fileWriter *util.FileWriter
```
Thejas-bhat: i still think you should keep a fileWriter interface type and use it here

Likith101: I don't really see the need to do this. It isn't an exported field for one, and I do not see a future where we have more implementations for the writers and readers.

search/searcher/base_test.go
```diff
@@ -67,8 +67,16 @@ func initTwoDocs(twoDocIndex index.Index) {
 		panic(err)
 	}
 	batch := index.NewBatch()
+
+	// Make a copy of the documents to prevent modification of the
+	// original slice across multiple tests
+	docs := []document.Document{}
 	for _, doc := range twoDocIndexDocs {
-		batch.Update(doc)
+		docs = append(docs, *doc)
```
Thejas-bhat: are these changes necessary? like which test uses this?

index/scorch/scorch.go
```diff
@@ -1066,13 +1120,218 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+	return nil
+}
+
+// waitTillFileCleanup blocks until the given files are cleaned up by the persister or merger or
+// returns an error after a timeout. It does so by checking the index directory every 5 seconds
+// for the presence of the given files, and returns once they are no longer present.
+func (s *Scorch) waitTillFileCleanup(filePaths []string) error {
+	ticker := time.NewTicker(5 * time.Second)
```
Thejas-bhat, line 1464: is this something that'll be made configurable?

util/file_callbacks.go
```diff
@@ -0,0 +1,120 @@
...
+	processor func(data []byte) ([]byte, error)
+}
+
+func NewFileReader(id string, context []byte) (*FileReader, error) {
+	rv := &FileReader{
+		id: id,
+	}
+
+	if ReaderHook != nil {
+		var err error
```
Thejas-bhat, line 97: maybe this extra var can be dropped

index_impl.go
```diff
@@ -479,6 +482,50 @@ func (i *indexImpl) Search(req *SearchRequest) (sr *SearchResult, err error) {
...
+	if cidx, ok := i.i.(IndexWithCallbacks); ok {
+		cIds, err := cidx.FileWriterIDsInUse()
+		if err != nil {
+			return nil, err
+		}
+		for k := range cIds {
+			ids[k] = struct{}{}
+		}
+	} else {
+		ids[""] = struct{}{}
```
Thejas-bhat: can you please explain the purpose of this else block? wouldn't it be used some reserved string for this operation marking the underlying index isn't supporting?

index_meta.go
Thejas-bhat: true i understand that you don't see a future implementation, but I think across packages its better to use an interface to component and it uses the things in a more modular fashion right?

util/file_callbacks.go
```diff
@@ -0,0 +1,120 @@
...
+var ReaderHook func(id string, context []byte) (
+	func(data []byte) ([]byte, error), error)
+
+// Register callbacks with zapv17. This should be called in init() of the package
+// that defines the callbacks, after the callbacks have been set. This allows
+// zapv17 to use the callbacks when creating file readers and writers.
+// This should be called before any indexes are opened or created that will
+// use the callbacks.
+func RegisterFileCallbacks() {
+	zapv17.WriterHook = WriterHook
```
Thejas-bhat: Since the exact same callbacks are used in both the zapx and bleve side, I think you should just do this init/registry on the bleve_index_api package and in both zap and bleve packages invoke the set hooks. this would avoid redundant setting of the callback values. so in the places where the WriterHook() is being invoked, you can do index.WriterHook().

index/scorch/scorch.go
```diff
@@ -1066,13 +1120,218 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+	atomic.AddUint64(&s.stats.TotFileMergePlan, 1)
+
+	// create a default merge plan incase some of the segments can actually be merged
+	mergePlan, err := mergeplan.Plan(segsToCompact, mergePlannerOptions)
+	if err != nil {
+		atomic.AddUint64(&s.stats.TotFileMergePlanErr, 1)
+		return fmt.Errorf("merge plan creation err: %v", err)
+	}
+
+	segDictionary := make(map[uint64]bool)
```
Thejas-bhat: since this is like a set, wouldnt it better to use the `map[uint64]struct{}` way of doing things?

index/scorch/scorch.go
```diff
@@ -1066,13 +1120,218 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+	if err != nil {
+		return err
+	}
+	s.rootLock.Lock()
+
+	// tag all segments with the callback ids for removal for
+	// force merging.
+	segsToCompact := make([]mergeplan.Segment, 0)
+	filePaths := make([]string, 0)
+	for _, segmentSnapShot := range s.root.segment {
```
Thejas-bhat: nit: alternate naming like seg/snapshot/ss perhaps?

search/searcher/base_test.go
Likith101: At one point I saw test failures and I had debugged to be caused by this. But now, i do not see it. Will revert it

index/scorch/scorch.go
Likith101: We are also using the set to track if the segments are part of the default merge plan and if not, then we add it as a separate task. So bool is needed here for code simplicity

index/scorch/scorch.go
Likith101: For now, no. Ideally, this is something that will happen regardless of the wait here. The main goal is to safely block until the persister has done its job. So, I do not see a value in making this configurable. 

Additionally, incase the persister has not done this job, we will eventually reach this exact same function to wait without redoing all of the work.

util/file_callbacks.go
Likith101: Line 102 requires this var to be declared earlier.

index_impl.go
Likith101: The empty string indicates unencrypted data. If we encounter an underlying index which does not implement IndexWithCallbacks, we are assuming here that it is using unencrypted data.

index_meta.go
Likith101: I still think this is a bit redundant. I will let the others weigh in here before making this change

util/file_callbacks.go
Likith101: Makes sense, I'll move these to bleve_index_api

index_impl.go
abhinavdangeti: @Likith101 Always allow the author of the comment to mark a conversation resolved - I agree w/ Thejas on using a reserved key here - for better readability.

index_meta.go
abhinavdangeti: On further though maybe interfaces is a good idea - for a future where users can add their own reader/writer hooks?

index/scorch/merge.go
```diff
@@ -96,10 +97,23 @@ OUTER:
...
+				var err error
 				// lets get started
-				err := s.planMergeAtSnapshot(ctrlMsg.ctx, ctrlMsg.options,
-					ourSnapshot)
+				// if there is no plan, then we are doing a regular merge at
+				// napshot based on the current state of the index. If there is
+				// a plan, then we are doing a force merge based on the segments
+				// specified in the plan.
+				if ctrlMsg.plan == nil {
+					err = s.planMergeAtSnapshot(ctrlMsg.ctx, ctrlMsg.options,
```
Thejas-bhat: i feel like this can be simplified into a single 

```
err = s.planMergeAtSnapshot(ctrlMsg, ourSnapshot)
```

and within the func, you can check for the ctrlMsg.plan and if its nil, generate the plan using the current snapshot and its segment (the current logic) otherwise just continue to execute the resultPlan.tasks. 

```
func (s *Scorch) planMergeAtSnapshot(ctrlMsg, snapshot) {
   resultMergePlan = ctrl.plan
   if resultMergePlan == nil {
      resultMergePlan <- mergeplan.Plan(persistedSnapshots, ctrl.options)
   }
   
    cw := newCloseChWrapper(s.closeCh, ctx)
	defer cw.close()

	go cw.listen()
    for _, task = resultMergePlan.tasks {
       // start performing merge on the task
    }
}
```

index/scorch/merge.go
```diff
@@ -372,7 +397,8 @@ func (s *Scorch) planMergeAtSnapshot(ctx context.Context,
 
 			atomic.AddUint64(&s.stats.TotFileMergeZapBeg, 1)
 			prevBytesReadTotal := cumulateBytesRead(segmentsToMerge)
-			newDocNums, _, err := s.segPlugin.MergeUsing(segmentsToMerge, docsToDrop, path,
+			var newDocNums [][]uint64
```
Thejas-bhat, line 405: why is this a new var? please check if this can be avoided by the above refactor suggestion.

edit: maybe this is needed since you're using the doneCh in ctx

index/scorch/scorch.go
```diff
@@ -1150,17 +1153,63 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
 				continue
 			}
 			snapshot := snapshots.Bucket(k)
+			metaBucket := snapshot.Bucket(util.BoltMetaDataKey)
```
Thejas-bhat, line 1184: since it looks like every bolt write/read is something now subjected to the file writer/reader operation, why not rootBolt be a struct that wraps the boltDB within it and it implements the exact same methods of bolt APIs. This struct will also have the writer and reader registered in it, so when you do any bolt.API() call it'll do an addtional filewriter/reader operation - in my opinion this way we will have much better inheritance and encapsulation of the writer and reader operations. Similar to how index_meta encapsulates the file writer and reader

@Likith101, @abhinavdangeti please let me know if this is something you guys agree and are open to implementing this

search/searcher/base_test.go
```diff
@@ -67,6 +67,7 @@ func initTwoDocs(twoDocIndex index.Index) {
 		panic(err)
 	}
 	batch := index.NewBatch()
+
```
Thejas-bhat: remove extra line

go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/blevesearch/bleve/v2
 
-go 1.24.0
+go 1.25.0
```
Thejas-bhat, line 3: @abhinavdangeti since the go.mod is being updated, do you think its better if we have the latest go version (1.26) instead of 1.25?

index_impl.go
```diff
@@ -493,6 +496,56 @@ func (i *indexImpl) Search(req *SearchRequest) (sr *SearchResult, err error) {
 	return i.SearchInContext(context.Background(), req)
 }
 
+// returns the set of file callback writer ids in use by the index
+func (i *indexImpl) FileWriterIDsInUse() (map[string]struct{}, error) {
+	ids := map[string]struct{}{}
+	ids[i.meta.fileReader.Id()] = struct{}{}
```
Thejas-bhat: nit: init with the meta filereader info

`ids := map[string]struct{}{i.meta.fileReader.Id(): {}}`

index/scorch/merge.go
Likith101: Yes, I believe this is needed because of that doneCh

index/scorch/scorch.go
Likith101: I believe this is something we can look into. I do agree with all of the positives, but it will be a significant effort to implement this currently

go.mod
capemox: +1, the new GC is great and the lower CGO call overhead may help a bit with knn queries.

index_meta.go
```diff
@@ -49,11 +63,52 @@ func openIndexMeta(path string) (*indexMeta, error) {
...
 	var im indexMeta
+	var fileReader util.FileReader
 	err = util.UnmarshalJSON(metaBytes, &im)
 	if err != nil {
-		return nil, ErrorIndexMetaCorrupt
+		if len(metaBytes) < 4 {
+			return nil, ErrorIndexMetaCorrupt
+		}
+
+		pos := len(metaBytes) - 4
```
capemox, line 95: I don't understand this part of the code from here on, can you add some commentary?

index/scorch/scorch.go
```diff
@@ -1192,13 +1246,215 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+	for _, seg := range segsToCompact {
+		if segDictionary[seg.Id()] {
+			mergePlan.Tasks = append(mergePlan.Tasks, &mergeplan.MergeTask{
+				Segments: []mergeplan.Segment{seg},
+			})
+		}
+	}
+
+	// trigger the merge with the force merge plan
+	s.forceMergeRequestCh <- &mergerCtrl{
```
Thejas-bhat: if the root changes by the time this forceMergeRequest is picked, won't the plan be sorta invalid/wrong since that was created on an older snapshot? the existing forceMerge doesn't have this issue because gives the specific mergePlanOptions not the plan itself and the merger is responsible for computing the plan from the _current root snapshot_. this way the ForceMerge() API doesn't have to retry or something. 

I think what can be done is utilize the ctx that you've created here, which can have a callback that computes the plan, and invoke the callback from the ctx at the planMergeSnapshot() level since that's always passed with the latest root snapshot during that iteration. 

please let me know if this makes sense

index_meta.go
```diff
@@ -110,6 +179,68 @@ func (i *indexMeta) CopyTo(d index.Directory) (err error) {
...
+	metaBytes, err := util.MarshalJSON(i)
+	if err != nil {
+		return err
+	}
+
+	i.fileWriter, err = util.NewFileWriter([]byte(indexMetaPath))
+	if err != nil {
+		return err
+	}
+
```
Thejas-bhat: nit: i understand everyone has their own style of writing code, but i feel like having a line break is supposed to provide a logical separation between code snippets, for eg 
- here i think the entire block from 187-198 can be considered as one block of code without line breaks. 
- down below, all the writes have a line break between them, from what i've seen in the bleve and zap codebase all these writes are techincally part of a single logical unit of writing things out and we don't keep the line breaks in between them. 

i know i press on these line breaks quite a bit xD but its just so that your code looks even cleaner, nothing more than that :) so maybe keeping that "logical unit" in mind, you can remove these kinda line breaks?

index_meta.go
Likith101: Done

index_impl.go
Likith101: Done

util/file_callbacks.go
Likith101: Done

index/scorch/scorch.go
Likith101: Makes sense

index_meta.go
Likith101: Apart from the logical blocks, I also add a line break whenever a loop or if statement ends unless it is strictly tied together with the statement/loop. It makes the structure better and the code cleaner.

I do agree that these nits are coming down to coding preferences and I do not have that much of an attachment to the way I add these line breaks as it is instinctual to me at this point to add breaks like these, but I just wanted to highlight that it does follow a logical pattern. 

Feel free to point out the line breaks as needed and I'll factor in those reviews as well

index/scorch/merge.go
Likith101: Please resolve comment if there are no more questions

index_meta.go
Likith101: Done

index/scorch/scorch.go
```diff
@@ -1220,13 +1274,218 @@ func (s *Scorch) updateBolt(fieldInfo map[string]*index.UpdateFieldInfo, mapping
...
+		return err
+	}
+
+	s.rootLock.Lock()
+	// create a done channel to ensure success of merge
+	ctx := context.Background()
+	doneCh := make(chan error)
+	ctx = context.WithValue(ctx, mergeDoneKey, doneCh)
+
+	// ROLLBACKING IS DISABLED FOR THIS OPERATION
```
abhinavdangeti: Quick nit :)
Refactor to - `PARTIAL ROLLBACK* WILL NOT BE SUPPORTED`

abhinavdangeti: @Thejas-bhat would you weigh in on this - still a bit uncomfortable with this.

Thejas-bhat: Yeah I've been trying to understand this part of code today. I'm guessing we'd temporarily _force_ the index to have only one index snapshot in the background which is like a violation of contract that the index should _always_ maintain `numSnapshotsToKeep` snapshots, especially when there are enough snapshots in the system. I think we should be re-encrypting things over here since we need to adhere to the existing contracts as well.

Likith101: I'll try to explain the scenario the best I can. Rollbacking works because of sequence number changes. Let's say there are 2 snapshots currently present. One with sequence number 200 and one with sequence number 300. In this case, it is useful to hold on to the older snapshot incase of any KV related issues causing us to go back to any sequence number between 200 and 300.

Now, coming to this particular drop operation, there are a couple more assumptions we need to make. Drop Keys is mostly called as soon as a key has been rotated and re-encrypted or when encryption is enabled/disabled or when the key itself has been replaced (this is different from rotation). In all of these operations, as soon as the option is toggled, we are issued a key drop request to get rid of the data that is used by the prior key material (empty key id incase of encryption disabling/enabling).

With this being the nature of this request, the first assumption we will make is that almost all of the data needs to be rewritten. There are other situations possible, but this is the most common one.

Secondly, when working with any index that is sufficiently large, we will end up encountering multiple segments/zap files that are unable to be merged. We instead opt to rewrite them using different merge tasks. With the way our current merger and persister work, each merge task execution will trigger the creation of a new snapshot. Therefore we introduce a bunch of new snapshots for the same operation.

Now, coming to the situation, assuming there are two snapshots with sequence numbers 200 and 300 both being present. It is incredibly complicated to ensure the survival of both of these snapshots independently.

- We will not be able to hold any data older data because it is a breach of contract for drop keys. Even if we held on to it, we will not be able to read it after rolling back because the key material would have been purged after the operation (by cbauth)
- We will not be able to hold on to the sequence number difference as we will likely introduce a bunch of newer snapshots all of which will have the latest sequence number

Thus, I opted to temporarily force the index to only hold one snapshot during the operation

abhinavdangeti: > Secondly, when working with any index that is sufficiently large, we will end up encountering multiple segments/zap files that are unable to be merged. 

I can't say I follow this.

> We will not be able to hold on to the sequence number difference as we will likely introduce a bunch of newer snapshots all of which will have the latest sequence number.

And this - introducing new segments with the same content from older snapshot and binding it to a newer snapshot should not change sequence numbers. We're not really merging the segments are we - we're simply opening them and re-encrypting them with new keys before writing back.

I understand there is complexity involved in binding segments to an older snapshot than the root .. so for now - I'm ok with tabling this and introduce a separate PR if needed* after a conversation on Thursday.

go.mod
abhinavdangeti: Ya let's do this separately - across all projects 👍🏼

PR #2211, opened 2025-07-25
refactor: use slices.Equal to simplify code

houpo-bob:
In the Go 1.21 standard library, a new function has been introduced that enhances code conciseness and readability. It can be find  [here](https://pkg.go.dev/slices@go1.21.1#Equal).

### DIFF
--- a/search/search.go
+++ b/search/search.go
@@ -17,6 +17,7 @@ package search
 import (
 	"fmt"
 	"reflect"
+	"slices"
 	"sort"
 
 	"github.com/blevesearch/bleve/v2/size"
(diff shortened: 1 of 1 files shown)


CascadingRadium (dismissed):

search/search.go
```diff
@@ -40,18 +41,6 @@ func init() {
 
 type ArrayPositions []uint64
 
-func (ap ArrayPositions) Equals(other ArrayPositions) bool {
```
abhinavdangeti, line 43: I would recommend _not_ dropping this API and instead replacing it's content with ..

```
return slices.Equal(ap, other)
```

abhinavdangeti (changes requested):
Thanks for this contribution @houpo-bob .

However it seems this change is incomplete. There're references to ArrayPositions' `Equals` method.

search/search.go
houpo-bob: Apologize for my error.  Your suggestions are great!

Modified.

CascadingRadium (approved):

abhinavdangeti (approved):

PR #2218, opened 2025-08-20
MB-60401: Support RRF for hybrid search

capemox:
This PR introduces hybrid search into bleve. This comes with 2 major additions: the `fusion` package and the `rescorer` interface.
- The `fusion` package is purely algorithmic to apply hybrid search strategies (for this time, only reciprocal rank fusion) to a `DocumentMatchCollection`. Additional algorithms can be added to the package.
- The `rescorer` interface provides the compatibility layer to enable hybrid search with bleve. This is done with the concrete type `fusionRescorer` which applies the algorithms in `fusion` to the search pipeline.


abhinavdangeti (commented):
Yet to actually review, but couple quick points ..
- Should rename `fusion` to better represent what it's contents are without even going in. So maybe your idea of `rescore` was better?
- Copyrights to use `2025` :)

capemox:
- `fusion` currently only holds algorithms to perform fusion, which right now is RRF, meanwhile `rescorer` is an interface I defined in bleve. Specifically, `FusionRescorer` is a concrete type I defined that uses the algorithms defined in `fusion` and makes it work with bleve (found in `rescorer.go`). I hope this makes sense!
- Will update the copyrights :)

index_alias_impl.go
```diff
@@ -914,7 +950,7 @@ func hitsInCurrentPage(req *SearchRequest, hits []*search.DocumentMatch) []*sear
 
 // MultiSearch executes a SearchRequest across multiple Index objects,
 // then merges the results.  The indexes must honor any ctx deadline.
-func MultiSearch(ctx context.Context, req *SearchRequest, preSearchData map[string]map[string]interface{}, indexes ...Index) (*SearchResult, error) {
+func MultiSearch(ctx context.Context, req *SearchRequest, preSearchData map[string]map[string]interface{}, doHybridSearch bool, rescorer rescorer, indexes ...Index) (*SearchResult, error) {
```
abhinavdangeti: Let's do something about this growing list of params and constant API signature change here.
```
type multiSearchParams struct {
    preSearchData   map[string]map[string]interface{}
    doHybridSearch bool
    rescorer               rescorer
}

func MultiSearch(ctx context.Context, req *SearchRequest, indexes ...Index, msp multiSearchParams) {
    ...
```

rescorer.go
```diff
@@ -0,0 +1,116 @@
...
+// The main use of Rescorer is in hybrid search; all the individual scores
+// for query and knn are combined using Rescorer
+type rescorer interface {
+	prepareSearchRequest()
+	restoreSearchRequest()
+	rescore(*SearchResult)
+}
+
+const (
+	ReciprocalRankFusionStrategy string = "rrf"
```
abhinavdangeti: No need for explicit type `string`

search_knn_test.go
```diff
@@ -1701,3 +1701,507 @@ func TestNumVecsStat(t *testing.T) {
 		}
 	}
 }
+
+func getHybridSearchIndex(path string) (Index, error) {
```
abhinavdangeti: Would you move the new tests and supporting utility methods to a new `hybrid_search_fusion_test.go` (or a better name of your choice). You will certainly be adding tests for `rsf` and the other one as well when you're ready - might be good to use a new file for all pertinent tests.

Also refactor `getHybridSearchIndex` to `createAHybridSearchIndex`

search_knn_test.go
```diff
@@ -1701,3 +1701,507 @@ func TestNumVecsStat(t *testing.T) {
...
+		{
+			"color":        "light sky blue",
+			"colorvect_l2": []float32{135, 206, 250},
+		},
+	}
+
+	return documents
+}
+
+func getHybridSearchRequest() *SearchRequest {
```
abhinavdangeti: Ditto - `createAHybridSearchRequest`

search_knn.go
```diff
@@ -213,6 +237,12 @@ func (r *SearchRequest) UnmarshalJSON(input []byte) error {
 		r.KNNOperator = knnOperatorOr
 	}
 
+	if isHybridSearch(r) {
+		if r.KNNOperator == knnOperatorAnd {
+			return fmt.Errorf("knn operator 'and' is not compatible with hybrid search")
```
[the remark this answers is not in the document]
abhinavdangeti: Ignore `Copilot`'s review here.

I don't think the case here is that all hybrid search is not available and just rank fusion is not. If that is the case I'f refactor `isHybridSearch` to `scoreFusionReqired` or something better that better paints the picture.

Either way, we need sufficient commentary here explaining the situation.

abhinavdangeti (changes requested):
Few comments/concerns.

p.s. approach copilot's review with extra care.

rescorer.go
```diff
@@ -0,0 +1,116 @@
...
+	// for all the knn queries, restore boost values
+	r.restoreKnnRequest()
+}
+
+func (r *fusionRescorer) rescore(sr *SearchResult) {
+	var fusionResult fusion.FusionResult
+
+	switch r.req.Score {
+	case ReciprocalRankFusionStrategy:
+		fusionResult = fusion.ReciprocalRankFusion(
```
Thejas-bhat: is there a reason why you're specifically instantiating `fusion.FusionResult` instead of using a pointer to that type? because over here you'd be doing a redundant copy of the instance generated within the function and also creating garbage within the function. 

the golang way would be to return a pointer to a struct instance from the function and define a `fusion *fusion.FusionResult` above such that this assignment would be a pointer change instead of a copy by value

rescorer.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"github.com/blevesearch/bleve/v2/fusion"
+	"github.com/blevesearch/bleve/v2/search/query"
+)
+
+// Rescorer is applied after all the query and knn results are obtained.
+// The main use of Rescorer is in hybrid search; all the individual scores
+// for query and knn are combined using Rescorer
+type rescorer interface {
```
Thejas-bhat: i think i'm missing somehting, but can you explain why an interface is necessary over here? do you see there being a different implementation altogether of these methods in the future? because the functions defined and implemented look pretty straightforward and generic for multiple rescoring strategies. 

also i think defining an interface at the bleve level (which has a lot of code); although it does hide it from the user, it doesn't necessarily hide the different implementations within the codebase itself since the non-scoring related constructs is also exposed to this interface and thereby making it a bit more vulnerable.

rescorer.go
capemox: There isn't a reason for this, you're right. I'll use a pointer type to avoid creating garbage.

fusion/rrf.go
```diff
@@ -0,0 +1,141 @@
...
+			Hits:     hits,
+			Total:    0,
+			MaxScore: 0.0,
+		}
+	}
+
+	// Create a map of document ID to a slice of ranks.
+	// The first element of the slice is the rank from the FTS search,
+	// and the subsequent elements are the ranks from the KNN searches.
+	docRanks := make(map[string][]int)
```
Likith101, line 42: Initialise this array here itself to avoid lines 53, 85 and also the check on 81. All of the hits will ideally require an entry in this map and the length of the int array is always numKNNQueries + 1.

Likith101 (changes requested):

fusion/rrf.go
capemox: Makes sense! Will do this.

fusion/fusion.go
```diff
@@ -0,0 +1,107 @@
...
+			return 1
+		}
+
+		return 0
+	}
+}
+
+func getFusionExplAt(hit *search.DocumentMatch, i int, value float64, message string) *search.Explanation {
+	if hit.Score == 0.0 {
+		// No query component here; explanation doesn't contain it, hence decrement i
```
abhinavdangeti: I'm not sure I quite follow this - perhaps comment needs to explain how there is no query component.

Also I see that this utility method is used in one place - so I think we can just situate this content where it's needed.

fusion/fusion.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	return &search.Explanation{
+		Value: value,
+		Message: message,
+		Children: []*search.Explanation{hit.Expl.Children[i]},
+	}
+}
+
+func finalizeFusionExpl(hit *search.DocumentMatch, explChildren []*search.Explanation) {
```
abhinavdangeti: Ditto, no need for separate utility method - given that it's being called once.

search/search.go
```diff
@@ -377,6 +377,7 @@ type SearchContext struct {
 	DocumentMatchPool *DocumentMatchPool
 	Collector         Collector
 	IndexReader       index.IndexReader
+	FusionRescoring   bool
```
abhinavdangeti: Bad name :), would advise `ScoreFusion` or something better.

fusion/fusion.go
capemox: So if a hit is only knn and has no query component, the query explanation is absent. The query explanation also comes first in the list of explanations. So if we wanted to get the `i`'th explanation (a knn explanation if `i` > 0), we need to decrement `i` if there is no query component, else you'd get index out of bounds. I'll add a better comment for this :)

The way to get the fusion explanation is the same for the other fusion methods, so that's why its in a separate function. I can maybe make it part of the rrf function itself maybe for now, and make a separate function out of it in the PRs for RSF and DBSF if you'd like?

fusion/fusion.go
capemox: This function too would be used in the other fusion methods, should I include it in rrf for now?

search/search.go
capemox: Sure, will change! I use naming similar to this in many other places, should I change those too?

This is the most accurate naming though - it uses fusion and it does rescoring

fusion/fusion.go
abhinavdangeti: Ah if there's future additional purpose for this method, then ok - you can leave it where it is. Better commentary though.

fusion/fusion.go
abhinavdangeti: All good then, you can leave it where it is.

search/search.go
abhinavdangeti: I would recommend changing then - and keep the naming consistent and as "grep-able" as possible.

rescorer.go
```diff
@@ -0,0 +1,110 @@
...
+	origBoosts []float64
+}
+
+// Stores information about the hybrid search into FusionRescorer.
+// Also mutates the SearchRequest by:
+// - Setting boosts to 1: top level boosts only used for rescoring
+// - Setting From and Size to 0 and ScoreWindowSize
+func (r *rescorer) prepareSearchRequest() {
+	if r.req.Params.ScoreRankConstant == nil {
+		src := 60
```
CascadingRadium: 60 must be a global bleve config

`
var DefaultScoreRankConstant = 60
`

search.go
```diff
@@ -654,3 +667,60 @@ func isMatchAllQuery(q query.Query) bool {
...
+// Additional parameters in the search request. Currently only being
+// used for hybrid search parameters.
+type Params struct {
+	ScoreRankConstant *int `json:"score_rank_constant,omitempty"`
+	ScoreWindowSize   *int `json:"score_window_size,omitempty"`
+}
+
+func ParseParams(r *SearchRequest, input []byte) (*Params, error) {
+	if len(input) == 0 {
+		src := 60
```
CascadingRadium: again - make all the defaults point to one global var

search.go
```diff
@@ -654,3 +667,60 @@ func isMatchAllQuery(q query.Query) bool {
...
+		} else if *params.ScoreWindowSize < r.Size {
+			return nil, fmt.Errorf("score window size must be greater than or equal to Size (%d)", r.Size)
+		}
+	} else {
+		sws := r.Size
+		params.ScoreWindowSize = &sws
+	}
+
+	return &params, nil
+}
```
CascadingRadium: very messy code
avoid using pointers to int
```

var (
	DefaultScoreRankConstant = 60
)

// Additional parameters in the search request. Currently only being
// used for hybrid search parameters.
type Params struct {
	ScoreRankConstant int `json:"score_rank_constant"`
	ScoreWindowSize   int `json:"score_window_size"`
}

func NewDefaultParams(size int) *Params {
	return &Params{
		ScoreRankConstant: DefaultScoreRankConstant,
		ScoreWindowSize:   size,
	}
}

func (p *Params) UnmarshalJSON(input []byte) error {
	var temp struct {
		ScoreRankConstant *int `json:"score_rank_constant"`
		ScoreWindowSize   *int `json:"score_window_size"`
	}
	if err := util.UnmarshalJSON(input, &temp); err != nil {
		return err
	}
	if temp.ScoreRankConstant != nil {
		p.ScoreRankConstant = *temp.ScoreRankConstant
	}
	if temp.ScoreWindowSize != nil {
		p.ScoreWindowSize = *temp.ScoreWindowSize
	}
	return nil
}

func (p *Params) Validate(minimumSize int) error {
	if p.ScoreRankConstant < 1 {
		return fmt.Errorf("score rank constant must be greater than 0")
	}
	if p.ScoreWindowSize < minimumSize {
		return fmt.Errorf("score window size must be greater than or equal to Size (%d)", minimumSize)
	}
	return nil
}

func ParseParams(r *SearchRequest, input []byte) (*Params, error) {
	params := NewDefaultParams(r.Size)
	if len(input) == 0 {
		return params, nil
	}
	err := util.UnmarshalJSON(input, params)
	if err != nil {
		return nil, err
	}
	// validate params
	err = params.Validate(r.Size)
	if err != nil {
		return nil, err
	}
	return params, nil
}
```

- if len(input) = 0 -> rv = defaultParams
- else -> overwrite the default params with custom values
- validate the final params

search.go
```diff
@@ -654,3 +667,60 @@ func isMatchAllQuery(q query.Query) bool {
...
+
+func validateScore(r *SearchRequest) error {
+	if r.Score == "" || r.Score == "none" || IsFusionRescoringRequired(r) {
+		return nil
+	}
+
+	return fmt.Errorf("invalid score field \"%s\": must be one of \"\", \"none\", \"%s\"", r.Score, ReciprocalRankFusionStrategy)
+}
```
CascadingRadium: use bleve_index_api for the valid `score` content, put below in bleve_index_api

```
const (
    ScoreDefault       = ""
	ScoreNone          = "none"
	ScoreRRF           = "rrf"
)

var SupportedScoreValues = map[string]int{
    ScoreDefault:       0,
	ScoreNone:          1,
	ScoreRRF:           2
}
```
now
```
func validateScore(r *SearchRequest) error {
   if _, exists := index.SupportedScoreValues[req.Score]; !exists {
		return fmt.Errorf("invalid score config")
	}
   return nil
}
```

search.go
```diff
@@ -654,3 +667,60 @@ func isMatchAllQuery(q query.Query) bool {
 	_, ok := q.(*query.MatchAllQuery)
 	return ok
 }
+
+// Checks if the request is hybrid search. Currently supports: RRF.
+func IsFusionRescoringRequired(req *SearchRequest) bool {
+	switch req.Score {
+	case ReciprocalRankFusionStrategy:
```
CascadingRadium: put this constant in bleve_index_api
```
case index.ScoreRRF
```
ref my other comment

search.go
```diff
@@ -317,6 +317,19 @@ func (r *SearchRequest) Validate() error {
...
 
+	err = validateScore(r)
+	if err != nil {
+		return err
+	}
+
+	if IsFusionRescoringRequired(r) {
+		if r.SearchAfter != nil || r.SearchBefore != nil {
+			return fmt.Errorf("cannot use search after or search before with score fusion")
+		} else if so := (search.SortOrder{&search.SortScore{Desc: true}}); r.Sort != nil && !reflect.DeepEqual(r.Sort, so) {
```
CascadingRadium: drop the else if and keep the allowedSort as a global sort object to avoid GC churn

```
var AllowedFusionSort = search.SortOrder{&search.SortScore{Desc: true}}
```

and ur code

```

// Inline validation for fusion rescoring compatibility
if IsFusionRescoringRequired(r) {
	if r.SearchAfter != nil || r.SearchBefore != nil {
		return fmt.Errorf("cannot use search after or search before with score fusion")
    }
	
	// Fusion rescoring requires either no sort or score descending sort
	if r.Sort != nil {
		if !reflect.DeepEqual(r.Sort, AllowedFusionSort) {
			return fmt.Errorf("fusion rescoring requires sort to be empty or descending by score")
		}
	}
}
```

search_knn.go
```diff
@@ -189,6 +192,27 @@ func (r *SearchRequest) UnmarshalJSON(input []byte) error {
...
 	}
 
+	if temp.Params == nil {
+		if IsScoreFusionRequired(r) {
+			// If params is not present and it is requires rescoring, assign
+			// default values
+			src := 60
+			sws := r.Size
+			params := Params{ScoreRankConstant: &src, ScoreWindowSize: &sws}
+			r.Params = params
```
CascadingRadium: ref my other comment

```
r.Params = NewDefaultParams(r.Size)
```

search_knn.go
```diff
@@ -67,6 +67,8 @@ type SearchRequest struct {
 
 	PreSearchData map[string]interface{} `json:"pre_search_data,omitempty"`
 
+	Params Params `json:"params,omitempty"`
```
CascadingRadium: lazy naming, keep the field name separated from the type name:

```
RequestParameters   Params `json:"params,omitempty"`
```

or vice versa and also use a pointer here why the concrete type?

search_knn.go
```diff
@@ -213,6 +237,12 @@ func (r *SearchRequest) UnmarshalJSON(input []byte) error {
 		r.KNNOperator = knnOperatorOr
 	}
 
+	if IsScoreFusionRequired(r) {
+		if r.KNNOperator == knnOperatorAnd {
+			return fmt.Errorf("knn operator 'and' is not compatible with score fusion")
+		}
+	}
+
```
CascadingRadium: shift this to the validation func

```
func validateKNN(req *SearchRequest) error {
```

search.go
abhinavdangeti: Good suggestion @CascadingRadium. But, we need not list these constants in `bleve_index_api`, just here in `bleve` is good - if you can't find a good location for it, perhaps adding a separate constants.go and relocating all misplaced constants to this file would be a good idea.

rescorer.go
abhinavdangeti: ➕

search.go
abhinavdangeti: Let's situate these within bleve itself, but constants are a good recommendation.

search.go
capemox: I'll keep the constants within bleve for now, within `search.go`.

index_alias_impl.go
```diff
@@ -912,9 +978,17 @@ func hitsInCurrentPage(req *SearchRequest, hits []*search.DocumentMatch) []*sear
 	return hits
 }
 
+// Extra parameters for MultiSearch
+type multiSearchParams struct {
+	preSearchData map[string]map[string]interface{}
+	doScoreFusion bool
+	rescorer      *rescorer
```
abhinavdangeti, line 977: Just a quick thought - do we really a separate attribute for `doScoreFusion`? Will `rescorer != nil` not suffice?

index_impl.go
```diff
@@ -562,6 +562,25 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
+		// Since the score fusion key is not set, check if it is a hybrid search.
+		// If it is, then set doScoreFusion. This indicates that the rescoring
+		// must happen locally within this index, and there are no other
+		// indexes/aliases involved.
+		doScoreFusion = IsScoreFusionRequired(req)
+
+		// new context will be used in internal functions to collect data
+		// as suitable for hybrid search. Rescorer is used for rescoring
+		// using fusion algorithms.
+		if doScoreFusion {
```
abhinavdangeti: Related to comment above, it seems we can simply set rescorer with `IsScoreFusionRequired(..)` and use `rescorer != nil`  for decision making everywhere else.

abhinavdangeti (commented):
Looking good @capemox , want to make a few more passes over the next few days.

index_alias_impl.go
capemox: This is to make it more readable. At the alias level, I agree it may not be strictly required, but it makes the logic at the index_impl much more readable, rather than just `recorer != nil`: https://github.com/blevesearch/bleve/blob/a35f1f6516b511a953ee35c25771d9613821809f/index_impl.go#L665-L672
Since I did this for index_impl, I also decided to do it for the alias, that's it. Should I go ahead and remove it?

rescorer.go
```diff
@@ -136,6 +148,7 @@ func (r *rescorer) mergeDocs(sr *SearchResult) {
 
 func newRescorer(req *SearchRequest) *rescorer {
 	return &rescorer{
-		req: req,
+		req:      req,
+		restored: false,
```
abhinavdangeti: restored will be assigned false by default, no need for this line :)

search.go
```diff
@@ -317,6 +331,23 @@ func (r *SearchRequest) Validate() error {
 		return err
 	}
 
+	err = validateScore(r)
```
abhinavdangeti: I feel we can skip score validation - if the user sets it right great, otherwise we ignore (in line with current behavior). I see it's a constant lookup validation but may not necessary. `SupportedScoreValues` can be dropped too.

search.go
```diff
@@ -654,3 +691,86 @@ func isMatchAllQuery(q query.Query) bool {
 	_, ok := q.(*query.MatchAllQuery)
 	return ok
 }
+
+// Checks if the request is hybrid search. Currently supports: RRF.
+func IsScoreFusionRequired(req *SearchRequest) bool {
```
abhinavdangeti: IsScoreFusionRequested*

Also, don't see a need for this method to be exported - is there?

abhinavdangeti:
@capemox  If the user requests for score fusion when there are no `knn` bits at all - I suppose we will still do what is requested, have you ensured the "explain" message works ok in the situation?

search.go
capemox: Because score takes a fixed set of values now, wouldn't it be better to have validation now?

capemox:
@abhinavdangeti yep, the explain message will still work properly in those cases.

search.go
capemox: Yep, this method will be needed in cbft for convertToBleveSearchRequest and decodeBleveSearchRequest

search.go
abhinavdangeti: I think it's ok to not do this validation and ignore the request if it doesn't match any unsupported values - let's save on everywhere possible considering this is a very hot path.

search.go
abhinavdangeti: Cool, let's rename the method to `IsScoreFusionRequested` though.

fusion/rrf.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Pre-assign rank lists to each candidate document
+	for _, hit := range hits {
+		docRanks[hit.ID] = make([]int, numKNNQueries+1)
+	}
+
+	// Only a max of `window_size` elements need to be counted for. Stop
+	// calculating rank once this threshold is hit.
+	sort.Slice(hits, func(a, b int) bool {
+		return scoreSortFunc()(hits[a], hits[b]) < 0
```
CascadingRadium, line 52: add the function here itself, callback makes it hard to read.

fusion/rrf.go
```diff
@@ -0,0 +1,139 @@
...
+		})
+
+		// Update the ranks of the documents in the docRanks map.
+		for j, hit := range knnDocs {
+			docRanks[hit.ID][i+1] = j + 1
+
+			if j == windowSize-1 {
+				// No need to calculate ranks from here
+				break
+			}
```
CascadingRadium: can just be 

```
for j := range windowSize
```

fusion/rrf.go
```diff
@@ -0,0 +1,139 @@
...
+			// Skip if Score is 0, since that means the document was not
+			// found as part of FTS, and only in KNN.
+			docRanks[hit.ID][0] = i + 1
+		}
+
+		if i == windowSize-1 {
+			// No need to calculate ranks from here
+			break
+		}
+	}
```
CascadingRadium: ditto

```
for i := range windowSize
```

fusion/rrf.go
```diff
@@ -0,0 +1,139 @@
...
+	// Create a map of document ID to a slice of ranks.
+	// The first element of the slice is the rank from the FTS search,
+	// and the subsequent elements are the ranks from the KNN searches.
+	docRanks := make(map[string][]int)
+
+	// Pre-assign rank lists to each candidate document
+	for _, hit := range hits {
+		docRanks[hit.ID] = make([]int, numKNNQueries+1)
+	}
+
```
CascadingRadium, line 48:
```
	limitFTS := windowSize
	if len(hits) < windowSize {
		limitFTS = len(hits)
	}
```
use limitValue here 
if lenHits < windowSize we safeguard below

index_alias_impl.go
```diff
@@ -653,7 +682,7 @@ func preSearch(ctx context.Context, req *SearchRequest, flags *preSearchFlags, i
 // if the request is satisfied by just the preSearch result,
 // finalize the result and return it directly without
 // performing multi search
-func finalizeSearchResult(req *SearchRequest, preSearchResult *SearchResult) *SearchResult {
+func finalizeSearchResult(ctx context.Context, req *SearchRequest, preSearchResult *SearchResult, doScoreFusion bool, rescorer *rescorer) *SearchResult {
```
CascadingRadium: you can just send `rescorer` and not the `doScoreFusion` as its duplicate purpose.

if `rescorer` is not nil then you can rescore

fusion/fusion.go
```diff
@@ -0,0 +1,102 @@
...
+		} else if i.HitNumber > j.HitNumber {
+			return 1
+		}
+
+		return 0 // Equal scores and HitNumbers
+	}
+}
+
+func scoreSortFunc() func(i, j *search.DocumentMatch) int {
+	return func(i, j *search.DocumentMatch) int {
```
maneuvertomars: Could we use shared comparator for all score-based sorting (descending score, ascending HitNumber). Used by both scoreSortFunc and scoreBreakdownSortFunc.

```
// comparatorBuilder creates a comparator given a score extraction function.
func comparatorBuilder(getScore func(*search.DocumentMatch) float64) func(i, j *search.DocumentMatch) int {
    return func(i, j *search.DocumentMatch) int {
        iScore := getScore(i)
        jScore := getScore(j)

        // Compare scores (descending)
        if iScore > jScore {
            return -1
        } else if iScore < jScore {
            return 1
        }

        // Tie-breaker: HitNumber ascending
        if i.HitNumber < j.HitNumber {
            return -1
        } else if i.HitNumber > j.HitNumber {
            return 1
        }

        return 0
    }
}
// Use for breakdown score
func scoreBreakdownSortFunc(idx int) func(i, j *search.DocumentMatch) int {
    return comparatorBuilder(func(m *search.DocumentMatch) float64 {
        if m.ScoreBreakdown != nil {
            if score, ok := m.ScoreBreakdown[idx]; ok {
                return score
            }
        }
        return 0.0
    })
}

// Use for main score
func scoreSortFunc() func(i, j *search.DocumentMatch) int {
    return comparatorBuilder(func(m *search.DocumentMatch) float64 {
        return m.Score
    })
}
```

index_alias_impl.go
```diff
@@ -682,7 +711,19 @@ func finalizeSearchResult(req *SearchRequest, preSearchResult *SearchResult) *Se
...
+	// rescore if fusion flag is set
+	if doScoreFusion {
+		// All hits here are only knn hits, so add in FusionKnnHits
+		preSearchResult.FusionKnnHits = preSearchResult.Hits
+		// Since no FTS hits, set Hits to nil
+		preSearchResult.Hits = nil
+		rescorer.rescore(preSearchResult)
+		rescorer.restoreSearchRequest()
+	}
+
```
CascadingRadium, line 717: this piece of code is called when there is no hybrid search (aka pure KNN search) what is the purpose of rescoring in this context?
you can just remove this

index_alias_impl.go
```diff
@@ -227,6 +228,25 @@ func (i *indexAliasImpl) SearchInContext(ctx context.Context, req *SearchRequest
...
+		// new context will be used in internal functions to collect data
+		// as suitable for fusion. Rescorer is used for rescoring
+		// using fusion algorithms.
+		if doScoreFusion {
+			ctx = context.WithValue(ctx, search.ScoreFusionKey, true)
+			rescorer = newRescorer(req)
+			rescorer.prepareSearchRequest()
+			defer rescorer.restoreSearchRequest()
+		}
+	}
```
CascadingRadium: here as well, if its NOT a hybrid search (aka pure KNN search) then we need not actually have to do the rescoring at all. Even if the user mentions `rrf` scoring. 

add a check here to ensure that if the req.Query is a MatchNone or nil then we skip redundant `rrf` computation 


perhaps streamline the code into a helper method that 
- checks the ctx avail
- checks the value of `score`
- checks if the Query is matchNone or nil
- if any check fails then we just return the original request
- else we return the new request which is prepared by the rescorer

clean API:

```
func SetupScoreFusion(ctx context.Context, req *SearchRequest) (context.Context, *ScoreFusion) {
```

rescorer.go
```diff
@@ -0,0 +1,153 @@
...
+	req *SearchRequest
+
+	// Stores the original From, Size and Boost parameters from the request
+	origFrom   int
+	origSize   int
+	origBoosts []float64
+
+	// Flag variable to make sure that restoreSearchRequest is only run once
+	// when it is deferred
+	restored bool
```
Thejas-bhat, line 41: i didn't understand what this flag variable is supposed to do. here's what i gather, the instance of this rescorer is created at the partition level (at a single bleve index level) right? so when the `defer restoreSearchRequest()` is called, only a single thread (corresponding to the partition/bleve index) is invoking it right and i see that down the code path its again invoked - maybe that's a redundant call? i might be missing somehting here, let mek now

index_impl.go
```diff
@@ -632,10 +652,22 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
+		if !skipKNNCollector && requestHasKNN(req) {
+			knnHits, err = i.runKnnCollector(ctx, req, indexReader, false)
+			if err != nil {
+				return nil, err
+			}
+		}
+	} else {
+		// if score fusion, run collect if doScoreFusion
+		// if not, then knn hits must come from presearchdata (may be nil too)
+		if doScoreFusion && requestHasKNN(req) {
```
Thejas-bhat: another doubt here, is there a case ever you can set the doScoreFusion to true in a non-knn request? because logically i'd expect that if doScoreFusion = true then it automatically implies that request has knn (simpler check that too)

index_impl.go
```diff
@@ -859,6 +891,17 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
...
 	}
 
+	if contextScoreFusionKeyExists {
+		rv.FusionKnnHits = knnHits
+	}
+
+	// rescore if fusion flag is set
+	if doScoreFusion {
+		rescorer.rescore(rv)
+		rescorer.restoreSearchRequest()
```
Thejas-bhat, line 896: based on my comment in rescorer.go please see if this is redundant or not (given that there's a defer restore...() above for the same doScoreFusion = true condition)

search/collector/topn.go
```diff
@@ -246,6 +246,7 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
 	if hc.size+hc.skip > PreAllocSizeSkipCap {
 		backingSize = PreAllocSizeSkipCap + 1
 	}
+
```
Thejas-bhat: nit: i'd suggest not adding too many whitespace lines to the existing code because they already follow a certain format and this would tend to add more than necessary whitespace lines in the codebase.

index_alias_impl.go
abhinavdangeti: Perhaps, in both locations if possible.

rescorer.go
capemox: the normal flow of the code is:
1. prepareSearchRequest()
2. Perform the search
3. restoreSearchRequest()
4. hitsInCurrentPage() - restoreSearchRequest has to run before this to restore values of From and Size for pagination

In case an error comes up, there is an early return. 1 will execute but the search request won't be restored, so will have the modified From, Size and boost values. So, the defer statement to ensure restoreSearchRequest is always run. In case the request is successful, restoreSearchRequest is already run and so the defer is useless, and so the restored flag provides an early exit.

The issue is that restoreSearchRequest must run before hitsInCurrentPage always, else pagination will be incorrect. So we cannot just always defer restoreSearchRequest.

index_impl.go
capemox: Yep, in case there is no knn you can still do score fusion, it just doesn't have any knn component to it.

index_impl.go
capemox: Yes, as I mentioned above, restoreSR() must run before hitsInCurrentPage(). the defer is only for early return in case of errors.

search/collector/topn.go
capemox: Yep, will fix these

fusion/rrf.go
capemox: This function is also used in rsf and dbsf, so there will be code duplication. Is this ok?

fusion/rrf.go
capemox: Not sure what you mean here.

index_alias_impl.go
capemox: Ok sure, will do

index_alias_impl.go
capemox: Even if it is pure knn search, there can still be multiple knn queries of different types with cosine, l2_distance etc so rescoring will still be useful.

index_alias_impl.go
capemox: Same as my above comment: we can still have multiple knn queries, so score fusion is still useful in those cases.

fusion/rrf.go
abhinavdangeti: No avoid the code duplication as much as possible.
Adding these kind of methods in a separate util.go within `fusion/.` may look better.

fusion/rrf.go
```diff
@@ -0,0 +1,131 @@
...
+		for _, hit := range hits {
+			if _, ok := hit.ScoreBreakdown[i]; ok {
+				knnDocs = append(knnDocs, hit)
+			}
+		}
+
+		// Sort the documents based on their score for this KNN query.
+		sort.Slice(knnDocs, func(a, b int) bool {
+			return scoreBreakdownSortFunc(i)(knnDocs[a], knnDocs[b]) < 0
```
abhinavdangeti, line 78: Ditto, if this method's going to be used elsewhere too.

fusion/fusion.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"github.com/blevesearch/bleve/v2/search"
+)
+
+type FusionResult struct {
+	Hits     search.DocumentMatchCollection
+	Total    uint64
+	MaxScore float64
+}
```
Likith101: Add an empty line at the end of files. Git complains about the formatting otherwise.

fusion/util.go
```diff
@@ -0,0 +1,96 @@
...
+		Children: []*search.Explanation{hit.Expl.Children[i]},
+	}
+}
+
+func finalizeFusionExpl(hit *search.DocumentMatch, explChildren []*search.Explanation) {
+	hit.Expl.Children = explChildren
+
+	hit.Expl.Value = hit.Score
+	hit.Expl.Message = "sum of"
+}
```
Likith101: Same here, add an empty line

fusion/fusion.go
capemox: This would make the functions (which are already pretty complex) even more complicated, so I'll skip doing this

search_knn.go
```diff
@@ -189,6 +192,24 @@ func (r *SearchRequest) UnmarshalJSON(input []byte) error {
...
+		// parameters.
+		if IsScoreFusionRequested(r) {
+			params, err := ParseParams(r, temp.RequestParams)
+			if err != nil {
+				return err
+			}
+			r.RequestParams = params
+		}
+	}
+
```
Likith101, line 210: Switch the order of the If statements

```
If IsScoreFusionRequested(r) {
    If temp.RequestParams == nil {
        .....
    } else {
        .....
    }
}
```

search.go
```diff
@@ -654,3 +674,78 @@ func isMatchAllQuery(q query.Query) bool {
...
+	case ScoreRRF:
+		return true
+	default:
+		return false
+	}
+}
+
+// Additional parameters in the search request. Currently only being
+// used for hybrid search parameters.
+type Params struct {
```
Likith101: Perhaps rename this to RequestParams? Params seem too generic of a name

Likith101 (approved):

search.go
```diff
@@ -420,6 +420,14 @@ func (r *SearchRequest) SetSearchBefore(before []string) {
 	r.SearchBefore = before
 }
 
+// AddParams adds a RequestParams field to the search request
+func (r *SearchRequest) AddParams(scoreRankConstant, scoreWindowSize int) {
```
abhinavdangeti: This is an exported API - so in our interest to keep signature from changing in future ..

```
func (r *SearchRequest) AddParams(params RequestParams) {
	r.Params = &params
}
```

abhinavdangeti (approved):

Likith101 (approved):

Issue #2225, opened 2025-09-10
[feature request] OpenAPI Spec for the Query JSON


ajroetker:
It'd be real handy to have an OpenAPI spec for the Query JSON language (maybe even generate the go types from the spec using https://github.com/go-openapi/spec). I'm trying to build a query builder for bleve queries and being able to generate the typescript models from the spec would make it really easy to generate the builder from.

Locally I just had an LLM create an OpenAPI spec from all the files in the directory and I played around with generating the go types from the spec and wrapping those types with the functions supported on those types in the query package and could definitely push up my change if the maintainers were down for the idea.

PR #2226, opened 2025-09-11
Adding openai spec for bleve queries

abhinavdangeti:
Addresses: https://github.com/blevesearch/bleve/issues/2225


ajroetker:
@abhinavdangeti This matches the one I've been using, my only note is that it would be more convenient from a auto-generated code perspective to remove the test api endpoints from the path. I'd be happy to work on using the go generated models from the spec in bleve and open sourcing my bleve query builder components built on this one this PR goes in! Thanks for hopping on this so quick.

Thejas-bhat (approved):

Thejas-bhat (approved):

PR #2227, opened 2025-09-21
MB-60401: RSF and DBSF fusion

capemox:
Adds RSF and DBSF fusion algorithms for hybrid search.
Pending: End to end tests with rescorer


search.go
```diff
@@ -52,12 +52,16 @@ const (
 	ScoreDefault = ""
 	ScoreNone    = "none"
 	ScoreRRF     = "rrf"
+	ScoreRSF     = "rsf"
+	ScoreDBSF    = "dbsf"
```
abhinavdangeti, line 56: Would you add commentary here and I suppose in the previous PR as well - expanding what the acronyms stand for and a link to any online material diving into detail on each of these algorithms.

fusion/rsf_dbsf.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	mean := sum / float64(len(scores))
+
+	// Calculate standard deviation
+	var variance float64
+	for _, score := range scores {
+		variance += math.Pow(score-mean, 2)
+	}
+	variance = variance / float64(len(scores))
+	stdDev := math.Sqrt(variance)
```
Thejas-bhat, line 78: i realize that this is a math-intensive library especially with these mean calculation etc. so, could you check out https://github.com/gonum/gonum (its quite popular and actively maintained)? maybe benchmark these math computes between the two implementations to see if we can save CPU cycles since this is a super hot path

abhinavdangeti: If we're looking into bringing a new third party dependency - let's vet it's licensing.
Also, this can be a separate change.

PR #2237, opened 2025-11-06
(feat) add `vfs`: storage abstraction layer for Scorch

ajroetker:
`vfs` is a new storage abstraction layer that decouples Scorch's filesystem operations from the core indexing logic. This enables Scorch to use different storage backends (filesystem, S3, etc.) without modification.

- Directory interface abstracting filesystem operations
- FSDirectory: Drop-in replacement using local filesystem
- Backward compatible with existing Scorch indexes
- Allows custom storage backend implementations
- Gets closer to enabling cloud-native and distributed architectures

I've also created a demonstration of the efficacy of the abstraction using an s3 backed minio client with local caching if that's of interest.


abhinavdangeti:
Thanks for the contribution @ajroetker . Pinging the team to weigh in on this.

PR #2242, opened 2025-11-11
(feat) Add prefix and regex filtering for term facets

ajroetker:
This change adds support for filtering facet terms by prefix and/or regex pattern. Which is useful for faceting during `search_as_you_type` style faceting for terms that match what is in the searchbox, aggregated by field (for books I can facet facet authors, titles etc in an autosuggest a bit cleaner now).

- Adds `TermPrefix` and `TermPattern` fields to FacetRequest
- Updated `NewTermsFacetBuilder` to return error
- Only matching terms trigger string allocation from `[]byte`

### DIFF
--- a/search.go
+++ b/search.go
@@ -152,6 +152,9 @@ type FacetRequest struct {
 	TermPattern    string           `json:"term_pattern,omitempty"`
 	NumericRanges  []*numericRange  `json:"numeric_ranges,omitempty"`
 	DateTimeRanges []*dateTimeRange `json:"date_ranges,omitempty"`
+
+	// Compiled regex pattern (cached during validation)
+	compiledPattern *regexp.Regexp
 }
 
 // NewFacetRequest creates a facet on the specified
@@ -175,12 +178,13 @@ func (fr *FacetRequest) SetRegexFilter(pattern string) {
 }
 
 func (fr *FacetRequest) Validate() error {
-	// Validate regex pattern if provided
+	// Validate regex pattern if provided and cache the compiled regex
 	if fr.TermPattern != "" {
-		_, err := regexp.Compile(fr.TermPattern)
+		compiled, err := regexp.Compile(fr.TermPattern)
 		if err != nil {
 			return fmt.Errorf("invalid term pattern: %v", err)
 		}
+		fr.compiledPattern = compiled
 	}
 
 	nrCount := len(fr.NumericRanges)
--- a/index_impl.go
+++ b/index_impl.go
@@ -856,13 +856,18 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 					facetBuilder.SetPrefixFilter(facetRequest.TermPrefix)
 				}
 
-				// Compile and set regex filter if provided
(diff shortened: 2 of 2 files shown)


search.go
```diff
@@ -147,6 +147,8 @@ type numericRange struct {
 type FacetRequest struct {
 	Size           int              `json:"size"`
 	Field          string           `json:"field"`
+	TermPrefix     string           `json:"term_prefix,omitempty"`
```
CascadingRadium, line 151: Can you please add Setter methods for API completeness.

```
func (fr *FacetRequest) SetPrefixFilter(prefix string)
func (fr *FacetRequest) SetRegexFilter(pattern string)
```

Also add regex pattern validation in

```
func (fr *FacetRequest) Validate() error {
```

search/facet/facet_builder_terms.go
```diff
@@ -30,25 +32,45 @@ func init() {
...
+	regex       *regexp.Regexp
+	termsCount  map[string]int
+	total       int
+	missing     int
+	sawValue    bool
 }
 
-func NewTermsFacetBuilder(field string, size int) *TermsFacetBuilder {
-	return &TermsFacetBuilder{
+func NewTermsFacetBuilder(field string, size int, prefix, pattern string) (*TermsFacetBuilder, error) {
```
CascadingRadium: I think the API signature should not be changed, since it’s an exported API and would introduce a breaking change.
You can add setters for the required fields instead. Also, please include the regex field directly in the struct: this way, at `L851` of `index_impl.go`, you can compile and set it directly if a pattern is present in the FacetRequest, and return an error before creating the `TermsFacetBuilder` itself, thereby avoiding an API change.

```
func (fr *TermsFacetBuilder) SetPrefixFilter(prefix string)
func (fr *TermsFacetBuilder) SetRegexFilter(regex *regexp.Regexp)
```

search/facet/facet_builder_terms.go
```diff
@@ -63,8 +85,22 @@ func (fb *TermsFacetBuilder) Field() string {
...
 func (fb *TermsFacetBuilder) UpdateVisitor(term []byte) {
+	// Fast prefix check on []byte - zero allocation
+	if len(fb.prefixBytes) > 0 && !bytes.HasPrefix(term, fb.prefixBytes) {
+		fb.total++
+		return
+	}
+
+	// Fast regex check on []byte - zero allocation
+	if fb.regex != nil && !fb.regex.Match(term) {
+		fb.total++
```
CascadingRadium: would it be correct to have `fb.total = total terms visited here`? even if the term doesn't pass the filter?

search/facet/facet_builder_terms.go
ajroetker: Fixed!

search.go
ajroetker: Fixed!

search/facet/facet_builder_terms.go
ajroetker: I moved it to the top of the function and added a comment about why it's needed!

PR #2246, opened 2025-11-13
MB-60401: Optimizations for score fusion

capemox:
Optimizes rrf and rsf.
- In place updates of scores instead of extra maps
- Extra sorts removed

Original benchmarks:
```
BenchmarkRescorerRRF/rrf/small-14         	   28918	     41116 ns/op	   59655 B/op	     325 allocs/op
BenchmarkRescorerRRF/rrf/medium-14        	    5394	    208945 ns/op	  250594 B/op	    1277 allocs/op
BenchmarkRescorerRRF/rrf/large-14         	    1411	    823237 ns/op	 1129037 B/op	    5085 allocs/op
BenchmarkRescorerRSF/rsf/small-14         	   41151	     29468 ns/op	   41060 B/op	      21 allocs/op
BenchmarkRescorerRSF/rsf/medium-14        	    9566	    128250 ns/op	  158735 B/op	      29 allocs/op
BenchmarkRescorerRSF/rsf/large-14         	    2215	    541930 ns/op	  743451 B/op	      84 allocs/op
```

Optimized benchmarks:
```
BenchmarkRescorerRRF/rrf/small-14         	   58382	     20003 ns/op	   13642 B/op	      10 allocs/op
BenchmarkRescorerRRF/rrf/medium-14        	   12789	     91741 ns/op	   47990 B/op	      10 allocs/op
BenchmarkRescorerRRF/rrf/large-14         	    2982	    387546 ns/op	  299895 B/op	      36 allocs/op
BenchmarkRescorerRSF/rsf/small-14         	   61227	     19480 ns/op	   11595 B/op	       9 allocs/op
BenchmarkRescorerRSF/rsf/medium-14        	   13282	     89511 ns/op	   39795 B/op	       9 allocs/op
BenchmarkRescorerRSF/rsf/large-14         	    3184	    409432 ns/op	  266798 B/op	      35 allocs/op
```

### DIFF
--- a/fusion/rrf.go
+++ b/fusion/rrf.go
@@ -50,6 +50,15 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
 		limit = windowSize
 	}
 
+	// precompute rank+scores to prevend additional division ops later
+	var rankReciprocals []float64
+	if limit > 0 {
+		rankReciprocals = make([]float64, limit)
+		for i := range rankReciprocals {
+			rankReciprocals[i] = 1.0 / float64(rankConstant+i+1)
+		}
+	}
+
 	// init explanations if required
 	var fusionExpl map[*search.DocumentMatch][]*search.Explanation
 	if explain {
@@ -67,7 +76,7 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
 			}
 			rank := i + 1
 			if explain {
-				contrib := ftsWeight / float64(rankConstant+rank)
+				contrib := ftsWeight * rankReciprocals[i]
 				expl := getFusionExplAt(
 					hit,
 					0,
@@ -77,7 +86,7 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
 				fusionExpl[hit] = append(fusionExpl[hit], expl)
 				hit.Score = contrib
 			} else {
-				hit.Score = ftsWeight / float64(rankConstant+rank)
+				hit.Score = ftsWeight * rankReciprocals[i]
 			}
 		}
 		for i := limit; i < len(hits); i++ {
@@ -100,16 +109,15 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
 			continue
 		}
 
-		sortDocMatchesByBreakdown(hits, queryIdx)
-
 		weight := weights[queryIdx+1]
 		rank := 0
+		sortDocMatchesByBreakdown(hits, queryIdx)
 		for _, hit := range hits {
 			if _, ok := scoreBreakdownForQuery(hit, queryIdx); !ok {
 				break
 			}
 			rank++
-			contrib := weight / float64(rankConstant+rank)
+			contrib := weight * rankReciprocals[rank-1]
 			if explain {
 				expl := getFusionExplAt(
 					hit,
--- a/fusion/rrf.go
+++ b/fusion/rrf.go
@@ -17,27 +17,9 @@ package fusion
 
 import (
 	"fmt"
-	"sort"
 
 	"github.com/blevesearch/bleve/v2/search"
 )
-
-// sortDocMatchIdxsByScore orders the provided index slice in-place based on the
-// corresponding hit scores, breaking ties by `HitNumber`.
-func sortDocMatchIdxsByScore(hits search.DocumentMatchCollection, idxs []int) {
-	if len(idxs) < 2 {
-		return
-	}
-
-	sort.Slice(idxs, func(a, b int) bool {
-		left := hits[idxs[a]]
-		right := hits[idxs[b]]
-		if left.Score == right.Score {
-			return left.HitNumber < right.HitNumber
-		}
-		return left.Score > right.Score
-	})
-}
 
 // formatRRFMessage builds the explanation string for a single component of the
 // Reciprocal Rank Fusion calculation.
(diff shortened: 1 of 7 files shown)


fusion/rrf.go
```diff
@@ -36,90 +40,113 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
...
+					formatRRFMessage(ftsWeight, rank, rankConstant),
+				)
+				fusionExpl[hit] = append(fusionExpl[hit], expl)
+				hit.Score = contrib
+			} else {
+				hit.Score = ftsWeight * rankReciprocals[i]
 			}
 		}
+		for i := limit; i < len(hits); i++ {
+			hits[i].Score = 0.0
```
abhinavdangeti, line 87: Can you throw in a comment here and below in the else clause on why we should do this? Can we not simply truncate the hits slice instead to just drop these hits whose score is being overwritten to 0?

fusion/rsf.go
```diff
@@ -36,119 +41,126 @@ func RelativeScoreFusion(hits search.DocumentMatchCollection, weights []float64,
...
-			// create and replace new explanation
-			expl := getFusionExplAt(
-				hit,
-				0,
-				tempRsfScore,
-				formatRSFMessage(weights[0], tempRsfScore, min, max),
-			)
-			explMap[hit.ID] = append(explMap[hit.ID], expl)
+		for i := ftsLimit; i < len(hits); i++ {
+			hits[i].Score = 0.0
```
abhinavdangeti, line 92: ditto

abhinavdangeti (commented):
Great numbers there^ @capemox , while I don't have any significant objections - I have put in a couple of inline comments that deserve your attention.

fusion/rrf.go
capemox: Before fusion, `Score` holds the FTS score and `ScoreBreakdown` holds the knn scores. During RRF/RSF, the final scores are placed in the `Score` field, so if a hit goes beyond the `score_window_size`, the FTS score of that is is inconsequential and so will be set 0. If the doc has a knn score though, it will get added to the score field.

I tried working with maps to prevent this loop, but that led to more memory and latency instead.

fusion/rsf.go
capemox: Same as above

rescorer.go
```diff
@@ -95,28 +95,26 @@ func (r *rescorer) restoreSearchRequest() {
...
-		fusionResult = &res
 	case ScoreRSF:
-		res := fusion.RelativeScoreFusion(
+		fusionResult = fusion.RelativeScoreFusion(
 			mergedHits,
 			r.origBoosts,
 			r.req.Params.ScoreWindowSize,
 			numKNNQueries(r.req),
 			r.req.Explain,
 		)
```
abhinavdangeti, line 117: Do we not not need to handle a default case here? We'll be returning 0, 0, 0 in the event scoring scheme is unsupported.

fusion/rrf.go
abhinavdangeti: Ok fair, let's add a comment here and below explaining what you just said for why we're resetting the score.

rescorer.go
capemox: `IsScoreFusionRequested` will handle this:
https://github.com/blevesearch/bleve/blob/58e81c8cc22d5aaefb399c14d209d29a496c809e/search.go#L709-L717

If score is anything else it defaults to normal hybrid search.

abhinavdangeti (dismissed):

fusion/rrf.go
```diff
@@ -17,16 +17,20 @@ package fusion
...
 	return fmt.Sprintf("rrf score (weight=%.3f, rank=%d, rank_constant=%d), normalized score of", weight, rank, rankConstant)
 }
 
-// ReciprocalRankFusion performs a reciprocal rank fusion on the search results.
+// ReciprocalRankFusion applies Reciprocal Rank Fusion across the primary FTS
+// results and each KNN sub-query. Ranks are limited to `windowSize` per source,
+// weighted, and combined into a single fused score, with optional explanation
+// details.
 func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64, rankConstant int, windowSize int, numKNNQueries int, explain bool) FusionResult {
 	if len(hits) == 0 {
```
CascadingRadium: i see `len(hits)` being called many times in this method:
consider - 

```
	n := len(hits)
	if n == 0 {
		return FusionResult{Hits: hits, Total: 0, MaxScore: 0}
	}
```

fusion/rrf.go
```diff
@@ -36,90 +40,110 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
...
-	sort.Slice(hits, func(a, b int) bool {
-		return scoreSortFunc()(hits[a], hits[b]) < 0
-	})
-	// Only consider top windowSize docs for rescoring
-	for i := range min(windowSize, len(hits)) {
-		if hits[i].Score != 0.0 {
-			// Skip if Score is 0, since that means the document was not
-			// found as part of FTS, and only in KNN.
-			docRanks[hits[i].ID][0] = i + 1
+	// precompute rank+scores to prevent additional division ops later
```
CascadingRadium, line 46: // Precompute reciprocals

fusion/rrf.go
```diff
@@ -36,90 +40,110 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
...
+	var rankReciprocals []float64
+	if limit > 0 {
```
CascadingRadium: this check of limit > 0 seems redundant.
limit can only be 0 if len(hits) > 0 and windowSize = 0. Can u just 
have a combined check in L31?
```
	n := len(hits)
	if n == 0 || windowSize == 0 {
		return FusionResult{Hits: nil, Total: 0, MaxScore: 0}
	}
```

CascadingRadium: u can just have below then

```
	rankReciprocals := make([]float64, limit)
	for i := 0; i < limit; i++ {
		rankReciprocals[i] = 1.0 / float64(rankConstant+i+1)
	}
```

fusion/rrf.go
```diff
@@ -36,90 +40,110 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
...
-				knnDocs = append(knnDocs, hit)
+	// Calculate fts rank+scores
+	if limit > 0 {
+		ftsWeight := weights[0]
+		for i := 0; i < limit; i++ {
+			hit := hits[i]
+			originalScore := hit.Score
+			if originalScore == 0.0 {
+				break
+			}
```
CascadingRadium: originalScore is unused -> stick to 

```
		if hit.Score == 0 {
			break
		}
```

fusion/rrf.go
```diff
@@ -36,90 +40,110 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
...
+					hit,
+					0,
+					contrib,
+					formatRRFMessage(ftsWeight, rank, rankConstant),
+				)
+				fusionExpl[hit] = append(fusionExpl[hit], expl)
+				hit.Score = contrib
+			} else {
+				hit.Score = ftsWeight * rankReciprocals[i]
 			}
```
CascadingRadium: simplify and remove `rank := i + 1` as u can send `i+1` directly

```
		contrib := ftsWeight * rankReciprocals[i]

		if explain {
			expl := getFusionExplAt(
				hit,
				0,
				contrib,
				formatRRFMessage(ftsWeight, i+1, rankConstant),
			)
			fusionExpl[hit] = append(fusionExpl[hit], expl)
		}
		hit.Score = contrib
	
```

fusion/rrf.go
```diff
@@ -36,90 +40,110 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
...
+	// Code from here is to calculate knn ranks and scores
+	// iterate over each knn query and calculate knn rank+scores
+	for queryIdx := 0; queryIdx < numKNNQueries; queryIdx++ {
+		limit := len(hits)
+		if windowSize >= 0 && windowSize < limit {
+			limit = windowSize
+		}
+		if limit == 0 {
+			continue
+		}
```
CascadingRadium: this is redundant right. u already calculated limit once above

CascadingRadium: also limit == 0 can never occur here as u ensure limit > 0 as per my above comment of early exiting when window size = 0.

fusion/rrf.go
```diff
@@ -36,90 +40,110 @@ func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64
...
-			docRanks[knnDocs[j].ID][i+1] = j + 1
+		weight := weights[queryIdx+1]
+		rank := 0
+		sortDocMatchesByBreakdown(hits, queryIdx)
+		for _, hit := range hits {
+			if _, ok := scoreBreakdownForQuery(hit, queryIdx); !ok {
+				break
+			}
+			rank++
+			contrib := weight * rankReciprocals[rank-1]
```
CascadingRadium:
```
        contrib := weight * rankReciprocals[rank]
        rank++
```

fusion/rrf.go
```diff
@@ -17,110 +17,122 @@ package fusion
...
+// ReciprocalRankFusion applies Reciprocal Rank Fusion across the primary FTS
+// results and each KNN sub-query. Ranks are limited to `windowSize` per source,
+// weighted, and combined into a single fused score, with optional explanation
+// details.
 func ReciprocalRankFusion(hits search.DocumentMatchCollection, weights []float64, rankConstant int, windowSize int, numKNNQueries int, explain bool) FusionResult {
-	if len(hits) == 0 {
+	nHits := len(hits)
+	if nHits == 0 || windowSize == 0 {
 		return FusionResult{
 			Hits:     hits,
```
CascadingRadium: if windowSize is 0 then the FusionResult should not have any hits right?
```
return FusionResult{
			Hits:     nil,
```

capemox: Missed this, thanks

rescorer.go
CascadingRadium: this creates 3 structs
- var fusionResult fusion.FusionResult
- return value inside fusion.ReciprocalRankFusion
- fusionResult as the capture value (copies the return value)

Can u not have `fusion.ReciprocalRankFusion` and `fusion.RelativeScoreFusion` to return a pointer to a struct?

abhinavdangeti (approved):

CascadingRadium (approved):

PR #2249, opened 2025-11-20
Updates to query-openapi-spec for boolean `filter` clause, vectors.md

abhinavdangeti:
- Improved OpenAPI schema definitions for BooleanQuery, ConjunctionQuery, and DisjunctionQuery to properly specify that query clauses can be single queries or arrays
- Added version v2.5.6 to the FAISS compatibility table in vectors.md


docs/query-openai-spec.yaml
```diff
@@ -49,9 +49,9 @@ info:
 
     ## Common Properties
 
-  All queries support:
-    - **boost**: Query boost value (float, default 1.0)
-    - **field**: Field to search (string, optional)
+    All queries support:
+      - **boost**: Query boost value (float, default 1.0)
```
ajroetker: Fyi ** here and the following line breaks parsing for some openapi spec parsers

CascadingRadium (approved):

PR #2252, opened 2025-11-26
Minor code fixes

CascadingRadium:
- `ScoreBreakdown` map was always getting created, and not being reused. When recycling the `DocumentMatch`, clear out the score breakdown map and reuse it when needed.
- Fix dropped error when creating pre-filter searcher.
- Close KNN and Pre-filter Searchers when done.

### DIFF
--- a/search/query/boolean.go
+++ b/search/query/boolean.go
@@ -15,7 +15,6 @@
 package query
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


search_knn.go
```diff
@@ -416,6 +419,11 @@ func (i *indexImpl) runKnnCollector(ctx context.Context, req *SearchRequest, rea
 			return nil, err
 		}
 		knnFilterResults[idx] = filterColl.EligibleSelector()
+		// Close the filter searcher, as we are done with it.
+		err = filterSearcher.Close()
```
abhinavdangeti, line 427: Better to defer this upon creation after the err handling.

search_knn.go
```diff
@@ -449,7 +466,7 @@ func (i *indexImpl) runKnnCollector(ctx context.Context, req *SearchRequest, rea
 	if err != nil {
 		return nil, err
 	}
-	return knnHits, nil
+	return knnHits, err
```
abhinavdangeti: Leaving this as `return knnHits, nil` is more readable.

abhinavdangeti (approved):

Likith101 (approved):

PR #2253, opened 2025-12-01
(feat) Add dynamic templates for customizing dynamic field mappings

ajroetker:
This implements Elasticsearch-style dynamic templates that allow users to define rules for how dynamically detected fields are mapped. This allows indexes where schema not be know ahead of time to dynamically use the full power of bleve mappings, for instance wrappers could define rules for creating geo fields based on pathing or field name. 

Templates can match based on:

- Field name patterns (using glob matching with * and **)
- Full path patterns (supporting ** for multi-level matching)
- Detected data type (string, number, boolean, date)
- Exclusion patterns (unmatch, path_unmatch)

Notes:
- Templates are defined at the DocumentMapping level via DynamicTemplates field
- Templates are inherited from parent mappings but can be overridden
- First matching template wins (evaluated in order)
- Templates support all FieldMapping options (analyzer, store, index, etc.)
- Global dynamic settings (StoreDynamic, IndexDynamic, DocValuesDynamic) are applied as defaults when templates don't explicitly set them

Example usage:
```go
mapping := NewIndexMapping()
mapping.DefaultMapping.AddDynamicTemplate(
    NewDynamicTemplate("keyword_fields").
        MatchField("*_keyword").
        MatchType("string").
        WithMapping(&FieldMapping{
            Type:     "text",
            Analyzer: "keyword",
        }),
)
```

Uses github.com/bmatcuk/doublestar/v4 for glob pattern matching with ** support.


abhinavdangeti:
Hey @ajroetker - thanks for this contribution again. I believe it's a useful feature - let me add it to v2.6.0 for now. However, after the team evaluates all the work we're currently tracking - we may reorganize some into separate releases as needed.

PR #2255, opened 2025-12-02
MB-55637: Added new error types in scorch

capemox:
This error is called from the merger or persister when creation of a new zap file on disk is unsuccessful.

This helps filter out persistence errors and handle them in a different way compared to the other async errors.

### DIFF
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -107,7 +107,7 @@ const (
 type ScorchError struct {
 	Source  string
 	ErrMsg  string
-	ErrType error
+	ErrType ScorchErrorType
 }
 
 func (e *ScorchError) Error() string {
@@ -121,7 +121,7 @@ func (e *ScorchError) Unwrap() error {
 	return e.ErrType
 }
 
-func NewScorchError(source, errMsg string, errType error) error {
+func NewScorchError(source, errMsg string, errType ScorchErrorType) error {
 	return &ScorchError{
 		Source:  source,
 		ErrMsg:  errMsg,


index/scorch/scorch.go
```diff
@@ -98,6 +98,16 @@ func (e *AsyncPanicError) Error() string {
 	return fmt.Sprintf("%s panic when processing %s", e.Source, e.Path)
 }
 
+// AsyncPersistError is an error that occurs when there is an issue with persisting segments to disk
+// It is also passed to scorch asyncErrorHandler in case of a persistence issue
+type AsyncPersistError struct {
```
abhinavdangeti: Wondering if we could do one better - and call this OutOfDiskError or something and fire only when `errors.Is(err, syscall.ENOSPC)` or whichever the error is that GO returns.

abhinavdangeti: Whether you choose to do this or not, ought to drop the "Async" word from the name above :)

capemox: The issue here is that bbolt doesn't return error codes, so checks like `errors.Is(err, syscall.ENOSPC)` always fail. Instead bbolt returns an error string only.

We _can_ string match on the error string, but that's very fragile.

abhinavdangeti: Ya let's not do string matches.

index/scorch/merge.go
```diff
@@ -110,7 +110,9 @@ OUTER:
 						ctrlMsg = nil
 						break OUTER
 					}
-					s.fireAsyncError(fmt.Errorf("merging err: %v", err))
+					s.fireAsyncError(&PersistError{
+						errMsg: fmt.Sprintf("merging err: %v", err),
+					})
```
Likith101: You could have helper functions to use this error

Something like
```
func NewPersistError(format string, args ...{}) error {
    return fmt.Errorf("%w", &PersistError{
		errMsg: fmt.Sprintf(format, args...),
	})
}
```

index/scorch/scorch.go
```diff
@@ -88,14 +89,39 @@ type Scorch struct {
...
-type AsyncPanicError struct {
-	Source string
-	Path   string
+// ErrType values for ScorchError
+var (
+	ErrAsyncPanic   = errors.New("async panic error")
+	ErrPersist      = errors.New("persist error")
+	ErrCleanup      = errors.New("cleanup error")
+	ErrOptionsParse = errors.New("options parse error")
+)
```
CascadingRadium, line 103: can be done as you are embedding the error in a string so better to keep a new type that impl's the `Error()` interface

```go
type ScorchErrType string

func (t ScorchErrType) Error() string { return string(t) }

const (
    ErrAsyncPanic   = ScorchErrType("AsyncPanic")
    ErrPersist      = ScorchErrType("Persist")
    ErrCleanup      = ScorchErrType("Cleanup")
    ErrOptionsParse = ScorchErrType("OptionsParse")
)
```

index/scorch/scorch.go
```diff
@@ -88,14 +89,39 @@ type Scorch struct {
...
-func (e *AsyncPanicError) Error() string {
-	return fmt.Sprintf("%s panic when processing %s", e.Source, e.Path)
+// Lets the onAsyncError function verify what type of
+// error is fired using errors.Is(...). This lets the function
+// handle errors differently.
+func (e *ScorchError) Unwrap() error {
+	return e.ErrType
+}
+
+func NewScorchError(source, errMsg string, errType error) error {
```
CascadingRadium: you can have the following API signature, why are you taking errMsg as a string, take it as error and handle the string internally, more robust

```go
func NewScorchError(source string, msg error, t ScorchErrType) error {
    return &ScorchError{
        Source:  source,
        ErrMsg:  msg.Error(),
        ErrType: t,
    }
}
```

index/scorch/persister.go
```diff
@@ -95,10 +95,11 @@ type notificationChan chan struct{}
 func (s *Scorch) persisterLoop() {
 	defer func() {
 		if r := recover(); r != nil {
-			s.fireAsyncError(&AsyncPanicError{
-				Source: "persister",
-				Path:   s.path,
-			})
+			s.fireAsyncError(NewScorchError(
+				"persister",
```
CascadingRadium: keep a const's for errSource

```go
const (
    SourcePersister  = "persister"
    SourceMerger     = "merger"
    SourceIntroducer = "introducer"
)
```

index/scorch/merge.go
```diff
@@ -110,7 +115,12 @@ OUTER:
 						ctrlMsg = nil
 						break OUTER
 					}
-					s.fireAsyncError(fmt.Errorf("merging err: %v", err))
+
+					s.fireAsyncError(NewScorchError(
+						"merger",
+						fmt.Sprintf("merging err: %v", err),
```
CascadingRadium, line 123: you can just pass error here `fmt.Errorf("merging err: %w", err))`
and make NewScorchError do the unwrap

index/scorch/introducer.go
```diff
@@ -50,10 +50,11 @@ type epochWatcher struct {
 func (s *Scorch) introducerLoop() {
 	defer func() {
 		if r := recover(); r != nil {
-			s.fireAsyncError(&AsyncPanicError{
-				Source: "introducer",
-				Path:   s.path,
-			})
+			s.fireAsyncError(NewScorchError(
+				"introducer",
+				fmt.Sprintf("path: %s", s.path),
```
CascadingRadium: should we include `r` the panic value in the error message ? like itll be useful for debugging i feel. `r` will be whatever argument was given to `panic()`

like if `panic("CR")` then r is `CR`

index/scorch/scorch.go
```diff
@@ -88,14 +88,51 @@ type Scorch struct {
...
 
-// AsyncPanicError is passed to scorch asyncErrorHandler when panic occurs in scorch background process
-type AsyncPanicError struct {
-	Source string
-	Path   string
+const (
+	SourceIntroducer = "introducer"
+	SourcePersister  = "persister"
+	SourceMerger     = "merger"
+)
```
abhinavdangeti: Quick nit :)

Let's declare each of these consts in the files they're pertinent to.
Is there a reason these need to be exported? Simply `const introducer = "introducer"` etc. would suffice?

abhinavdangeti (approved):
@capemox 👍🏼

PR #2258, opened 2025-12-03
MB-69641: Add support for retrieval of multi-vector documents

CascadingRadium:
- KNN search can now be performed on documents that contain a field with multiple vectors (e.g., vectors in JSON object arrays or arrays of vectors). For such documents, we select only the best-scoring vector and discard the rest, assigning the document the similarity score of that best vector.
- Added unit tests to verify this behavior.
- Fixed a bug where using a filter query that marked every document as ineligible in a multi-vector query could corrupt the internal state of the KNN collector.
- Fixed a bug where a KNN query was incorrectly marked as non-fieldable due to not adhering to the `FieldableQuery` interface.
- Requires: https://github.com/blevesearch/zapx/pull/348

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/blevesearch/zapx/v13 v13.4.2
 	github.com/blevesearch/zapx/v14 v14.4.2
 	github.com/blevesearch/zapx/v15 v15.4.2
-	github.com/blevesearch/zapx/v16 v16.2.7
+	github.com/blevesearch/zapx/v16 v16.2.8
 	github.com/couchbase/moss v0.2.0
 	github.com/spf13/cobra v1.8.1
 	go.etcd.io/bbolt v1.4.0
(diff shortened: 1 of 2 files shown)


search/collector/knn.go
```diff
@@ -95,6 +95,61 @@ func (c *collectStoreKNN) Final(fixup collectorFixup) (search.DocumentMatchColle
...
+// knnMerger is a preprocessing layer on top of collectStoreKNN that merges
+// duplicate document matches before adding them to the underlying KNN store.
+// Since KNN searchers may return the same document multiple times (with different
+// scoreBreakdown values for each KNN query), we must merge these to retain the
+// best score per KNN query before adding them to the KNN store.
+// NOTE: This implementation assumes documents arrive in increasing order of their
+// internal IDs, which is guaranteed by all searchers in bleve.
+type knnMerger struct {
+	// curr holds the current document match being accumulated during the merge process.
+	curr *search.DocumentMatch
```
abhinavdangeti: Are we certain hits will come in sorted by name? As we seem to be comparing only the previous document with the current.

CascadingRadium: yes, this is a critical detail that must be satisfied by all searchers. this assumption is made in hierarchy search as well, and was documented as a requirement when the `ConjunctionSearcher` was first introduced. More here : https://github.com/blevesearch/bleve/blob/knnDup/index/scorch/README.md

Likith101 (approved):

PR #2259, opened 2025-12-03
MB-69654: Fix vector field alias validation

CascadingRadium:
- Fix vector field alias validation logic to check fully qualified field names instead of just the name of the field mapping.
- Fix race condition that could occur due to validation method modifying the vector field mapping, which could be a shared object used in multiple document mapping objects. We now ensure that the field mapping is a read only object, and we pick the defaults during processing instead of writing the defaults into the field mapping object.
- Disallow vector alias creation if the `vector index optimization` flag is different.


abhinavdangeti:
@CascadingRadium unit test(s) please.

PR #2260, opened 2025-12-04
MB-69655: Fix vector normalization to handle multi-vectors correctly

CascadingRadium:
- When indexing multi-vector fields (e.g., `[[3,0,0], [0,4,0]]`) with `cosine` similarity, normalization was incorrectly applied to the entire flattened array instead of each sub-vector independently, resulting in degraded similarity scores.
- Added `NormalizeMultiVector(vec, dims)` that normalizes each sub-vector separately, fixing scores for multi-vector documents (e.g., score now correctly returns 1.0 instead of 0.6 for exact matches).


abhinavdangeti:
Ah ok, let's merge this only after that goes in then.

mapping/mapping_vectors.go
```diff
@@ -292,11 +296,33 @@ func validateVectorFieldAlias(field *FieldMapping, path []string,
...
+func NormalizeMultiVector(vec []float32, dims int) []float32 {
+	if len(vec) == 0 || dims <= 0 || len(vec)%dims != 0 {
+		return vec
+	}
+	// Single vector - delegate to NormalizeVector
+	if len(vec) == dims {
+		return NormalizeVector(vec)
+	}
+	// Multi-vector - make a copy to avoid modifying the original
+	result := slices.Clone(vec)
```
abhinavdangeti, line 322: If we make this copy here, we should avoid re-cloning within NormalizeVector again in this path.

abhinavdangeti: Oh nevermind, you're bypassing `NormalizeVector`.

PR #2262, opened 2025-12-05
Fix String API for SearchResult

CascadingRadium:
- `SearchResult.String()` previously printed only the total hit count unless the original `SearchRequest` was present (requiring `request.Explain = true`)
- The new implementation removes this dependency and always prints full hit details when available.
- Output now consistently includes IDs, scores, fragments, fields, and decoded sort values.

Before patch

```bash
$ go run -tags=vectors main.go 
Search Result: 
 1 matches, took 617.5µs
```

After patch

```bash
$ go run -tags=vectors main.go 
Search Result: 
 1 matches, took 548.125µs
    1. example (0.403426)
        text
                hello from united states
```

### DIFF
--- a/search.go
+++ b/search.go
@@ -591,7 +591,7 @@ func (sr *SearchResult) String() string {
 			fmt.Fprintf(rv, "%d matches, took %s\n", sr.Total, sr.Took)
 		}
 	} else {
-		fmt.Fprintf(rv, "No matches, took %s\n", sr.Took)
+		fmt.Fprintf(rv, "No matches\n")
 	}
 	if len(sr.Facets) > 0 {
 		fmt.Fprintf(rv, "Facets:\n")


search.go
```diff
@@ -571,26 +571,50 @@ func (sr *SearchResult) Size() int {
...
+		for otherFieldName, otherFieldValue := range hit.Fields {
+			if _, ok := hit.Fragments[otherFieldName]; !ok {
+				rv += fmt.Sprintf("\t%s\n", otherFieldName)
+				rv += fmt.Sprintf("\t\t%v\n", otherFieldValue)
+			}
+		}
+		if len(hit.DecodedSort) > 0 {
+			rv += "\t_sort: ["
+			for i, v := range hit.DecodedSort {
+				if i > 0 {
```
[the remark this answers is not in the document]
abhinavdangeti: Good suggestion, `for k, v`

search.go
```diff
@@ -571,26 +571,50 @@ func (sr *SearchResult) Size() int {
 }
 
 func (sr *SearchResult) String() string {
-	rv := ""
+	// Helper to format one hit
+	formatHit := func(i int, hit *search.DocumentMatch, start int) string {
```
abhinavdangeti: Don't see the need for this being a closure - wouldn't it be faster to declare this outside as a utility?

abhinavdangeti (approved):

search.go
```diff
@@ -572,25 +572,22 @@ func (sr *SearchResult) Size() int {
...
 	rv := ""
+	// main header
 	if sr.Total > 0 {
-		if sr.Request != nil && sr.Request.Size > 0 {
-			rv = fmt.Sprintf("%d matches, showing %d through %d, took %s\n", sr.Total, sr.Request.From+1, sr.Request.From+len(sr.Hits), sr.Took)
+		switch {
+		case sr.Request != nil && sr.Request.Size > 0:
+			start := sr.Request.From
+			end := sr.Request.From + len(sr.Hits)
+			rv = fmt.Sprintf("%d matches, showing %d through %d, took %s\n", sr.Total, start+1, end, sr.Took)
```
capemox: Instead of `%d matches` which sounds wrong for 1 hit (`1 matches`), maybe can have `count: %d` instead?

search.go
```diff
@@ -572,25 +572,22 @@ func (sr *SearchResult) Size() int {
...
-				for otherFieldName, otherFieldValue := range hit.Fields {
-					if _, ok := hit.Fragments[otherFieldName]; !ok {
-						rv += fmt.Sprintf("\t%s\n", otherFieldName)
-						rv += fmt.Sprintf("\t\t%v\n", otherFieldValue)
-					}
-				}
+				rv += formatHit(i, hit, start)
+			}
+		case sr.Request == nil:
+			rv = fmt.Sprintf("%d matches, took %s\n", sr.Total, sr.Took)
```
capemox: Here too

search.go
abhinavdangeti: @capemox That would be syntax change, perhaps best retaining it to what it was.

search.go
```diff
@@ -572,25 +572,22 @@ func (sr *SearchResult) Size() int {
 
 func (sr *SearchResult) String() string {
 	rv := ""
+	// main header
```
Likith101: Unnecessary comment

search.go
```diff
@@ -572,25 +572,22 @@ func (sr *SearchResult) Size() int {
...
 	if sr.Total > 0 {
-		if sr.Request != nil && sr.Request.Size > 0 {
-			rv = fmt.Sprintf("%d matches, showing %d through %d, took %s\n", sr.Total, sr.Request.From+1, sr.Request.From+len(sr.Hits), sr.Took)
+		switch {
+		case sr.Request != nil && sr.Request.Size > 0:
+			start := sr.Request.From
```
Likith101: init to start + 1

search.go
```diff
@@ -572,25 +572,22 @@ func (sr *SearchResult) Size() int {
...
-						rv += fmt.Sprintf("\t\t%s\n", fragment)
-					}
-				}
-				for otherFieldName, otherFieldValue := range hit.Fields {
-					if _, ok := hit.Fragments[otherFieldName]; !ok {
-						rv += fmt.Sprintf("\t%s\n", otherFieldName)
-						rv += fmt.Sprintf("\t\t%v\n", otherFieldValue)
-					}
-				}
+				rv += formatHit(i, hit, start)
```
capemox: Instead of the repeated string concat in these places, will be better to use `strings.Builder`

Likith101: change to use only the hit and hit number and avoid start + i + 1 inside the function

PR #2268, opened 2025-12-22
MB-59633: Improve performance of Geospatial Search

CascadingRadium:
- Reuse `DocumentMatch` objects in `FilteringSearcher` to avoid triggering `DocumentMatchPoolTooSmall` when documents fail the filter function and are not recycled, which previously caused repeated `DocumentMatch` allocations.
- Move the `DocValueVisitor` closure out of the hot path to avoid per-document heap allocations caused by closure escape; the closure is now allocated once and reused, eliminating `runtime.newobject` calls in the critical loop.
- Avoid unnecessary slice creation in the `DocValuesVisitor`.
- Reduce garbage created when sorting by `GeoDistance` by avoiding unnecessary string and slice creation and by reusing byte buffers.

### DIFF
--- a/index/scorch/snapshot_index.go
+++ b/index/scorch/snapshot_index.go
@@ -796,10 +796,6 @@ func (is *IndexSnapshot) documentVisitFieldTermsOnSegment(
 	// Filter out fields that have been completely deleted or had their
 	// docvalues data deleted from both visitable fields and required fields
 	filterUpdatedFields := func(fields []string) []string {
-		// fast path: if no updatedFields just return the input
-		if len(is.updatedFields) == 0 {
-			return fields
-		}
 		filteredFields := make([]string, 0, len(fields))
 		for _, field := range fields {
 			if info, ok := is.updatedFields[field]; ok &&
@@ -811,8 +807,10 @@ func (is *IndexSnapshot) documentVisitFieldTermsOnSegment(
 		return filteredFields
 	}
 
-	fields = filterUpdatedFields(fields)
-	vFields = filterUpdatedFields(vFields)
+	if len(is.updatedFields) > 0 {
+		fields = filterUpdatedFields(fields)
+		vFields = filterUpdatedFields(vFields)
+	}
 
 	var errCh chan error
 


index/scorch/snapshot_index.go
```diff
@@ -832,8 +834,8 @@ func (is *IndexSnapshot) documentVisitFieldTermsOnSegment(
 		}
 	}
 
-	if ssvOk && ssv != nil && len(vFieldsFiltered) > 0 {
-		dvs, err = ssv.VisitDocValues(localDocNum, fieldsFiltered, visitor, dvs)
+	if ssvOk && ssv != nil && len(vFields) > 0 {
```
abhinavdangeti, line 837: As far as readability goes, it'd be good to check for `len(fields) > 0` if you're to be using `fields` within the conditional. Any situation `fields` could be empty when `vFields` isn't?

search/searcher/search_geopointdistance.go
```diff
@@ -113,27 +113,33 @@ func boxSearcher(ctx context.Context, indexReader index.IndexReader,
 	return boxSearcher, nil
 }
 
-func buildDistFilter(ctx context.Context, dvReader index.DocValueReader, field string,
+func buildDistFilter(ctx context.Context, dvReader index.DocValueReader,
 	centerLon, centerLat, maxDist float64) FilterFunc {
+	// reuse the following for each document match that is checked using the filter
+	var lons, lats []float64
+	var found bool
+	dvVisitor := func(_ string, term []byte) {
```
abhinavdangeti, line 121: Can we extract this method out as a separate function into a `search/searcher/geo_utils.go` and use it here and above in `buildRectFilter` and below in `buildPolygonFilter`. There's avoidable code duplication.

CascadingRadium: Extracting `dvVisitor` into a shared function in `geo_utils.go` isn’t a good fit here.
`dvVisitor` is a closure over mutable state (`lons`, `lats`, `found`) that’s owned by the filter builder and reused across every document: the returned `FilterFunc` clears and reuses those slices and the found flag, then passes a visitor that updates them. That state is an intrinsic part of the filter's lifecycle, not a standalone `visitor` that can live in a util.

index/scorch/snapshot_index.go
CascadingRadium: This is what the existing code does, I have just refactored the variable names.

`vFields` is set of fields with docValues enabled = `A`.
`fields` is the set of fields for which docValues requested = `B`.
`cFields` is the subset of `fields` for which there are no docValues enabled but requested = `B - A`

This way the code ensures that we get the docValues of both 
- `B ∩ A` via `VisitDocValues` 
- `B - A` via the `cachedDocs` and `cFields`
which add up to `B`

Hence the check for `vFields > 0` is necessary here, it is there to check if `B ∩ A` is non nil, the actual `B ∩ A` is calculated inside the `VisitDocValues` API.

search/searcher/search_geopointdistance.go
abhinavdangeti: Fair enough.

index/scorch/snapshot_index.go
abhinavdangeti: Ack.

abhinavdangeti (approved):

PR #2272, opened 2026-01-16
MB-27666: Fix nested mode handling for field-implicit queries

CascadingRadium:
- Fix nested-mode search for field-implicit queries (`match_all`, `docID`): These queries previously filtered nested documents at the searcher level, but since `ExtractFields` returned no fields, the nested collector was never activated. In compound queries (e.g., Boolean filters), this caused a level mismatch—searcher filtering vs nested-level execution—resulting in zero matches.
- Unify filtering at collector level: Introduced `HasID()` and `HasAll()` flags in `FieldSet` to track field-implicit queries. Removed searcher-level filtering from `MatchAllSearcher`. 
- Nested collector now correctly activates, ensuring consistent behavior regardless of query composition.
- Fix an edge case where usage of `match_all query` as a sub-clause in a `NestedConjunctionSearcher` could cause a panic.
- Fix a case where root document fields would get excluded from the `_all` field even when included.
- Fix outdated information in `hierarchy.md`
- Add unit tests for Hierarchical Nested Vector Search.

### DIFF
--- a/search_knn_test.go
+++ b/search_knn_test.go
@@ -33,6 +33,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/blevesearch/bleve/v2/analysis/analyzer/keyword"
 	"github.com/blevesearch/bleve/v2/analysis/lang/en"
 	"github.com/blevesearch/bleve/v2/index/scorch"
 	"github.com/blevesearch/bleve/v2/mapping"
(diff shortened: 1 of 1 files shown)


search/util.go
```diff
@@ -278,24 +278,78 @@ type BM25Stats struct {
 	FieldCardinality map[string]int `json:"field_cardinality"`
 }
 
-// FieldSet represents a set of queried fields.
-type FieldSet map[string]struct{}
+// FieldSet represents a set of queried fields and related helper methods.
+type FieldSet interface {
```
abhinavdangeti: Don't quite see the need* for this interface - why not just make the argument be of type Struct.

CascadingRadium: 👍🏻

search/query/query.go
```diff
@@ -502,6 +502,11 @@ func ExtractFields(q Query, m mapping.IndexMapping, fs search.FieldSet) (search.
 				break
 			}
 		}
+	case *MatchAllQuery, *DocIDQuery:
+		if fs == nil {
+			fs = search.NewFieldSet()
+		}
+		fs.SetGlobal()
```
abhinavdangeti: MatchAll and DocID both go to the `_id` internal field - so they're not technically "field-agnostic".

search/util.go
```diff
@@ -279,23 +279,58 @@ type BM25Stats struct {
...
+}
+
+// AddField adds a field to the set.
+func (fs *FieldSet) AddField(field string) {
+	fs.fields[field] = struct{}{}
+}
+
+// SetGlobal sets the field-agnostic flag.
+func (fs *FieldSet) SetGlobal() {
+	fs.global = true
```
abhinavdangeti: Poor naming choice for field-agnostic flag - our definition of "field agnostic" search is going to the `_all` composite field and yet you're setting this for MatchAll and DocID which go to the `_id` internal field.

search/util.go
```diff
@@ -279,23 +279,58 @@ type BM25Stats struct {
...
+}
+
+// Global returns true if the set contains any field-agnostic queries,
```
abhinavdangeti: Need to refine this naming to avoid confusion here.

search/query/query.go
CascadingRadium: thanks for pointing this out, I think I can simplify the PR a lot more now.

abhinavdangeti (approved):

PR #2279, opened 2026-01-29
MB-62182: Implement File transfer APIs

Thejas-bhat:
- Utilizes the new file transfer APIs to able to transfer the trained index file from one bleve index to another. This is helpful in a multi-partition scenario with a single datasource - where only one bleve index is trained on a set of vectors and then transfer the generated file to the other bleve indexes. 
- The file transfer is currently focused around the trained index file but the protocol is such that its future extendable for other files if needed.
- The main control flow here is that the application needs to implement the `IndexDirectory` interface which maintains a reference to the destination bleve index. 
    - Now, when the `srcIndex.CopyFile(index.TrainedIndexFileName, directory)` is invoked, it internally fetches the writer on the destination using `directory.GetWriter()`
    - Then, as part of the CopyFile it does an `io.Copy()` of the specific file and currently this is focused only for trained index file.
    - This is followed by an update bolt transaction + commit and sync in the destination index by invoking `directory.SetPathInBolt(fileName, filePath)`. This is mandatory since bolt is reflective of the on-disk information.

### DIFF
--- a/index.go
+++ b/index.go
@@ -353,13 +353,6 @@ type IndexCopyable interface {
 	CopyTo(d index.Directory) error
 }
 
-// IndexFileCopyable is an index supporting the transfer of a single file between
-// two indexes
-type IndexFileCopyable interface {
-	SetPathInBolt(key []byte, value []byte) error
-	CopyFile(file string, d index.IndexDirectory) error
-}
-
 // FileSystemDirectory is the default implementation for the
 // index.Directory interface.
 type FileSystemDirectory string
--- a/index/scorch/scorch.go
+++ b/index/scorch/scorch.go
@@ -79,8 +79,6 @@ type Scorch struct {
 	persisterNotifier        chan *epochWatcher
 	rootBolt                 *bolt.DB
 	asyncTasks               sync.WaitGroup
-	// not a real searchable segment
-	centroidIndex *SegmentSnapshot
 
 	trainer trainer
 
--- a/index/scorch/train_vector.go
+++ b/index/scorch/train_vector.go
@@ -263,16 +263,19 @@ func (t *vectorTrainer) getInternal(key []byte) ([]byte, error) {
 
 func (t *vectorTrainer) getCentroidIndex(field string) (interface{}, error) {
 	// return the coarse quantizer of the centroid index belonging to the field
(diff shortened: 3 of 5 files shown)


index/scorch/scorch.go
```diff
@@ -112,6 +115,10 @@ type trainer interface {
 	loadTrainedData(*bolt.Bucket) error
 	// to fetch the internal data from the component
 	getInternal(key []byte) ([]byte, error)
+
+	// file transfer operations
```
abhinavdangeti: Let's be specific about which file transfer operations.

index/scorch/persister.go
```diff
@@ -909,14 +913,14 @@ func (s *Scorch) loadFromBolt() error {
...
-		if trainer := initTrainer(s, s.config); trainer != nil {
-			s.trainer = trainer
-			trainerBucket := snapshots.Bucket(util.BoltTrainerKey)
-			err := s.trainer.loadTrainedData(trainerBucket)
-			if err != nil {
-				return err
-			}
+		// try init trainer with the mapping details
+		if s.trainer == nil {
+			s.trainer = initTrainer(s, s.config)
```
maneuvertomars: could this panic? I see initTrainer() returns nil when fastmerge is off, but we still call s.trainer.loadTrainedData(...) further ahead. May be safer to do something like:
`if s.trainer == nil {
    s.trainer = initTrainer(s, s.config)
}
if s.trainer != nil {
    err := s.trainer.loadTrainedData(trainerBucket)
    if err != nil {
        return err
    }
}`

Thejas-bhat: yeah thanks, this bit just needed a rebase

index.go
```diff
@@ -353,6 +353,13 @@ type IndexCopyable interface {
 	CopyTo(d index.Directory) error
 }
 
+// IndexFileCopyable is an index supporting the transfer of a single file between
+// two indexes
+type IndexFileCopyable interface {
```
abhinavdangeti: Why this - I don't see it being used anywhere?

abhinavdangeti: For `cbft` I suppose.

abhinavdangeti: Can cbft also not use `index.IndexFileCopyable`?

Thejas-bhat: yeah true, this is not needed

PR #2282, opened 2026-01-31
MB-62985: Forcing cosine similarity for backing indexes within bq

Likith101:
- Forcing cosine similarity for binary indexes during indexing and query times
- Added relevant test cases

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,9 +5,9 @@ go 1.24.0
 require (
 	github.com/RoaringBitmap/roaring/v2 v2.14.4
 	github.com/bits-and-blooms/bitset v1.24.2
-	github.com/blevesearch/bleve_index_api v1.3.2
+	github.com/blevesearch/bleve_index_api v1.3.3-0.20260311083042-62a5343ee2d1
 	github.com/blevesearch/geo v0.2.5
-	github.com/blevesearch/go-faiss v1.0.27
+	github.com/blevesearch/go-faiss v1.0.28-0.20260311083224-baa72a4cd87f
 	github.com/blevesearch/go-metrics v0.0.0-20201227073835-cf1acfcdf475
 	github.com/blevesearch/go-porterstemmer v1.0.3
 	github.com/blevesearch/goleveldb v1.0.1
@@ -25,7 +25,7 @@ require (
 	github.com/blevesearch/zapx/v14 v14.4.3
 	github.com/blevesearch/zapx/v15 v15.4.3
 	github.com/blevesearch/zapx/v16 v16.3.1
-	github.com/blevesearch/zapx/v17 v17.0.2
+	github.com/blevesearch/zapx/v17 v17.0.3-0.20260311100439-7e7434b4f844
 	github.com/couchbase/moss v0.2.0
 	github.com/spf13/cobra v1.10.2
 	go.etcd.io/bbolt v1.4.0
--- a/go.sum
+++ b/go.sum
@@ -2,12 +2,12 @@ github.com/RoaringBitmap/roaring/v2 v2.14.4 h1:4aKySrrg9G/5oRtJ3TrZLObVqxgQ9f1zn
 github.com/RoaringBitmap/roaring/v2 v2.14.4/go.mod h1:oMvV6omPWr+2ifRdeZvVJyaz+aoEUopyv5iH0u/+wbY=
 github.com/bits-and-blooms/bitset v1.24.2 h1:M7/NzVbsytmtfHbumG+K2bremQPMJuqv1JD3vOaFxp0=
 github.com/bits-and-blooms/bitset v1.24.2/go.mod h1:7hO7Gc7Pp1vODcmWvKMRA9BNmbv6a/7QIWpPxHddWR8=
-github.com/blevesearch/bleve_index_api v1.3.2 h1:y4VLXBF7nQR01CvF+QzmCJKMpVPCLp1CJ5FsRSZXzRE=
-github.com/blevesearch/bleve_index_api v1.3.2/go.mod h1:xvd48t5XMeeioWQ5/jZvgLrV98flT2rdvEJ3l/ki4Ko=
+github.com/blevesearch/bleve_index_api v1.3.3-0.20260311083042-62a5343ee2d1 h1:k2Q/e830m+fHx/4djya4UFef/jCV53sr0s97n9ergEc=
+github.com/blevesearch/bleve_index_api v1.3.3-0.20260311083042-62a5343ee2d1/go.mod h1:xvd48t5XMeeioWQ5/jZvgLrV98flT2rdvEJ3l/ki4Ko=
 github.com/blevesearch/geo v0.2.5 h1:yJg9FX1oRwLnjXSXF+ECHfXFTF4diF02Ca/qUGVjJhE=
 github.com/blevesearch/geo v0.2.5/go.mod h1:Jhq7WE2K6mJTx1xS44M2pUO6Io+wjCSHh1+co3YOgH4=
-github.com/blevesearch/go-faiss v1.0.27 h1:7cBImYDDQ82WJd5RUZ1ie6zXztCsC73W94ZzwOjkatk=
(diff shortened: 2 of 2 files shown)


document/field_vector.go
```diff
@@ -114,6 +114,10 @@ func NewVectorFieldWithIndexingOptions(name string, arrayPositions []uint64,
 	// skip freq/norms for vector field
 	options |= index.SkipFreqNorm
 
+	if vectorIndexOptimizedFor == index.IndexOptimizedForBinary {
+		similarity = index.CosineSimilarity
```
CascadingRadium, line 121: - adding this here is a mistake, as the normalization happens upstream in the `processVector` and the `processVectorBase64` APIs.

```go
func (fm *FieldMapping) processVector(propertyMightBeVector interface{},
	pathString string, path []string, indexes []uint64, context *walkContext) bool {
	vector, ok := processVector(propertyMightBeVector, fm.Dims)
	// Don't add field to document if vector is invalid
	if !ok {
		return false
	}
	// Apply defaults for similarity and optimization if not set
	similarity := fm.Similarity
	if similarity == "" {
		similarity = index.DefaultVectorSimilarityMetric
	}
	vectorIndexOptimizedFor := fm.VectorIndexOptimizedFor
	if vectorIndexOptimizedFor == "" {
		vectorIndexOptimizedFor = index.DefaultIndexOptimization
	}
	// Binary optimization only supports cosine similarity
	if vectorIndexOptimizedFor == index.IndexOptimizedForBinary {
		similarity = index.CosineSimilarity
	}
```

- Add validation in the fieldAliasValidation step to check for this override.

```go
func validateVectorFieldAlias(field *FieldMapping, path []string,
	fieldAliasCtx map[string]*FieldMapping) error {
	// fully qualified field name
	pathString := encodePath(path)
	// check if field has a name set, else use path to compute effective name
	effectiveFieldName := getFieldName(pathString, path, field)
	// Compute effective values for validation
	effectiveSimilarity := field.Similarity
	if effectiveSimilarity == "" {
		effectiveSimilarity = index.DefaultVectorSimilarityMetric
	}
	effectiveOptimizedFor := field.VectorIndexOptimizedFor
	if effectiveOptimizedFor == "" {
		effectiveOptimizedFor = index.DefaultIndexOptimization
	}
	if effectiveOptimizedFor == index.IndexOptimizedForBinary {
		effectiveSimilarity = index.CosineSimilarity
	}
```

- Add relevant unit tests for ensuring override is done, in `TestVectorFieldAliasValidation` and in `search_knn_test.go`

document/field_vector.go
```diff
@@ -114,6 +114,10 @@ func NewVectorFieldWithIndexingOptions(name string, arrayPositions []uint64,
 	// skip freq/norms for vector field
 	options |= index.SkipFreqNorm
 
+	if vectorIndexOptimizedFor == index.IndexOptimizedWithBivfFlat {
+		similarity = index.CosineSimilarity
```
abhinavdangeti, line 121: Add commentary here, that this is for the backing index - as as default to hamming distance with Bivf.

mapping/mapping_vectors.go
```diff
@@ -151,6 +151,11 @@ func (fm *FieldMapping) processVector(propertyMightBeVector interface{},
 	if vectorIndexOptimizedFor == "" {
 		vectorIndexOptimizedFor = index.DefaultIndexOptimization
 	}
+	// BIVF-Flat index needs cosine similarity for correct scoring
+	// regardless of the similarity metric specified in the mapping.
```
abhinavdangeti: Update to indicate that this is for the backing flat (ditto above)

mapping/mapping_vectors.go
```diff
@@ -185,6 +190,11 @@ func (fm *FieldMapping) processVectorBase64(propertyMightBeVectorBase64 interfac
...
```
abhinavdangeti: ditto

mapping/mapping_vectors.go
```diff
@@ -234,6 +244,10 @@ func validateVectorFieldAlias(field *FieldMapping, path []string,
 	if effectiveOptimizedFor == "" {
 		effectiveOptimizedFor = index.DefaultIndexOptimization
 	}
+	if effectiveOptimizedFor == index.IndexOptimizedWithBivfFlat && field.Dims%8 != 0 {
+		return fmt.Errorf("field: '%s', vector dimensions for bivf-flat must be a multiple"+
```
abhinavdangeti: Update this to the same message - use a `const` for here and below. Maybe use - "incompatible vector dimensionality for BQ".

mapping/mapping_vectors.go
```diff
@@ -289,6 +303,11 @@ func validateVectorFieldAlias(field *FieldMapping, path []string,
 			effectiveOptimizedFor,
 			reflect.ValueOf(index.SupportedVectorIndexOptimizations).MapKeys())
 	}
+	// BIVF-Flat index optimization requires vector dimensions to be a multiple of 8
+	if effectiveOptimizedFor == index.IndexOptimizedWithBivfFlat && field.Dims%8 != 0 {
+		return fmt.Errorf("field: '%s', invalid vector dimension: %d for optimization with BIVF-Flat,"+
+			" dimension should be a multiple of 8", effectiveFieldName, field.Dims)
```
abhinavdangeti, line 307: See comment above.

search/query/knn.go
```diff
@@ -84,6 +84,9 @@ func (q *KNNQuery) Searcher(ctx context.Context, i index.IndexReader,
 	if q.K <= 0 || len(q.Vector) == 0 {
 		return nil, fmt.Errorf("k must be greater than 0 and vector must be non-empty")
 	}
+	if fieldMapping.VectorIndexOptimizedFor == index.IndexOptimizedWithBivfFlat {
+		similarityMetric = index.CosineSimilarity
```
abhinavdangeti, line 91: Related to comment earlier.

mapping/mapping_vectors.go
```diff
@@ -303,9 +301,9 @@ func validateVectorFieldAlias(field *FieldMapping, path []string,
 			effectiveOptimizedFor,
 			reflect.ValueOf(index.SupportedVectorIndexOptimizations).MapKeys())
 	}
-	// BIVF-Flat index optimization requires vector dimensions to be a multiple of 8
+	// bivf-flat's primary indexes requires vector dimensionality to be a multiple of 8
 	if effectiveOptimizedFor == index.IndexOptimizedWithBivfFlat && field.Dims%8 != 0 {
-		return fmt.Errorf("field: '%s', invalid vector dimension: %d for optimization with BIVF-Flat,"+
+		return fmt.Errorf("field: '%s', incompatible vector dimensionality for BIVF-FLAT: %d,"+
```
abhinavdangeti: @Likith101 this should be part of validation alone, while processing data - we _will_ simply ignore vectors that do not qualify.

document/field_vector.go
```diff
@@ -114,6 +114,13 @@ func NewVectorFieldWithIndexingOptions(name string, arrayPositions []uint64,
 	// skip freq/norms for vector field
 	options |= index.SkipFreqNorm
 
+	// bivf-flat indexes only supports hamming distance for the primary
```
CascadingRadium: this override is already added in `processVector` and `processVectorBase64` right? the additional override here feels redundant and fragile as well, 
since if we ever call `NewVectorFieldWithIndexingOptions` from outside the `processVector` and `processVectorBase64` APIs, we will do the override in this API but not normalize the vector

mapping/mapping_vectors.go
CascadingRadium: This will be bad UX in my opinion, we should have both the checks ideally.
- invalidate index creation request upon wrong dims
- reject ingest of vectors which do not match the specified dims in the validated index mapping (we do this already)

document/field_vector.go
Likith101: Similarity is still a field mapping option. It only makes sense to have it here as well as opposed to not having it, instead of downstream trying to figure out what metric to use for a bivf-flat index.

capemox (approved):

CascadingRadium (approved):

PR #2295, opened 2026-03-17
Unit test to repro index corruption

CascadingRadium:


index/scorch/scorch_test.go
```diff
@@ -2810,3 +2810,262 @@ func TestPersistorMergerOptions(t *testing.T) {
...
+				t.Fatalf("expected to find docID %s in segment, but not found", docID)
+			}
+		}
+	}
+	// persist the snapshot, which will create a new snapshot with 2 persisted segments
+	// we have to listen to the merge channel to know when the merge is done and the new snapshot is ready
+	var errCh = make(chan error, 1)
+	var doneCh = make(chan struct{})
+	go func() {
+		_, err := s.persistSnapshotMaybeMerge(snapshot, po)
```
Thejas-bhat, line 2918: this operation will create a snapshot with 1 segment right? because the 2 segments have been merged into 1 as part of the same flush operation. and we wait for this merge to finish at 2930.

index/scorch/scorch_test.go
```diff
@@ -2810,3 +2810,262 @@ func TestPersistorMergerOptions(t *testing.T) {
...
+	if lastSeg.deleted != nil && lastSeg.deleted.GetCardinality() != 0 {
+		t.Fatalf("expected no deletions in last segment, got %d", lastSeg.deleted.GetCardinality())
+	}
+	// the merge data should not be nil and should have 2 segments to merge
+	if mergeData == nil {
+		t.Fatal("expected merge data, got nil")
+	}
+	// now we apply the merge, which should result in a new snapshot
+	// with 1 segment that has both deletions applied
+	s.introduceMerge(mergeData)
```
Thejas-bhat, line 2992: this API just introduces the snapshot (with 1 merged segment, 2 of them going away) but with the new root's picture in mind right? at this point we know that the new root has 3 segments (2 old segments for doc1-4, 1 with doc5). the introduce merge basically changes the root to have 2 segments (1 merged segment doc1-4 which has deleted bitmap having 1 and 3; and 1 segment with doc5). the persister after it gets the notification that this introduction is successful (basically the downstream code of [this point](https://github.com/blevesearch/bleve/blob/0e05d0a710030f9895c5f0ab6ef5e7db6012839d/index/scorch/persister.go#L501)) tries to update bolt info but its view is strictly limited to what its supposed to have been according to the old snapshot when it started the process - which is why we set the deleted var to nil imo. the new deleted + doc 5 info is part of a newer snapshot which will update the bolt in the next iteration.

please let me know if that makes sense.

Thejas-bhat: also if the index is closed at this point, FTS handles these deletes by streaming the mutations from KV since the var `snapshot` (the old one) doesn't have the seq no. details related to the deletes + doc 5. when the index is open back, the `snapshot` is loaded into system and the deletes + doc 5 will steamed again. 


i further checked the root snapshot's live size right before the `idx.Close()` the value is = 3 (doc2, doc4 and doc5) which is as per expectation.

PR #2307, opened 2026-04-01
bug: use `index.TrainedIndexCallbackFn` signature 

Thejas-bhat:
- use the same func signature in the both zapx and bleve for fetching the trained index to avoid panics caused by differing func signatures

### DIFF
--- a/index/scorch/train_vector.go
+++ b/index/scorch/train_vector.go
@@ -79,7 +79,7 @@ func (t *vectorTrainer) trainLoop() {
 		t.parent.asyncTasks.Done()
 	}()
 	// initialize stuff
-	t.parent.segmentConfig[index.TrainedIndexCallback] = t.getCentroidIndex
+	t.parent.segmentConfig[index.TrainedIndexCallback] = index.TrainedIndexCallbackFn(t.getCentroidIndex)
 	path := filepath.Join(t.parent.path, index.TrainedIndexFileName)
 	for {
 		select {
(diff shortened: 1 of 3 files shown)


abhinavdangeti:
go mod updates.

index/scorch/train_vector.go
```diff
@@ -79,7 +79,7 @@ func (t *vectorTrainer) trainLoop() {
 		t.parent.asyncTasks.Done()
 	}()
 	// initialize stuff
-	t.parent.segmentConfig[index.TrainedIndexCallback] = t.getCentroidIndex
+	t.parent.segmentConfig[index.TrainedIndexCallback] = index.TrainedIndexCallbackFn(t.getCentroidIndex)
```
abhinavdangeti, line 82: Is typecast better than type assertion here?

Thejas-bhat: okay let me change it

abhinavdangeti: Make sure to test it then.

Thejas-bhat: actually the func t.getCentroidIndex isn't an interface to do a type assertion - so i'm thinking a typecast is needed?

abhinavdangeti: Agreed, we shouldn't do a type assert. But is a typecast needed at all - would you check?

Thejas-bhat: no i think typecast is a necessity since the type `index.TrainedIndexCallbackFn` is different from the func t.getCentroidIndex (which is of tyep `func(string) (interface{}, error)`) which was the cause of panics

abhinavdangeti (approved):

PR #2309, opened 2026-04-10
MB-71375: Bolt Wrappers for File Callbacks

Likith101:
- Wrap bolt.DB, bolt.TX and bolt.Bucket with impls
 - Override all bucket, transaction and key functions currently in use
 - Use a map to track keys and bucket names that require file callback handling
 - Process values internally before returning data outside the wrappers

### DIFF
--- a/util/bolt.go
+++ b/util/bolt.go
@@ -1,0 +1,167 @@
+//  Copyright (c) 2026 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


util/bolt.go
```diff
@@ -0,0 +1,167 @@
...
+	if bucket == nil {
+		return nil
+	}
+	return &BoltBucketImpl{
+		name:   string(name),
+		Bucket: bucket,
+	}
+}
+
+func (b *BoltBucketImpl) GetBucket(name []byte) *BoltBucketImpl {
```
abhinavdangeti, line 95: Let's rename this to `Bucket(..)` and replace the member attribute to `Bkt`.

abhinavdangeti (commented):
Looks so much better now @Likith101 👍🏼 .
One comment - let's keep the redirect API names the same, change the member attributes to avoid that collision.

util/bolt.go
Likith101: Changing the member attribute's name would cause the wrapper to fail for all bolt.Bucket functions/attributes that we have not explicitly redirected.

For example, Cursor() would now have to be defined for BoltBucketImpl as well for it to work. Along with all other bucket functions we are using right now.

abhinavdangeti: Ack 👍🏼

abhinavdangeti (approved):

PR #2311, opened 2026-04-14
MB-71383: Pass GPU error callbacks into zapx

capemox:
Introduces a mechanism for passing GPU error callbacks when instantiating a new index:
- User defined callbacks are passed into scorch config
- These callbacks are then put into the segment config
- `zapx` can then call them on GPU vectors search related errors


CascadingRadium:
Add the bleve level stat `TotGPUCloneFailed` in stats.go here and update it using the already existing `FieldStatsReporter`. The GPU detailing should not leak into the bleve layer.

PR #2316, opened 2026-04-20
MB-71383: Expose vector field stats in scorch

capemox:
Expose per-field vector index stats, which shows if the GPU index is in CPU memory or GPU memory.
Requires:
- https://github.com/blevesearch/scorch_segment_api/pull/77
- https://github.com/blevesearch/zapx/pull/400


CascadingRadium (dismissed):
good and clean solution 👍🏻

abhinavdangeti (approved):

PR #2317, opened 2026-04-20
MB-71516: Add unit test to reproduce result inconsistency issue

CascadingRadium:
- Add a unit test that reproduces the issue fixed by https://github.com/blevesearch/zapx/pull/401

### DIFF
--- a/search_knn_test.go
+++ b/search_knn_test.go
@@ -2988,82 +2988,66 @@ func TestKNNNullParams(t *testing.T) {
 }
 
 func TestVectorIndexExhaustion(t *testing.T) {
-	tmpIndexPath := createTmpIndexPath(t)
-	defer cleanupTmpIndexPath(t, tmpIndexPath)
+	for optimization := range index.SupportedVectorIndexOptimizations {
+		t.Run(optimization, func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


Thejas-bhat (commented):
but isn't this better to test out at the zap side itself?

search_knn_test.go
```diff
@@ -2986,3 +2986,84 @@ func TestKNNNullParams(t *testing.T) {
 		}
 	}
 }
+
+func TestVectorIndexExhaustion(t *testing.T) {
```
abhinavdangeti, line 2990: TestBivfVectorIndexExhaustion*

abhinavdangeti (commented):
Agree with @Thejas-bhat here, that it'd be good to add a simple unit test in zapx as well. Leaving this here is fine too 👍🏼

search_knn_test.go
CascadingRadium: This test is applicable to applicable to all index optimizations, hence I have named it so. Have extended it to cover any future optimizations as well.

capemox (dismissed):

Likith101 (dismissed):

abhinavdangeti (approved):

PR #2320, opened 2026-04-22
Updates to docs/vectors.md for v2.6.0

abhinavdangeti:

PR #2322, opened 2026-04-23
Update docs/vectors.md with BQ details

Likith101:


Likith101:
Not fully sure why this is the case, but the prettifier I use changed the formatting in a few extra places. It doesn't seem like anything changed visually so I didn't exclude it from the PR

abhinavdangeti:
@Likith101 would you rebase this over https://github.com/blevesearch/bleve/pull/2320. Also it seems we're touching a lot of lines we don't need to here, would you fix that.

Likith101:
I do not think an additional change is necessary here. #2320 seems to cover BQ and rabitq at a similar level of depth as compared to the rest of the vector types.

PR #2344, opened 2026-06-08
MB-71969 Propagate custom_filter/custom_score callback errors

maneuvertomars:
CustomFilterFunc and CustomScoreFunc now return an error so a failing
per-hit callback (e.g. a JavaScript exception from an embedder's custom
script) aborts the search and surfaces via SearchStatus.Errors, instead
of silently dropping the hit or falling back to the original score."

### DIFF
--- a/search/searcher/search_custom_filter.go
+++ b/search/searcher/search_custom_filter.go
@@ -32,8 +32,10 @@ func init() {
 
 // CustomFilterFunc decides whether a hit (with doc-value fields populated)
 // should be kept. Unlike FilterFunc it does not receive a SearchContext since
-// custom-query callbacks only need the DocumentMatch.
-type CustomFilterFunc func(d *search.DocumentMatch) bool
+// custom-query callbacks only need the DocumentMatch. A non-nil error aborts
+// the search so the failure can be surfaced to the caller rather than silently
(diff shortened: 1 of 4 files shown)


search/searcher/search_custom_filter.go
```diff
@@ -71,7 +73,11 @@ func (f *CustomFilterSearcher) Next(ctx *search.SearchContext) (*search.Document
...
 		}
-		if f.accept(next) {
+		keep, ferr := f.accept(next)
+		if ferr != nil {
+			return nil, ferr
+		}
+		if keep {
 			return next, nil
 		}
 		next, err = f.child.Next(ctx)
```
CascadingRadium, line 84: shouldn't we not be recycling the object when we discard the document?

We do this in FilteringSearcher

https://github.com/blevesearch/bleve/blob/f5f8138b24b5d5e7b91c94a209f4f7991d83aa8a/search/searcher/search_filter.go#L65

Same in `Advance` API as well.

maneuvertomars: Makes sense. Thanks

CascadingRadium (approved):

Thejas-bhat (approved):

PR #2352, opened 2026-06-23
MB-71967: Fix for SearchRequest.fields override by custom_*.fields

maneuvertomars:


capemox (dismissed):

search/searcher/search_custom_filter.go
```diff
@@ -72,10 +72,15 @@ func (f *CustomFilterSearcher) Size() int {
 func (f *CustomFilterSearcher) Next(ctx *search.SearchContext) (*search.DocumentMatch, error) {
 	next, err := f.child.Next(ctx)
 	for next != nil && err == nil {
+		// The fields loaded below are UDF input only; remember the hit's prior
+		// fields so we can restore them and avoid leaking the UDF's internal
+		// fields into (or overriding) SearchRequest.Fields in the response.
+		priorFields := next.Fields
```
CascadingRadium: Isn't `next.Fields` a map? If we modify `next.Fields` in `loadDocValuesOnHitWithTypes`, we will be modifying the reference which will be pointed to by `priorFields`, i think a better solution here would be to make `loadDocValuesOnHitWithTypes` return its own map, which we control here - to keep the ownership principle.

CascadingRadium (changes requested):

search/searcher/search_custom_filter.go
```diff
@@ -96,10 +102,15 @@ func (f *CustomFilterSearcher) Advance(ctx *search.SearchContext, ID index.Index
...
 		return nil, nil
 	}
-	if err = loadDocValuesOnHitWithTypes(adv, f.dvReader, f.indexReader, f.fieldTypes); err != nil {
-		return nil, err
+	// See Next: put the loaded fields on the hit only for the callback.
+	priorFields := adv.Fields
+	udfFields, lerr := loadDocValuesOnHitWithTypes(adv, f.dvReader, f.indexReader, f.fieldTypes)
+	if lerr != nil {
+		return nil, lerr
 	}
```
CascadingRadium, line 109: perhaps transfer all the `priorFields := adv.Fields` lines to just above `adv.Fields = udfFields`? So that it is more evident.

CascadingRadium (approved):

Likith101 (approved):

PR #2367, opened 2026-07-02
MB-71114 : Fix for KNN scoring details

maneuvertomars:
Fix for KNN scoring details when a doc matches both text and vector branches in a root level hybrid query

### DIFF
--- a/search_knn.go
+++ b/search_knn.go
@@ -479,8 +479,7 @@ func setKnnHitsInCollector(knnHits []*search.DocumentMatch, coll *collector.TopN
 		mergeFn := func(ftsMatch *search.DocumentMatch, knnMatch *search.DocumentMatch) {
 			// Boost the FTS score using the KNN score
 			ftsMatch.Score += knnMatch.Score
-			// Capture the merged explanation; MergeWith doesn't always mutate
-			// the receiver, so ignoring the return value drops the KNN part.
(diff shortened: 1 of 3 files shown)


search_knn.go
```diff
@@ -479,8 +479,9 @@ func setKnnHitsInCollector(knnHits []*search.DocumentMatch, coll *collector.TopN
 		mergeFn := func(ftsMatch *search.DocumentMatch, knnMatch *search.DocumentMatch) {
 			// Boost the FTS score using the KNN score
 			ftsMatch.Score += knnMatch.Score
-			// Combine the FTS explanation with the KNN explanation, if present
-			ftsMatch.Expl.MergeWith(knnMatch.Expl)
+			// Capture the merged explanation; MergeWith doesn't always mutate
```
CascadingRadium: Hey @maneuvertomars, thanks for catching this. I must have inadvertently caused this regression when i introduced the `MergeWith` API. 

LGTM, can you please retain the older comment though?

CascadingRadium: Also i think this holds in the 8.0.x-couchbase branch as well. Could you perhaps raise a back port commit too?

CascadingRadium (dismissed):

CascadingRadium (approved):

Likith101 (approved):

PR #2368, opened 2026-07-02
MB-71114 : Fix for KNN scoring details when a doc matches both text a…

maneuvertomars:
Fix for KNN scoring details when a doc matches both text and vector branches in a root level hybrid query


CascadingRadium (approved):
please merge

PR #2376, opened 2026-07-09
bug: Close bolt on error when updating index

CascadingRadium:
- When updating an index with an updated mapping, we open bolt via scorch's `OpenMeta` API but we never close it on error. 
- The error case can happen if the user has given an invalid mapping or if the updated mapping does not satisfy the criterion required for index update.
- Scorch's `Close()` API handles the dual case:
    - If only bolt is opened but not scorch, it closes only bolt.
    - If both bolt and scorch is opened, it closes both.
- Fix the issue by always closing scorch on error, irrespective of whether it was opened using `OpenMeta` or `Open` or `OpenMeta + Open`.

### DIFF
--- a/index_impl.go
+++ b/index_impl.go
@@ -213,9 +213,6 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
 		if !ok {
 			return nil, fmt.Errorf("updated mapping present for unupdatable index")
 		}
-
-		// Load the meta data from bolt so that we can read the current index
-		// mapping to compare with
 		err = ui.OpenMeta()
 		if err != nil {
 			return nil, err
--- a/index_impl.go
+++ b/index_impl.go
@@ -225,12 +225,12 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
 		if err != nil {
 			return nil, err
 		}
-		defer func(rv *indexImpl) {
-			if !rv.open {
-				rv.i.Close()
-			}
-		}(rv)
 	}
+	defer func(rv *indexImpl) {
+		if !rv.open {
+			rv.i.Close()
+		}
+	}(rv)
 
 	// now load the mapping
 	indexReader, err := rv.i.Reader()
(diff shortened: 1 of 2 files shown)


index_impl.go
```diff
@@ -213,9 +213,6 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
 		if !ok {
 			return nil, fmt.Errorf("updated mapping present for unupdatable index")
 		}
-
-		// Load the meta data from bolt so that we can read the current index
-		// mapping to compare with
```
Likith101, line 218: Why remove this comment?

index_impl.go
```diff
@@ -269,7 +266,8 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
 			return nil, err
 		}
 
-		fieldInfo, err := DeletedFields(im, um)
+		var fieldInfo map[string]*index.UpdateFieldInfo
+		fieldInfo, err = DeletedFields(im, um)
```
Likith101, line 270: Feels unnecessary

index_impl.go
```diff
@@ -269,7 +266,8 @@ func openIndexUsing(path string, runtimeConfig map[string]interface{}) (rv *inde
 			return nil, err
 		}
 
-		fieldInfo, err := DeletedFields(im, um)
+		var fieldInfo map[string]*index.UpdateFieldInfo
```
capemox, line 269: Why do we need declare the var separately?

index_impl.go
CascadingRadium: felt redundant

index_impl.go
CascadingRadium: its because we shadow the `err` variable in L151. By using `:=`, we basically ensure that the `err` variable in L151 is not shadowed.

index_impl.go
CascadingRadium: its because we shadow the `err` variable in L151. By using `:=`, we basically ensure that the `err` variable in L151 is not shadowed.

Likith101 (approved):

capemox (approved):

PR #2379, opened 2026-07-09
MB-71131: Support file writer callbacks for trained index

Thejas-bhat:
- implement the `fileWriterIDs` and `removeFileWriterIDs` APIs which fetch and remove the file writer IDs/keys corresponding to the trained segment. 
- refactor the request handling + ack mechanism of the trainLoop()
- update the destination index's bolt with the trained sample count, for consistent stats

### DIFF
--- a/index/scorch/train_vector.go
+++ b/index/scorch/train_vector.go
@@ -56,6 +56,10 @@ type vectorTrainer struct {
 	// the data vectors, returns trained centroid layout
 	trainedIndex *SegmentSnapshot
 	trainCh      chan *trainRequest
+	// closed by trainLoop when it exits (via close, training completion, or
+	// error). Producers select on it so they never block on a request that the
+	// trainer will never consume/ack.
+	doneCh chan struct{}
 }
 
 const IndexTrainedWithFastMerge = "vector_index_fast_merge"
@@ -68,6 +72,7 @@ func initTrainer(s *Scorch, config map[string]interface{}) *vectorTrainer {
 				parent:  s,
 				config:  maps.Clone(s.config),
 				trainCh: make(chan *trainRequest, 1),
+				doneCh:  make(chan struct{}),
 			}
 			// update the parent scorch config with the trainer's callback to fetch the trained index
 			s.segmentConfig[index.TrainedIndexCallback] = index.TrainedIndexCallbackFn(trainer.getTrainedIndex)
@@ -86,7 +91,6 @@ func moveFile(sourcePath, destPath string) error {
 	return nil
 }
 
-// update with the new writer API
 func (t *vectorTrainer) persistToBolt(trainReq *trainRequest) error {
 	tx, err := t.parent.rootBolt.Begin(true)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


index/scorch/train_vector.go
```diff
@@ -429,3 +441,125 @@ func (t *vectorTrainer) updateBolt(snapshotsBucket *util.BoltBucketImpl, key []b
...
+				if _, ok := ids[string(oldWriterID)]; ok {
+					err = trainerBucket.Put(util.BoltMetaDataFileWriterIDKey, []byte(writer.Id()), writer)
+					if err != nil {
+						return err
+					}
+				}
+			}
+		}
+		return nil
+	})
```
Likith101: If I am understanding this correctly, we have a trainer bucket within the top level snapshots bucket in bolt.
Within the trainer bucket, we write
- file path
- train complete
- num samples trained

I do not think these values need to be encrypted and so, essentially we have a full bucket that does not have need for encryption. If I am right here, then that means this entire step can be skipped and as can L538-L562.

Additionally, that means we can skip L246-L249 and create the reader using an empty string instead. Since we make use of loadSnapshot api, it will require a non nil reader, so making an empty reader should suffice.

index/scorch/train_vector.go
```diff
@@ -429,3 +441,125 @@ func (t *vectorTrainer) updateBolt(snapshotsBucket *util.BoltBucketImpl, key []b
...
+	trainedIndexPath := filepath.Join(t.parent.path, index.TrainedIndexFileName)
+	if encryptedIndex, ok := t.trainedIndex.segment.(segment.SegmentWithCallbacks); ok {
+		if _, ok := ids[encryptedIndex.CallbackId()]; ok {
+			_, _, err := t.parent.segPlugin.MergeUsing([]segment.Segment{t.trainedIndex.segment},
+				[]*roaring.Bitmap{nil, nil}, trainedIndexPath+".tmp", t.parent.closeCh, nil, t.config)
+			if err != nil {
+				return err
+			}
+
+			t.trainedIndex.segment.Close()
```
Likith101: In case the process panics or crashes at this stage, we will need some sort of cleanup for the temp file at startup. You can either choose to replace the existing file with the temp if you find it or you may choose to remove the temp. Either way should work fine

index/scorch/train_vector.go
```diff
@@ -429,3 +441,125 @@ func (t *vectorTrainer) updateBolt(snapshotsBucket *util.BoltBucketImpl, key []b
...
+		return nil
+	})
+
+	// invoke the merge API and let the zap side of things handle the update of writer IDs and also
+	// the data in the trained index.
+	trainedIndexPath := filepath.Join(t.parent.path, index.TrainedIndexFileName)
+	if encryptedIndex, ok := t.trainedIndex.segment.(segment.SegmentWithCallbacks); ok {
+		if _, ok := ids[encryptedIndex.CallbackId()]; ok {
+			_, _, err := t.parent.segPlugin.MergeUsing([]segment.Segment{t.trainedIndex.segment},
+				[]*roaring.Bitmap{nil, nil}, trainedIndexPath+".tmp", t.parent.closeCh, nil, t.config)
```
Likith101: The path should be trainedIndexPath+"temp"

Likith101 (changes requested):

index/scorch/train_vector.go
Thejas-bhat: > Additionally, that means we can skip L246-L249 and create the reader using an empty string instead. Since we make use of loadSnapshot api, it will require a non nil reader, so making an empty reader should suffice

I see, and in the case of an empty reader the data gets returned as is right? and also, in the loadSegment bit, there are certain bits in bolt (like the BoltDeletedKey) that are supposed to go through encryption path right? don't we need to pass in a valid writerID for that? so in which case we'll wanna generate the writerID while updating bolt's trainer bucket as well right?

Likith101: In case of an empty reader, the data gets returned as is.

Although there are certain bits in bolt that require encryption, none of them are being used for the trainer bucket. Let's take the deleted bitmap for example. We do not actually store one within the trainer bucket at all right? Same for all other keys that require encryption. So I think it is safe to do all operations within with empty readers and writers.

Likith101 (dismissed):

index/scorch/train_vector.go
```diff
@@ -130,112 +135,172 @@ func (t *vectorTrainer) persistToBolt(trainReq *trainRequest) error {
...
+			t.trainedIndex = nil
+			t.m.Unlock()
+			return fmt.Errorf("error closing previous trained index: %v", err)
+		}
+		if err := moveFile(path+".tmp", path); err != nil {
+			t.trainedIndex = nil
+			t.m.Unlock()
+			return fmt.Errorf("error renaming trained index: %v", err)
+		}
+		t.m.Unlock()
```
capemox: At this point, `prev` is closed and in the happy path prev (and hence `t.trainedIndex`) now points to a closed index. When you unlock the mutex, there's a chance a reader can still attempt to read the closed segment until the mutex is locked later in the function.

capemox (changes requested):

maneuvertomars (dismissed):

capemox (approved):

Likith101 (approved):

PR #2381, opened 2026-07-13
MB-72489: standalone query hot-path optimizations

capemox:
Five standalone performance optimizations on query hot paths, each independently benchmarked (Apple M4 Pro, benchstat). No functional/API changes.

### Changes

- **Fast-path `prepareDocumentMatch` + guard `adjustDocumentMatch` for non-KNN queries** — skip KNN adjustment and the always-false prepare branches on simple score-sorted queries. `Top{K}of{N}Scores` geomean **-1.8%**.
- **Close orphaned sub-searchers after unadorned disjunction optimization** *(correctness fix)* — `newDisjunctionSearcher` never closed the original N sub-searchers after the `disjunction:unadorned` optimization, so their TermFieldReaders were never released and search-accounting was skipped. Now closed immediately after the optimization succeeds.
- **Skip empty-bitmap work in `disjunction:unadorned` `Finish()`** — remove a dead O(segments × terms) loop and reuse empty/1-hit iterator singletons instead of allocating. Also lets the empty sentinel compose in nested disjunction/conjunction optimizations.
- **Specialized score-descending comparator for the top-N collector** — detect the common single-score-descending sort at construction time and compare `Score` directly instead of walking the generic sort-order path. Collector geomean **-11.6%** for score-sorted queries.
- **`MergeFieldTermLocations` fast-path + `DocumentMatch.Reset` nil-guard** — early-return when no match carries field-term locations (**-49%**), and skip `clear()` on a nil score breakdown (**-13%**). Both run per scored candidate.

Benchmarks for each change are included in the diff. Marked draft pending review.


search/search.go
```diff
@@ -235,8 +235,12 @@ func (dm *DocumentMatch) Reset() *DocumentMatch {
 	}
 	// remember the score breakdown map
 	scoreBreakdown := dm.ScoreBreakdown
-	// clear out the score breakdown map
-	clear(scoreBreakdown)
+	// clear out the score breakdown map; nil-guard because clear(nil) still
```
CascadingRadium: restore older comment

search/util.go
```diff
@@ -54,6 +54,9 @@ func MergeFieldTermLocations(dest []FieldTermLocation, matches []*DocumentMatch)
 			n += len(dm.FieldTermLocations)
 		}
 	}
+	if n == len(dest) {
+		return dest // fast path: no constituent has field term locations to merge
```
CascadingRadium: redundant comment

search/merge_reset_bench_test.go
```diff
@@ -0,0 +1,31 @@
+package search
+
+import "testing"
+
+// BenchmarkMergeFieldTermLocationsNoLocs: common no-highlighting case — several
```
CascadingRadium: this file is unnecessary

search/searcher/search_disjunction.go
```diff
@@ -69,6 +69,15 @@ func newDisjunctionSearcher(ctx context.Context, indexReader index.IndexReader,
 			rv, err := optimizeCompositeSearcher(ctx, "disjunction:unadorned",
 				indexReader, qsearchers, options)
 			if err != nil || rv != nil {
```
CascadingRadium:
```go
if err != nil {
    return nil, err
} 
if rv != nil {
	for _, s := range qsearchers {
			_ = s.Close()
		}
     return rv, nil
}
```

index/scorch/disj_unadorned_bench_test.go
```diff
@@ -0,0 +1,96 @@
...
+			}
+		}
+		benchDisjIdx = idx
+	})
+	return benchDisjIdx
+}
+
+var sinkOptimized index.Optimized
+
+func BenchmarkDisjunctionUnadornedFinish(b *testing.B) {
```
CascadingRadium: unnecessary file.

index/scorch/empty.go
```diff
@@ -38,4 +41,12 @@ func (e *emptyPostingsIterator) ResetBytesRead(uint64) {}
 
 func (e *emptyPostingsIterator) BytesWritten() uint64 { return 0 }
 
+// Implement OptimizablePostingsIterator so that anEmptyPostingsIterator can
```
CascadingRadium: redundant comment
also have you verified that the inteface is satisfied by all the iterator types used? We have the empty iterator (this one), zapx posting iterator and 2 more in unadorned.go.

index/scorch/optimize.go
```diff
@@ -355,16 +337,26 @@ func (o *OptimizeTFRDisjunctionUnadorned) Finish() (rv index.Optimized, err erro
...
 		}
 
+		// Fast path: no hits in this segment — reuse the zero-alloc empty sentinel.
+		if len(actualBMs) == 0 && len(docNums) == 0 {
+			oTFR.iterators[i] = anEmptyPostingsIterator
+			continue
+		}
+
+		// Fast path: exactly one 1-hit doc with no bitmaps.
+		if len(actualBMs) == 0 && len(docNums) == 1 {
```
CascadingRadium: seems like we can have this in L360?

search/collector/topn.go
```diff
@@ -127,9 +133,38 @@ func NewNestedTopNCollectorAfter(size int, sort search.SortOrder, after []string
 func newTopNCollector(size int, skip int, sort search.SortOrder, nr index.NestedReader) *TopNCollector {
 	hc := &TopNCollector{size: size, skip: skip, sort: sort}
 
-	hc.store = getOptimalCollectorStore(size, skip, func(i, j *search.DocumentMatch) int {
-		return hc.sort.Compare(hc.cachedScoring, hc.cachedDesc, i, j)
-	})
+	// Compute once up-front; comparator and store creation need them.
+	hc.neededFields = sort.RequiredFields()
```
CascadingRadium: expand 

```go
hc := &TopNCollector{
    size:
    skip:
    sort:
    cachedScoring:
    cachedDesc:
    store:
```
also bring in `hc.needDocIds` as equal to `sort.RequiresDocID()` within the expansion.

search/collector/topn.go
```diff
@@ -127,9 +133,38 @@ func NewNestedTopNCollectorAfter(size int, sort search.SortOrder, after []string
...
+	hc.cachedScoring = sort.CacheIsScore()
+	hc.cachedDesc = sort.CacheDescending()
+
+	// Specialize for the common case: single score-descending sort.
+	// SortOrder.Compare iterates a slice and checks two bool flags per call,
+	// adding ~40% overhead to every heap comparison. The direct float64 path
```
CascadingRadium: drop the comment and shift the callback to `sort.go` alongside the impl for `hc.sort.Compare(`. 

Also add a new API `getOptimalCollectorCompare` that returns the comparator to use. see

search/collector/topn.go
```diff
@@ -385,9 +420,11 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
 	if hc.nestedStore != nil {
 		currRoot := hc.nestedStore.Current()
 		if currRoot != nil {
-			err = hc.adjustDocumentMatch(searchContext, reader, currRoot)
-			if err != nil {
-				return err
+			if hc.knnHits != nil {
+				err = hc.adjustDocumentMatch(searchContext, reader, currRoot)
```
CascadingRadium: there's a check to see if knnHits is nil inside `adjustDocumentMatch`. You can drop that check now that we do the check outside

search/collector/topn.go
```diff
@@ -468,6 +505,18 @@ func (hc *TopNCollector) adjustDocumentMatch(ctx *search.SearchContext,
 func (hc *TopNCollector) prepareDocumentMatch(ctx *search.SearchContext,
 	reader index.IndexReader, d *search.DocumentMatch, isKnnDoc bool) (err error) {
 
+	// Fast path: score-sorted queries with no field loading, no KNN, no docID needs.
+	// Skips all conditional branches that are always false in this common case.
+	if hc.fastPrepare && !isKnnDoc {
```
CascadingRadium: this seems fragile. Can we instead have a 2 step prepare instead? 
The first step is mandatory and adds the things you have added here (setting hit number and sort etc) as a separate method. 

We can scaffold the check `hc.fastPrepare` outside then.

also add the check to see if fast prepare is possible in a new method

hit -> basicPrepare -> if !fastPrepare (prepareDocumentMatch) -> rest.

capemox:
@CascadingRadium still WIP, will add some more commits and resolve your comments later.

search/collector/topn.go
```diff
@@ -447,60 +482,57 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
 
 var sortByScoreOpt = []string{"_score"}
 
+// adjustDocumentMatch merges any KNN hit corresponding to d into d. Callers
+// must only invoke it when hc.knnHits != nil (checked at the call sites).
 func (hc *TopNCollector) adjustDocumentMatch(ctx *search.SearchContext,
```
Thejas-bhat: since this func is specific to KNN hit adjustment, it might be better to rename this to something like adjustKNNDocumentMatch() or similar

index/scorch/empty.go
capemox: yes, have checked

index/scorch/optimize.go
capemox: I don't fully understand, can you explain what you mean?

CascadingRadium: L360 is

```go
		} else {
			bm = roaring.New()
		}
```
which basically is equivalent to `if len(actualBMs) == 0`

So you can perhaps have the `len(docNums) == 1` \ `len(docNums) == 0` as a nested if block there.

search/collector/topn.go
```diff
@@ -158,17 +171,24 @@ func newTopNCollector(size int, skip int, sort search.SortOrder, nr index.Nested
...
-	}
-	hc.neededFields = sort.RequiredFields()
-	hc.cachedScoring = sort.CacheIsScore()
-	hc.cachedDesc = sort.CacheDescending()
-
 	return hc
 }
 
+// getOptimalCollectorCompare returns the comparator the collector should use for
+// its sort order. It specializes the overwhelmingly common single
```
CascadingRadium: please avoid redundant comments like these everywhere, the function is self descriptive. Perhaps shift the function near the `getOptimalCollectorStore` method?

search/collector/topn.go
```diff
@@ -447,60 +482,57 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
 
 var sortByScoreOpt = []string{"_score"}
 
-func (hc *TopNCollector) adjustDocumentMatch(ctx *search.SearchContext,
+// adjustKNNDocumentMatch merges any KNN hit corresponding to d into d. Callers
```
CascadingRadium: ditto remove comment

search/collector/topn.go
```diff
@@ -447,60 +482,57 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
...
+		// we have a knn hit corresponding to this document
+		hc.hybridMergeCallback(d, knnHit)
+		// remove this knn hit from the map as it's already
+		// been merged
+		delete(hc.knnHits, d.ID)
 	}
 	return nil
 }
 
+// basicPrepare performs the bookkeeping every collected hit needs regardless of
```
CascadingRadium: ditto, comment is bigger than the method

CascadingRadium:
hey @capemox, maybe the fuzzy stuff can be put into a different PR? 
Since this PR mostly deals with minor refactors and micro optimizations, it would be beneficial to save the fuzzy PR for later.

search/collector/topn.go
capemox: yep, there are a lot of verbose ai comments lol, will remove them all

index/scorch/unadorned.go
```diff
@@ -184,6 +198,23 @@ type ResetablePostingsIterator interface {
...
+// reuse rewinds them; the empty iterator is stateless and intentionally opts out
+// (ResetIterator is invoked via an optional type assertion).
+var (
+	_ segment.OptimizablePostingsIterator = (*emptyPostingsIterator)(nil)
+	_ segment.OptimizablePostingsIterator = (*unadornedPostingsIteratorBitmap)(nil)
+	_ segment.OptimizablePostingsIterator = (*unadornedPostingsIterator1Hit)(nil)
+
+	_ ResetablePostingsIterator = (*unadornedPostingsIteratorBitmap)(nil)
+	_ ResetablePostingsIterator = (*unadornedPostingsIterator1Hit)(nil)
+)
```
CascadingRadium: this is redundant?

index/scorch/unadorned.go
```diff
@@ -168,6 +168,20 @@ func (i *unadornedPostingsIterator1Hit) BytesWritten() uint64 {
 
 func (i *unadornedPostingsIterator1Hit) ResetBytesRead(uint64) {}
 
+// Implement OptimizablePostingsIterator so a 1-hit iterator produced by one
```
CascadingRadium: redundant comments here

search/collector/topn.go
```diff
@@ -81,6 +82,11 @@ type TopNCollector struct {
 	hybridMergeCallback search.HybridMergeCallbackFn
 
 	nestedStore *collectStoreNested
+
+	// fastPrepare is true when a hit needs only basicPrepare plus the shared
+	// score sort value (see canFastPrepare) — set once in Collect after loadID
+	// is known. Applies only to score-sorted queries with no field-loading needs.
```
CascadingRadium: redundant comments

search/collector/topn.go
```diff
@@ -447,60 +476,46 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
...
+func (hc *TopNCollector) basicPrepare(d *search.DocumentMatch) {
+	hc.total++
+	d.HitNumber = hc.total
+	if d.Score > hc.maxScore {
+		hc.maxScore = d.Score
+	}
+}
+
+func (hc *TopNCollector) canFastPrepare() bool {
+
```
CascadingRadium: nit: extra newline here

search/collector/topn.go
```diff
@@ -521,6 +553,25 @@ func (hc *TopNCollector) prepareDocumentMatch(ctx *search.SearchContext,
...
+	if hc.facetsBuilder != nil {
+		err = hc.visitFieldTerms(reader, d, func(field string, term []byte) {
+			hc.facetsBuilder.UpdateVisitor(field, term)
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	hc.basicPrepare(d)
```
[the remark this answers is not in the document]
CascadingRadium: i dont think we should be adding this line here. Put it outside like how we prepare the regular doc matches.

basicPrepare()
knnPrepare()

search/collector/topn.go
Thejas-bhat: no, i think its better to leave some in-line comment for quick ref, about what case fastPrepare is supposed to help with. so i've moved it from here to the place where its set

search/collector/topn.go
```diff
@@ -406,7 +436,7 @@ func (hc *TopNCollector) Collect(ctx context.Context, searcher search.Searcher,
 		// we may have some knn hits left that did not match any of the top N tf-idf hits
 		// we need to add them to the collector store to consider them as well.
 		for _, knnDoc := range hc.knnHits {
-			err = hc.prepareDocumentMatch(searchContext, reader, knnDoc, true)
+			err = hc.prepareKnnDocumentMatch(searchContext, reader, knnDoc)
```
CascadingRadium: also maybe rename to `pr/epareKNNDocumentMatch` to match `adjustKNNDocumentMatch`

Thejas-bhat (approved):

CascadingRadium (approved):

PR #2382, opened 2026-07-17
MB-72489: fuzzy/regexp candidate-collection optimizations

capemox:
Split out from #2381 per review — the fuzzy/regexp candidate-collection optimizations, as their own PR. No functional/API changes.

### Changes

- **Trim fuzzy candidate-collection overhead** (`search/searcher/search_fuzzy.go`): skip the per-candidate dedup map when the field has no synonyms (the dictionary iterator already yields each term once; the map only exists to de-dup synonym terms), replace the O(n²) `prefixTerm` rune-concat with a zero-alloc slice, and hoist repeated `ctx.Value` lookups.
- **Omit postings-count read for fuzzy/regexp candidate collection** (`index/scorch/snapshot_index*.go`): fuzzy/regexp collectors use only the term + edit distance and discard `DictEntry.Count`, yet the segment iterator read each candidate's postings list to compute it. A new opt-in `AutomatonIteratorOmitCount` path skips that read, wired via a graceful type-assertion fallback. `FieldDict`/`FieldDictPrefix`/`FieldDictRange` still populate counts.

**Depends on** blevesearch/zapx#436 — the omit-count optimization activates once the `zapx/v17` dependency is bumped to a release that includes `AutomatonIteratorOmitCount`. Until then the wiring falls back to the count-reading path (the wiring test skips), so this is safe to merge independently.

### Benchmark

`BenchmarkFuzzyCandidateCollection` (scorch index, 1000 terms / 3000 docs, fuzziness-2 query matching 280 candidates), original → both commits, Apple M4 Pro, count=12 via benchstat:

| | sec/op | B/op | allocs/op |
|---|---|---|---|
| micro-opts | −5.9% | −3.8% | −1.3% (988→975) |
| omit-count wiring | −9.7% | −1.1% | −29.1% (975→691) |
| **combined** | **−15.0%** | **−4.8%** | **−30.1%** (988→691) |

All p=0.000, n=12. Verified with the full fuzzy + synonym search suites and `-race`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/blevesearch/zapx/v14 v14.4.3
 	github.com/blevesearch/zapx/v15 v15.4.3
 	github.com/blevesearch/zapx/v16 v16.3.4
-	github.com/blevesearch/zapx/v17 v17.2.0
+	github.com/blevesearch/zapx/v17 v17.2.1
 	github.com/couchbase/moss v0.2.0
 	github.com/spf13/cobra v1.10.2
 	go.etcd.io/bbolt v1.4.0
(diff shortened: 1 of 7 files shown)


steveyen (dismissed):
lgtm

Thejas-bhat (approved):

steveyen (approved):

PR #2385, opened 2026-07-27
MB-72489: top-N collector optimizations 

maneuvertomars:
Top-N collector optimizations:

**Changes:**
- Replace `container/heap` with an inline ternary heap in the top-N collector — shallower siftDown, no interface dispatch. Collector benchmarks: ~14% faster (a hand-rolled binary heap measures ~10%, benchstat, n=10).
- `collectStoreHeap.Final()`: extract results via one in-place `sort.Slice` + sequential copy instead of repeated heap pops.
- `DocumentMatch.Reset()`: zero only the fields not restored afterwards, instead of wiping the whole struct.
- Early-stop for `score=none` + `Size`: stop scanning once `Size+From` hits are collected, when nothing depends on unseen matches (no facets, KNN, `SearchAfter`, nested, or field sort). Adds `SearchResult.TotalRelation` (`"eq"`/`"gte"`) since `Total` becomes a lower bound for such requests.
- Remove the unused `collectStoreList` type.


search/collector/list_test.go
```diff
@@ -0,0 +1,218 @@
...
+var errTestFixup = errors.New("fixup error")
+
+// noFixup is a no-op fixup function used throughout the list tests.
+var noFixup collectorFixup = func(*search.DocumentMatch) error { return nil }
+
+// TestCollectStoreListRoundTrip verifies that add() keeps elements in
+// ascending-score order (Front=worst, Back=best) and that Final(0, ...) returns
+// them best-first.  This exercises the core insertion-sort invariant of the
+// linked-list store, which had 0% coverage before this test.
+func TestCollectStoreListRoundTrip(t *testing.T) {
```
Thejas-bhat: this file change can be dropped because we don't use collectStoreList anymore, optionally i think its better to clean up that type in the collector package as well

search/collector/heap.go
```diff
@@ -15,85 +15,126 @@
...
 	"github.com/blevesearch/bleve/v2/search"
 )
 
+// collectStoreHeap is a min-heap of DocumentMatches where the root (heap[0])
+// is always the *worst* result (lowest score / highest sort key).  Popping the
+// root discards the worst element, so AddNotExceedingSize keeps the best k
+// documents at all times.
+//
+// Implemented as a ternary heap (3 children per node) instead of the standard
+// library's binary heap.  Height is log₃(n) vs log₂(n), so siftDown is
```
Thejas-bhat, line 29: i'm a bit skeptical about the ternary heap bit because the gains don't necessarily look good enough to justify having a custom heap implementation replace the stable and industry-standard container/heap implementation?

search/search.go
```diff
@@ -244,19 +240,39 @@ func (dm *DocumentMatch) Reset() *DocumentMatch {
...
 		descendants[i] = descendants[i][:0]
 	}
-	// idiom to copy over from empty DocumentMatch (0 allocations)
-	*dm = DocumentMatch{}
-	// reuse the []byte already allocated (and reset len to 0)
+	// Zero only the fields that are NOT restored below.  Nil/empty guards on
+	// pointer/string/map fields avoid triggering a GC write barrier for nil→nil
+	// stores, saving ~5 cycles per field across the many millions of Reset calls
+	// (common case: no explain, no fragments, no field highlights).
+	if dm.Index != "" {
```
Thejas-bhat: since this is a reset operation, should we be safe to zero out the following fields without the nil checks? also the comments feel redundant?

search/collector/topn.go
```diff
@@ -692,6 +710,22 @@ func (hc *TopNCollector) Total() uint64 {
 	return hc.total
 }
 
+// SetEarlyStop bounds the scan: Collect() stops pulling from the searcher once n
```
Thejas-bhat: redundant comments, its probably better to add a concise comment at the caller side

search.go
```diff
@@ -675,6 +685,10 @@ func (sr *SearchResult) Merge(other *SearchResult) {
 	sr.Status.Merge(other.Status)
 	sr.Hits = append(sr.Hits, other.Hits...)
 	sr.Total += other.Total
+	if other.TotalRelation == TotalRelationGte {
+		// Any constituent whose Total is a lower bound makes the merged Total one.
```
Thejas-bhat: remove this comment

index_impl.go
```diff
@@ -1036,16 +1053,23 @@ func (i *indexImpl) SearchInContext(ctx context.Context, req *SearchRequest) (sr
 		req.SearchAfter = nil
 	}
 
+	totalRelation := TotalRelationEq
+	if coll.EarlyStopped() {
+		// Total is a lower bound: the early-stop bounded scan stopped before
```
Thejas-bhat: remove this comment

search/collector/topn.go
```diff
@@ -84,6 +84,15 @@ type TopNCollector struct {
 	nestedStore *collectStoreNested
 
 	fastPrepare bool
+
+	// earlyStopN, when > 0, bounds the scan: once this many root hits have been
```
Thejas-bhat: these comments can be dropped and concisely explained at the func call sites

Thejas-bhat (changes requested):
please tag the associated MB-72489

search/collector/heap.go
maneuvertomars: I checked it using the repo's collector benchmarks (10 runs each, benchstat). A custom written binary heap is ~10% faster than container/heap, and the ternary version in this PR is ~14% faster, so most of the win is from dropping the heap.Interface indirection, but ternary still adds a few percent on top.

Like most of the advantage comes from dropping the heap.Interface, not from the arity. But given we custom write at all, ternary adds a consistent ~3 to 9% more (largest at k=1000) for essentially zero extra complexity. For example, Lucene made the same call that its top-k collector uses a custom written PriorityQueue rather than java.util.
Having said so, yes the advantage is not significant enough but what we can do here is rather relying on heap.Interface of Go, we can custom write our binary heap.

search/collector/topn.go
```diff
@@ -85,14 +85,8 @@ type TopNCollector struct {
...
 
-	// earlyStopN, when > 0, bounds the scan: once this many root hits have been
-	// collected, Collect() stops pulling from the searcher. Valid only when result
-	// identity does not depend on unseen docs (score="none", no facets, no field
-	// sort, no KNN, no nested, no SearchAfter). Set via SetEarlyStop before Collect.
-	earlyStopN int
-	// earlyStopped is set when Collect() actually broke out via earlyStopN, so the
-	// caller can report TotalRelation="gte" (Total is then a lower bound).
-	earlyStopped bool
+	earlyStopN   int  // when > 0, Collect() stops after this many hits (see SetEarlyStop)
```
Thejas-bhat: i think you can drop these comments since the func comments are quite sufficient

Thejas-bhat:
also, i feel like the PR description is way more verbose than necessary. can you please remove the redundant information that's there, and also update it with the latest patch in mind (removing the collectorList impl for eg)

Thejas-bhat (approved):

steveyen (approved):
perf-GAR!!

PR #2387, opened 2026-07-28
MB-72535: Cap leaf term searchers per index per search

Samsonnyyeet:
Add MaxTermSearchersKey: count every leaf term searcher (at the single choke point newTermSearcherFromReader) and fail the search once the count exceeds the limit.

The cap is per leaf index, not global: within a single index it spans the whole query tree, but across an IndexAlias each index is counted independently, so the effective ceiling scales with the number of indexes.

Unlike DisjunctionMaxClauseCount, which is enforced per fan-out node, this bounds the total across a single index's query tree rather than at any one node.

### DIFF
--- a/search/term_limit.go
+++ b/search/term_limit.go
@@ -1,0 +1,63 @@
+//  Copyright (c) 2026 Couchbase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 		http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package search
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+)
+
+// termSearchersCounter tracks the running total of leaf term searchers created
+// while searching a single index and the limit beyond which construction fails.
+type termSearchersCounter struct {
+	limit int
+	count atomic.Uint64
+}
+
+// ContextWithTermSearchersCounter installs a fresh counter for the limit at
+// MaxTermSearchersKey, when that limit is positive. It is called once per index
+// search (indexImpl.SearchInContext), so the counter is scoped to that index:
+// within a single index it spans the whole query tree (all fan-out nodes share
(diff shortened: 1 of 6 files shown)


search/term_limit.go
```diff
@@ -0,0 +1,61 @@
...
+	c, ok := ctx.Value(termSearchersCounterKey).(*termSearchersCounter)
+	if !ok || c == nil {
+		return nil
+	}
+	if n := c.count.Add(1); n > int64(c.limit) {
+		return fmt.Errorf("TooManyTermSearchers [%d > bleveMaxTerms,"+
+			" which is set to %d]", n, c.limit)
+	}
+	return nil
+}
```
Likith101: New line missing at the end of the file

search/term_limit.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+	"fmt"
+	"sync/atomic"
+)
+
+// termSearchersCounter tracks the running total of leaf term searchers created
+// for a single search and the limit beyond which construction fails.
+type termSearchersCounter struct {
+	limit int
+	count atomic.Int64
```
Likith101: Atomic Uint64 can be used here right? SInce this value does not go below 0

search/term_limit_test.go
```diff
@@ -0,0 +1,74 @@
...
+	t.Run("nil context", func(t *testing.T) {
+		if err := RecordTermSearcher(nil); err != nil {
+			t.Fatalf("nil ctx should be a no-op, got: %v", err)
+		}
+		//nolint:staticcheck // deliberately exercising nil-ctx tolerance
+		if got := ContextWithTermSearchersCounter(nil); got != nil {
+			t.Fatalf("nil ctx should be returned unchanged")
+		}
+	})
+}
```
Likith101: Same here

term_limit_integration_test.go
```diff
@@ -0,0 +1,138 @@
...
+		if strings.Contains(err.Error(), "TooManyClauses") {
+			t.Fatalf("should be the global term-count check, not the per-node clause check: %v", err)
+		}
+
+		// Same query, limit 8 -> allowed (8 is not > 8).
+		if err := runSearch(conjunctionOfDisjunctions(2, 4), 8); err != nil {
+			t.Fatalf("8 term searchers at limit 8 should succeed, got: %v", err)
+		}
+	})
+}
```
Likith101: Ditto

capemox:
@Samsonnyyeet is `bleveMaxTerms` supposed to be global (for the whole search request) or per leaf index? It looks like it's the latter, and you should change the commit message to be explicit about this.

Likith101 (approved):

Thejas-bhat (approved):

PR #2398, opened 2026-09-16
Replace prealloc-passing with Release()-based pooling for term postings

capemox:
## Summary

Part of the bitpack-simd effort (zapx blevesearch/zapx#447, scorch_segment_api companion PR): replaces the prealloc-passing pattern in `TermFieldReader()` and `prepareField()` with pooled reuse.

- `rv.postings[i]`/`rv.iterators[i]` used to be handed back in as a same-shaped "prealloc" hint on every lookup, but a recycled reader's shape is frequently wrong (a mix of 1-hit and general-encoded terms landing on the same field), so the hint was often discarded and allocated over anyway.
- Both call sites now call `releasePostings()` -- a type assertion against the new, optional `segment.Releasable` capability -- on a value they're done with, and fetch a fresh one with `prealloc` left `nil`, letting the segment implementation's own pool decide what to hand back.
- `segment.Releasable` is an opt-in capability, the same idiom as the existing `segment.OptimizablePostingsIterator`: values that don't implement it (e.g. the unadorned/synthetic iterators built by query optimization) make `releasePostings()` a no-op, unchanged from before.
- `go.mod` bumped to the `scorch_segment_api` `bitpack-simd` branch, which adds `Releasable` plus `MergeablePostingsList`/`MergeablePostingsIterator` as optional interfaces (not required additions to `PostingsList`/`PostingsIterator`, so zapx v11-v17 keep compiling unchanged).

## Test plan

- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./...` clean
- [x] `go test -race ./index/scorch/...` clean


capemox:
Putting this off for a future release -- reverting the split/pooling approach in zapx. Closing rather than merging.

